PR #3, opened 2017-12-17
ltoutput.dtx: missing latexrelease guard

aminophen:
I found an empty `\@addtonextcol` block inside latexrelease.sty;

~~~~ tex
\IncludeInRelease{2015/01/01}
  {\@addtonextcol}{float order in 2-column}%
\EndIncludeInRelease
~~~~

If this block is unintended, my docstrip guard fix will add one.


davidcarlisle:
I think you are right thanks

FrankMittelbach:
I will merge that in (but locally) as it will need a few additional documentation mods (and I need think about ho to do them best

PR #60, opened 2018-08-05
Fix using UTF-8 sequences that start with "F4

mirabilos:
There’s an off-by-one error in setting up \UTFviii@four@octets
for the codepoints. However, due to the too permissible nature
of the definition checks, \parse@XML@charref must additionally
be changed to reject too large values upon definition attempt.

Taken from https://github.com/mirabilos/tex-unicodedomino

### DIFF
--- a/base/utf8ienc.dtx
+++ b/base/utf8ienc.dtx
@@ -417,6 +417,7 @@
 %
 % Handle the single byte control characters.
 % \changes{v1.2a}{2018/03/24}{Loop over C0 controls added}%
+% \changes{v1.2e}{2018/08/05}{Fix "F4 lead byte}%
 % C0 controls are valid UTF-8 but defined to give the ``Character not defined error''
 % They may be defined with |\DeclareUnicodeCharacter|.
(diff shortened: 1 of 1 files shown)


davidcarlisle:
Thanks for this, the F4 is certainly wrong thanks (it looks like I added it
(incorrectly) when re-arranging the code to add additional checks earlier
this year)

David

davidcarlisle:
Thanks, I have moved the test for > 10FFFF slightly as otherwise it leaves the internal macros "half done" and so if you scrolled past that error  it would redefine whatever command was last defined, (for example in the github-0060 test just checked in, U+10FFFF definition would be over-written and defined to make the TOO BIG warning. As checked in, it makes the error message but if you scroll past that it sets up the 4-byte definition as before.

mirabilos:
David Carlisle dixit:

>Thanks, I have moved the test for > 10FFFF slightly as otherwise it
>leaves the internal macros "half done" and so if you scrolled past

Oh, okay, thanks!

PR #61, opened 2018-08-05
cosmetic: show invalid byte in hex, not decimal

mirabilos:
Bytes and byte sequences are customarily given hexadecimal; this also eases debugging in most editors


josephwright:
TeX uses hex numbers as `"...`, so if this is changing I'd favour that over `0x...`.

mirabilos:
OK, changed to use `"XX`

mirabilos:
I’ll continue this in #62 since they share another helper function

PR #63, opened 2018-08-05
cosmetics: format Unicode codepoints properly

mirabilos:
Instead of just writing `U+` and the codepoint in hex, format them as `U+XXXX` for BMP codepoints and `U-XXXXXXXX` for astral planes.

### DIFF
--- a/base/testfiles/tlb-inputenc-001.tlg
+++ b/base/testfiles/tlb-inputenc-001.tlg
@@ -24,7 +24,7 @@ l. ...thorn U+00fe [^^c3^^be
 Your command was ignored.
 Type  I <command> <return>  to replace it with another command,
 or  <return>  to continue without it.
-! Package inputenc Error: Unicode character ^^cc^^81 (U+301)
+! Package inputenc Error: Unicode character ^^cc^^81 (U+0301)
 (inputenc)                not set up for use with LaTeX.
 See the inputenc package documentation for explanation.
 Type  H <return>  for immediate help.
@@ -33,7 +33,7 @@ l. ...comb acute U+0301 [^^cc^^81
                                 ]
 You may provide a definition with
 \DeclareUnicodeCharacter 
-! Package inputenc Error: Unicode character ^^f0^^9d^^94^^84 (U+1D504)
+! Package inputenc Error: Unicode character ^^f0^^9d^^94^^84 (U-0001D504)
 (inputenc)                not set up for use with LaTeX.
 See the inputenc package documentation for explanation.
 Type  H <return>  for immediate help.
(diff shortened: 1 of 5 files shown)


davidcarlisle:
Thanks, I've been meaning to fix the display of at least the small ones like U+7 for some time:-)

aminophen:
`\ifnum#1<1048576 U-000%`

Is this really wanted? I see Unicode Character "🦆" (Duck) represented as U+1F986, not U-0001F986. https://www.compart.com/en/unicode/U+1F986

aminophen:
See also:

- https://stackoverflow.com/questions/1273693/why-is-u-used-to-designate-a-unicode-code-point
- http://unicode.org/mail-arch/unicode-ml/y2005-m11/0060.html

mirabilos:
Hironobu Yamashita dixit:

>`\ifnum#1<1048576 U-000%`
>
>Is this really wanted?

Yes.

>I see Unicode Character "🦆" (Duck) represented as U+1F986, not
>U-0001F986. https://www.compart.com/en/unicode/U+1F986

That’s a recent, alternative, form. The original Unicode forms
were U+nnnn and U-nnnnnnnn, and people just forgot that U+ was
for BMP only when Emoji were introduced, because the masses were
stupid, so that form was eventually accepted.

It is still more correct and, to my eyes at least, much preferrable
to see U-0001F986 for astral plane codepoints, even if the limit
is now U-0010FFFF.

davidcarlisle:
On Sun, 30 Sep 2018 at 16:48, mirabilos <notifications@github.com> wrote:

> Hironobu Yamashita dixit:
>
> >`\ifnum#1<1048576 U-000%`
> >
> >Is this really wanted?
>
> Yes.
>
> >I see Unicode Character "🦆" (Duck) represented as U+1F986, not
> >U-0001F986. https://www.compart.com/en/unicode/U+1F986
>
> That’s a recent, alternative, form. The original Unicode forms
> were U+nnnn and U-nnnnnnnn, and people just forgot that U+ was
> for BMP only when Emoji were introduced, because the masses were
> stupid, so that form was eventually accepted.
>
> It is still more correct and, to my eyes at least, much preferrable
> to see U-0001F986 for astral plane codepoints, even if the limit
> is now U-0010FFFF.
>

I must admit I prefer the more compact   U+ form, I nearly reverted that
bit before accepting
the PR but decided to let it go . U-0001F986 is hopelessly verbose, I
wonder why you prefer it?
(actually I never understood the logic of changing between + and -
anyway:-)

David

davidcarlisle:
given that the mail on the unicode list saying U+ is officially allowed is dated from 13 years ago, this change isn't that recent:-) I think I may go back to U+ form (but keep the 0 padding for small numbers.

mirabilos:
David Carlisle dixit:

>(actually I never understood the logic of changing between + and -
>anyway:-)

It has to do with BMP and non-BMP, and the needing of surrogates
for 16-bit Unicode environments (such as Java™ or Microsoft® NT)
and the differing escapes (\uXXXX vs. \UXXXXXXXX in many langua‐
ges). It keeps things consistent to retain the U-00XXXXXX form.

davidcarlisle:
Just to confirm that I reverted to the U+ form, the Unicode standard makes it clear that that is (now) the intended use, only mentioning the U- form briefly in in a table of extended  BNF and consistently using U+ with  4 5 or 6 hex digits everywhere else.

See also the first section of :
https://www.unicode.org/versions/Unicode11.0.0/appA.pdf

PR #82, opened 2018-10-04
cosmetics: show invalid bytes in hex, not decimal

mirabilos:
This was originally part of https://github.com/latex3/latex2e/pull/62 which was rejected (for now, and with reason), but independent of the rejected part, and so still should be merged.

### DIFF
--- a/base/testfiles/tlb-utf8-undec-cp1252.tlg
+++ b/base/testfiles/tlb-utf8-undec-cp1252.tlg
@@ -9,7 +9,7 @@ l. ...a^^d6 ^^93
                x^^94 ^^df
 You may provide a definition with
 \DeclareUnicodeCharacter 
-! Package inputenc Error: Invalid UTF-8 byte 148.
+! Package inputenc Error: Invalid UTF-8 byte "94.
 See the inputenc package documentation for explanation.
 Type  H <return>  for immediate help.
  ...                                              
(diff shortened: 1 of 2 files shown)


mirabilos:
Travis fails unrelatedly: https://github.com/latex3/latex2e/compare/bcccc7c42356...b2b32f2af1d7 introduced a change without adapting the testsuite, causing https://travis-ci.org/latex3/latex2e/builds/437124178 to fail.

mirabilos:
The Travis failure is unrelated, as far as I can tell, and caused by a pre-existing bug.

josephwright:
@mirabilos The Travis-CI failure relates to changes in LuaTeX/`luaotfload`: the team (@davidcarlisle and I) are aware of it!

davidcarlisle:
Not really a bug just luatex changes again, I need to see if I can wrap the test suite to get the old log files back or whether we need to add (a lot of) new luatex-specific tlg files to match the new luatex log output, I'm hoping to avoid doing the latter....

davidcarlisle:
I don't think there is need to add another decimal-to-hex macro, we could simply use

`Invalid UTF-8 byte "\UTFviii@hexnumber{`#1}`

as the only difference is the test for numbers less than 16 which doesn't apply here (as they are not invalid UTF-8)

davidcarlisle:
Thanks slightly modified change adopted as noted above.

PR #83, opened 2018-10-06
rewrite to use a safer UTF-8 decoder

mirabilos:
This has two slight API changes:

 * `\UTFviii@check@continue` (an internal but documented macro) is gone
 * the calling convention of `\UTFviii@checkseq` changes from
  `\if\relax\expandafter\UTFviii@checkseq\string#1\relax\relax%` to
  `\if\relax\expandafter\UTFviii@checkseq\string#1\empty\relax%`

Please check (as you told me on the TeX.SE chat you do when introducing changes like these) that there are no external users of either `\UTFviii@check@continue` (there should not be any, as it only makes sense inside checkseq) or `\UTFviii@checkseq` (in which case this would need to be changed).

The internal, _un_documented macros `\UTFviii@cleanup` and `\dec@de@UTFviii` are also gone; these only made sense within `\decode@UTFviii` and therefore should not have any callers either.

### DIFF
--- a/base/testfiles/github-0060.tlg
+++ b/base/testfiles/github-0060.tlg
@@ -17,15 +17,17 @@ MAX VALUE!!
 ===110FFF
 TOO BIG!!
 ===110FFE
-! Package inputenc Error: Unicode character ^^f4^^90^^bf^^be (U+110FFE)
-(inputenc)                not set up for use with LaTeX.
+! Package inputenc Error: Invalid UTF-8 byte sequence.
 See the inputenc package documentation for explanation.
 Type  H <return>  for immediate help.
  ...                                              
 l. ...[^^f4^^90^^bf^^be
                       ]% trying to be U+110FFE, too high, undeclared
-You may provide a definition with
-\DeclareUnicodeCharacter 
+The document does not appear to be in UTF-8 encoding.
+Try adding \UseRawInputEncoding as the first line of the file
+or specify an encoding such as \usepackage [latin1]{inputenc}
+in the document preamble.
+Alternatively, save the file in UTF-8 using your editor or another tool
 ===bad 2 byte
 ! Package inputenc Error: Invalid UTF-8 byte sequence.
 See the inputenc package documentation for explanation.
--- a/base/utf8ienc.dtx
+++ b/base/utf8ienc.dtx
@@ -292,7 +292,7 @@
 % Test if the sequence is invalid UTF-8 or valid UTF-8 but without
 % a \LaTeX\ definition.
 %    \begin{macrocode}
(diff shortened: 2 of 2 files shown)


davidcarlisle:
Thanks for this, but my initial reaction is that this is showing diminishing returns, it is adding a lot of code for very minor edge case of malfomred UTF-8. The overwhelmingly most common reason for malformed UTF-8 sequences is that the file isn't UTF-8 at all, so the important thing is to flag an error, detailed byte level checking isn't that useful.

The one test case that changed  the new code saves one loop, in that currently you get a "not set up" error but if you try to declare that number you get an error that the number is out of range. With this change you get an error that the initial byte sequence is invalid, which isn't clearly an improvement for most users. Either way it's just changing the error message, not catching something that was previously not an error.

I'll leave open for a bit to allow review from others.

davidcarlisle:
> It *does* newly catch non-minimal-encoding malformed characters, such as using "C1 "80 for "40 (@). 

well not really as (without poking into the internals, and if users do that they can circumvent all checks) the only way to declare a character is by number and that will use the canonical minimal utf8 sequence so you should always get a "not set up" error for  bad sequences, or as in the case of C1 *0 you already get

`! Package inputenc Error: Invalid UTF-8 byte "C1.`

mirabilos:
David Carlisle dixit:

>sequence so you should always get a "not set up" error for bad

I’ve been able to actually set up bad sequences.

>sequences, or as in the case of C1 *0 you already get
>
>`! Package inputenc Error: Invalid UTF-8 byte "C1.`

Yesssss, which is why I wrote

>>… (This is just an example of
>>non-minimal encoding, not one of the actual things caught, there
>>are others, and the C0/C1 case is easy to catch, but something
>>in the three- or four-byte encoding forms isn’t.)

davidcarlisle:
> I’ve been able to actually set up bad sequences.
by defining the internal csname by hand, or do you mean that it's possible without defining internal commands to have a file with a bad utf-8 sequence to pass without error? I can't currently see that.

mirabilos:
>by defining the internal csname by hand, or do you mean that it's
>possible without defining internal commands to have a file with a bad
>utf-8 sequence to pass without error? I can't currently see that.

I think it was possible, but I would have to try and reproduce it
again. I’m currently somewhat deep in other contexts…

davidcarlisle:
Thanks again for this series  of PRs they definitely improved the UTF-8 handling, however as noted in comments above, this final one is adding delicate code that is only changing which error message is generated. Especially as UTF-8 is now enabled by default in LaTeX so this would affect the code paths in every latex document,  it is not clear the benefits are enough to justify the risks, we plan to close with no action.

PR #110, opened 2019-01-09
utf8ienc.dtx: Declare ⟨ (U+27E8), ⟩ (U+27E9)

adunning:
Although ⟨ (U+27E8, `\textlangle`) and ⟩ (U+27E9, `\textrangle`) are 'mathematical angle brackets', Unicode recommends them for text use in place of U+2329 and U+232A. These characters are deprecated and canonically decompose to U+3008 and U+300B, in the CJK Symbols and Punctuation block. See <https://unicode.org/consortium/utc-minutes/UTC-088-200108.html>.

### DIFF
--- a/base/testfiles/sx237381.tlg
+++ b/base/testfiles/sx237381.tlg
@@ -77,6 +77,8 @@ File: ts1enc.dfu ....-..-.. v... UTF-8 support for inputenc
    defining Unicode char U+25E6 (decimal 9702)
    defining Unicode char U+25EF (decimal 9711)
    defining Unicode char U+266A (decimal 9834)
+   defining Unicode char U+27E8 (decimal 10216)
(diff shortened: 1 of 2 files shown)


adunning:
I am assuming that the definitions of U+2329 and U+232A should stay in the file for backwards compatibility, although they're not technically correct. In practice those characters are sometimes used where U+27E8 and U+27E9 are intended, but I've never actually seen a file that uses them for U+3008 and U+300B.

davidcarlisle:
@adunning  don't worry too much about the tests, if we make the change it's easy enough to bulk update the test results locally, travis is set to stop at the first failure so you just get told one failure at a time.

adunning:
Ah, thank you!

And yes, I'm not sure what the consortium were thinking when they defined the characters in this way, but we seem to be stuck with this situation.

PR #169, opened 2019-08-19
[doc] add missing macro envs for \newtoks and \newfam

muzimuzhi:


FrankMittelbach:
we don't any longer add anything (pull or direct update) to master we now only work in the develop branch and merge that with master when we release to CTAN. So in the future, please make pull requests against develop (or one of its feature branches if that makes more sense). We will update the guidelines accordingly. I'll therefore include your suggestion there and close the pull.

muzimuzhi:
Thanks. I do hesitated when opened this pr to master branch.

PR #245, opened 2020-01-08
Fixdpc8

u-fischer:
Hopefully resolves https://github.com/davidcarlisle/dpctex/issues/8.

### DIFF
--- a/required/tools/longtable.dtx
+++ b/required/tools/longtable.dtx
@@ -1213,7 +1213,7 @@
 %
 % \begin{macro}{\LT@start}
 % % \changes{v4.13}{2020/01/07}
-%      {fix issue https://github.com/davidcarlisle/dpctex/issues/8 }
+%      {fix issue https://github.com/davidcarlisle/dpctex/issues/8}
 % This macro starts the process of putting the table on the current
 % page. It is not called until either a "\\" or "\endlongtable" command
 % ends a chunk, as we do not know until that point which of the four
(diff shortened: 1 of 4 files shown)


required/tools/longtable.dtx
```diff
@@ -1223,7 +1225,7 @@
 %    \begin{macrocode}
 \def\LT@start{%
   \let\LT@start\endgraf
-  \endgraf\penalty\z@\vskip\LTpre
+  \endgraf\penalty\z@\vskip\LTpre\endgraf
```
FrankMittelbach: I think the additional \endgraf is weird enough to deserve an explanation, after all we are in vertical mode at this point. This is to get the page builder executed I presume? If so th should be mentioned.

u-fischer: @FrankMittelbach  The problem is that in the next line \dimen@ is set to \pagetotal but without the \endgraf this doesn't take the \LTpre into account as \pagetotal is updated only after the \endgraf and so the calculation is wrong.

FrankMittelbach: I understand (this is triggering the page builder as I assumed) but I think it should be mentioned/documented that it is there for this reason

FrankMittelbach:
your call. I decided that I'm not adding anything in unless it was distributed as -dev or is a hotfix that is just needed (or is a fix for a -dev issue).

davidcarlisle:
> 
> 
> your call. I decided that I'm not adding anything in unless it was distributed as -dev or is a hotfix that is just needed (or is a fix for a -dev issue).

I wouldn't do the other lt fixes at this time, but this is just an \endgraf that usually does nothing and in the cases where it changes the behaviour the current code is clearly wrong, so I'm minded to merge (probably by hand rather than accept this pr  and resolving the conflicts in the tlmgr setup

davidcarlisle:
Main change to lt updated from this branch to develop and change log updated there so closing this pr

PR #250, opened 2020-01-17
Improve indentation

dr-scsi:
* base/doc/fntguide.tex: Set fill-column to 72 and re-indent the
source, delete white spaces.  No content-wise changes.

Hi all,

I'm thinking about suggesting (as time permits) some updates to `fntguide.tex` which track the latest changes to the kernel.  Before getting into that, I wrote a a style file for AUCTeX in order to support `ltxguide.cls`, opened the file in Emacs and made some re-indentation of the code to 72 chars, deleted unnecessary white spaces etc.  This is the reformatted result with not a single content change for review.

Any comments welcome.

### DIFF
--- a/base/doc/fntguide.tex
+++ b/base/doc/fntguide.tex
@@ -1,206 +1,203 @@
 % \iffalse meta-comment
 %
-% Copyright (C) 1993-2019 
+% Copyright (C) 1993-2019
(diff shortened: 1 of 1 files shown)


FrankMittelbach:
thanks

PR #266, opened 2020-02-06
use \Ucharcat for \symbol under XeTeX

### LINKED ISSUE #124: Patch `\symbol` under XeTeX (or fix upstream)
## Brief outline of the bug

In XeTeX `\Uchar` should be used instead of `\char` in some cases.
More details here: https://github.com/wspr/unicode-math/issues/499

```latex
\RequirePackage{latexbug}
\documentclass{article}
\begin{document}
\font\x="[STIX2Math.otf]"\relax\x
{$\tracingall\symbol{"1D400}$}
\end{document}
```

```
\symbol #1->\char #1\relax 
#1<-"1D400
{\char}
./small.tex:8: Bad character code (119808).
<to be read again> 
                   \relax 
l.8 {$\tracingall\symbol{"1D400}
```

wspr:
closes #124


base/changes.txt
```diff
@@ -4,6 +4,17 @@ completeness or accuracy and it contains some references to files that
...
 =======================================================================
 
+#########################
+# 2020-02-07 PL X Release
+#########################
+
+2020-02-07  Will Robertson  <will.robertson@latex-project.org>
+
+	* ltfssini.dtx:
+	XeTeX-specific definition for \symbol to avoid bug in maths mode (gh/124)
```
zauguin: Is there any reason why this is specifically a bug in math mode? If I'm not mistaken the same issue should appear in other modes.

wspr: No I haven't tried to trace the issue in the engine. Long story short, `\char"1D400\relax` is fine in text mode but fails in maths mode.

PR #323, opened 2020-04-15
Make U+00AD active for Unicode engines

zauguin:
Make U+00AD an active alias for `\-` on LuaTeX and XeTeX. This gives U+AD the proper behavior and also improves compatibility with non-Unicode engines.

(The PR by itself doesn't actually has any visible effect right now because the babel hyphenation code globally resets the catcode, but I think that should be changed in babel directly.)
 
# Internal housekeeping

## Status of pull request

- Feedback wanted 

- Ready to merge

- [ ] Test file(s) added
- [X] `ltnewsX.tex` updated

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -11,6 +11,11 @@ are not part of the distribution.
 	* ltxref.dtx:
 	added \@currentcounter to \refstepcounter (gh/300)
 
+2020-05-06  Marcel Krüger    <Marcel.Krueger@latex-project.org>
+
+  * ltspace.dtx:
+  Make Unicode softhyphen U+00AD active and defined as \-.
+
 2020-05-02  Johannes Braams  <texniek at braams.xs4all.nl>
 
 	* ltexpl.dtx: Created aliases for two expl3 macros in order to use
--- a/base/doc/ltnews32.tex
+++ b/base/doc/ltnews32.tex
@@ -517,6 +517,17 @@ documentation~\cite{32:babel}.
 
 
 
+\subsection{Add support for Unicode soft hyphens}
+
+For a long time, the UTF-8 option for \pkg{inputenc} made the Unicode
+soft hyphen character (U+00AD) an alias for the \LaTeX\ soft hyphen
+|\-|. The Unicode engines \XeTeX{} and \LuaTeX{} behaved
+different though: They either ignored U+00AD or interpreted it as an
+unconditional hyphen. This inconsistency is fixed now and \LaTeX{}
+always treats \texttt{U+00AD} as |\-|.
+
+
+
 \begin{thebibliography}{9}
 
 \fontsize{9.3}{11.3}\selectfont
--- a/base/ltspace.dtx
+++ b/base/ltspace.dtx
@@ -32,7 +32,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{ltspace.dtx}
-             [2020/04/21 v1.3m LaTeX Kernel (spacing)]
+             [2020/05/06 v1.3n LaTeX Kernel (spacing)]
 % \iffalse
 \documentclass{ltxdoc}
 \GetFileInfo{ltspace.dtx}
@@ -1361,6 +1361,18 @@
 % \end{macro}
 % \end{macro}
 % \end{macro}
+%
+% \changes{v1.3n}{2020/05/06}{Made softhyphen active in TU engines}
+%
+%   For Unicode engines, make the Unicode soft hyphen an active
+%   character defined as \cs{-}.
+%
+%    \begin{macrocode}
+\ifx\Umathcode\@undefined\else
+  \catcode "AD=13
+  \def^^ad{\-}
+\fi
+%    \end{macrocode}
 %
 % \begin{macro}{\obeycr}
 % \begin{macro}{\restorecr}


car222222:
Please do not do this before we have a general plan for dealing with such active chracters without giving each an arbitrary hard-coded definition that precludes there use for any purpose other than a particular typesetting methodology and conventions.

FrankMittelbach:
If you want I can resolved the conflicts and merge.

zauguin:
> I take it back :-) I didn't realize what you meant yesterday about being robust. I now saw you used \DeclareRobustCommand and that will most likely fail with ligatures and kerns as you feared. Instead
> \DeclareTextCommandDefault needs to be used.
> And we probably should add at least 1 test that checks if the character if used kills kerning (which it shouldn't).

While adding a test here, I realized that the issue exists on a much deeper level: The TeX primitives `\-` and `\discretionary` already suppress ligatures and kerning in all traditional engines (where "traditional" means non-Lua). So while preserving them would be nice, I don't really see a way yet to archive that. Of course `\DeclareTextCommandDefault` is still useful if the command is redefined to be empty. But there is another issue: `\DeclareTextCommandDefault` breaks ligatures/kerning when the command is used for the first time with a specific encoding (because it defines the encoding specific command first). There is not really a point in encoding specific commands here because a softhyphen isn't outputencoding specific, so this seems like a useless restriction. Is there a variant of `\DeclareTextCommandDefault` which always preserves ligatures/kerning?

wspr:
I actually think that using the output encoding functionality (i.e., LICRs and related TextCommand stuff) is a red herring here. Generally, I think that soft hyphens don’t end up in output, unless you’re creating HTML, and in that case I think you would want soft hyphens in all feasible locations not just explicitly marked ones.

IMO for TeX purposes the soft hyphen Unicode char is only sensible to think about as an input char, and needs to be defined using a different framework than LaTeX’s font encoding technology. Unless I am mistaken, all of the \textXYZ symbol commands are essentially assuming that they will result in some font-glyph-related typesetting output.

When it appears in input text, the soft hyphen char can either be interpreted as “misleading input via copy/paste, ignore” or “sophisticated user really means soft hyphen, use \-“. Where \- isn’t intended to insert a literal soft hyphen Unicode glyph into a PDF but rather act as a typesetting directive to indicate a discretionary hyphenation point that may or may not be acted upon.

Which in my mind brings us back to the “shorthand” idea where we are really acting at an "input encoding" level. I’d probably propose something like \charXYZ as a naming convention here, where you might say

    \def\charsofthyphen{}% OR
    \def\charsofthyphen{\-}

and have an active character definition roughly like I proposed in #703 to define the input char to be a protected macro defined in terms of \charsofthyphen which can then be customised as needed by the user.

To go back to previous examples that have been discussed elsewhere, we might separately have something like an active em-dash character which is defined in terms of

    \def\charemdash{\textemdash}

which “unites” the shorthand idea with the font encoding LICR idea.

But I could be entirely wrong here — I’m not at all convinced I’m on the right track :) This is just what seems sensible to me right now...

zauguin:
> I actually think that using the output encoding functionality (i.e., LICRs and related TextCommand stuff) is a red herring here. Generally, I think that soft hyphens don’t end up in output, unless you’re creating HTML, and in that case I think you would want soft hyphens in all feasible locations not just explicitly marked ones. IMO for TeX purposes the soft hyphen Unicode char is only sensible to think about as an input char, and needs to be defined using a different framework than LaTeX’s font encoding technology. Unless I am mistaken, all of the \textXYZ symbol commands are essentially assuming that they will result in some font-glyph-related typesetting output. When it appears in input text, the soft hyphen char can either be interpreted as “misleading input via copy/paste, ignore” or “sophisticated user really means soft hyphen, use \-“. Where \- isn’t intended to insert a literal soft hyphen Unicode glyph into a PDF but rather act as a typesetting directive to indicate a discretionary hyphenation point that may or may not be acted upon.

 I tend to agree here. The entire idea of the text commands being able to react to output encodings does  seem a bit out of place when talking about soft hyphens because there is no real encoded output.

> Which in my mind brings us back to the “shorthand” idea where we are really acting at an "input encoding" level. I’d probably propose something like \charXYZ as a naming convention here, where you might say \def\charsofthyphen{}% OR \def\charsofthyphen{\-} and have an active character definition roughly like I proposed in #703 to define the input char to be a protected macro defined in terms of \charsofthyphen which can then be customised as needed by the user.

I think `\text...` can still make sense because the command, even if it not really is a text command, is used basically in the same way acts like one in most circumstances. I wouldn't mind a different name though. In any case, I don't think having the character being a protected macro and `\charsofthyphen` expandable would be the right way. First, this would probably only really work on Unicode engines (I might be wrong here, but I think that "protected active UTF-8 chars" do not work with utf8 inputenc). Second it would make `\charsofthyphen` a really odd macro. It would no longer correspond to the behaviour of a real softhyphen in an input but instead be expandable but most importantly, writing this into a file would result in an actual U+00AD being written there.

In my opinion, especially invisible Unicode characters like U+AD which are interpreted by TeX in some way should become visible in most output streams to give users a chance to determine where the changes came from.

So I think for these characters the right approach might be to have a stripped down version of TextCommands (CharCommands?), which appear as part of text handling but are not in any way related to output encodings or any other kind of actual glyph output. This would be protected in a way similar to TextCommands, while the actual active character would then be expandable and expand to this CharCommand.

> To go back to previous examples that have been discussed elsewhere, we might separately have something like an active em-dash character which is defined in terms of \def\charemdash{\textemdash} which “unites” the shorthand idea with the font encoding LICR idea. But I could be entirely wrong here — I’m not at all convinced I’m on the right track :) This is just what seems sensible to me right now...

car222222:
One small comment on names: I do not think that the collection of commands of the form \text... was ever thought of as being somehow formally defined or even necessarily having common properties.

car222222:
A completely different question: is there a particular reason for adding this to the file ltspace.dtx?
It would seem to fit better in the section on hyphens etc. in ltdefns.

zauguin:
@car222222 
> A completely different question: is there a particular reason for adding this to the file ltspace.dtx?
> It would seem to fit better in the section on hyphens etc. in ltdefns.

I had a really hard time trying to fond any fitting file and then decided for `ltspace.dtx` because the description

> This section deals with spacing, and line- and page-breaking.

seemed fitting, given that this is basically a linebreaking control character.
(I didn't find the discretionary hyphenation part of ltdefns then, but that feels odd anyway. I would be ok with moving this there, but maybe it would make more sense to move the disretionary hyphenation part to ltspace)

car222222:
> maybe it would make more sense to move the disretionary hyphenation part to ltspace
This would make sense too.

I did notice that the Unicode documentation does make it clear that its (historically determined) name us misleading as its semantics are not at all tied to hyphen characters/glyphs.  A better name would indeed have been SOFT LINEBREAK.

car222222:
Going back to the motivating problem: am I right that this was, at least in part, motivated by problems from unwanted (incorrect, even:-) 00AD’s getting into .tex files?
To avoid the need for tedious editing of these file (including the regal of invisible characters!) it would be good to provide a command to set them (locally) as ‘ignored’ (probably by defining 
\textsofthyphen to ‘empty’) . 
 [As @zauguin wrote above: 
Of course \DeclareTextCommandDefault is still useful if the 
command is redefined to be empty.]

car222222:
One more point about the name: the documentation (from 2015) in utf8ienc.dtx suggests that 00AD might well (if it had then been named) have been treated like \nobreakspace : i.e., dropping the \text ... because these two, as documented, are “not really characters(sic)” — I think that the documenter really meant  “not really glyphs” :’-).

wspr:
Just trying to get back to the short term issue here. Currently utf8ienc.dtx contains:

    %<all,t1,ot1,ly1>\DeclareUnicodeCharacter{00AD}{\-}

After these discussions I actually wonder if that is the right approach — if I copy/paste from HTML into a TeX document I probably want to ignore them, and it seems unlikely that users will WANT to switch from `\-` to 00AD as a convenient way to manually mark hyphenation points.

But operating on the assumption that we shouldn't change things unless there are really good reasons, I think @zauguin's approach of making OOAD active and defined in terms of `\-` is currently the best option to take; it solves the current issue at hand.

Deciding on an active character interface that can be useful for other chars as well should definitely be left on the table but that can supersede this current work if/when it appears. There's no point holding up the current patch because of theoretical improvements that we may or may not be making down the line.

So I would say we have something like the following timeline:

1. Make 00AD active defined to `\-` (this pull request)
2. Design a perfect way of setting up active characters in general (babel, etc.)
3. Use the new approach to redefine our definitions of 00AD, perhaps then changing the behaviour from `\-` to <*ignore*>.

car222222:
@wspr wrote:

1.  Make 00AD active defined to \- (this pull request)
Actually this PR currently uses \textsofthyphen as the definition (which could be simply \softhyphen .
Also, there are severe problems with \-, described above by Marcel.
So not done like this and maybe not a good idea at this stage .

2.  Design a perfect way of setting up active characters in general (babel, etc.)
Dealt with elsewhere (see emails re classification etc.): first need to put them into (subsub)classes that will need at least two types of set-up (for robust and non-robust==input-shorthand types).
But there is (almost:-) only one character just like ‘softhyphen’, so we do not need anything general right now.

3.  Use the new approach to redefine our definitions of 00AD, perhaps then changing the behaviour from \- to <ignore>.
There is (and probably will not be) any such behaviour to change from.
Also, a method to ignore them is needed right now so cannot wait for any any possible knew approach’.
Note that this is probably best dine by redefining (text)softhyphen, see above under: Going back to the motivating . . .

wspr:
Whether its an amendment to this pull request or a new one I don’t mind. I’m not really in favour of \textsofthyphen though — I think that it would be confusing to redefine \textsofthyphen to {} to ignore incoming 00AD chars in the source. IMO \textsofthyphen would be what you might type if \- was redefined in a local environment but you still wanted the discretionary hyphen.

My point (perhaps made badly) was that at this stage I would favour no indirection at all and just hard code in the mapping from 00AD to \- to be consistent with pdfTeX. If we call it \textsofthyphen or \charsofthyphen or \softhypen now we are stuck with it, and I’d rather we make that decision when we are speaking more generally about how to handle active characters.

car222222:
OK, Will!  I understand your reasoning now, so thanks for the explanation.

Having read the Unicode docs about it, I now think that we should not perpetuate the misnaming of this character.  It is possible that there were good reasons for this name when it was first used but the semantics of this character have now changed.  

It really is just indicating a possible intra-word breakpoint and not saying anything about what glyphs may or may not be used to render it.

I also agree that fixing a unique LICR-style name may not be a good idea but I cannot see a way to assign synonyms!

car222222:
I agree that we should decide and act on 00AD quickly in some way that supports both somehow inserting a discretionary and inserting nothing (ignore it).

Marcel suggested that we also look now at the zero-width space characters as they are similar with respect to being in some sense ignorable since they do not necessarily require that anything is sent to the font (or to a text-shaper).

car222222:
One advantage of somehow associating a unique internal name with every active character of Class 1 or 2 (which, as Frank said) are very similar is that this supports them being written out ‘verbatim’ to a Unicode character steam.  I do not think that we can simply ignore this type of support, even if a use-case suggests that certain ‘internal characters‘ are ignored or replaced by some other character string when output to such a file.

zauguin:
Given that there is no agreement for any other model, I reverted this to the original form of expanding U+AD to `\-` for now until a more general solution is found. (Basically the suggestion of @wspr)
This will solve the most important issue in the short term, without restricting the possibilities of a future extension which will add more flexibility. While this does not easily support inserting nothing instead of a discretionary, users who need that either have to redefine the active U+AD or wait for the extended version in the future.

car222222:
@zauguin  I understand you going ahead with this but I wish to note under this issue the following points: 

1.  We have established the uniqueness of this Unicode code-point: nothing else has exactly the same properties.  Therefore it does not need to fit completely any ‘general solution’.

2.   It has been reported that these characters are sometimes introduced erroneously into input files.  Thus there seems to be a need to ignore them (at least locally) at some stage of formatting some documents.

3.  It may be essential to use this code-point within PDF text streams.  
The PDF Reference 1.7 states that:
The producer of a Tagged PDF document must distinguish explicitly between soft and hard hyphens so that the consumer does not have to guess which type a given character represents.

zauguin:
> @zauguin I understand you going ahead with this but I wish to note under this issue the following points:
> 
>   1. We have established the uniqueness of this Unicode code-point: nothing else has exactly the same properties.  Therefore it does not need to fit completely any ‘general solution’.

I agree, but I still think that seeing how a solution for other characters will look like will be important to decide how a potential indirection for U+00AD is handled. Even if not the same system is used, it could at least be somewhat similar.

> 
>   2. It has been reported that these characters are sometimes introduced erroneously into input files.  Thus there seems to be a need to ignore them (at least locally) at some stage of formatting some documents.

I agree, but I don't see a nice interface for this which doesn't restrict us later on. (Also we might want to set U+AD to catcode `ignore` in such cases, especially since the decision between "should be seen as `\-`" or "should be ignored" does depend only on the (source of the) input, not on the context where it is used. So in this case, it might actually be desirable to do this either through catcodes or at least in an expandable way.

But that's discuss that in a separate issue.

> 
>   3. It may be essential to use this code-point within PDF text streams.
>      The PDF Reference 1.7 states that:
>      The producer of a Tagged PDF document must distinguish explicitly between soft and hard hyphens so that the consumer does not have to guess which type a given character represents.

This happens to be one of my main motivations to get U+AD in the input sorted out: Currently all Unicode engines discard U+AD in the PDF stream because otherwise weird hyphens appear in the middle of words. This also means that they can not easily be used as recommended in the PDF Reference. If the input handling here is sorted out, `luaotfload` (and maybe at some later stage XeTeX) can be changed to keep U+AD in the stream, such that the relevant hyphenchar parameters can be changed to insert U+AD instead of a traditional hyphen. Then LuaTeX hyphenation would automatically introduce the right character. But for this, we first need to be sure that a U+AD glyph node isn't just an (accidental?) U+AD from the input.

wspr:
On 8 May 2020, at 6:01 pm, Marcel Krüger <notifications@github.com> wrote:
> 
> If the input handling here is sorted out, luaotfload (and maybe at some later stage XeTeX) can be changed to keep U+AD in the stream, such that the relevant hyphenchar parameters can be changed to insert U+AD instead of a traditional hyphen. Then LuaTeX hyphenation would automatically introduce the right character.

There’s a wrinkle to this I think. U+AD is not a hyphen-like glyph in any font I’ve seen, so while the internal (tagged?) representation of the PDF text would need to use U+AD, the actual typeset output still needs to have a regular hyphen (-) glyph.

Or is this essentially what you mean above, Marcel?

zauguin:
> > If the input handling here is sorted out, luaotfload (and maybe at some later stage XeTeX) can be changed to keep U+AD in the stream, such that the relevant hyphenchar parameters can be changed to insert U+AD instead of a traditional hyphen. Then LuaTeX hyphenation would automatically introduce the right character.
>
> There’s a wrinkle to this I think. U+AD is not a hyphen-like glyph in any font I’ve seen, so while the internal (tagged?) representation of the PDF text would need to use U+AD, the actual typeset output still needs to have a regular hyphen (-) glyph. Or is this essentially what you mean above, Marcel?

Do you have an example for such a font? In all fonts I looked at, U+AD looks identical to U+2D (aka. the hyphen):

```
\documentclass{article}
\usepackage{fontspec}
\newcommand\showAD[1]{%
  #1:&U+00AD: \fontspec[RawFeature=-invisible]{#1}\char"AD&U+002D: \fontspec[RawFeature=-invisible]{#1}\char"2D\\
}
\begin{document}
\begin{tabular}{lll}
\showAD{lmroman10-regular.otf}
\showAD{texgyrebonum-regular.otf}
\showAD{GFSBaskerville.otf}
\showAD{LibertinusSans-Regular.otf}
\showAD{GaramondLibre-Regular.otf}
\showAD{Roboto-Medium.otf}
\showAD{NotoSans-Regular.ttf}
\end{tabular}
\end{document}
```
image

wspr:
Oh! I clearly wasn’t looking hard enough. Thanks, this looks great (well in LuaTeX, but not XeTeX).

I don’t actually have fontforge installed at the moment and was going on the output of a GUI programme I have. I wonder if OpenType fonts that vary the style of the hyphen with +case are smart enough to do the same thing for U+AD… :)

khaledhosny:
Using the glyph for U+00AD at line break might not be a wise idea, since AFAIK no widely used application does this (they either use U+002D HYPHEN-MINUS or try U+2010 HYPHEN then U+002D), so the result might be untested (like having kerning or other OpenType interactions with the other hyphen glyphs and overlooking U+002D).

Also, AFAIK U+002D is really a soft hyphen. U+200B ZERO WIDTH SPACE is used for indicating optional line breaks where hyphenation is undesirable.

u-fischer:
> Also, AFAIK U+002D is really a soft hyphen.

No the pdf reference is quite clear here:

 
> In tagged PDF, the visible hyphen that is introduced through the incidental division of a word at the end of a line but which would not be present otherwise, may be represented as a soft hyphen, mapped to the Unicode value U+00AD, in one of the ways described in 14.8.2.6, "Unicode mapping in tagged PDF".
> NOTE 1 The soft hyphen character represented by the Unicode value U+00AD is distinct from an ordinary hard hyphen, whose Unicode value is U+002D.
> The writer of a tagged PDF document shall distinguish explicitly between soft and hard hyphens so that a PDF processor can unambiguously determine which type a given character represents.

wspr:
I share Khaled’s concern that many fonts *may* not have set up U+AD for actual typesetting in the same way as U+2D, although Marcel showed an example of it working in LuaTeX so I’m happy enough it can work in theory and we could use further LuaTeX tweaks to adjust kerning etc to match U+2D if a font was non-compliant.

In terms of the tagged PDF side of things, can’t the typesetting be done with U+2D but the underlying representation of the text use U+AD? Or would one simply omit the hyphenation entirely in the underlying text representation?

(I may have my terminology a bit garbled sorry.)

zauguin:
Basically there are three cases: (Of course U+2D could be replaced by U+2010 in the following)

  1. U+2D and U+AD behave basically the same in a font. Then it doesn't matter which is chosen for visual representation but U+AD interacts better with PDF tagging.
  2. The font designer thought that a hyphen inserted by hyphenation should look slightly different than a "strong hyphen", so U+AD looks different than U+2D on purpose. That might be a bit odd, but that's a design question so we should respect it.

  3. The font is weird and the U+AD character does not behave properly. Given the typical consistentency of Open/TrueType fonts, this might be a quite common case. Then it might be best to just use U+2D as the actual glyph, but we still want to tag it as U+AD to help the PDF viewer. So we still have to keep track that this hyphen represents a soft hyphen. Especially in LuaTeX, the easiest way of doing this is to still use U+AD and then replace the code shortly before shaping. (That also allows this to be a font instead of language level distinction) Then adding the right tagging to mark this as U+AD should be handled by the regular tagging code automatically. (It would be nice to avoid triggering this for well-behaving fonts though because it can introduce quite some overhead in the PDF file compared to just having a separate glyph)

khaledhosny:
> > Also, AFAIK U+002D is really a soft hyphen.
> 
> No the pdf reference is quite clear here:
> 
> > In tagged PDF, the visible hyphen that is introduced through the incidental division of a word at the end of a line but which would not be present otherwise, may be represented as a soft hyphen, mapped to the Unicode value U+00AD, in one of the ways described in 14.8.2.6, "Unicode mapping in tagged PDF".
> > NOTE 1 The soft hyphen character represented by the Unicode value U+00AD is distinct from an ordinary hard hyphen, whose Unicode value is U+002D.
> > The writer of a tagged PDF document shall distinguish explicitly between soft and hard hyphens so that a PDF processor can unambiguously determine which type a given character represents.

I don’t see how this contradicts what I said; the character that has to be copied from the PDF does not need to match the glyph being used. You can (and should) use the hyphen glyph, and embed U+00AD in the tagged text. Thus is exactly how the various glyphs for an Arabic character  are mapped back to the same Unicode code point.

khaledhosny:
From https://unicode.org/reports/tr14/#SoftHyphen

> Unlike U+2010 HYPHEN, which always has a visible rendition, the character U+00AD SOFT HYPHEN (SHY) is an invisible format character that merely indicates a preferred intraword line break position.

So it is plausible that fonts can just have a blank zero width glyph there, and it would still work basically everywhere else. The fact that most fonts don’t do this is probably related to the legacy of character sets, but I wouldn’t rely on this to be true forever or for all fonts.

car222222:
Thanks, @khaledhosny for the further information from unicode.org.

The PDF (1.7) manual contains several statements concerning soft/hard hyphens and it is not clear to me if these are just ‘not well-aligned’ with current Unicode practice, or even possibly contradictory?

PDF1.7 is  of course dated long before the more recent Unicode reports.  
I do not know whether PDF2.0 has anything that is better aligned with the current Unicode ideas.

car222222:
@wspr wrote: 
In terms of the tagged PDF side of things, can’t the typesetting be done with U+2D but the underlying representation of the text use U+AD? Or would one simply omit the hyphenation entirely in the underlying text representation?

Yes, it is possible to do all such things, whilst also (we hope) conforming to the both the requirements/guidelines of Unicode and (tagged/structured) PDF.  But some of these possibilities are not well,described as ‘simple’, either in the (LaTeX/lua) code needed to produce the conformant PDF or in the PDF ‘mark-up’that is needed. 

>  (I may have my terminology a bit garbled sorry.)
No worries! . . . much what is written in this area is anyway pretty vague in its usage.

u-fischer:
> I don’t see how this contradicts what I said;

I was contradicting the part that "002D is a soft hyphen" and the idea to use a zero width space. From the pdf terminology 002D is a hard hyphen, and to indicate a soft hyphen 00AD is needed. So we have either the option to use 00AD directly (if the glyph looks ok, which is does in many cases as @zauguin has shown above) or 002D (or some other glyph) with /ActualText 00AD.

car222222:
@u-fischer  But how practical is it to add “ /ActualText 00AD “ for every broken/hyphenated word?  Or to do this on a font-by-font basis, i.e., only for those fonts where it is needed?

@khaledhosny   I am not trying to be argumentative about this but it is important to note that the quote: 
“the character U+00AD SOFT HYPHEN (SHY) is an invisible format character that merely indicates a preferred intraword line break position” 
is not obviously an absolute/general statement.  
Rather it comes from a description of the Unicode Line-breaking Algorithm, which is an algorithm that is not used by tex engines (in their normal set-up).

khaledhosny:
> Thanks, @khaledhosny for the further information from unicode.org.
> 
> The PDF (1.7) manual contains several statements concerning soft/hard hyphens and it is not clear to me if these are just ‘not well-aligned’ with current Unicode practice, or even possibly contradictory?

PDF was never good for text extraction, it makes no clear distinction between characters and glyphs so I’d always take whatever it says in this area with a big grain of salt.

khaledhosny:
> > I don’t see how this contradicts what I said;
> 
> I was contradicting the part that "002D is a soft hyphen" and the idea to use a zero width space. From the pdf terminology 002D is a hard hyphen, and to indicate a soft hyphen 00AD is needed. So we have either the option to use 00AD directly (if the glyph looks ok, which is does in many cases as @zauguin has shown above) or 002D (or some other glyph) with /ActualText 00AD.

U+00AD is an invisible control character, and that is exactly the reason why you would want U+00AD the charter copied for an automatically inserted hyphen as it would make invisible unless line is broken at it.

U+00AD is always invisible, unless the line is broken at it and at this point a hyphen glyph is inserted. The question is where to take that hyphen glyph from. All application I know use a glyph other than that of U+00AD, so if TeX does otherwise it will be an outlier with potential breakage because virtually no fonts are tested against TeX engines (mine included).

car222222:
@khaledhosny 

TL;DR  ‘Since you can’t beat them, join them!’  
LaTeX is currently used pretty much only PDF, so we decided to fix the PDF it produces (and hopefully thus improve various aspects of PDF and irs documentation and consumption).

I agree with this:

>  PDF was never good for text extraction, 

But for current LaTeX, our major (and almost the sole) target output is PDF! 
Also (unfortunately?)  the LaTeX Team are now, for various reasons, fully committed to rectifying such problems in PDFs produced by LaTeX.  This will be done in cooperation with Adobe engineers, with whom we plan soon to have direct contact.

>  it makes no clear distinction between characters and glyphs

The reference does show how to do this, but it is not always very clear about the details.
Also, I think that many PDF consumers do not follow the specification.

One of our aims is to make LaTeX produce good ‘tagged and structured PDF’, which includes making it comply with the pdf reference standards and hence support smart cut-and-paste (assuming if course fully compliant consumer apps).

Adobe engineers are interested in this concept for several reasons, of which one is that LaTeX is already recognised by them as a major source of real-world PDF documents, and LaTeX is an obvious source for ‘structured’ PDFs.  Thus, when this project is complete there will be a widely available collection of ‘correctly encoded’ (including soft-hyphens) and ‘tagged/structured’ PDFs available that can, inter alia, be used in the education of producers of consumer apps to explain how to handle such ‘perfect’ PDFs.

>  so I’d always take whatever it says in this area with a big grain of salt.

You can do that, but we cannot.  However we can investigate and maybe fix the necessity of salt, probably by dining with Adobe:-).

khaledhosny:
OK.

PR #326, opened 2020-04-22
correct unicode value of \=y

u-fischer:
there was a small typo, \=Y and \=y pointed to the same value. 

## Status of pull request

- Ready to merge

- [ ] Test file(s) added
- [ x] Relevant `\changes` entries in source included
- [ ] `ltnewsX.tex` updated

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -1,9 +1,9 @@
-2020-04-22 Ulrike Fischer <Ulrike.Fischer@latex-project.org>
+2020-04-22  Ulrike Fischer <Ulrike.Fischer@latex-project.org>
 
 	* ltoutenc.dtx (tuenc.def):
 	corrected unicode value of \=y from "0232 to "0233
 
-2020-04-21 Frank Mittelbach <Frank.Mittelbach@latex-project.org>
+2020-04-21  Frank Mittelbach <Frank.Mittelbach@latex-project.org>
 
 	* ltspace.dtx (subsection{Horizontal space (and breaks)}):
 	Support calc syntax with \hspace (gh/152)
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -1,4 +1,9 @@
-2020-04-21  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
+2020-04-22 Ulrike Fischer<Ulrike.Fischer@latex-project.org>
+
+	* ltoutenc.dtx (tuenc.def):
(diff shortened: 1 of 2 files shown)


wspr:
Looks good to me. I would agree an entry in ltnews isn't needed, and I doubt there's much benefit to a test file here (?). So I would say good to go.

wspr:
Oh, Ulrike if you leave out the space in `[ x]` it will appear as a proper checkbox :)

base/changes.txt
```diff
@@ -1,4 +1,9 @@
-2020-04-21  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
+2020-04-22 Ulrike Fischer<Ulrike.Fischer@latex-project.org>
```
wspr, line 1: space before email address for consistency I guess

base/ltoutenc.dtx
```diff
@@ -3601,7 +3603,7 @@
 \xdef\@fontenc@load@list{\@fontenc@load@list
   \@elt{\csname opt@fontenc.sty\endcsname}}
 %    \end{macrocode}
-%    
+%
```
wspr, line 3606: thanks for including this -- we should purge all trailing whitespace! :)

u-fischer: Hm. I don't think that it was me, so either git or windows should get the praise.

base/changes.txt
FrankMittelbach: The 2 spaces in front of my name or David's name are automatically placed there by the workflow we use in emacs (pressing C-c c and getting most of the line autoinserted in the right changes file), so removing it for consistency is probably counterproductive as 99% of the changes (at the moment are done by us 2 :-).  So I would rather add a space in front of Ulrike's name if we care about consistency

wspr: Yep, that's kind of what I meant although I missed there were two spaces :)
Ulrike has now added a space before her address

FrankMittelbach: Well, she added one after her name but removed the second space(s) from the line with my name. Not that it matters to me, I was just explaining that whenever I generate such a line it will have those spaces in as they are auto-generated

u-fischer: @FrankMittelbach  yes, sorry, some "autocorrecting space stuff" from my editor  kicked in (that was also what remove the one trailing space later one), but I didn't really realize that it would remove spaces when I touch the line. I readded both now.

PR #339, opened 2020-05-23
DeclareFontEncoding: always set a default substitution

DamienRobert:
If a font encoding 'Foo' is declared but `\DeclareFontSubstitution` is
never set up on this encoding, then switching to this font encoding will
never update the substitution defaults.

For instance the default substitution shape may be 'n' if 'OT1' was the
encoding used before, or 'it' if 'OML' was the encoding used before.

Fix this by setting `\D@Foo` in `\DeclareFontEncoding{Foo}{}{}` too. This
uses the current values of the global `\default@family` and so on, which
at this point have been set by the last `\DeclareErrorFont` command.

In `fonttext.ltx` all encodings specified before `\DeclareErrorFont`
have a `\DeclareFontSubstitution`, so it does not matter that the global
`\default@family` was not yet specified.

# Internal housekeeping

## Status of pull request

- Feedback wanted

### DIFF
--- a/base/ltfssbas.dtx
+++ b/base/ltfssbas.dtx
@@ -495,6 +495,11 @@
 %   support cyrillic integration (pr/2988)}
 %    \begin{macrocode}
   \xdef\LastDeclaredEncoding{#1}%
+  \expandafter\edef\csname D@#1\endcsname{%
+       \def\noexpand\default@family{\default@family}%
+       \def\noexpand\default@series{\default@series}%
+       \def\noexpand\default@shape {\default@shape}%
+       }%
   }
 \@onlypreamble\DeclareFontEncoding@
 %    \end{macrocode}


FrankMittelbach:
Damien

> This is what I call a |vanity bug|, ie an easy and not critical bug to 
> fix, so a nice opportunity to contribute to latex development. I have 
> several such |vanity bug| fixes in |git| itself for instance :)

oh I do appreciate that. I was just saying there is (for the better or 
worse much more to even such a simple bug

> I indeed saw the checklist when doing my pull request, but I am 
> motivated to update my pull request to fulfill them all (when I have 
> time), and in particular to add tests (this is a nice occasion to learn 
> about the new lua testing framework of latex).

well I would appreciate that, but that checklist was more written for us 
internally, we don't quite expect others to be able to do that (which is 
one of the reasons why pull requests usually don't work). E.g. writing a 
test file that actually integrates into the test system is non-trivial 
and doing the rollback stuff is a headache (even for us)

if you like to give it a try, feel free, and I take it from there.
> 
> I know that as a maintainer you don't have much time to waste on vanity 
> bug fixes, so if you prefer that I just send a minimal test file, I'll 
> understand.

alternatively, just giving me a test file that I massage to go into the 
test system would also be fine.

FrankMittelbach:
Looking at this again I wonder if it should be done or whether it is better to document (if not already there) that there should be a substitution declared when a new encoding is declared. My point is that you are not much better off when you define a default substitution because the default family may not exist in that encoding (quite likely actually) and in that case you just end up by a different route at ErrorFont don't you?

josephwright:
After discussion, we feel that the explicit declaration is the best approach here.

PR #350, opened 2020-07-01
Document relocation of textcomp macros into kernel

dr-scsi:
Dear all,

this should be the final PR for `textcomp` macros now being part of the kernel.  I also added the available macros as I'm not aware of them being described in any other free manual.  Again, any comments welcome and thanks for your patience.


FrankMittelbach:
looks good to me, thanks. Do you mind lifting you alias dr-scsi and tell me your name for an acknolegement in the source? Or am I supposed to recognise it? If so mea culpa, no bell rings.

PR #358, opened 2020-07-28
make a number of commands short again that got \long by mistake

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [ ] Test file(s) added (not needed I think)
- [ ] `ltnewsX.tex` updated (n/a)

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -1,3 +1,4 @@
+
 ================================================================================
 This file lists changes to the LaTeX2e files in reverse chronological order of
 publication (therefore the dates might be out of sequence if there are hotfixes).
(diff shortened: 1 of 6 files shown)


PhelypeOleinik:
I think perhaps `\begin` could also be fixed (even if overwritten later) in `ltmiscen.dtx`...

Also there's a blank line added at the top of `changes.txt` (probably no harm either way).

Other than these, it looks good to me.

FrankMittelbach:
> I think perhaps `\begin` could also be fixed (even if overwritten later) in `ltmiscen.dtx`...
> 
> Also there's a blank line added at the top of `changes.txt` (probably no harm either way).
> 
> Other than these, it looks good to me.

once you are done with your changes my intention is to move most of lthooks and friends into the right places so only lthooks will remain bit the being code and other stuff migrates. This is why I didn't touch \begin, right now

PR #361, opened 2020-07-31
Switch from Umathchar to Umathcode for Unicode detection

### LINKED ISSUE #279: Both \Umathcode and \Umathchar are used to determine the unicode-capability
The ltnews31 says that, 
 - engines which are fully Unicode-capable must provide three primitives: `\Uchar`, `\Ucharcat`, and `\Umathcode`
 - it has become standard practice to check for Unicode-aware engines by using the existence of the `\Umathcode` primitive.

But in latex2e kernel, both `\Umathcode` and `\Umathchar` are used to determine the unicode-capability, and `\Umathchar` is even not one of the three required primitives.

Is this on purpose?

zauguin:
Fixes #279. Given that this is only for internal consistency and should never lead to any change in behavior and also doesn't change the documentation, I had the feeling that `\changes` lines would not provide any benefit and just distract from more useful changes. I'm not sure how we handle such cases though, so I can add them if you feel that they are useful.

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- n/a Test file(s) added
- [ ] Relevant `\changes` entries in source included -- see above
- n/a `ltnewsX.tex` updated

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -7,6 +7,11 @@ completeness or accuracy and it contains some references to files that
 are not part of the distribution.
 ================================================================================
 
+2020-08-01  Marcel Krüger     <Marcel.Krueger@latex-project.org>
+
+	* fontdef.dtx, inputenc.dtx, ltfinal.dtx, ltluatex.dtx:
+	More consistently use \Umathcode to check for Unicode-aware engines. (gh/279)
+
 2020-07-27  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
 
 	* ltpage.dtx (section| command.):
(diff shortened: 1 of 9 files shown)


FrankMittelbach:
Am 01.08.20 um 01:01 schrieb Marcel Krüger:
> Fixes #279 <https://github.com/latex3/latex2e/issues/279>. Given that 
> this is only for internal consistency and should never lead to any 
> change in behavior and also doesn't change the documentation, I had the 
> feeling that |\changes| lines would not provide any benefit and just 
> distract from more useful changes. I'm not sure how we handle such cases 
> though, so I can add them if you feel that they are useful.

I agree that adding changes entrys in the sources looks like overkill 
(that would be similar to putting in changes entries if we fix a typo.

But there should perhaps be one-liner in changes.txt saying that this 
got normalized (you may have done that, not checked)

FrankMittelbach:
>  But there should perhaps be one-liner in changes.txt saying that this got normalized (you may have done that, not checked)

ah you said so 3 lines later ...

Issue #352, opened 2020-07-04
[Localization] Add `japanese` Option to varioref.sty


cmplstofB:
## Brief outline of the enhancement
Hello.
Could you add a new language option `japanese` to varioref.sty?
the following code will make it works.
Because of grammatical reason, \vrefrange command has \reftextlabelrange and \vpagerefrange in a reversed order.

```latex
\DeclareOption{japanese}
  {\vref@addto\extrasjapanese{%
    \def\reftextfaceafter {\reftextvario{見開き}{次}頁}%
    \def\reftextfacebefore{\reftextvario{見開き}{前}頁}%
    \def\reftextafter     {\reftextvario{直後の}{次}頁}%
    \def\reftextbefore    {\reftextvario{直前の}{前}頁}%
    \def\reftextcurrent   {\reftextvario{この}{現}頁}%
    \def\reftextfaraway#1{\pageref{#1}頁}%
    \def\reftextpagerange#1#2{\pageref{#1}頁から\pageref{#2}頁}%
    \def\reftextlabelrange#1#2{\ref{#1}から\ref{#2}}%
  }%
  \AtEndOfPackage{%
  \renewcommand\vrefrange[3][\reftextcurrent]{%
  \vpagerefrange[{#1}]{#2}{#3}の\reftextlabelrange{#2}{#3}%
  }}}
```

FrankMittelbach:
don't really think it can be done this way. If I'm not mistaken that this would disallow using more than one language as the different order in \vrefrange would be permanent. So have to think about this a bit, but yes glad to add a Japanese language option.

FrankMittelbach:
@aminophen can I trouble you to have a look over the the fix for #352 and see if this correctly works including when mixing Japanese with, say English, thanks.  If you prefer I can add it into the dev format first, then you see it there via the CTAN route.

aminophen:
@FrankMittelbach Thanks; I will ask some more users in Japan if they have any comments on the current implementation and/or translations.

FrankMittelbach:
good, thanks. It is now on its way to CTAN in the dev release so in a day or two it should be usable via uplatex-dev

FrankMittelbach:
It should be now available if the dev format is used, e.g., `pdflatex-dev file`

cmplstofB:
@FrankMittelbach Hello. I'm sorry for the late response.

[...]

PR #369, opened 2020-08-11
[Localization] Modify Japanese localization

cmplstofB:
Fix for https://github.com/latex3/latex2e/issues/352#issuecomment-663046365


FrankMittelbach:
applied differently

PR #370, opened 2020-08-12
Fixcontentsline

u-fischer:
# Internal housekeeping

## Status of pull request

This changes addcontentsline so that it always writes 4 argument, to improve compability with hyperref.

- Feedback wanted 

- [ x] Test file(s) added (old testfiles adapted)
- [ x] Relevant `\changes` entries in source included
- [ x] `ltnewsX.tex` updated


JohannesBraams:
I would at least need to look at the NTG document classes...

u-fischer:
@FrankMittelbach  well it is marked as "feedback wanted" and naturally we can move it to later. 
But I don't think that the change can really break something, it is only dangerous if someone redefined contentsline so that it looks behind three arguments and gets confused by the additional brace group. 

On the other side there are a number of packages and classes which manipulate contentsline and fight (or leave hyperref) fight with the different arguments numbers. E.g. titlesec contains

    \def\ttl@gobblecontents#1#2#3{\ignorespaces}%

and hyperref contains

```
\@ifpackageloaded{titletoc}{%
  \def\ttl@gobblecontents#1#2#3#4{\ignorespaces}%
}{}
```

It will take quite some time anyway to clean up hyperref of such patches, but it would be easier to communicate if I wouldn't have to write "define contentsline now so that it takes three argument without hyperref and four with,  but take care that it also works if we decide to change it to four argument in a future latex".

PR #372, opened 2020-08-16
Picture lengths

davidcarlisle:
Extension to picture mode to allow eTeX length expressions in picture mode coordinate. Non-length numbers are accepted as before, being a multiple of `\unitlength`. 

Similar extensions have been provided previously by Heiko Oberdiek's `picture` package but the implementation here is different (just using etex expression syntax, and directly integrated into the source rather than by applying patch commands).

Typical use:

```
\put(.5\textwidth,.5\textheight){...}
```


base/doc/ltnews32.tex
```diff
@@ -297,6 +298,20 @@ \subsection{Support \pkg{calc} in various kernel commands}
 %
 \githubissue{152}
 
+\subsection{Support \eTeX\ length expressions in \texttt{picture} coordinates}
+
+Picture mode coordinates specified with \texttt{(\_,\_)} previously
+accepted multiples of \cs{unitlength}. They now accept \eTeX\ length
```
FrankMittelbach, line 304: now also allow

base/ltboxes.dtx
```diff
@@ -376,15 +381,38 @@
...
+%<latexrelease>        \let\mb@l\relax\let\mb@r\relax
+%<latexrelease>      \else
+%<latexrelease>        \expandafter\let\csname mb@\reserved@a\endcsname\relax
+%<latexrelease>      \fi}%
+%<latexrelease>    \mb@t
+%<latexrelease>    \hb@xt@ #1\unitlength{\mb@l #4\mb@r}%
+%<latexrelease>    \mb@b
+%<latexrelease>    \kern\z@}}
+%<latexrelease>\EndIncludeInRelease
+%<*2ekernel>
```
FrankMittelbach, line 410: I would keep the comment at the place where it belongs, i.e. 3 line above

base/ltpictur.dtx
```diff
@@ -237,34 +293,98 @@
...
+%<latexrelease>\IncludeInRelease{2020/10/01}%
+%<latexrelease>                 {\@multiput}{default units}%
 \long\def\@multiput(#1,#2)#3#4{%
   \@killglue\@multicnt #3\relax
   \@whilenum \@multicnt >\z@\do
     {\raise\@ydim\hb@xt@\z@{\kern\@xdim #4\hss}%
      \advance\@multicnt\m@ne
-     \advance\@xdim#1\unitlength\advance\@ydim#2\unitlength}%
+     \@defaultunitsset{\advance\@xdim}{#1}\unitlength
+     \@defaultunitsset{\advance\@ydim}{#2}\unitlength}%
```
FrankMittelbach, line 370: what that usage documented in the definition?

base/testfiles/tlb-picture-01.lvt
```diff
@@ -0,0 +1,42 @@
...
+\START
+
+\begin{picture}(200,4\zzz)
+
+\loop
+\addtolength\zzzb{40pt}%
+\put(\zzzb+30pt,\zzzb+40pt){\line(1,0){150pt}}
+\put(\zzzb+30pt,\zzzb+40pt){\line(0,1){60}}
+\put(\zzzb+30pt,\zzzb+\zzz){\line(1,0){20}}
+\put(\zzzb+50pt,\zzzb+\zzz){\line(1,-4){10}}
```
FrankMittelbach, line 24: there should also be a test file testing the version with plain numbers and posibly with mixtures

davidcarlisle:
> 
> 
> I wonder what this does to pict2e package --- have you tested?

One reason for making this a public PR was so that I could reference it in a PR to pict2e. (As we didn't have  a public issue for it, just internal discussion)

The current situation is that pict2e isn't broken but it removes the features, just accepting numbers.
The same  use of `\@defaultunitsset` would extend pict2e in the same way.

base/testfiles/tlb-picture-01.lvt
davidcarlisle: > 
> 
> there should also be a test file testing the version with plain numbers and posibly with mixtures

Yes although there were already half a dozen picture environments in tests, plus more in the documentation using the existing syntax so those tests not breaking already covers the existing syntax to some extent.

davidcarlisle:
Matching PR at pict2e

https://github.com/rolfn/pict2e/pull/6

base/ltpictur.dtx
davidcarlisle: I intended to document it but haven't yet it seems. The alternative is to document that only a register is used as `#1` and do this in two steps, default units to `\@tempdimc` then add that to `\@xdim`  but for an internal call this seems neater.

base/doc/ltnews32.tex
davidcarlisle: done

base/ltpictur.dtx
davidcarlisle: documented use of `\advance`

base/ltboxes.dtx
davidcarlisle: I moved it to above the matching kern in the current code not the latexrelease block.

PR #374, opened 2020-08-20
\ShowCommand

PhelypeOleinik:
# Internal housekeeping

The implementation for `\ShowCommand` is done.  What I'd like feedback on is the formatting of the output.  The behaviour is more or less the same as doing `\show` in all the internals of a command, except for a few small changes to add clarity (I hope), and TeX's prompt only appears when the command is `\show`n completely (that is, I only use `\show` or `\showtokens` in the last macro). Currently

 - For a command defined with `\DeclareRobustCommand` (and without an optional argument), like `\frac`, it shows:

    ```latex
    > \frac=robust macro:
    ->\protect \frac  .
    
    > \frac =\long macro:
    #1#2->{\begingroup #1\endgroup \over #2}.
    ```
   which looks the same as doing `\show\frac` and `\show\frac␣`, except that for the first one it says `robust macro` rather than just `macro`.  

 - For a command defined with `\newcommand` and an optional argument, like `\renewcommand\foo[2][***]{\bar{#1}{#2}}`, it shows:

    ```latex
    > \foo=robust macro:
    ->\@protected@testopt \foo \\foo {***}.
    
    > \\foo=\long macro:
    > default #1=***.
    [#1]#2->\bar {#1}{#2}.
    ```
   which also looks the same as doing `\show` twice, but for the first one it also says `robust`, and the second one shows `> default #1=***.`

- For a command defined with `\DeclareRobustCommand` _and_ an optional argument, the output looks like three `\show`s.

Is that format good? Another option, more compact, would be to only show the actual definition and omit `\foo␣` and `\\foo`, like:
```
> \foo=robust \long macro:
> default #1=***.
[#1]#2->\bar {#1}{#2}.
```

It looks better, but is a “half truth”.

Any other option?

## Status of pull request

- Feedback wanted 

- [x] Test file(s) added
- [x] `ltnewsX.tex` updated


car222222:
@PhelypeOleinik wrote: 
Is that format good? Another option, more compact

I think that the longer form looks pretty good, that much detail is essential for anyone who really needs to do a \\ShowCommand.

PhelypeOleinik:
> But I agree that complicated coding just to get rid of that may not be that useful. On the other hand we might want to consider a general mechanism in expl3 to mimic the interaction loop which could then be applied in all such cases by calling it --- but that is a separate issue for a different time.

Just for the fun of it, I started [`tex-interaction`](https://github.com/latex3/latex3/tree/tex-interaction) over at `latex3`.  All the actions available at TeX's `?` prompt should be available with `\sys_interact:`, though some don't yet work precisely like TeX does (formatting may be slightly off, `I` may trip on catcode changes, `E` may not report the correct file name and will not open an editor and I'm not sure all the `\errhelp` printed by `H` are correct yet). Comments welcome

We can probably use `\sys_interact:` here to get almost the same behaviour of `\show` for `\ShowCommand`, but I'd rather wait until that implementation is properly finished and tested.  Meanwhile I think we can leave it here as it is, or implement a light parsing of Frank's suggestion for `X` (probably very few people use anything other than `X` or `<return>` at the prompt; maybe `I`).  Suggestions?

FrankMittelbach:
the obvious merge issues afterI merged lthooks2 back into develop, but once that's fixed I guess this could be merged too, right?

PhelypeOleinik:
@FrankMittelbach Right. I'll sort out the merge later today. I'll have a PR for `\MakeRobust` branching from this one

PR #375, opened 2020-08-21
Lthooks2

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

there is still work on documentation needed and some commands with 2e names should get changed to internal ones with expl3 names but both can be done as a next step

- [x] Test file(s) added
- [ ] `ltnewsX.tex` updated - not applicable


u-fischer (approved):

FrankMittelbach:
actually it is not yet good as ltfilehook is missing a lot of the necessary includeinrelease stuff :-( currently working through that.

car222222:
> Frank Mittelbach wrote (approx.:-):
> 271f0e9 don't use \str_use:c on a tl :-)

The last time I looked there was either \tl_use:c or maybe \use:c (or even both) which maybe could be 'used' here?

But maybe they do not get the expansion requirements here quite right!

Chris

FrankMittelbach:
Am 25.08.20 um 08:45 schrieb Chris Rowley:
>     Frank Mittelbach wrote (approx.:-):
>     271f0e9
>     <https://github.com/latex3/latex2e/commit/271f0e9af3114072ec98282f2a20fcc1d5982866>
>     don't use \str_use:c on a tl :-)
> 
> The last time I looked there was either \tl_use:c or maybe \use:c (or 
> even both) which maybe could be 'used' here?
> 
> But maybe they do not get the expansion requirements here quite right!

\str_usec = \tl_use:c if it is used on "non-strings" that is the issue. 
As a result the tl is expanded and not just its top-level(which would be 
the case if it is a string. For performance reasons \str_use:N doesn't 
do a detokenize it assumes it only gets a str variable as its argument.

But the problem is by now resolved through correct coding.

Issue #355, opened 2020-07-07
Extensions not appended to graphics when file exists


joel-coffman:
## Brief outline of the bug

Not sure if this issue is a bug or simply inconsistent behavior due to kernel changes related to the handling of file names. When including an image and the file name contains a dot (`.`), known extensions (e.g., .png) are automatically appended *unless* a file with the "base" name exists when using LaTeX2e <2020-02-02> patch level 2. For example, including the image `latex-project-logo.svg.png` fails when the file `latex-project-logo.svg` exists. In contrast, prior versions compile successfully when both files exist.

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{article}

% Any preamble code goes here
\usepackage{graphicx}


\begin{document}

  % Demonstration of issue here
  \includegraphics{latex-project-logo.svg}
  
\end{document}
```

Compilation fails with the following error message:

> LaTeX Error: Unknown graphics extension: .svg.

To be clear, the directory contains the following files:
* latex-project-logo.svg
* latex-project-logo.svg.png
* mwe.tex

Due to kernel changes related to the handling of file names, a slight change is needed to compile this document for older kernels:

~~~ patch
-\includegraphics{latex-project-logo.svg}
+\includegraphics{{{latex-project-logo.svg}}}
~~~

With this change, the same document compiles on the following systems (pdfTeX and LaTeX versions copied from output):
* Ubuntu 18.04: pdfTeX, Version 3.14159265-2.6-1.40.18 (TeX Live 2017/Debian); LaTeX2e <2017-04-15>
* Ubuntu 19.10: pdfTeX, Version 3.14159265-2.6-1.40.20 (TeX Live 2019/Debian); LaTeX2e <2018-12-01>
* Debian 10: pdfTeX, Version 3.14159265-2.6-1.40.19 (TeX Live 2019/dev/Debian); LaTeX2e <2018-12-01>

## Log file (required) and possibly PDF file  

[mwe.log](https://github.com/latex3/latex2e/files/4885622/mwe.log)

[latex-project-logo.svg.gz](https://github.com/latex3/latex2e/files/4885645/latex-project-logo.svg.gz) (compressed due to GitHub's restrictions on types of attachments)
![latex-project-logo.svg.png](https://user-images.githubusercontent.com/8376231/86801974-cb99b480-c031-11ea-987b-42d85893a965.png)

joel-coffman:
[...]

PR #383, opened 2020-08-30
Extension appended to graphics when file exists (#355)

PhelypeOleinik:
This PR fixes issue #355.

Previously, if the files
```
latex-project-logo.svg
latex-project-logo.svg.eps
```
existed, and the user did `\includegraphics{latex-project-logo.svg}` expecting `graphics` to automatically detect and add the `.eps` extension, it would instead raise the error
```
! LaTeX Error: Unknown graphics extension: .svg.
```
if running with PDF output, or
```
! LaTeX Error: Cannot determine size of graphic in latex-project-logo.svg (no BoundingBox).
```
with DVI output, after trying to read the file as `.eps`.

The changes in this PR check if the extension (or any part of the file name after the last `.`) is a known graphics extension, and if it is not, then it tries appending one of the known extensions before falling back to `.eps` or raising the `unknown extension` error.

The old code would do (pseudocode):
```
\if (file has no extension)
  <try adding an extension>
\fi
```
and with the change in this PR it does:
```
\if (file has no extension) or (if the extension is unknown)
  <try adding an extension>
\fi
```
so the change should not affect existing (working) documents.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [x] `ltnews32.tex` updated


car222222:
Many thanks for this, especially the very clear explanation (including pseudo-code!).

FrankMittelbach:
looks good to me.

davidcarlisle (approved):
Haven't as much time as I'd like for full testing but this seems OK and shouldn't break any non error documents as far as I can see.

PR #396, opened 2020-09-30
Lua callback for shipout box preparation

zauguin:
A Lua callback which can be used for luacolor and similar code to apply changes indicated using Attributes directly before shipout, making sure that even content added in hooks is picked up.

Uses ltluatex's callback functionality instead of a TeX hook in order to be usable from Lua code. Currently no sorting is supported (well luatexbase adds a priority system, but that's rarely usable in practice) but this coul be added more generally to the callback system.

Lua code can use this with `luatexbase.add_to_callback('pre_shipout_filter', ...)`, e.g.

```
luatexbase.add_to_callback('pre_shipout_filter', function(head)
  -- Do something with the list node passed in head
  print("Shipping out list node ", head)

  return true
end, 'my_pre_shipout_code')
```

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [ ] Test file(s) added
- [x] `ltnewsX.tex` updated

### DIFF
--- a/base/ltshipout.dtx
+++ b/base/ltshipout.dtx
@@ -364,7 +364,7 @@
 %
 %
 %
-% \subsection{Provided Lua\Tex\ callbacks}
+% \subsection{Provided Lua\TeX\ callbacks}
 % 
 %  \begin{variable}{pre_shipout_filter}
 %    Under Lua\TeX, additionally the \texttt{pre_shipout_filter} callback is
--- a/base/ltshipout.dtx
+++ b/base/ltshipout.dtx
@@ -363,6 +363,20 @@
 %    names as given below.
 %
 %
+%
+% \subsection{Provided Lua\Tex\ callbacks}
+% 
+%  \begin{variable}{pre_shipout_filter}
+%    Under Lua\TeX, additionally the \texttt{pre_shipout_filter} callback is
+%    provided. The signature is
+%    \begin{verbatim}
+%     function(<node> head)
+%       return true | false | <node> newhead
+%     end
+%    \end{verbatim}
+%    This callback is invoked after all \TeX-level hooks described above.
+%  \end{variable}
+%
 % \subsection{Information counters}
 %
 %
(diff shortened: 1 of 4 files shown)


base/ltshipout.dtx
```diff
@@ -279,6 +279,20 @@
 %    names as given below.
 %
 %
+%
+% \subsection{Provided Lua\Tex\ callbacks}
```
FrankMittelbach, line 283: \TeX

base/ltshipout.dtx
```diff
@@ -542,6 +556,33 @@
...
+          if~not (result == true) then~
+            setbox(\the \l_shipout_box, result~or~nil)
+          end~
+        end
+      }
+    }
+    \protected \luadef \@@_finalize_box: \the \@@_finalize_box:
+  } {
+    \cs_set_eq:NN \@@_finalize_box: \scan_stop:
+  }
```
FrankMittelbach, line 581: As a one-off I thin this would be fine, but if we expect other such callback to come up we might want to think about how to encapsulate that kind of setup differently (just a thought)

base/ltshipout.dtx
zauguin: Fixed

u-fischer:
With the new callback the following works fine. What is the best way to test if the callback is there? 

~~~~
\documentclass{book}
\AddToHook{shipout/background}{\put(7cm,-10cm){\color{blue}\rule{1cm}{2cm}back}}
\AddToHook{shipout/foreground}{\put(9cm,-10cm){\color{yellow}\rule{1cm}{2cm}fore}}
\usepackage{luacolor}
%disable standard luacolor code
\RemoveFromHook{shipout/before}[luacolor]
%add new callback: How to test the callback?
\directlua{%
 luatexbase.add_to_callback('pre_shipout_filter',
 function(head) oberdiek.luacolor.process('ShipoutBox') return true  end,"luacolor")}%

\AddToHook{shipout/background}{\put(8cm,-12cm){\color{red}\rule{1cm}{2cm}back}}
\AddToHook{shipout/foreground}{\put(10cm,-12cm){\color{green}\rule{1cm}{2cm}fore}}
\begin{document}

\color{red} abc

\end{document}
~~~~

hvoss49:
> Am 03.10.2020 um 20:15 schrieb Ulrike Fischer <notifications@github.com>:
> 
> 
> With the new callback the following works fine. What is the best way to test if the callback is there?
> 
> 

if callback.find(<name>) then … else … end

u-fischer:
@hvoss49  that doesn't work with this callback. I already tries it, it is not in the callback.list table.

davidcarlisle:
should be in luatexbase.callbacktypes I think: that table is exposed and
added to when you create a callback

zauguin:
> With the new callback the following works fine. What is the best way to test if the callback is there?

As David said:
```
if luatexbase.callbacktypes.pre_shipout_filter then ... end
```

base/ltshipout.dtx
zauguin: I agree.

At this point I didn't copy the pseudo-primitive infrastructure from expl3, but that would already handle the `\newluafunction`, `everyjob` construction, `get_functions_table` and `luadef` parts. (It would also probably reuires  this to go to some Lua file or making some interfaces public.

This leaves the `create_callback` (which can't really be simplified, especially since other uses might need default functions here) and the actual `call(...)` / result handling which are quite specific to the individual usecase.

So I guess we could just adopt the expl3 model if necessary.

PR #436, opened 2020-11-25
Integrate xparse as ltcmd

josephwright:
# Internal housekeeping

## Status of pull request

- Under development

- [ ] Test file(s) added
- [ ] `ltnewsX.tex` updated


josephwright:
If we want the full history of `xparse`, I could do that by adding a temporary file, then renaming/'massaging' to `ltcmd`: thoughts?

josephwright:
I'm going to re-do the code here: something is not right as this is failing.

Issue #422, opened 2020-10-31
LaTeX2e 2020-10-01: when loaded by relative path, pkg options are ignored


muzimuzhi:
## Brief outline of the bug

For example, when package `mypkg.sty` located in `./style/mypkg.sty` is loaded by `\usepackage[key1]{style/mypkg}`,
 - until LaTeX2e 2020-02-02, package option `key1` is processed, but
 - from LaTeX2e 2020-10-01, package option `key1` is never processed.

I know that loading a package by using relative path is a bad practice, but some thesis templates (for example, [mohuangrui/ucasthesis](https://github.com/mohuangrui/ucasthesis)) use this trick to make the use of template portable.

## Minimal example showing the bug

Directory structure:
```tex
.
├── sty-option-test.tex
└── style
    └── mypkg.sty
```

```latex
% \RequirePackage{latexbug} % third-party file "mypkg.sty" required
\documentclass{article}
\usepackage[key1]{style/mypkg}

\begin{document}
content
\end{document}
```

```latex
% this is file "mypkg.sty" located in sub-dir "./style"
\ProvidesPackage{mypkg}

\DeclareOption{key1}{\PackageError{mypkg}{error text}{}}
\ProcessOptions\relax
```

## Log file (required) and possibly PDF file  

The following log files are produced with applying
```latex
\tracingall
\usepackage[key1]{style/mypkg}
\fail
```
[sty-option-test-20201001.log](https://github.com/latex3/latex2e/files/5469266/sty-option-test-20201001.log), with LaTeX2e <2020-10-01> patch level 2
[sty-option-test-20200202.log](https://github.com/latex3/latex2e/files/5469267/sty-option-test-20200202.log), with LaTeX2e <2020-02-02> patch level 5, the one currently provided by overleaf

muzimuzhi:
Perhaps the template author(s) should use something like
```tex
\makeatletter
\def\input@path{{./style}}
\makeatother

\usepackage[key1]{mypkg}
```

PhelypeOleinik:
Thanks!  I have to check carefully, but I'm almost certain this is caused by the same as #421 (thus the same temporary patch should work), and apparently came to life when I fixed #414.

muzimuzhi:
> Perhaps the template author(s) should use something like
> 
> ```
> \makeatletter
> \def\input@path{{./style}}
> \makeatother
> 
> \usepackage[key1]{mypkg}
> ```

[...]

Issue #426, opened 2020-11-19
Warning on provided package should ignore `./` and raise no error about the load sequence `before \documentclass`


SieDa:
It should be the same if you load a class or style of a  package with  
`\documentclass{./pkgDir/Class} ` or `\documentclass{pkgDir/Class} `
`\usepackage{./pkgDir/Style}` or `\usepackage{./pkgDir/Style}`, 
which you need in the preambel to load `Class.cls` and `Style.sty` from` pkgDir/`.

In the class or style file you have to determin for one or the other in \ProvidesClass{...}.
 
`./` should be ignored to raise no warning if you specify the short form in the class or style without `./` . 
1. LaTeX Warning: You have requested package `./pkgDir/Style',
               but the package provides `pkgDir/Style'.

This is confusing, as it is loaded and means the same. I recognized this whily analyzing the issue related to it, which is not yet fixed: 

2. `! Package xkeyval Error: xkeyval loaded before \documentclass.`

I store my custom package file in a subdirectory `/pkgDir/` by handling that as a git submodule in different projects in different versions. 
By that I don't care if the project needs an update, due to package changes, till I update my project ressources to fit to the latest package changes or just run a different branch for special purposes. 
The second error message raises because of the `/pkgDir/` usage. 

It looks like the commit 
[commit/978bf8f68](https://github.com/latex3/latex2e/commit/978bf8f6823a6119e84f7dc0c0831f24eb45b8b3) does not fix that. 

I run that in a new installation of TexLive in Win10 and there is no minimum example, as all the rest is ignored after the first lines of code to load my packages - it looks like it just happend during the last weeks, as I work that way since several years.

PhelypeOleinik:
Thanks for the report.  This is partly a duplicate of #422, to which I am working on a fix as we speak.  We expect to make a patch release to fix that issue some time next week, so watch out for that.

Now addressing the different points in your report:

> It should be the same if you load a class or style of a package with
> `\documentclass{./pkgDir/Class} ` or `\documentclass{pkgDir/Class} `

Yes, and so should `\documentclass{./pkgDir/../pkgDir///.././pkgDir/Class}`, but that would lead to a rabbit hole of parsing, far beyond what's considered fun :-)

> `./` should be ignored to raise no warning if you specify the short form in the class or style without `./` .
> 
>     1. LaTeX Warning: You have requested package `./pkgDir/Style', but the package provides `pkgDir/Style'.

[...]

Issue #421, opened 2020-10-31
xkeyval package raises error after recent update of ltxbase


kia999:
## Brief outline of the bug
The xkeyval package which erroneously detects that it is called before `\documentclass` in case if a custom document class is loaded from a subfolder. This error appears after recent update of `miktex` made on 28.10.2020. Short investigation has shown that changed files in miktex bundle are `ltxbase` and `unicode-data`. It was also noted that \ProcessOptions command does not see options in case if custom package is loaded from a subfolder.

## Minimal example showing the bug

```
%% !Mode:: "TeX:RU:UTF-8"
%%
%% this is patch proposed in
%%https://tex.stackexchange.com/questions/568887/xkeyval-package-raises-error-after-recent-update-of-ltxbase?noredirect=1#comment1434051_568887
%%
%\makeatletter
%\def\tmpa{2020-10-01}%
%\ifnum0\ifx\tmpa\fmtversion 1\fi\ifnum\patch@level=2 1\fi=11
%  \def\@pass@ptions#1#2#3{%
%    \@expl@@@filehook@set@curr@file@@nNN
%      {\@expl@@@filehook@resolve@file@subst@@w #3.#1\@nil}%
%        \reserved@a\reserved@b
%    \expandafter\xdef\csname opt@\reserved@a\endcsname{%
%      \@ifundefined{opt@\reserved@a}\@empty
%        {\csname opt@\reserved@a\endcsname,}%
%      \zap@space#2 \@empty}%
%    \ifx\reserved@b\@empty\else
%      \expandafter\let\csname opt@\reserved@b/\reserved@a\expandafter
%        \endcsname\csname opt@\reserved@a\endcsname
%    \fi}
%\else
%  \PackageError{patch}{Invalid patch}{}%
%\fi
%\makeatother

\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{Common/testBook} % this results in error when calling xkeyval
%\documentclass{testBook}       % this works fine

\usepackage{xkeyval}
\begin{document}
    Something to say\ldots
\end{document}
[test-xkeyval.log](https://github.com/latex3/latex2e/files/5468491/test-xkeyval.log)
[test-xkeyval.pdf](https://github.com/latex3/latex2e/files/5468496/test-xkeyval.pdf)

```

PR #438, opened 2020-11-25
Hotfix/gh422

PhelypeOleinik:
This PR changes how LaTeX treats file identifiers internally.  Now a file is only known by its base name and extension, and the path is used only when required by I/O operations.  This solves several problems of options not passing correctly due to the lack/missing path in `\opt@<base>.<ext>` token list, and also avoid that a package is mistakenly loaded twice if requested with different paths (even if they point to the same location in the disk).

This fixes issue #422, addresses #421 (though that is actually fixed by `xkeyval`), and addresses in part #426 (but some aspects of #426 are left as “by-design”).

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [ ] `ltnewsX.tex` updated&mdash;does not apply: hotfix


xexeo:
This actually seems to affect \@ifclassloaded and \@ifpackageloaded too, since it will not recognize the class/package by its name if it is in a subdirectory.
Does the correction fix this too?

PhelypeOleinik:
@xexeo I'm not sure which case you mean exactly, but this document (assuming there is a folder called `path` in the same directory as the `.tex` file):
```latex
\begin{filecontents}{path/pkg-A.sty}
\typeout{Loading pkg-A}
\end{filecontents}
\begin{filecontents}{path/pkg-B.sty}
\typeout{Loading pkg-B}
\end{filecontents}

\documentclass{article}
\makeatletter

\usepackage{path/pkg-A}

\def\input@path{{path/}}
\usepackage{pkg-B}

\@ifpackageloaded{pkg-A}
  {\typeout{pkg-A loaded!}}{}

\@ifpackageloaded{pkg-B}
  {\typeout{pkg-B loaded!}}{}

\begin{document}
\end{document}
```
prints to the terminal:
```
(./path/pkg-A.sty
Loading pkg-A
) (./path//pkg-B.sty
Loading pkg-B
)
pkg-A loaded!
pkg-B loaded!
```
which means both packages were loaded correctly, either using an explicit path or `\input@path`, and both were detected by `\@ifpackageloaded`.

If I try this same example with an older version of LaTeX I get:
```
(./path/pkg-A.sty
Loading pkg-A
) (./path/pkg-B.sty
Loading pkg-B
)
pkg-B loaded!
```
so the file loaded with the explicit path was not detected.

Do you have a specific issue? If so, can you make a compilable example?

PR #446, opened 2020-12-02
Integrate xparse as ltcmd

josephwright:
# Internal housekeeping

## Status of pull request

- Under development

- [x] Test file(s) added
- [ ] `ltnewsX.tex` updated


josephwright:
This is a new version of the PR: I've fixed the issues with the previous one, so it's now mergeable and should pass tests.

FrankMittelbach:
I'm currently reading usrguide3 and have a bunch of updates coming up. I also replaced @PhelypeOleinik with @davidcarlisle  for a second review (of code)as Phelype should do his thesis! not  code work.

base/build.lua
```diff
@@ -109,6 +109,7 @@ typesetfiles   =
     "ltx3info.tex",
     "modguide.tex",
     "usrguide.tex",
+    "usrguide3.tex",
```
FrankMittelbach, line 112: maybe we don't want to wait adding usrguide3 until next release?

base/doc/usrguide3.tex
```diff
@@ -0,0 +1,677 @@
...
+% \fi
+% Filename: usrguide3.tex
+
+\documentclass{ltxguide}
+
+\title{\LaTeX3 methods for authors}
+\author{\copyright~Copyright 2020, \LaTeX\ Project Team.\\
+   All rights reserved.}
+
+\date{Released 2020-12-02}
```
FrankMittelbach, line 37: if we wait until next year then this date is bogus

FrankMittelbach (commented):
while the integration of the code should have in 2021 I think the guide should  perhaps go out with the next PL

base/doc/usrguide3.tex
josephwright: Well yes, it's simply today, as it's in the Git repo now :)

josephwright:
> I'm currently reading usrguide3 and have a bunch of updates coming up. I also replaced @PhelypeOleinik with @davidcarlisle for a second review (of code)as Phelype should do his thesis! not code work.

Fine with me: I put down @PhelypeOleinik as he's been active on this area recently. But anyone 'sensible' is fine with me :)

josephwright:
> while the integration of the code should have in 2021 I think the guide should perhaps go out with the next PL

That would be easy enough: it's just a documentation file and we could simply cherry-pick the one commit (or squash in some edits, if required).

base/ltexpl.dtx
```diff
@@ -33,7 +33,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{ltexpl.dtx}
-             [2020/09/06 v1.2f LaTeX Kernel (expl3-dependent code)]
+             [2020/12/02 v1.3 LaTeX Kernel (expl3-dependent code)]
```
FrankMittelbach, line 36: The 2e convention is throughout x.y<letter> so either 1.2g or 1.3a

base/ltexpl.dtx
```diff
@@ -188,18 +188,26 @@
...
-         \fi
-      }%
+      {\input expl3.ltx }%
   }
 %<latexrelease>\EndIncludeInRelease
 %<latexrelease>
+%    \end{macrocode}
+% To support roll-forward for the case where \textsf{xparse} is fully
+% integrated into the kernel, we do not need to repeate the complex test
+% above as we can simply look for the marker command.
```
FrankMittelbach, line 198: and that work also for rolling back to that date?

FrankMittelbach (approved):
If we are sure that rolloback to 2020-10-01 works (don't think we have a test that tests that) then I think the code part is fine

FrankMittelbach:
Am 02.12.20 um 21:25 schrieb Joseph Wright:
 > +\date{Released 2020-12-02}
 >
 > Well yes, it's simply today, as it's in the Git repo now :)
 >

but it says released :-) it would be fine if we move it to master and 
release it within the next 2 weeks and we might as well do that imho

base/ltexpl.dtx
josephwright: Er, I *think* so

josephwright:
> Am 02.12.20 um 21:25 schrieb Joseph Wright: +\date{Released 2020-12-02} Well yes, it's simply today, as it's in the Git repo now :) but it says released :-) it would be fine if we move it to master and release it within the next 2 weeks and we might as well do that imho

I was assuming that once approved we'd alter the dates, rebase and merge ;)

FrankMittelbach:
Maybe I overlooked it, but is it anywhere stated that "o" and the like can deal with nested brackets? ie that `\foo[ \bar[baz] X]` works correctly without the need to hide the inner brackets?

Also I think there could some level of information comparing the old `\newcommand` to the new method and some background why and when the former may be preferable.

josephwright:
> Maybe I overlooked it, but is it anywhere stated that "o" and the like can deal with nested brackets? ie that `\foo[ \bar[baz] X]` works correctly without the need to hide the inner brackets?

I took the text from the existing `xparse` docs ... perhaps we've never actually documented that. (By 'we' I mean 'me': I wrote that code.)

josephwright:
> Also I think there could some level of information comparing the old `\newcommand` to the new method and some background why and when the former may be preferable.

OK, I'll think about something

FrankMittelbach:
Am 02.12.20 um 21:59 schrieb Joseph Wright:
> I took the text from the existing |xparse| docs ... perhaps we've never 
> actually documented that. (By 'we' I mean 'me': I wrote that code.)

I know :-) about the "we" here. Anyway it is a cool feature and worth 
documenting.

FrankMittelbach:
sorry, I thought I was on my branch but I was on the origin :-(

base/doc/usrguide3.tex
```diff
@@ -0,0 +1,677 @@
...
+processors is a somewhat advanced topic, (or at least a less commonly used
+feature) and is covered in Section~\ref{sec:cmd:processors}.
+
+\subsection{Creating document commands and environments}
+
+\begin{decl}
+  |\NewDocumentCommand|     \arg{cmd} \arg{arg spec} \arg{code} \\
+  |\RenewDocumentCommand|   \arg{cmd} \arg{arg spec} \arg{code} \\
+  |\ProvideDocumentCommand| \arg{cmd} \arg{arg spec} \arg{code} \\
+  |\DeclareDocumentCommand| \arg{cmd} \arg{arg spec} \arg{code} \\
```
stone-zeng, line 190: The final `\\` should be suppressed:
```suggestion
  |\DeclareDocumentCommand| \arg{cmd} \arg{arg spec} \arg{code}
```

and in Line 232, 548 it's the same.

josephwright: Frank spotted them :)

josephwright: Thanks for looking over the code :)

josephwright:
I have moved the user documentation to https://github.com/latex3/latex2e/pull/449 as a hotfix, so this PR is now only about the code part. There will be a bit of cleanup as a result when we merge back

josephwright:
Is anything else needed here? Or are we OK to merge?

FrankMittelbach:
> Is anything else needed here? Or are we OK to merge?

want me to resolve the conflicts?

josephwright:
> > Is anything else needed here? Or are we OK to merge?
> 
> want me to resolve the conflicts?

If you feel like it: I was mainly wondering if we are the stage where that can be done.

FrankMittelbach:
> > > Is anything else needed here? Or are we OK to merge?
> > 
> > 
> > want me to resolve the conflicts?
> 
> If you feel like it: I was mainly wondering if we are the stage where that can be done.

changed my mind ... I'll leave it up to you :-) but otherwise yes I think we should at some point in the near future

PhelypeOleinik:
> > > > Is anything else needed here? Or are we OK to merge?
> > > 
> > > 
> > > want me to resolve the conflicts?
> > 
> > 
> > If you feel like it: I was mainly wondering if we are the stage where that can be done.
> 
> changed my mind ... I'll leave it up to you :-) but otherwise yes I think we should at some point in the near future

I pushed a fix for latex3/latex3#839 and I hope to add `\NewCommandCopy` and `\ShowCommand` support to `xparse` tomorrow, so I'll do that also in the `latex3` repo (so it gets to TL 2020 before it's frozen), then I'll pull the changes here then I can do the merging.

PhelypeOleinik:
I merged develop back into this branch and solved the conflicts.

What about the test files that are still in the `latex3` repo?  These should probably be imported too, no?

And what will be the home of the “rest” of `xparse`?  I'm thinking maybe `required/tools/xparse`?

PhelypeOleinik:
> What about the test files that are still in the `latex3` repo? These should probably be imported too, no?
> 
> And what will be the home of the “rest” of `xparse`? I'm thinking maybe `required/tools/xparse`?

I think the best approach here would be to keep the deprecated argument types in the `ltcmd.dtx` source, and generate `xparse.sty` from that in `base`, so all of `xparse` would be “just moved” without awkward splitting of sources and tests.  Anyone against this?

FrankMittelbach:
> I think the best approach here would be to keep the deprecated argument types in the `ltcmd.dtx` source, and generate `xparse.sty` from that in `base`, so all of `xparse` would be “just moved” without awkward splitting of sources and tests. 
 Anyone against this?

It seems the most pragmatic solution even though it kind of feels wrong to add deprecated stuff that we really don't recomment people to use to base. On the other hand it is as it is and in tools it would be much better, so yes from my perspective fine (means updating manifest)

josephwright:
My plan had been to leave the stub `xparse` as-is in `l3packages`: it's lived there or the former `xpackages` _for ever_. Is there a pressing reason to move the odds and ends?

PhelypeOleinik:
> My plan had been to leave the stub `xparse` as-is in `l3packages`: it's lived there or the former `xpackages` _for ever_. Is there a pressing reason to move the odds and ends?

Not _pressing_, but it would separate the sources, which is a bit awkward, and (assuming test files are moved to the 2e side, which I think they should, since development of most `xparse` will be there) it would require splitting the test files in “core” and “deprecated” parts, so I think the less error prone approach would be to keep everything together.

josephwright:
> > My plan had been to leave the stub `xparse` as-is in `l3packages`: it's lived there or the former `xpackages` _for ever_. Is there a pressing reason to move the odds and ends?
> 
> Not _pressing_, but it would separate the sources, which is a bit awkward, and (assuming test files are moved to the 2e side, which I think they should, since development of most `xparse` will be there) it would require splitting the test files in “core” and “deprecated” parts, so I think the less error prone approach would be to keep everything together.

That was rather the point :) If you look at the branch in latex3 where I sketched things out, my plan was to strip `xparse.dtx` down to the deprecated stuff only, which then loads on to of the kernel part if it's available. (Either way, we will need a frozen `xparse-XXXX-XX-XX.sty` for the case where a user has an older kernel.)

My plan from the get-go was to separate the sources: `ltcmd.dtx` is _not_ the same as `xparse.dtx`, as the latter should be what generates `xparse.sty`. Even if you want to put `xparse` in tools, you still need two source files (one in base, one in tools).

FrankMittelbach:
On the whole I guess I favor Jospeh's approach. Seems cleaner overall.

> it would require splitting the test files in “core” and “deprecated” parts, so I think the less error prone approach would be to keep everything together.

hat shouldn't be really an obstacle? copy them all over and run them and then look at the few that fail and check if they fail because they test deprecated stuff, if so move them back.

PhelypeOleinik:
> On the whole I guess I favor Jospeh's approach. Seems cleaner overall.
> 
> > it would require splitting the test files in “core” and “deprecated” parts, so I think the less error prone approach would be to keep everything together.
> 
> hat shouldn't be really an obstacle? copy them all over and run them and then look at the few that fail and check if they fail because they test deprecated stuff, if so move them back.

Not _that_ clean because the `xparse` test files test all features in the same file, so the deprecated parts would have to be cut out of the tests.  Not impossible of course.  What I really don't like is splitting the sources in two (@josephwright that's why I changed my mind from `tools` to `base`, as that would require just moving files over).  But the deprecated types are a rather isolated part of the code, unlikely to be developed any further, so I guess it's not that bad to split them...

davidcarlisle:
I would also favour an end result where only the non deprecated parts are in the core distribution and the other parts are in an extension package distributed separately, even if that causes some work re-arranging the tests. I think otherwise the message that was intended to be given by splitting xparse is somewhat lost.

PhelypeOleinik:
Okay, so splitting the sources it is.  I will bring the test files over and merge

josephwright:
I think I split out the 'just the necessary' tests in my branch in latex3: my thinking was that we really should only need to test the 'additions'.

PhelypeOleinik:
I think this one is ready to merge into `develop` as well.

The checklist mentions an ltnews entry, but this move from `xparse` to `ltcmd` is mainly internal, and there is already a “Providing `xparse` in the format” in ltnews32, so I think it's not needed here.

FrankMittelbach:
> I think this one is ready to merge into `develop` as well.

I'll take another look later today

 
> The checklist mentions an ltnews entry, but this move from `xparse` to `ltcmd` is mainly internal, and there is already a “Providing `xparse` in the format” in ltnews32, so I think it's not needed here.

agreed. Not needed this time.

base/testfiles-ltcmd/ltcmd000.lvt
```diff
@@ -0,0 +1,15 @@
+%
+% Copyright (C) 2017,2021 LaTeX Project
```
FrankMittelbach, line 2: as I said elsewhere I think copyright lines on .lvt files are kind of overkill and just mean one needs to look after them :-) ... not hurting though, but I would suggest to just drop them.

FrankMittelbach (approved):
guess this is ready to go (after fixing the channges.txt merge problem. My commands could be addressed first or later (or not at all)

base/build.lua
```diff
@@ -155,7 +155,7 @@ indexstyle = "source2e.ist"
 
 -- Allow for TU and other test configurations
 checkconfigs = {"build","config-1run","config-TU","config-legacy","config-lthooks",
-                "config-lthooks2"}
+                "config-lthooks2","config-ltcmd"}
```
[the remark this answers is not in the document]
PhelypeOleinik: My opinion is not too strong in favour of one or the other (neither do I see any downsides for either), but I went for a separate set of tests because it makes it much easier to “test changes to `ltcmd`”.  If I'm working on some change to `ltcmd` I can just run `l3build check -c config-ltcmd` and the relevant (or more likely to break) tests will be run.  Also because we also separated the tests for `lthooks`, which makes it easier if I want to change something specific to that module (of course that doesn't exempt me from running the full test suite, but I can triage first).

base/testfiles-ltcmd/ltcmd000.lvt
PhelypeOleinik: Dilemmas :)  Since these dates are only the dates the files are modified, not `<creation date>-<now>`, it doesn't require manual changing each year/release.  But I for one tend to forget updating these if I change the test file, so I'd favour removing them too.  If nobody objects, I'll remove then

PhelypeOleinik:
> guess this is ready to go (after fixing the channges.txt merge problem.

Great!  But before I merge this one, I want to merge the `latex3` side of this undertaking, because as I said, the `latex3` side is standalone and should work with both new and old LaTeX, whereas this PR was not tested with the current `master` `xparse`.

Is the `latex3` side good to go?

PR #451, opened 2020-12-05
[base/config-filenames.lua] Fix `checkengines` table

SamuelMarks:
Unbalanced quotes, as per https://www.lua.org/pil/3.6.html they need to be balanced, and you probably want this balancing (assumed typo).

# Internal housekeeping

## Status of pull request

- Ready to merge

### DIFF
--- a/base/config-filenames.lua
+++ b/base/config-filenames.lua
@@ -1,5 +1,5 @@
 -- Tests for non Unicode engines
 
-checkengines = {"pdftex,"xetex","luatex""}
+checkengines = {"pdftex","xetex","luatex"}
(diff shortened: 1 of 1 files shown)


josephwright:
As things are currently working, I guess this PR should be to develop not to master? (In the sense it's not super-urgent.)

FrankMittelbach:
Normally I would say yes @josephwright,  but in this particular case and given that we wil have soon a PL3 we might as well put it in master (which during the PL3 is merged back to develop).

josephwright:
OK, merging in that case :)

PR #456, opened 2020-12-15
Hotfix/gh455 firstaid

FrankMittelbach:
# Internal housekeeping

## Status of pull request
- Ready to merge and to push to CTAN

- [x] Test file(s) added
- [n/a ] Relevant `\changes` entries in source included (not done for fileaid as temporary)
- [ n/a] `ltnewsX.tex` updated

@josephwright could you push to CTAN perhaps?


u-fischer:
I get an invalid object message from xdvipdfmx when compiling with xelatex:

~~~~
\documentclass{article}
\usepackage{pgfmorepages}
\pgfpagesuselayout{2 on 1}
\begin{document}

abc\newpage abc
\end{document}
~~~~

~~~~
xdvipdfmx:fatal: typecheck: Invalid object type: 9 6 (line 1786)

No output PDF file written.

Error 1 (driver return code) generating output;
file test-utf8.pdf may not be valid.
SyncTeX written on test-utf8.synctex.gz.
~~~~

FrankMittelbach:
> Actually I get the same with pgfpages, so it is not specific to the hotfix.

that's what I thought, can't see that this corrections is related.

josephwright:
@u-fischer Your example is fine in TL'19 so something is not right ...

u-fischer:
> > Actually I get the same with pgfpages, so it is not specific to the hotfix.
> 
> that's what I thought, can't see that this corrections is related.

Yes, but it looks as if something with the fix in general is not ok, it compiles if I remove the firstaid code from the hooks with `\RemoveFromHook{file/after/pgfmorepages.sty}[firstaid]` or `\RemoveFromHook{file/after/pgfpages.sty}[firstaid]`.  (then naturally we are back to the bug in tex.sx question which triggered this in first place). And it also compiles in texlive 2019. 

Should I open an extra issue for this?

FrankMittelbach:
> Yes, but it looks as if something with the fix in general is not ok, it compiles if I remove the firstaid code from the hooks with `\RemoveFromHook{file/after/pgfmorepages.sty}[firstaid]` or `\RemoveFromHook{file/after/pgfpages.sty}[firstaid]`. (then naturally we are back to the bug in tex.sx question which triggered this in first place). And it also compiles in texlive 2019.
> 
> Should I open an extra issue for this?

I can't replicate your findings. I added your example as another test file and at least locally it compiles with all engines correctly. Will upload and see if that is also the case with Travis.

u-fischer:
> I can't replicate your findings. I added your example as another test file and at least locally it compiles with all engines correctly. Will upload and see if that is also the case with Travis.

Interesting. On windows I have a newer xdvipdfmx then the average texlive user (as Akira updates it also between the releases), I will try to check if this matter.

FrankMittelbach:
> Interesting. On windows I have a newer xdvipdfmx then the average texlive user (as Akira updates it also between the releases), I will try to check if this matter.

as one can see (unless I made a mistake) your file runs also without issues on travis, so I guess the problem lies with Akira's update.

u-fischer:
> as one can see (unless I made a mistake) your file runs also without issues on travis, so I guess the problem lies with Akira's update.

it really looks like an xdvipdfmx problem. With older xdvipdfmx I can process the dvi/xdv without problems. I sent a mail to the dvipdfmx list. So the fix looks fine to me.

u-fischer (approved):

PR #462, opened 2020-12-27
Gh457

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [ n/a] Test file(s) added
- [ n/a] Version and date string updated in changed source files
- [x] `ltnewsX.tex` updated

### DIFF
--- a/base/lppl.md
+++ b/base/lppl.md
@@ -1,0 +1,197 @@
+# The LaTeX Project Public License
+
+
+LPPL Version 1.3c  2008-05-04 [markdown version]
+
+**Important notice:** *This is a markdown version of the LaTeX Project Public License (LPPL). As the markdown notation is often displayed differently on different devices (for example, showing 6.ii instead of 6b as a list label) it is only provided for convenince --- the legally binding version of the license is the plain text version in `lppl-1-3c.txt` available from [https://www.latex-project.org/lppl/](https://www.latex-project.org/lppl/).*
+
+
+Copyright 1999, 2002-2008, 2020 LaTeX3 Project
+
+Everyone is allowed to distribute verbatim copies of this license document, but modification of it is not allowed.
+
+
+
+
+## PREAMBLE
+
+The LaTeX Project Public License (LPPL) is the primary license under which the LaTeX kernel and the base LaTeX packages are distributed.
+
+You may use this license for any work of which you hold the copyright and which you wish to distribute. This license may be particularly suitable if your work is TeX-related (such as a LaTeX package), but it is written in such a way that you can use it even if your work is unrelated to TeX.
+
+The section `WHETHER AND HOW TO DISTRIBUTE WORKS UNDER THIS LICENSE`, below, gives instructions, examples, and recommendations for authors who are considering distributing their works under this license.
+
+This license gives conditions under which a work may be distributed and modified, as well as conditions under which modified versions of that work may be distributed.
+
+We, the LaTeX3 Project, believe that the conditions below give you the freedom to make and distribute modified versions of your work that conform with whatever technical specifications you wish while maintaining the availability, integrity, and reliability of that work. If you do not see how to achieve your goal while meeting these conditions, then read the document `cfgguide.tex` and `modguide.tex` in the base LaTeX distribution for suggestions.
+
+
+## DEFINITIONS
+
+In this license document the following terms are used:
+
+### `Work`
+Any work being distributed under this License.
+
+### `Derived Work`
+Any work that under any applicable law is derived from the Work.
+
+### `Modification`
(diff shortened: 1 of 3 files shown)


base/lppl.md
```diff
@@ -0,0 +1,197 @@
...
+
+     c. No information in the Derived Work implies that any persons, including (but not limited to) the authors of the original version of the Work, provide any support, including (but not limited to) the reporting and handling of errors, to recipients of the Derived Work unless those persons have stated explicitly that they do provide such support for the Derived Work.
+
+     d. You distribute at least one of the following with the Derived Work:
+
+          1. A complete, unmodified copy of the Work; if your distribution of a modified component is made by offering access to copy the modified component from a designated place, then offering equivalent access to copy the Work from the same or some similar place meets this condition, even though third parties are not compelled to copy the Work along with the modified component;
+
+          2. Information that is sufficient to obtain a complete, unmodified copy of the Work.
```
PhelypeOleinik, line 85: GitHub's markdown rendering (which as far as I know follows the markdown spec, so I think it can be trusted) understands these two items as blocks of code.  Removing four spaces before the `1.` and `2.` seems to fix the rendering (I didn't check the spec though).

base/lppl.md
```diff
@@ -0,0 +1,197 @@
...
+
+If the Work is unmaintained, you can become the Current Maintainer of the Work through the following steps:
+
+ 1. Make a reasonable attempt to trace the Current Maintainer (and the Copyright Holder, if the two differ) through the means of an Internet or similar search.
+
+ 2. If this search is successful, then enquire whether the Work is still maintained.
+
+     a. If it is being maintained, then ask the Current Maintainer to update their communication data within one month.
+
+     b. If the search is unsuccessful or no action to resume active maintenance is taken by the Current Maintainer, then announce within the pertinent community your intention to take over maintenance. (If the Work is a LaTeX work, this could be done, for example, by posting to comp.text.tex.)
```
PhelypeOleinik, line 127: Maybe make `comp.text.tex` a link.  Most people don't know what that is nowadays :)

Not sure to what extent that is allowed in a legal text though...

base/lppl.md
muzimuzhi: > An indented code block is composed of one or more indented chunks separated by blank lines. An indented chunk is a sequence of non-blank lines, each indented _four or more spaces_.

https://github.github.com/gfm/#indented-code-blocks

> Four spaces indent (of a list item) gives a code block:

https://github.github.com/gfm/#example-267

FrankMittelbach: crampdown has some specialties though. The above is not intended to be a code blockit is intended to be a sub-list within one item and you find that without the spaces it is not formatted as such but the numbering goes wrong.

on the other hand the numbering goes wrong anyway in places, which is why the thing is sitting there without being finished. I was getting more or less convinced that it is not a good idea to put the license into markdown unless it is actually faithfully representing the original (which I was up to now not able to achieve.

base/lppl.md
FrankMittelbach: there is no link in the license, but I guess we could make it one here (but on the other hand it gives more prominence to that example than it warrants as that wouldn't be then best way to announce your intention these days).

base/lppl.md
PhelypeOleinik: I did a small research and it doesn't seem to be possible to get the rendered output _consistently_ correct.  With pure markdown I can see three options: 1) we trick markdown into believing that there is no list, so `1. whatever` is rendered in HTML as `<p>1. whatever</p>`, which kind of defeats the purpose of markdown but keeps the text faithful to the legal text; 2) keep the bottom-level list as a list, and do 1) for nested lists, so
```markdown
1. whatever
  a. nested
    1. nested nested
```
becomes in HTML
```html
<ol>
<li>
<p>whatever</p>
<p>a. nested</p>
<p>1. nested nested</p>
</li>
</ol>
```
which looks kinda better but very far from good; or 3) format everything as nested lists:
```markdown
1. whatever
   1. nested
      1. nested nested
```
which will keep the semantics in the rendered HTML:
```html
<ol><li>whatever
  <ol><li>nested
    <ol><li>nested nested</li></ol>
  </li></ol>
</li></ol>
```
but will rely on the implemented renderer to decide how to label each level (GitHub labels as 1., i., a.; stackexchange labels as 1., 1., 1.).

We could bite the bullet and write HTML directly into markdown, but even that seems to be ignored (GitHub) or not at all supported (stackexchange). This:
```html
<ol type="1"><li>whatever
  <ol type="a"><li>nested
    <ol type="1"><li>nested nested</li></ol>
  </li></ol>
</li></ol>
```
renders here, regardless of what I used in `type`, as:
<ol type="1"><li>whatever
  <ol type="a"><li>nested
    <ol type="1"><li>nested nested</li></ol>
  </li></ol>
</li></ol>

So for a text that needs cross-referencing like this one, markdown doesn't seem to be a good option.  You either have control on the meaning or on the appearance. Not both...  
Seems eerily similar to MS Word

PR #492, opened 2021-02-07
update doc to V3

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development

- [x] Test file(s) added
- [x] `ltnewsX.tex` updated


FrankMittelbach:
Things still to do 
- kvoptions should be replaced 
- ltnews documentation
- some more code documentation
but otherwise ready to test with other packages I would say.

josephwright:
You looking for a review?

u-fischer:
- We just splitted hypdoc out of oberdiek do handle the "fix a bug/defficiency in hypdoc (should be fixed there eventually)" from 61d93fa  so this can be remove from doc if a few days when the update is on ctan.
-  doc now loads by default hypdoc, but without option. This means that this errors with an option clash:
~~~
\documentclass[a4paper]{ltxdoc}
\usepackage[numbered]{hypdoc}
~~~

This is used by a number of documentations, so it would probably better if doc would load `\RequirePackage[numbered]{hypdoc}` and test if hypdoc has already been loaded.

- l3build install installs also a `ltxdoc.cfg`. Shouldn't that be only a support file for our documentation?

davidcarlisle (commented):
oh no ``\stopeventually` had its name changed:-)

` \RequirePackage{hypdoc}` This is causing issues in the ho-tex packages (and so probably elsewhere) as the package gets loaded with options so hard to avoid option clashes depending on package order.  As the package is really a list of patches to doc then ideally (not necessarily in the first release) we'd just load hyperref and have doc being naturally (optionally) hyperref aware. Then hypdoc could be made a legacy stub package. Shorter term this could perhaps be delayed in an end of preamble hook to allow it to be loaded with options later.

Perhaps it doesn't matter (as it's only temporary until can switch to l3keys) but using kvoptions  means that the base release can no longer bootstrap itself without outside dependencies. In practice we have been using hyperref (at least) for years but the code could (in theory) always test and avoid it if not there,

Other than that, looks good

FrankMittelbach:
> oh no ``\stopeventually` had its name changed:-)

yeah sorry ... but I thought I should mention I understand the false friend after 3 decades :-)

> ` \RequirePackage{hypdoc}` This is causing issues in the ho-tex packages (and so probably elsewhere) as the package gets loaded with options so hard to avoid option clashes depending on package order. As the package is really a list of patches to doc then ideally (not necessarily in the first release) we'd just load hyperref and have doc being naturally (optionally) hyperref aware. Then hypdoc could be made a legacy stub package. Shorter term this could perhaps be delayed in an end of preamble hook to allow it to be loaded with options later.

@u-fischer suggested to load it with more options but (stop)eventually it should not be loaded at all but code just automatically made hyperref-aware. Loading it late is another option, one or the other should be done for now.

> Perhaps it doesn't matter (as it's only temporary until can switch to l3keys) but using kvoptions means that the base release can no longer bootstrap itself without outside dependencies. In practice we have been using hyperref (at least) for years but the code could (in theory) always test and avoid it if not there,

it is on the list (the V3 code was written long time ago but never integrated). The base can still bootstrap itself only the test that involve doc do not work without adding "search=true" and the documentation then needs kvoptions but that doesn't compile without a lot of additional packages that are not part of base (or tools) anyway so not much change to before. But yes, this should move to l3keys ... just waying for @josephwright coming up with the 2e interfaces :-)

> Other than that, looks good

good. Just detected a bug in index generation in l3build setup that showed up because of the more structured approach doc now has (probably all our indexes are wrong in the documentation (off by a page or so) and we just didn't notice

dbitouze:
> Just detected a bug in index generation in l3build setup that showed up because of the more structured approach doc now has (probably all our indexes are wrong in the documentation (off by a page or so) and we just didn't notice

Why not make use of `Latexmk`? It makes you sure that the document is stabilized if possible (otherwise it throws a clear warning).

FrankMittelbach:
> > Just detected a bug in index generation in l3build setup that showed up because of the more structured approach doc now has (probably all our indexes are wrong in the documentation (off by a page or so) and we just didn't notice
> 
> Why not make use of `Latexmk`? It makes you sure that the document is stabilized if possible (otherwise it throws a clear warning).

main reason is that this generates a dependency  (for example to perl) that we don't want on l3build and in practice there is no need for it. The fact that there was a coding error in the way l3build did its determination could have happened in latexmk as well (and in either case it shouldn't have happened)

josephwright:
The idea is that we _know_ the pattern needed for typesetting: the odd mistake in setting up what we need doesn't mean we don't start from that - the aim is to avoid any guessing.

base/alltt.dtx
```diff
@@ -104,7 +104,7 @@
 %  The glyph at the position of the $ in an italic font: \textit{$}.
 %    \end{alltt}
 %
-%  \StopEventually{}
+%  \MaybeStopHere{}
```
josephwright, line 107: I've always thought it's more (literally) `\StopPrehaps` or more fluidly `\ConditionalStop`/`\PotentialStop`

josephwright: @u-fischer  suggests `\MaybeStop` :)

FrankMittelbach: fine by me

FrankMittelbach:
> This is used by a number of documentations, so it would probably better if doc would load `\RequirePackage[numbered]{hypdoc}` and test if hypdoc has already been loaded.

@u-fischer solved differently in the end (to avoid automatically adding numbers suddenly).

base/letter.dtx
```diff
@@ -178,7 +178,7 @@
 % \maketitle
 % \tableofcontents
 %
-% \StopEventually
+% \MaybeStop
```
josephwright, line 181: No `{}`?

base/ltcmdhooks.dtx
```diff
@@ -1394,3 +1394,11 @@
...
 % \Finale
 %
+
+
+\endinput
+
+%%% Local Variables: 
+%%% mode: latex
+%%% coding: utf-8-unix
+%%% End:
```
josephwright, line 1404: Stray local lines?

base/ltoutenc.dtx
```diff
@@ -3744,3 +3744,10 @@
 % \Finale
 %
 \endinput
+
+
+
+%%% Local Variables: 
+%%% mode: latex
+%%% coding: utf-8-unix
+%%% End:
```
josephwright, line 3753: Sigh ...

josephwright (changes requested):

base/doc/ltnews35.tex
```diff
@@ -146,6 +146,43 @@ \subsection{???}
 
 \section{Code improvements}
 
+\subsection{\class{ltxdoc} gets \option{nocfg} option}
+
+\emph{document}
```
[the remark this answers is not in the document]
FrankMittelbach: that's why it has a placeholder :-) yes that is needed eventually

base/ltcmdhooks.dtx
FrankMittelbach: no not really. I need them because the files  contain characters that confuse emails and the file is not opened as utf8 but some strange mixed encoding. They are behind \endinput, so harmless but I coule alternatively, write them as
```
^^A  Needed for emacs
^^A
^^A  Local Variables: 
^^A  mode: latex
^^A  coding: utf-8-unix
^^A  End: 
```
if that is preferred --- I seem to need them in my setup.

base/letter.dtx
FrankMittelbach: so true :-) but it was always missing ... it works because it picks up the following `\par`. Added a brace group

base/ltfssbas.dtx
```diff
@@ -2257,3 +2257,11 @@
...
 % \Finale
 %
+
+\endinput
+
+%%% Local Variables: 
+%%% mode: latex
+%%% coding: utf-8-unix
+%%% End: 
+
```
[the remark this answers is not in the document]
FrankMittelbach: same argument

base/lthooks.dtx
```diff
@@ -5582,3 +5582,9 @@
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 \endinput
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+
+%%% Local Variables: 
+%%% mode: latex
+%%% coding: utf-8-unix
+%%% End:
```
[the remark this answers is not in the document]
FrankMittelbach: same

base/ltoutenc.dtx
FrankMittelbach: :-)

base/doc/ltnews35.tex
```diff
@@ -146,6 +146,43 @@ \subsection{???}
...
+
+\subsection{\pkg{doc} upgraded to version~3}
+
+After roughly three decades the \pkg{doc} package gets a cautious
+uplift, as already announced at the TUG conference 2019\Dash changes
+to \pkg{doc} are obviously always done in a leisurely manner.
+
+Given that most documentation is nowadays viewed on screen,
+\pkg{hyperref} support is added and by default enabled (suppress it
+with option \option{nohyperref}) so the internal cross-reference are
```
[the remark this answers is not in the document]
FrankMittelbach: yes, that is supported too, I'll add a note to that effect in ltnews

josephwright (approved):

u-fischer:
I still get an option clash from 

~~~~
\documentclass[a4paper]{ltxdoc}
\usepackage[numbered]{hypdoc}
~~~~

should that be resolved or will it have to wait for the new option handling?

FrankMittelbach:
> ```
> 
> ```
> 
> 
>     
>   
> 
> should that be resolved or will it have to wait for the new option handling?

it should work now -- can you make a quick check in your setup?

u-fischer:
it seems to work now.

PR #506, opened 2021-02-17
ltpara (light) integration

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

final decision on IndentationBox can be done afterwards during the next days/weeks

- [n/a] Test file(s) added  - the whole suite is test
- [x] `ltnewsX.tex` updated


base/ltpara.dtx
```diff
@@ -0,0 +1,1147 @@
...
+%    \begin{macrocode}
+%  \bool_lazy_and:nnT
+%       { \mode_if_horizontal_p: }
+%       { \bool_not_p:n { \mode_if_inner_p: } }
+%       { ...
+%    \end{macrocode}
+%    This is inner loop so we code the above test with low-level \cs{if_mode...}s.
+%    \begin{macrocode}
+  \if_mode_horizontal:
+    \if_mode_inner: \else:
```
PhelypeOleinik, line 880: I think these two conditionals should be replaced by the slightly higher level ones `\mode_if_horizontal:TF` and `\mode_if_inner:TF` because should the user hook contain an unbalanced conditional, things will go off tracks (only these two need changing though, because of the user-level hook inside the conditional body; others should be fine).

base/ltpara.dtx
```diff
@@ -0,0 +1,1147 @@
...
+%    The we execute the public hook (which may add final typesetting
+%    material) followed by the kernel hook we need for adding tagging
+%    support. None of this is supposed to change the mode---at the
+%    moment we make only a very simple test for this, more devious
+%      changes go unnoticed, but too bad, that will then probably
+%      badly backfire.
+%    \begin{macrocode}
+         \hook_use:n{para/end}
+         \@kernel@after@para@end
+         \if_mode_horizontal:
```
PhelypeOleinik, line 911: This `\if_mode_horizontal:` would need changing as well because of `para/after`

PhelypeOleinik (approved):
Overall looks good.  I only suggest changing three conditionals to slightly higher level ones to avoid problems with unbalanced conditionals (the higher level ones are pretty optimised, so it shouldn't affect speed noticeably).

What I'm not sure about is the half integration now.  It just adds complication to `latexrelease` later, so I would suggest to put this in its final place in one go, unless there is a more pressing reason to half-integrate at this time

u-fischer (approved):

FrankMittelbach:
> 
> What I'm not sure about is the half integration now. It just adds complication to `latexrelease` later, so I would suggest to put this in its final place in one go, unless there is a more pressing reason to half-integrate at this time

I indent to have the full later (probably worked on during the freeze), it cuts across much more terrain and there is other work needed before the freeze.

base/changes.txt
```diff
@@ -6,6 +6,12 @@ completeness or accuracy and it contains some references to files that
 are not part of the distribution.
 ================================================================================
 
+2021-02-16  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
+
+	* ltpara.dtx: providing paragraph hooks,; full integration into
+	latex.ltx will happen at a later point in time. For documentation
```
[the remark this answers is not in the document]
FrankMittelbach: yes that was my plan (before May) ready for the last -dev release in April

josephwright (approved):

base/ltpara.dtx
FrankMittelbach: What kind of  unbalanced conditionals you see could happen here?

Even though they are optimized, it is still a difference between 2 primitive fast scan/jump ahead to reading the code several times as arguments to \use_ii:nnn and the like. (doesn't mean I can't get convinced to use \mode_if_horizontal:TF instead

base/changes.txt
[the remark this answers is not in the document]
PhelypeOleinik: Ah, okay, that will need no `latexrelease` entry, so it's okay

base/ltpara.dtx
FrankMittelbach: hmm, that `para/hook` could be more of a problem, so I guess you are right and I should use the higher-level interfaces

base/ltpara.dtx
josephwright: I guess the worry is someone might put `\if...` in the begin hook, `\fi` in the end hook, for excluding blocks of text.

FrankMittelbach: yes I see that (for excluding text that would be a pretty weird approach though :-))  but yes you could also have an env added that for some reason starts a conditional, so yes, Phelype has already convinced me

PhelypeOleinik: Any incomplete TeX conditional command would either throw an error in the wrong place or make the code do something else.  For example, `\AddToHook{para/after}{\fi}` errors with `Extra \else`, and `\AddToHook{para/after}{\else}` makes the kernel hook `\@kernel@after@para@after` be silently ignored (actually makes it only execute if in vmode, followed by the error).

Just to bring some numbers to the table: doing some light benchmarking consistently gives (for 100000 iterations of `\leavevmode\par`) 1.24 s for the raw conditional version and 1.39 s for the higher level version: a 12% relative increase.  Quite a lot.  On the other hand, it's 0.15 s longer for 100000 paragraphs, or 0,0015 ms per paragraph, so not that much in absolute times. Besides, doing the same test with `Lorem ipsum dolor sit amet, consectetur adipisicing elit.` instead of `\leavevmode` raises the times to 2.17 s and 2.34 s, respectively, for the same 100000 paragraphs.

PhelypeOleinik: Note that using `\if...\fi` like that to ignore a paragraph wouldn't work: the `\fi` would never be seen because it's hidden in the hook and would be skipped by the `\iffalse`.

PhelypeOleinik: What does save you a couple of cycles in both cases (for the 100000 paragraphs, 0.02 s in the raw conditional case and 0.10 s in the `\mode_if_...:TF` case is to hide the error message with
```latex
\cs_new_protected:Npn \__para_end_vertical_error:
  { \__kernel_msg_error:nnnn {para}{mode}{end}{horizontal} }
```
then use the single token `\__para_end_vertical_error:` in `\para_end:`.

FrankMittelbach: > What does save you a couple of cycles in both cases (for the 100000 paragraphs, 0.02 s in the raw conditional case and 0.10 s in the `\mode_if_...:TF` case is to hide the error message with
> 
> ```
> \cs_new_protected:Npn \__para_end_vertical_error:
>   { \__kernel_msg_error:nnnn {para}{mode}{end}{horizontal} }
> ```
> 
> then use the single token `\__para_end_vertical_error:` in `\para_end:`.

that's what I was kind of commenting on in the other thread about error messages, kist passing over them as longish token lists takes time. But this is perhaps something to generally look at  at least in places that are extremely often executed.

base/doc/ltnews33.tex
```diff
@@ -112,6 +112,31 @@ \section{Introduction}
 
 
 
+\section{Extending the hook concept to paragraphs}
+
+Largely triggered by the need for better control of paragraph text
+processing, in particular when producing tagged PDF output (see
+\cite{33:blueprint}), we have extended the paragraph processing of
+\LaTeX{} so that the kernel gains control both at the start and the
+end of each paragraph. This is is done in a manner that is (or should
```
muzimuzhi, line 121: typo: `This is is done` -> `This is done`

FrankMittelbach: thanks

jfbu:
Hi,package using privately some "dissecting paragraph technique via lastbox" and for this putting things in some vbox with the material ended by `\@@par` then `\analyselines` will get the user/package hooks executed which are destined to body of document. Maybe this can cause problems. For people maintaining code not using expl3 layer, there should be some "usable" alias for `\tex_par:D`. ``\ifdefined\@@@@par\@@@@par\else\@@par\fi`` would help people ensure their old `\@@par` code will behave exactly as in past independent of hooks in paragraphs actually typesetting things and not in boxes used to measure things.

FrankMittelbach:
@jfbu an example would be helpful (even though I understand what you are saying, I guess). I don't doubt that there may be some adjustments needed including perhaps building a paragraph where the hooks are totally disabled ( which may be better than accessing `\tex_par:D` )

jfbu:
```latex
% Mercredi 24 février 2021 à 09:49:28
\documentclass{article}

\makeatletter
\newbox\jfb@boxi
\newbox\jfb@boxii

% simple dissecting technique
% works in simple situations after \par (the TeX one)
\def\jfb@getwidths {%
    \unskip\unpenalty
    \setbox\jfb@boxii\lastbox
    \ifvoid\jfb@boxii
    \else
       \xdef\jfb@steps{\the\numexpr\jfb@steps+1}%
       \setbox\jfb@boxii\hbox{\unhbox\jfb@boxii}%
       \ifdim\jfb@maxwidth<\wd\jfb@boxii
          \xdef\jfb@maxwidth{\number\wd\jfb@boxii sp}%
       \fi
       \ifdim\jfb@minwidth>\wd\jfb@boxii
          \xdef\jfb@minwidth{\number\wd\jfb@boxii sp}%
       \fi
       \expandafter\jfb@getwidths
    \fi
}%

\def\getwidths#1{%
    \global\let\jfb@maxwidth\z@
    \global\let\jfb@minwidth\linewidth
    \gdef\jfb@steps{0}%
    \setbox\jfb@boxi
    \vtop{\raggedright\hyphenpenalty\z@\exhyphenpenalty\z@
          \doublehyphendemerits\z@\finalhyphendemerits\z@
          \strut #1\strut\@@par
          \jfb@getwidths
    }%
   \immediate\write128{%
    \jfb@steps\space lines.^^J%
    \@spaces Found max width: \the\dimexpr\jfb@maxwidth,
                   min width: \the\dimexpr\jfb@minwidth\relax
                   \space(linewidth: \the\linewidth)}%
}%
\makeatother

\def\samplenospace{aaaa}
\edef\Samplenospace{\samplenospace\samplenospace\samplenospace}
\edef\SamplenospacE{\Samplenospace\Samplenospace\Samplenospace}
\edef\sample{\samplenospace\space\Samplenospace\space\SamplenospacE\space
             \SamplenospacE\SamplenospacE\SamplenospacE\space\samplenospace\space}
%\show\sample
\begin{document}
\hfuzz\maxdimen
\getwidths{\sample\sample\sample\sample}

\makeatletter
\let\@@@@par\@@par
\def\paraafterhook{\vskip1sp\vskip1cm }
\def\@@par{\@@@@par\paraafterhook\@@@@par}%
\makeatother

\getwidths{\sample\sample\sample\sample}

\end{document}
```

Produces:
```
9 lines.
    Found max width: 540.00165pt, min width: 23.33339pt (linewidth: 345.0pt)
0 lines.
    Found max width: 0.0pt, min width: 345.0pt (linewidth: 345.0pt)
```

But I don't know what will be the guidelines for such hooks and what they are allowed to do.

In the case above a `\pleaselatexdonotexecuteparaafterhooktrue` toggle would be useful, (but even simpler would be for `\@@par` to keep its well-established original meaning so that one does not have to complicate code by checking for latex version).

As per the code above, I took it from some real-life problem with identifying overflowing lines in some verbatim+syntax highlighting like rendering, so that one can modify the input to let its mark-up add necessary breakpoints if needed, and do the typesetting with these additional breakpoints. I know this is simplistic but it works fine in this original context where one knows the input `#1` does not contain extra mark-up inserting things in the vertical list.

The other simpler remark, regarding rather "before" hooks: what if some user decides to count paragraphs via stepping a LaTeX counter? If such hooks are executed even in private dealings typesetting multiple times in private vboxes, results will be surprising.

In the past it has been known that `\@@par` or  for example `\@@input` are the TeX primitives. If this changes, legacy documents might break (although I am probably exaggerating here). But couldn't `\@@par` and `\endgraf` be left untouched and LaTeX kernel code use a `\latex@@par` for the mechanisms of lists etc..., which will be incorporating the hook mechanism? I am asking without thinking. The idea is that the kernel code would never use `\@@par` which is sanctified but replace it everywhere with its own, hook receptive, version.

jfbu:
In my use case I guess I can make sure code will work with current as future LaTeX via something such as:
```
\ifcsname tex_par:D\endcsname etc... \else\let\my@@par\@@par\fi
```
at start of "macros".

The hooks at start of paragraph are probably potentially less of a worry in the limited context of my dissection thing.

I don't know what tagged pdfs are, but the infrastructure to produce them should perhaps leave open the possibilities of playing (from inside LaTeX) with naked TeX boxes privately, which are not related to such or such output format and constraints.

FrankMittelbach:
> Produces:
> 
> ```
> 9 lines.
>     Found max width: 540.00165pt, min width: 23.33339pt (linewidth: 345.0pt)
> 0 lines.
>     Found max width: 0.0pt, min width: 345.0pt (linewidth: 345.0pt)
> ```
> 

If I run your code example through current LaTeX it produces the above and if I run it through the upcoming latex-dev nothing changes.

Of course, if somebody would go `\AddToHook{para/after}{\vskip1sp\vskip1cm }` then the code in the `para/after` hook would interfer with your backwards parsing.  So yes, it is probably advisable that for programming we also offer a hookless  version that doesn't run them.

> But I don't know what will be the guidelines for such hooks and what they are allowed to do.

guidelines are ltpara-doc but as with everything they probably will grow over time.


> In the case above a `\pleaselatexdonotexecuteparaafterhooktrue` toggle would be useful, (but even simpler would be for `\@@par` to keep its well-established original meaning so that one does not have to complicate code by checking for latex version).

no the latter would not be simpler. The main point is not the public hooks (they are there for user/package writer convenience) but the internal LaTeX private hooks that establish something similar to `\everypar` at the end of the paragraph. That should have been an engine feature but as it isn't, we added it on macro level.  If we would have kept  `\@@par`as it is then that control point would be unavailable in many many cases where we need it. On TeXLive there are close to 200 files that use `\@@par` and basically all of them use it  for typesetting and not for measuring like in your case. But even then, the internal control points do no harm, only explicit public hook code would interfer.

> As per the code above, I took it from some real-life problem with identifying overflowing lines in some verbatim+syntax highlighting like rendering, so that one can modify the input to let its mark-up add necessary breakpoints if needed, and do the typesetting with these additional breakpoints. I know this is simplistic but it works fine in this original context where one knows the input `#1` does not contain extra mark-up inserting things in the vertical list.

sure it worked (and still works) unless you get additional material that your parser didn't expect. and yes, because of applications like this we may want to offer a "non-typesetting" version that omits the hooks

> The other simpler remark, regarding rather "before" hooks: what if some user decides to count paragraphs via stepping a LaTeX counter? If such hooks are executed even in private dealings typesetting multiple times in private vboxes, results will be surprising.

sure but that's the case always and not in any way special to these hooks. If you put code with side effects in then you need to make sure that it isn't  run multiple times, counter changes are an obvious example.

> In the past it has been known that `\@@par` or for example `\@@input` are the TeX primitives. If this changes, legacy documents might break (although I am probably exaggerating here). 

I'm sure something will break, as that is basically unavoidable whenever something changes. But the goal was to abstractly keep `\@@par` the primitive (only with an extended definition) from the package code perspective.

> But couldn't `\@@par` and `\endgraf` be left untouched and LaTeX kernel code use a `\latex@@par` for the mechanisms of lists etc..., which will be incorporating the hook mechanism? I am asking without thinking. The idea is that the kernel code would never use `\@@par` which is sanctified but replace it everywhere with its own, hook receptive, version.

As I explained this is not for lists (for which LaTeX already had its controls in place in the past), it is for each an every typeset paragraph. So no ,by controlling the layer between engine and package code level we gain reasonable control throughout and, yes it means that for a very small number of packages there is a need for adjustment (and possibly for us to provide additional support for that). The alternative is to say, whenever you use one of the 200 packages that break well-tagged PDF because they use \@@par for typesetting then tough luck ... go and get their authors to change their package code.

FrankMittelbach:
> In my use case I guess I can make sure code will work with current as future LaTeX via something such as:
> 
> ```
> \ifcsname tex_par:D\endcsname etc... \else\let\my@@par\@@par\fi
> ```
> 
> at start of "macro package".

I would rather say that this is not an advisable approach. It is will be better if you use something that is made for this purpose . But yes it would work (as long as you do not use `\my@@par` for doing typesetting (as that would then break your documents going forward.
 
> The hooks at start of paragraph are probably potentially less of a worry in the limited context of my dissection thing.

for your case it is really only the para/after hook not even the para/end hook
 
> I don't know what tagged pdfs are, but the infrastructure to produce them should perhaps leave open the possibilities of playing (from inside LaTeX) with naked TeX boxes privately, which are not related to such or such output format and constraints.

as I said, we are aware of that being a possible need we have to cater for. What is needed there depends a bit on use cases which is why we push this out 2-3 month in advance.

As to what are tagged pdfs: take a look at https://www.latex-project.org/publications/indexbytopic/pdf/

FrankMittelbach:
> ```
> \makeatletter
> \let\@@@@par\@@par
> \def\paraafterhook{\vskip1sp\vskip1cm }
> \def\@@par{\@@@@par\paraafterhook\@@@@par}%
> \makeatother
> ```

One question: in your code is this to mimic `\AddToHook{para/after}` or does this has some deeper significance?

jfbu:
Le 24 févr. 2021 à 16:51, Frank Mittelbach <notifications@github.com> a écrit :

> 
> \makeatletter
> \let\@@@@par\@@par
> \def\paraafterhook{\vskip1sp\vskip1cm }
> \def\@@par{\@@@@par\paraafterhook\@@@@par}%
> \makeatother
> One question: in your code is this to mimic \AddToHook{para/after} or does this has some deeper significance?
> 
> —
> 
> 

no no this was only for proof of concept that it would break the \unskip\unpenalty \lastbox thing.

I had been reading ltpara.dtx in the previous evening and I mimicked the \AddToHook{para/after} to test in current latex what the hooks would give in my paragraph dissection context (which exists for real at Sphinx, so I was worrying about when some people will have LaTeX 2021 and add some package using hooks because they want for example paragraphs being separated by some ornaments perhaps not realizing paragraphs are everywhere; the input in my context is output from Pygments and is completely controlled). And I was a bit in a hurry so did not check if \addvspace{..} or \vspace{...}  are the latex things doing the double \vskip (and perhaps rather in the reversed order and a \z@skip).

FrankMittelbach:
Am 24.02.21 um 17:43 schrieb Jean-François B.:
> 
> no no this was only for proof of concept that it would break the 
> \unskip\unpenalty \lastbox thing.

so I thought, just wanted to make sure.

> I had been reading ltpara.dtx in the previous evening and I mimicked the 
> \AddToHook{para/after} to test in current latex what the hooks would 
> give in my paragraph dissection context (which exists for real at 
> Sphinx, so I was worrying about when some people will have LaTeX 2021 
> and add some package using hooks because they want for example 
> paragraphs being separated by some ornaments perhaps not realizing 
> paragraphs are everywhere; 

well, give them the tools and you get weird results :-).  Sure if people 
do not realize their code will fail, but this is like in other places 
too, if somebody tries globally using \everypar it would fail badly in 
many places and so it is with the hook code in general too if you use 
them without restriction.

The para hooks have been added, because there are some use cases in 
which they are interesting for package developers, but for using them 
uncontrolled, they are of little use. However, I doubt that this will 
cause problems, simply because it will not fail in a subtle but in a 
very clear way.

I guess all that is necessary in that respect is to make this very clear 
in the documentation (probably better than it is now).

blefloch:
> The para hooks have been added, because there are some use cases in
> which they are interesting for package developers, but for using them
> uncontrolled, they are of little use. However, I doubt that this will
> cause problems, simply because it will not fail in a subtle but in a
> very clear way.
> 
> I guess all that is necessary in that respect is to make this very clear
> in the documentation (probably better than it is now).


This discussion reminds me a bit of some code we had (maybe in l3galley or
similar) where \par was looking ahead to see by what it was followed.  I don't
remember details, but we could imagine checking if this is the last paragraph of
an \item in a list, for instance.

This is all very vague, sorry.

josephwright:
> This discussion reminds me a bit of some code we had (maybe in l3galley or similar) where \par was looking ahead to see by what it was followed. I don't remember details, but we could imagine checking if this is the last paragraph of an \item in a list, for instance. 

You are thinking of
```latex
\cs_new_protected:Npn \galley_par:
  {
    \s_@@_par_omit
    \exp_after:wN \@@_par_auxi: \exp:w \exp_end_continue_f:w
  }
\cs_new_protected:Npn \@@_par_auxi:
  {
    \peek_meaning:NTF \s_@@_par_omit
      { \@@_par_aux:N }
      { \@@_par_auxii: }
  }
\cs_new_protected:Npn \@@_par_aux:N #1
  {
    \str_if_eq:nnF {#1} { \s_@@_par_omit }
      {
        \@@_par_auxii:
        #1
      }
  }
\cs_new_protected:Npn \@@_par_auxii:
  {
    \mode_if_vertical:TF
      {
        \group_begin:
          \tex_par:D
        \group_end:
      }
  % Continues to insert stuff
```
which makes sure that we only insert material when we start a 'real' paragraph.

jfbu:
à propos

> >  no no this was only for proof of concept that it would break the \unskip\unpenalty \lastbox thing.
>
> so I thought, just wanted to make sure.

the

> ```
> \makeatletter
> \let\@@@@par\@@par
> \def\paraafterhook{\vskip1sp\vskip1cm }
> \def\@@par{\@@@@par\paraafterhook\@@@@par}%
> \makeatother
> ```

Should have been simply (I was too lazy to read how to use the hooks so emulated them, not exactly the same `\vskip`'s)
```
\AddToHook{para/after}{\vspace{1cm}}
```
and test the latex code with ``latex-dev`` of February/March TeXLive (pretest of 2021)

I believe the above usage is typical of what a LaTeX user might be thinking of doing.

PR #515, opened 2021-03-04
Gh415

### LINKED ISSUE #415: filecontents breaks with UTF-8 in filename if file exists
## Brief outline of the bug

When overwriting a file with UTF-8 characters, `filecontents` will break with:
```none
! Undefined control sequence.
\UseTextAccent  ...p \@firstofone \let \@curr@enc 
                                                  \cf@encoding \@use@text@en...
l.5 \begin{filecontents}[overwrite]{föö.tex}
                                              
?
```
whereas writing the file the first time works correctly.

```latex
\RequirePackage{latexbug}
\begin{filecontents}{föö.tex}
hi
\end{filecontents}
\begin{filecontents}[overwrite]{föö.tex}
hi again
\end{filecontents}
\stop
```

[foo.log](https://github.com/latex3/latex2e/files/5397676/foo.log)

PhelypeOleinik:
This PR fixes #415: overwriting a file with Unicode characters in its name with `filecontents`.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [x] `ltnewsX.tex` updated


car222222 (commented):
Is it essential at this stage to introduce l3 programming level code into ltnews.tex?

PhelypeOleinik:
Not _essential_ (I could use `\texorpdfstring` there) but I don't think it causes any trouble.  Does it?

car222222:
It just looked out-of-place, that’s all.  Actually it should probably all be hidden by something defined elsewhere (at a suitably low-level) for the whole process of using lowercase+small-caps.

PhelypeOleinik:
Yeah, likely we eventually need to reorder `ltnews` (it's kinda messy, each with its own different preamble, then a common one for all).  But I'd rather leave that to another task and not pollute this PR with that

PR #532, opened 2021-03-15
Hotfix/bigfoot fixing an allocation assumption

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [n/a ] Relevant `\changes` entries in source included
- [n/a] `ltnewsX.tex` updated

@josephwright once merged can you upload firstaid?

I will afterwards do the integration into develop


josephwright (approved):

required/firstaid/latex2e-first-aid-for-external-files.dtx
```diff
@@ -397,6 +397,45 @@
...
+%
+%    This firstaid therefore jumps over the problematical point by
+%    pushing the count allocation to a safe value if necessary.
+%    \begin{macrocode}
+\AddToHook{file/before/bigfoot.sty}{%
+   \ifnum\count10>188 %
+   \else
+     \global\count10=189 %
+   \fi
+%    \end{macrocode}
```
davidcarlisle, line 420: am I missing something here, but why 188 and why just check count 10?

`\newinsert` switches to stealling boxes from the float free list if

```
\ifnum\count10<\insc@unt
\ifnum\count11<\insc@unt
\ifnum\count12<\insc@unt
\ifnum\count14<\insc@unt
```

so don't you need to make count10 bigger than `insc@ount` (which is 198 initially with current kernel allocations)

FrankMittelbach: maybe you do  or maybe I do :-).  Point is, perpage, manyfoot and bigfoot (which loads the other two) do allocations and there is no reason to push them out. 198 is bigger than 188 last time I looked, so making a tests for 198 and pushing to that would make a much bigger gap than just going to 189 which is enough to get the first important insert taken from the float list. So basically what the test does is to ensure that the count register allocation is at least high enough that inserts will always be stolen from the floats. As count registers are close to that cutoff value that is all that is needed (assuming perpage/manyfoot/bigfoot are not suddenly dropping register allocations (which is rather unlikely). Anything wrong with that reasoning?

davidcarlisle: er, I suppose so OK let this go through it's only first-aid after all, the hardcoded 188 looked slightly strange, but it  is essentially \insc@ount - number of \newcount bigfoot does before it starts doing \newinsert, so would be wrong if another package does inserts and changes inscount, that is pretty unlikely though, to load another insert packeg. I guess \numexpr\insc@ount-10\relax would be safer?

FrankMittelbach: I agree the value is a but odd, but if there are other packages that allocate stuff it doesn't get wrong as far as I can see, as it is only a minimal value and so would then not apply and all is still good. But I guess going after the package is probably clearer
```
\AddToHook{file/after/bigfoot.sty}{%
   \ifnum\count10<\insc@unt %
     \global\count10=\insc@unt %
   \fi}
```

josephwright: Looks nice to me [other than we could drop those `%` as they are redundant ;)]

FrankMittelbach: yes and no. They are redundent, but they point out that the space is deliberate and not by mistake.

FrankMittelbach: so are we good with that version @davidcarlisle ?

davidcarlisle (approved):

required/firstaid/latex2e-first-aid-for-external-files.dtx
josephwright: @FrankMittelbach  My point was the version with `\ifnum\count10<\insc@unt %` doesn't need the `%` - no big deal, just a bit of fun really.

josephwright: I'll wait for a merge to main then do a release of first-aid.

FrankMittelbach: > @FrankMittelbach My point was the version with `\ifnum\count10<\insc@unt %` doesn't need the `%` - no big deal, just a bit of fun really.

right you are ... no longer

PR #542, opened 2021-04-13
updated and integrated ltpara

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

hopefully complete

- [x] Test file(s) added
- [x] `ltnewsX.tex` updated


u-fischer:
Is the sentence in the documentation for \Omitindent  that the box remains available for use still correct? It seems to contradict a bit the description in  the hooks (but perhaps I misunderstand the "for use" meaning).

car222222:
The code is thus:

%     The declaration \cs{para_omit_indent:} (or
%    \cs{OmitIndent}) changes that to do nothing.
%    \begin{macrocode}
\cs_new:Npn \para_omit_indent: {
  \box_gclear:N \g_para_indent_box
}

So I think that the sentence should be removed.  (I am not too sure what it meant even with the previous behaviour.)

The above phrase "changes that to do nothing" is also not very helpful.  Partly because the way the code works is rather more complex.  

I can deal with both of these when I go over all the documentation (when I get a 'round too-wit':-).

FrankMittelbach:
> So I think that the sentence should be removed. (I am not too sure what it meant even with the previous behaviour.)
> 
> The above phrase "changes that to do nothing" is also not very helpful. Partly because the way the code works is rather more complex.
> 
> I can deal with both of these when I go over all the documentation (when I get a 'round too-wit':-).

that loooks like leftover documentation where the code got updated but the documentation is still from the earlier implementation, sorry for that. So yes, that needs updating.

car222222:
Really just a note to myself!

This last sentence here also got left over:

% \begin{variable}{\IndentBox,\g_para_indent_box}
%    The box register holding the indentation box for the paragraph is
%    available for inspection (or changes) inside hooks. It remains
%    available even if the \cs{OmitIndent} command was
%    used; in that case it  will just not be automatically placed.

PR #547, opened 2021-04-20
Gh524

PhelypeOleinik:
This PR adds support for `\tracingstacklevels` and `\tracinglostchars=3` into `\tracingall/none` and `\traceon/off`.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [x] `ltnewsX.tex` updated


PhelypeOleinik:
@zauguin Should you have the time to take a look at my flaky Lua code, that would be great.  I was wondering if there is a way to define `\tracingstacklevels` in LuaTeX such that it behaves like a primitive register, so that `\meaning\tracingstacklevels` gives `\tracingstacklevels

base/doc/ltnews33.tex
```diff
@@ -503,6 +503,33 @@ \subsection{Small fix for rolling back prior to 2020-02-02}
...
+Later, in January 2021, Petr Olšák suggested yet another extension, a
+new primitive parameter \cs{tracingstacklevels} that, when positive (and
+when \cs{tracingmacros} is also positive), will print a visual
+indication of the macro nesting level in \TeX's tracing information.
+
+Both these changes were incorporated to \LaTeX's debugging macros
+\cs{tracingall} and \cs{tracingnone}, so when you use them, the new
+extensions are automatically activated/deactivated if your \TeX{}
+distributions has a recent enough engine.  An example document
+demonstrating these parameters is available in the linked GitHub issue.
```
FrankMittelbach, line 527: I'm not sure what the best strategy is for \tracinglostchars. It seems to me that for that parameter there is not a "single" default, i.e., you may want to unconditionally set it to 3 for a document.
In that case \traceoff or \tracingnone should not kill that setting, should it?

base/ltluatex.dtx
```diff
@@ -1399,12 +1399,14 @@ local callbacktypes = callbacktypes or {
...
 % \changes{v1.1j}{2019/10/22}{page\_objnum\_provider and process\_pdf\_image\_content classified data}
 % \changes{v1.1l}{2020/02/02}{page\_order\_index added}
+% \changes{v1.1t}{2021/04/18}{input\_level\_string added}
 %    \begin{macrocode}
   finish_pdffile            = data,
   finish_pdfpage            = data,
   page_objnum_provider      = data,
   page_order_index          = data,
   process_pdf_image_content = data,
+  input_level_string        = data,
```
zauguin, line 1409: s/data/exclusive/

The `data` type is for callbacks which take data as their first argument, potentially modify it a little and return it again. But ´input_level_string`takes a completely different data type as input and output.

base/ltfinal.dtx
```diff
@@ -242,7 +242,70 @@
...
+            else
+              return "\string~\string~"
+            end
+          else
+            return "\string\n"
+          end
+        else
+          return ""
+        end
+      end
```
zauguin, line 277: Looks good.

base/ltfinal.dtx
```diff
@@ -242,7 +242,70 @@
...
+%<latexrelease>
+%<latexrelease>\IncludeInRelease{0000/00/00}{trace_stack_levels}%
+%<latexrelease>                 {Lua trace_stack_levels function}%
+%<latexrelease>\ifx\directlua\@undefined
+%<latexrelease>\else
+%<latexrelease>  \directlua{%
+%<latexrelease>    if luatexbase.callbacktypes['input_level_string'] and %
+%<latexrelease>       luatexbase.in_callback('input_level_string','tracingstacklevels') then
+%<latexrelease>        luatexbase.remove_from_callback('input_level_string','tracingstacklevels')
+%<latexrelease>    end}%
```
zauguin, line 302: Maybe run this outside of the `\IncludeInRelease` block in order to always redefine the callback? Then we don't get issues later if we ever change the function and `latexrelease` doesn't pick up the changes.

base/ltfinal.dtx
```diff
@@ -242,7 +242,70 @@
...
+            end
+          else
+            return "\string\n"
+          end
+        else
+          return ""
+        end
+      end
+%<*latexrelease>
+      if luatexbase.callbacktypes['input_level_string'] and %
```
zauguin, line 279: This shouldn't hurt but since `ltluatex.lua` doesn't rollback the test shouldn't be necessary.

zauguin:
> @zauguin Should you have the time to take a look at my flaky Lua code, that would be great. I was wondering if there is a way to define `\tracingstacklevels` in LuaTeX such that it behaves like a primitive register, so that `\meaning\tracingstacklevels` gives `\tracingstacklevels` rather than `\count<something>`...



>  I was wondering if there is a way to define `\tracingstacklevels` in LuaTeX such that it behaves like a primitive register, so that `\meaning\tracingstacklevels` gives `\tracingstacklevels` rather than `\count<something>`...

Sadly there isn't, you would need `luametatex` for that.

base/doc/ltnews33.tex
PhelypeOleinik: Hm... Indeed!  Maybe save the current value of `\tracinglostchars` in `\tracingall` and restore in `\tracingnone`?

base/ltluatex.dtx
PhelypeOleinik: Thanks, I wondered which would be the right type here

base/ltfinal.dtx
PhelypeOleinik: My reasoning was that in case of rollforward someone might be using an `ltluatex` without that callback installed, so the test in the `latexrelease` version prevents adding to an undefined callback.  Right or did I miss something?

base/ltfinal.dtx
PhelypeOleinik: Good point.  Done in b26e7ed7

base/doc/ltnews33.tex
FrankMittelbach: It is a general problem also with other tracing parameters, e.g., if you set \showoutput and then somewhere you go \traceon ... \traceoff or \tracingnone settings arge gone. So perhaps it does need a general solution saving and restoring all values. We could implement something like that now or as a separate issue at some point in the future  (maybe that is better).

PhelypeOleinik: An approach similar to what `\ExplSyntaxOn/Off` do with catcodes should do:
```latex
\let\tracingnone\relax
\def\tracingall{%
  \edef\tracingnone{%
    \tracinglostchars=\the\tracinglostchars\relax
    \tracingmacros=\the\tracingmacros\relax
    ...
    \let\tracingnone\relax
  }
  \tracinglostchars=\thr@@
  \tracingmacros=\tw@
  ...
}
```

Then `\tracingnone` would normally be a no-op, and `\tracingall` would define `\tracingnone` to do the proper resetting.

I'll leave it as that for now and open a separate issue for this, to keep things well separated

FrankMittelbach:
Am 20.04.21 um 15:36 schrieb Phelype Oleinik:
> Then |\tracingnone| would normally be a no-op, and |\tracingall| would 
> define |\tracingnone| to do the proper resetting.
> 
> I'll leave it as that for now and open a separate issue for this, to 
> keep things well separated
> 

yes makes sense (both :-))

base/ltfinal.dtx
zauguin: I thought that when rolling forward, `ltluatex` is rolled forward too, but that probably depends on the user (if they install the `ltluatex.lua`or not). So keeping the test is probably safer.

base/ltfinal.dtx
```diff
@@ -242,7 +242,78 @@
...
+         not luatexbase.in_callback('input_level_string','tracingstacklevels') then
```
zauguin, line 291: No longer needed since the new version always removes it first.

zauguin (approved):

Issue #528, opened 2021-03-08
Unexpected outcome when using a robust command to specify a filename for `\input`


jfbu:
## Brief outline of the bug

I don't know is this expected:

```
(./testlatexinput2.tex
LaTeX2e <2020-10-01> patch level 4
L3 programming layer <2021-02-18>
(/usr/local/texlive/2021/texmf-dist/tex/latex/latexbug/latexbug.sty)
(/usr/local/texlive/2021/texmf-dist/tex/latex/base/article.cls
Document Class: article 2020/04/10 v1.4m Standard LaTeX document class
(/usr/local/texlive/2021/texmf-dist/tex/latex/base/size10.clo))
(/usr/local/texlive/2021/texmf-dist/tex/latex/l3backend/l3backend-pdftex.def)
(./testlatexinput2.aux)

! LaTeX Error: File `protectfilename.tex' not found.

Type X to quit or <RETURN> to proceed,
or enter new name. (Default extension: tex)
```

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}
\documentclass{article}

\begin{document}

\DeclareRobustCommand\foo{file}

\input{\foo name}
\end{document}

```

and the directory does contain a file with name `filename`.

## Log file (required) and possibly PDF file  

[testlatexinput2.log](https://github.com/latex3/latex2e/files/6100462/testlatexinput2.log)

FrankMittelbach:
What would you expect to happen?

Basically robust commands do not expand unless you are typesetting, so the fact that there is no need to "protect" the word "file" it is not really possible for LaTeX to look inside the definition and determine what to make of it. So it is rendered harmless somehow.

The actual result is of course rather useless and I agree curious, but not much can be done about that imo. If you use \raggedleft instead of \foo you get
```
! LaTeX Error: File `protectlet\@centercrrightskipz@skipleftskip@flushgluefinal
hyphendemerits=z@parindentz@parfillskipz@skipname.tex' not found.
```
which is even worse. Basically everything remotely looking like a character became one. So getting explicitly rid of the "protect" in the beginning wouldn't help.

Note that with the current -dev release it changes so it now tries to find
```
! LaTeX Error: File `foo name.tex' not found.
! LaTeX Error: File `raggedleft name.tex' not found.
```
which still doesn't make robust commands usable but seems a somewhat more consistent handling to me.

jfbu:
[...]

Issue #523, opened 2021-03-06
LaTeX2e: character per character filename parsing done by \InputIfFileExists (et al.)  looks costly to an ignorant external observer/user


jfbu:
## Brief outline of the question

Yesterday I investigated a minted issue. The minted package caches files using two md5sums in the basename so the basename is 64 characters long. Examining with `\tracingmacros1` I discovered what appeared to be thousands of expl3 macro expansions during `\minted@input` which does `\IfFileExists` and then `\input`.

I have now tested with `\InputIfFileExists` using current TL 2021-pretest pdflatex (not pdflatex-dev).

For filenames without extension it appears there is a cost of 125 macro calls per character in the filename:

>  Law of input for extension less filenames: (414 + 125 * (length of filename)) macro calls

For filenames with extension this appears to drop to 75 macro calls per character.

> Law of input for filenames with extension: (245 + 75 * (length of filename inclusive of extension)) macro calls

I understand there is something impolite in asking ; Wittgenstein as a kid realized politeness in society meant lying and opted for a life out of the expected pathways. But, I wonder if the character per character approach does not have alternatives.

I vaguely understand the parsing is due to checking for quote character and space character and perhaps the dot character (I have not looked).

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{article}

\newcommand\testinput[1]{%
\typeout{START TEST #1}%
  \tracingmacros1
  \tracingstacklevels2
  \InputIfFileExists{#1}{}
  \tracingmacros0
\typeout{STOP TEST #1}%
}
\begin{document}


% 1
\testinput{A}
% 539 macro calls

% 2
\testinput{AA}
% 664 macro calls

% 3
\testinput{AAA}
% 789 macro calls

% 4
\testinput{AAAA}
% 914 macro calls

% 8
\testinput{AAAAAAAA}
% 1414 macro calls

% 16
\testinput{AAAAAAAAAAAAAAAA}
% 2414 macro calls

% 64
\testinput{AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA}
% 8414 macro calls

Law of input for extension less filenames: (414 + 125 * (length of filename)) macro calls

% 1
\testinput{A.pygtex}
% 845 macro calls

% 2
\testinput{AA.pygtex}
% 920 macro calls

% 3
[...]

PR #550, opened 2021-04-20
Escape ~ in \input

PhelypeOleinik:
With the filename sanitisation change (latex3/latex3#869), `~` will now “just expand”, yielding `\nobreakspace {}` in a file name.  To preserve functionality, this adds `\edef~{\string~}` to `\set@curr@file` so that `~` is just `~` in a file name.

Related to #523 and #528 

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [n/a] `ltnewsX.tex` updated


PhelypeOleinik:
Wait a bit... Travis went rogue

FrankMittelbach:
> Wait a bit... Travis went rogue

I guess that's it for at least a day ... happens to me already all day. Only existing branches with a cache work, anything new fails

base/testfiles/github-0528.tlg
```diff
@@ -0,0 +1,15 @@
...
+hello :)
+)
+! Illegal parameter number in definition of \reserved@a.
+<to be read again> 
+                   }
+l. ...\texttt{\input{my{file}$with~sp&cials.tex#}}
+You meant to type ## instead of #, right?
+Or maybe a } was forgotten somewhere earlier, and things
+are all screwed up? I'm going to assume that you meant ##.
+(my{file}$with~sp&cials.tex#
```
FrankMittelbach, line 13: that's really the desired outcome, guess it probably is if we want to stay lean  and simple.

FrankMittelbach (approved):

base/testfiles/github-0528.tlg
PhelypeOleinik: Yes, I just added it as a test because it was mentioned in https://github.com/latex3/latex2e/issues/523#issuecomment-792623214, but to change that we'd need to change how `\text@command` works, so it's more of an assertion that it's working as intended

PR #582, opened 2021-06-07
Tl21-2 kernel msgs normalization (ltcmd mainly)

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development

- [x] Test file(s) added
- [ n/a] `ltnewsX.tex` updated

This is the result from looking at all core LaTeX error and warning messages in parallel and my attempt to rationalize them a bit. This is still WIP so feedback and corrections are welome.


base/ltcmd.dtx
```diff
@@ -1273,7 +1273,7 @@
       {
         \tl_if_empty:NF \l_@@_last_delimiters_tl
           {
-            \msg_error:nnxx { cmd } { invalid-after-optional-expandably }
+            \msg_error:nnxx { xparse } { invalid-after-optional-expandably }
```
josephwright, line 1276: I think a mistake!

base/ltcmd.dtx
```diff
@@ -3506,79 +3506,79 @@
...
+  { Bad~argument~specification~'#2'~for~#1. }
   {
     Expandable~commands~must~have~a~final~mandatory~argument~
     (or~no~arguments~at~all).~You~cannot~have~a~terminal~optional~
     argument~with~expandable~commands.
   }
-\msg_new:nnnn { cmd } { inconsistent-long }
-  { Inconsistent~long~arguments~for~expandable~command~'#1'. }
+\msg_new:nnnn { cmd } { long-short-mix }
+%  { Inconsistent~long~arguments~for~expandable~command~'#1'. }
```
josephwright, line 3558: A stray line from testing?

josephwright (approved):
Minor typo-level changes could be made, otherwise looks good

base/ltcmd.dtx
```diff
@@ -3611,143 +3613,155 @@
...
+%  { Two~'#2'~apply~to~the~same~argument~in~argument~specification~of~#1. }
   {
-    The~argument~specification~provided~has~two~markers~'#2'~applying~
-    to~the~same~argument;~these~are~redundant.
+    The~argument~specification~provided~for~#1~has~two~'#2'~markers~applied~
+    to~the~same~argument;~one~is~redundant.
   }
-\msg_new:nnnn { cmd } { unknown-argument-type }
-  { Unknown~argument~type~'#2'~for~the~#1. }
+\msg_new:nnnn { cmd } { unknown-argument-type } % should be unkown-arg-type but dep in xparse
```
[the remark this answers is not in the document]
FrankMittelbach: we probably can, there are 2 such places.

base/ltcmd.dtx
FrankMittelbach: why? The branch can only be taken (afais) if xparse got loaded, it is called if there was an l or u in the spec before.

josephwright: If the code is `xparse`-specific it shouldn't even be in `ltcmd`, other than the stub 'you need to load `xparse` message at the 'dead end'.

FrankMittelbach: true, you probably don't even need a stub because the calling function couldn't get called unles u and l are available. But this is the issue with xparse still in l3kernel, the coordination is a little arkward which is why I just made sure that the error message says xparse package

PhelypeOleinik: This also removes this particular message from the `cmd` namespace, so redirection, if used, won't affect this one message.  I don't think this improves things.  Rather, if this is only used by `xparse` it should be moved.

The move would have to be more or less synchronised, otherwise we'd have this same definition in two places and it would error (or check before defining in `xparse`).

base/ltcmd.dtx
```diff
@@ -3506,79 +3506,79 @@
 % Some messages intended as errors when defining commands/environments.
 %    \begin{macrocode}
 \msg_new:nnnn { cmd } { arg-after-body }
-  { In~the~definition~of~#1,~b~(body)~argument~must~be~last. }
+  { Invalid~argument~type~'#2'~in~#1. }
```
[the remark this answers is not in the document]
FrankMittelbach: Part of the reasons for rewrites is to keep error messages to a single line and explanations to multiple lines, the other is to standardizes on less error messages.
It is more important that a grep shows the problematic csname #1 than the part of the more detailed explanation.
I don't see that the message  as given is wrong the argument  type 'm' is invalid whether or not that is because it  is always invalid or invalid in that position.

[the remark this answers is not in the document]
FrankMittelbach: > This also removes this particular message from the cmd namespace, so redirection, if used, won't affect this one message. I don't think this improves things. Rather, if this is only used by xparse it should be moved.

It could get moved but that wouldn't change its behavior, would it? And as you say, it would need some arkward syncing without actual gain (as far as I can see) which is why I just changed the module. I don't mind keeping it "cmd" but it is a package error and for some reason the code was left in the kernel but it gets never executed unless the package is loaded.

base/ltcmd.dtx
PhelypeOleinik: > It could get moved but that wouldn't change its behavior, would it? And as you say, it would need some arkward syncing without actual gain (as far as I can see) which is why I just changed the module. I don't mind keeping it "cmd" but it is a package error and for some reason the code was left in the kernel but it gets never executed unless the package is loaded.

If it is left as `xparse` no, it would behave the same (i.e. redirection wouldn't work).  I'd dare say that `xparse` messages should remain as `cmd` because, for example, in `xparse.sty` it uses the `cmd` name: https://github.com/latex3/latex3/blob/95436ed690026255060de12ddd3cf113ad51265b/l3packages/xparse/xparse.dtx#L1081-L1099, so changing to `xparse` here feels a bit awkward.  I'd leave `cmd` here...

base/ltcmd.dtx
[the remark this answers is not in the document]
FrankMittelbach: ok, how about `{ Argument~type~'b'~must~be~last~in~#1. }` ?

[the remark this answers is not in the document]
PhelypeOleinik: Sounds good too.  I didn't remove `Invalid` because I thought you wanted that as the first word for sorting.

PhelypeOleinik (approved):

base/ltcmd.dtx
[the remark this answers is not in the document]
FrankMittelbach: only wanted the "Invalid" if the messages  are then identical, if not different odering (in line with other messages) is fine as long as it is short :-)

base/ltcmd.dtx
FrankMittelbach: changed back to cmd

zauguin (approved):

base/ltcmd.dtx
FrankMittelbach: now removed commented out lines

PR #589, opened 2021-06-16
Missing arguments added

dbitouze:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/lthooks.dtx
+++ b/base/lthooks.dtx
@@ -1497,7 +1497,7 @@
 %
 % \begin{function}{\BeforeBeginEnvironment}
 %   \begin{syntax}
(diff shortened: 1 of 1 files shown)


josephwright:
Could you also change the file date (but not version) for `lthooks.dtx`.

FrankMittelbach:
> Could you also change the file date (but not version) for `lthooks.dtx`.

normally I would agree @josephwright Joseph, but since there are pending updates in branches which change the date (and the version further) it is easier if we do this. So I would suggest @dbitouze leaves it as this and we merge and handle it at our end.

PR #615, opened 2021-07-12
Gh613

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [x] Test file(s) added


base/lterror.dtx
```diff
@@ -375,6 +375,58 @@ Type  H <return>  for immediate help.}%
...
+%<latexrelease>\IncludeInRelease{2021/11/15}%
+%<latexrelease>                 {\ClassNote}{Notes for classes/packages}%
+%    \end{macrocode}
+%
+%    \begin{macrocode}
+\def\ClassNote#1#2{%
+   \GenericWarning{%
+      (#1) \space\space\@spaces\@spaces
+   }{%
+      Class #1 Info: #2%
```
PhelypeOleinik, line 396: Shouldn't this be `Note` instead of `Info`?

base/doc/ltnews34.tex
```diff
@@ -146,6 +146,27 @@ \section{???}
...
+
+\LaTeX{} offers \cs{Packageerror} to signal errors that stop
+processing, \cs{PackageWarning} that generate a warning messageon the
+terminal, but continue with the processing and also \cs{PackageInfo}
+to provide some information that is only written to the \texttt{.log}
+file. What hasn't existed up to now is a way to provide some
+information on the terminal that is identifying itself as coming from
+a specific package but which isn't claiming to be a warning. Thus,
+packages that wanted to write to the terminal used \cs{PackageWarning}
+even though the information wasn't really warning the user.  for this
```
dbitouze, line 159: "for this" → "For this"

base/doc/ltnews34.tex
```diff
@@ -146,6 +146,27 @@ \section{???}
...
+we now have \cs{PackageNote} and \cs{PackageNotNoLine} that idntify
```
dbitouze, line 160: "idntify" → "identify"

PhelypeOleinik (approved):

base/lterror.dtx
FrankMittelbach: We could have a different word here. But my thought was it is just information one goes to the terminal+log the other just to the "log". Also not that at the moment the L3 interface also writes info (I think) in both cases

PhelypeOleinik: I thought of using `Note` instead because of the name of the command.  But indeed, `expl3`'s `\msg_note:nnn` says `Info`, so nevermind this comment

FrankMittelbach: I think it is ok to use the same string here.

PR #631, opened 2021-07-27
Gh625 (Simplify \RemoveFromHook behaviour) 

PhelypeOleinik:
This PR simplifies the behaviour of `\RemoveFromHook` to only remove code that already exists in the hook. 

# Internal housekeeping

## Status of pull request
 
- Under development

- [x] Test file(s) added


base/lthooks.dtx
```diff
@@ -321,14 +321,17 @@
 %    When the optional argument \meta{label} is not provided, the
 %    \meta{default label} is used (see section~\ref{sec:default-label}).
 %
-%    If the code for that \meta{label} wasn't yet added to the
-%    \meta{hook}, an order is set so that when some code attempts to add
-%    that label, the removal order takes action and the code is not
-%    added.
+%    If there is no code under the \meta{label} in the \meta{hook},
+%    or if the \meta{hook} does not exist, an error is raised when
```
FrankMittelbach, line 325: I wonder if this should be an error or just a warning (or not even that). I would probaby go for a warning only even with your explanation below

base/lthooks.dtx
```diff
@@ -1020,10 +1024,9 @@
 %   \end{syntax}
 %    Removes any code for \meta{hook} labeled \meta{label}.
 %
-%    If the code for that \meta{label} wasn't yet added to the
-%    \meta{hook}, an order is set so that when some code attempts to add
-%    that label, the removal order takes action and the code is not
-%    added.
+%    If there is no code under the \meta{label} in the \meta{hook},
+%    or if the \meta{hook} does not exist, an error is raised when
```
FrankMittelbach, line 1028: same here

base/lthooks.dtx
```diff
@@ -3116,27 +3146,13 @@
...
-%    Otherwise check if the label being removed exists in the code pool.
-%    If it does, just call \cs{@@_gremove_code_do:nn} to do the removal,
-%    otherwise mark it to be removed.
-%    \begin{macrocode}
-                \prop_get:cnNTF { g_@@_#1_code_prop } {#2} \l_@@_return_tl
-                  { \@@_gremove_code_do:nn }
-                  { \@@_mark_removal:nn }
-                      {#1} {#2}
+                \prop_gpop:cnNF { g_@@_#1_code_prop } {#2} \l_@@_return_tl
+                  { \msg_error:nnnn { hooks } { cannot-remove } {#1} {#2} }
```
FrankMittelbach, line 3155: strange space before `\l_@@retun_tl` (maybe only in the diff)

base/testfiles-lthooks/lthooks-001.tlg
```diff
@@ -109,6 +109,24 @@ All initialized (non-empty) hooks:
...
  xxx -> foobar\__hook_toplevel xxx \__hook_next xxx  
+! LaTeX hooks Error: Cannot remove chunk 'label3' from hook 'xxx' because it
+(hooks)              does not exist in that hook.
+Type <return> to continue.
+ ...                                              
+l. ...\hook_gremove_code:nn{xxx}{label3}
+                                        % errors because label3 doesn't exist
+LaTeX does not know anything more about this error, sorry.
+Try typing <return> to proceed.
+If that doesn't work, type X <return> to quit.
```
FrankMittelbach, line 120: if we keep it an error then there should also be a help text, but as I said I think warning is probably sufficient

FrankMittelbach (commented):
mainly random notes, not a real code review

base/lthooks.dtx
PhelypeOleinik: I have mixed feelings.  On the one hand, an error sounds more discouraging if someone wants to use `\RemoveFromHook` and is faced with the trouble of finding out which labels are there and which are not.  On the other hand, trying to remove a label that doesn't exist and then end up doing nothing is not really serious, so maybe not worth of an error...

I suppose I'll change to a warning then...

base/lthooks.dtx
PhelypeOleinik: I only see the space between `{#2}` and `\l_@@_return_tl`, but I suppose that's not what you mean?

base/lthooks.dtx
```diff
@@ -2766,11 +2762,17 @@
 % \subsection{Adding or removing hook code}
 %
 % \begin{macro}{\hook_gput_code:nnn}
-% \begin{macro}{\@@_gput_code:nnn,\@@_gput_code:nxv,\@@_hook_gput_code_do:nnn}
+% \begin{macro}{\@@_gput_code:nnn,\@@_hook_gput_code_do:nnn}
+%
+%    \begin{macrocode}
+%<latexrelease>\IncludeInRelease{2021/11/15}%
+%<latexrelease>                 {\hook_gput_code:nnn}{Do~not~queue~removals}
```
FrankMittelbach, line 2769: not sure we should really do a rollback or consider that a correction of the initial code that should have been this way day one. It complicates maintenance and I doubt that it has been used with the earlier meaning and if so a rollback would only be of limited use even then. Doing it wouldn't hurt that much but I think not doing would be fine as well.

support/test2e.tex
```diff
@@ -295,7 +295,24 @@
...
+% added to the shipout box, and since \RemoveFromHook no longer
+% removes code added in the future, we need to delay the removal
+% to "shipout/before".  We also need to redirect error messages
+% because in some tests, the "l3backend" chunk is not added.
+% The "voids" rule is no good here because "shipout/firstpage"
+% being nonempty is enough to add the extra box and kern to the
+% output.
+\ExplSyntaxOn
+\AddToHook{shipout/before}
+  {
```
FrankMittelbach, line 309: An here as well

FrankMittelbach (approved):
see comments

base/lthooks.dtx
FrankMittelbach: I was only showing strangely in the view I used I guess

car222222 (commented):
I am unsure what is meant by the end of this sentence, after the comma:

%    \cs{RemoveFromHook} should be used when you know exactly what
 %    labels are in a hook, usually added and removed within the same
 %    package.

car222222:
Looks good now!

PR #649, opened 2021-08-22
Fix typo

ivan-pan:
The file `amsldoc.tex` contains the typo `mathools`.


FrankMittelbach:
thanks;  I decided to  fix it manually (and not add another "revision" date, as that would end up being a very long list otherwise).

Issue #652, opened 2021-08-24
amsmath/hyperref: unary minus becomes binary at the begin of an equation


u-fischer:
## Brief outline of the bug
If hyperref is used with amsmath, an unary minus becomes a binary minus as the `refstepcounter` inserts a destination. The suggested fix is to insert an empty mathopen after the `\refstepcounter`. 

References: 
https://tex.stackexchange.com/q/235665/2388
https://tex.stackexchange.com/q/610904/2388

## Minimal example showing the bug

```latex
\RequirePackage{latexbug} %errors because of hyperref
\documentclass{article}
\usepackage{amsmath}
\usepackage{hyperref}
\begin{document}
\begin{equation}
  -1 = -1 
\end{equation}

\makeatletter
\def\incr@eqnum{\refstepcounter{equation}\ifmmode\mathopen{}\fi\let\incr@eqnum\@empty}
\makeatother

\begin{equation}
  -1 = -1
\end{equation}
\end{document}
```

![image](https://user-images.githubusercontent.com/4047173/130588125-9323f3fd-f75d-4179-98e2-d27e1aa14c19.png)

## Log file (required) and possibly PDF file  

[b-467.log](https://github.com/latex3/latex2e/files/7037891/b-467.log)

FrankMittelbach:
I'm not convinced this is the right approach. The reason why there is an issue with hyperref is that hyperref changes the definition of `\equation` and moves `\incr@eqnum` into the math part. When doing that it could at the same time add the open atom as well. Then the test for math etc isn't unnecessarily done in all the places where `\incr@eqnum` is used outside math.

If there are changes to amsmath I think it should remove the patching need for hyperref not support and "incomplete" patch.

Anyway the suggest push in #653 is incomplete as it misses out resetting the definition in `\@arrayparboxrestore`. But as I said I think it is better to alter `\equation`to explicitly add the open atom after `\incr@eqnum` because then you don't need to test for math (you know you are in math) and it is only where it is needed.

u-fischer:
Well I would love to move the whole hyperref patch to amsmath. Would this here be ok?

~~~~
\documentclass{article}
\usepackage{amsmath}
\usepackage{hyperref}
\begin{document}
\begin{equation}
  -1 = -1
\end{equation}

\makeatletter
\if@fleqn
\else
\long\def\equation{%
        \mathdisplay@push
        \st@rredfalse \global\@eqnswtrue
        \mathdisplay{equation}%
[...]

PR #653, opened 2021-08-24
guard refstepcounter with mathopen, issue #652

u-fischer:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added (I don't know how to make a sensible test file without hyperref)
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated (no sure if needed?)

### DIFF
--- a/required/amsmath/amsmath.dtx
+++ b/required/amsmath/amsmath.dtx
@@ -86,7 +86,7 @@ Bug reports can be opened (category \texttt{#1}) at\\%
 %    \end{macrocode}
 %
 %    \begin{macrocode}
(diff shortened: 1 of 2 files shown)


FrankMittelbach (changes requested):
As mentioned in #653 I think the change should be in `\equation` and that could (for now)  stay in hyperref or we could totally eliminate the need for patching and do it here. but not change `\incr@eqnum`.

Note that adding an open atom on the left is fine but not on the right as there it does influence spacing so putting it out in an uncontrolled way is a bit dangerous---who knows which macros have used `\incr@eqnum` besides amsmath itself.

PR #656, opened 2021-08-27
Gh648 [lthooks] Standardise generic hook names

### LINKED ISSUE #648: [lthooks] Reorder generic hook names
## Brief outline of the enhancement

When developing `lthooks` we ended up with generic hooks named in two different ways: with the variable part in the second position as in `env/<name>/before`, and in the end, as in `file/before/<name>`.  This difference makes it difficult to remember which is which, and makes for unnecessary complication in the code.

The plan here is to change all generic hooks to have the variable part in the second position in the name, so these hooks will be changed (hoping I didn't forget any):

| Old name                   | New name                   |
| -------------------------- | -------------------------- |
| `file/before/<file>`       | `file/<file>/before`       |
| `file/after/<file>`        | `file/<file>/after`        |
| `package/before/<package>` | `package/<package>/before` |
| `package/after/<package>`  | `package/<package>/after`  |
| `class/before/<class>`     | `class/<class>/before`     |
| `class/after/<class>`      | `class/<class>/after`      |
| `include/before/<file>`    | `include/<file>/before`    |
| `include/end/<file>`       | `include/<file>/end`       |
| `include/after/<file>`     | `include/<file>/after`     |

A deprecation cycle will be provided so that the old names still work for some time, to allow adjustments without compatibility issues.

PhelypeOleinik:
This PR closes #648.  The generic hook names are reordered so that the variable part is in the middle position (`package/foo/before` instead of `package/before/foo`) to match `env` and `cmd` hooks.  A deprecation fallback is provided so that the old names still work normally (with a warning) so that people have time to adjust.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


base/testfiles-lthooks/lthooks-018.lvt
```diff
@@ -15,8 +15,8 @@
 
 \START
 
-\AddToHook{env/before/quote}[my1]{\dotfill before 1\dotfill\par}
+\AddToHook{env/quote/before}[my1]{\dotfill before 1\dotfill\par}
 
-\LogHook{env/before/quote}
+\LogHook{env/quote/before}
```
FrankMittelbach, line 20: I wonder what this was supposed to test in the past. "env" always had its variable part in position 2  ... so maybe the test was supposed to test for an input error and should not be altered. After all, 019 has env/quote/before in the right order

FrankMittelbach (approved):
a few minor things noticed

base/ltfilehook.dtx
```diff
@@ -720,16 +720,10 @@
...
-%
-%  Furthermore, those named \texttt{.../after} or \texttt{.../end} are
-%  automatically declared as reversed hooks if filled with code, so this
-%  is also automatically taken care of.
-%
-%
+%  Three-part hooks starting with \texttt{file/} \texttt{include/},
+%  \texttt{class/} or \texttt{package/} and ending with \texttt{/before}
+%  or \texttt{/after} (or \texttt{/end} in case of \texttt{include/})
+%  are generic and will be allocated if code is added to them.
```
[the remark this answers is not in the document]
PhelypeOleinik: The documentation for each specific hook mentions whether it's reversed.  Though that second paragraph I removed by mistake.  I added it back.

base/testfiles-lthooks/lthooks-018.lvt
PhelypeOleinik: I tracked this test file down to commit 61460202 in the AccessiblePDF repository, which was related to `\typeout{\par}` (triggered by a `\ShowHook`), so the intent of the test didn't change.  It was just a typo back then.

FrankMittelbach: then maybe we should add a note to the test what it is testing :-) as it looks a little useless on first glance

PhelypeOleinik: I added a better explanation

Issue #658, opened 2021-08-31
Case and line breaking with \ij 


jbezos:
## Brief outline of the bug

There are two bugs related to `\ij` and `\IJ`:
* They aren't taken into account by `\MakeUppercase` and `\MakeLowercase`.
* With `OT1`, a line break is prevented if followed by a space.

As to `OT1`, I think the `\nobreak`s in `\ij` and `\IJ` are misplaced and they must be:
```latex
\DeclareTextCommand{\ij}{OT1}{%
  i\kern-0.02em\nobreak\hskip\z@skip j}
\DeclareTextCommand{\IJ}{OT1}{%
  I\kern-0.02em\nobreak\hskip\z@skip J}
```

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}
\documentclass{article}

% \usepackage[T1]{fontenc}

\begin{document}

\MakeUppercase{\ij} \MakeLowercase{\IJ} % Prints ij IJ

\hsize1pt

a b \ij{} c d
  
\end{document}
```

## Log file (required) and possibly PDF file  

```
This is pdfTeX, Version 3.141592653-2.6-1.40.23 (TeX Live 2021/W32TeX) (preloaded format=pdflatex 2021.8.30)  31 AUG 2021 10:10
entering extended mode
 restricted \write18 enabled.
 %&-line parsing enabled.
**xxx.tex
(./xxx.tex
LaTeX2e <2021-06-01> patch level 1
L3 programming layer <2021-06-18>
(c:/Aplicaciones/TeXLive/texmf-dist/tex/latex/latexbug/latexbug.sty
Package: latexbug 2021/06/28 v1.0k Bug-classification
)
(c:/Aplicaciones/TeXLive/texmf-dist/tex/latex/base/article.cls
Document Class: article 2021/02/12 v1.4n Standard LaTeX document class
(c:/Aplicaciones/TeXLive/texmf-dist/tex/latex/base/size10.clo
size10.clo 2021/02/12 v1.4n Standard LaTeX file (size option)
)
\c@part=\count182
\c@section=\count183
\c@subsection=\count184
\c@subsubsection=\count185
\c@paragraph=\count186
\c@subparagraph=\count187
\c@figure=\count188
\c@table=\count189
\abovecaptionskip=\skip47
\belowcaptionskip=\skip48
\bibindent=\dimen138
)
(c:/Aplicaciones/TeXLive/texmf-dist/tex/latex/l3backend/l3backend-pdftex.def
[...]

PR #661, opened 2021-09-07
fix for #658

FrankMittelbach:
# Internal housekeeping

## Status of pull request

This only implements the uc/lc handling because the nobreak in OT1 seems to be in the right place (it is a deficiency, but we know that OT1 is deficient) and it is better to prevent a break in a place where one would be allowed to allowing a break where it is definitely not right (e.g., between ij and a following punctuation). after all one can always add a `\allowbreak`but you can't undo an allowed break.

- Ready to merge

- [x] Test file(s) added

### DIFF
--- a/base/doc/ltnews34.tex
+++ b/base/doc/ltnews34.tex
@@ -541,6 +541,31 @@ will now correctly apply \cs{TrimSpaces} to both arguments.
 
 
 
+\subsection{Correct case changes for \cs{ij} and \cs{IJ}}
+
+
+The the ligatures \enquote{\ij} and \enquote{\IJ}, as used in Dutch,
+are only available in most \TeX{} fonts, if the commands \cs{ij} or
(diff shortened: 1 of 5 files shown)


car222222:
The fix looks good!

base/doc/ltnews34.tex
```diff
@@ -541,6 +541,31 @@ \subsection{Replicate argument processors for all embellishments}
 
 
 
+\subsection{Correct case changes for \cs{ij} and \cs{IJ}}
+
+
+The the ligatures \enquote{\ij} and \enquote{\IJ}, as used in Dutch,
```
muzimuzhi, line 547: typo: The <s>the</s> ligatures ...

mrpiggi:
The fix is actually for #658 not #568

car222222:
Corrected, in the title at least.  Thanks.

PR #671, opened 2021-09-18
Parhook skip

davidcarlisle:

# Internal housekeeping

## Status of pull request

- Ready to merge

- [X] Test file(s) updated

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,6 +6,12 @@ completeness or accuracy and it contains some references to files that
 are not part of the distribution.
 ================================================================================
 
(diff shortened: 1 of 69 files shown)


car222222 (changes requested):
I think I already reported this somewhere?

The documentation of when the guard skip is added is the opposite of what actually happens, here:

. . . only if not we add our skip, the guard skip

For now, I suggest changing this to:

. . . and if so we add a zero-length guard skip, which

FrankMittelbach (approved):
modulo suggested textual changes sugested by Chris

Issue #676, opened 2021-09-27
Reserve some mathgroups (\fam) for reuse on a per formula basis


FrankMittelbach:
## Brief outline of the enhancement

In pdftex  there are a maximum of 16 math families available. Those get allocated to symbol fonts and (when used) to math alphabets. If you need more than 16 in a single formula you are out of luck, but if you need different math alphabets in different formulas one could improve the current situation by reserving a number of slots for "local" use and restore them if used so that in the next formula they could be used for other math alphabets.

## Minimal example showing the current behaviour

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{article}

\DeclareSymbolFont{fooi}{OT1}{cmr}{m}{n}   %mimic installation of symbol fonts
\DeclareSymbolFont{fooii}{OT1}{cmr}{m}{n}
\DeclareSymbolFont{fooiii}{OT1}{cmr}{m}{n}
\DeclareSymbolFont{fooiv}{OT1}{cmr}{m}{n}
\DeclareSymbolFont{foov}{OT1}{cmr}{m}{n}
\DeclareSymbolFont{foovi}{OT1}{cmr}{m}{n}
\DeclareSymbolFont{foovii}{OT1}{cmr}{m}{n}
\DeclareSymbolFont{fooviii}{OT1}{cmr}{m}{n}
\DeclareSymbolFont{fooix}{OT1}{cmr}{m}{n} 
\DeclareSymbolFont{foox}{OT1}{cmr}{m}{n} % only two slots open for alphabets

\DeclareMathAlphabet\baz{OT1}{cmss}{m}{n}


\begin{document}

$\mathtt{T} < \textbf{X} = \baz{A} $      % now the last slot is taken

$\mathsf{B} $ % this now fails because we used \baz, \mathbf and \mathtt before
  
\end{document}
```

## Minimal example showing the desired new behaviour

Provide a counter "localmathalphabets" (or some better name) that reserves that many slots for local math alphabets. If they are used in a formula, undo their allocation after the formula has finished. Suggested default: 2 which seems a good compromise between flexibility and a slight degration of speed if you operate on the available limit. No speed loss if you are below the limit.

If the slots are allocated to symbol fonts then they are gone for good, e.g., adding another symbol font above would mean only 1 slot is free for reuse even if the default is higher.

PR #677, opened 2021-09-27
Solution for #676

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [x] Test file(s) added


base/ltfssdcl.dtx
```diff
@@ -454,25 +454,82 @@
...
+%
+%
+%  \begin{macro}{\c@localalphabets}
+% \changes{v3.0x}{2021/09/26}{Counter added for (gh/676)}
+%    To avoid hitting the ``no more math fams available'' limit of 16, we
+%    keep a defined number of math alphabets flexible/local. If we have to
+%    allocate any of those we roll back the allocation after the
+%    formula has ended, so the next formula can use other alphabets in
+%    the slot(s). This makes the processing a bit slower if you are
+%    working at the limit, but that is better than dying with out of memory.
```
car222222, line 467: Maybe add quotes:  ``out of memory‘‘

base/ltfssdcl.dtx
```diff
@@ -507,8 +569,138 @@
  \expandafter#1\ifx\math@bgroup\bgroup{#4}\else#4\fi
  }%
 }
+%    \end{macrocode}
+%
+%
+
+%  \begin{macro}{\freeze@math@version}
+% \changes{v3.0x}{2021/09/26}{Macro added for (gh/676)}
+%    This command stores the current state of the math version and set
```
car222222, line 578: set ==>  sets

base/ltfssdcl.dtx
```diff
@@ -507,8 +569,138 @@
...
+\cs_new:Npn\freeze@math@version #1 {
+%    \end{macrocode}
+%    Save the current \cs{mv@\meta{version}} code and the number of
+%    allocated mathgroups inside.
+%    \begin{macrocode}
+  \@font@info{Freeze~ math~ alphabet~ allocation~ in~ version~ #1}
+  \cs_gset_eq:cc { mv@#1@frozen }{ mv@#1 }
+  \tl_gset:cx { g__mv_frozen_ #1 _tl }{ \int_use:c { c@mv@#1 } }
+%    \end{macrocode}
+%    To do the reset the first time, we have wait until we are out of
```
car222222, line 592: ==>  Doing the reset for the first time, we must wait . . .

car222222 (commented):
Hope the code is OK too!

FrankMittelbach:
> Hope the code is OK too!

well it compiles the test suite and its own test files and works across TLC3 so can't be totally wrong :-)

car222222:
Hooray!

base/ltfssdcl.dtx
```diff
@@ -507,8 +569,138 @@
...
+%    \begin{macrocode}
+\ExplSyntaxOn
+\cs_new:Npn\freeze@math@version #1 {
+%    \end{macrocode}
+%    Save the current \cs{mv@\meta{version}} code and the number of
+%    allocated mathgroups inside.
+%    \begin{macrocode}
+  \@font@info{Freeze~ math~ alphabet~ allocation~ in~ version~ #1}
+  \cs_gset_eq:cc { mv@#1@frozen }{ mv@#1 }
+  \tl_gset:cx { g__mv_frozen_ #1 _tl }{ \int_use:c { c@mv@#1 } }
```
PhelypeOleinik, line 590: I think we should replace `__mv` by `__latexe`, or something like that, for “general” (not in a module of itself) `expl3` code in LaTeX2e, so we don't end up with hundreds of prefixes taken up by small pieces of code.

base/ltfssdcl.dtx
```diff
@@ -507,8 +569,138 @@
...
+%    \end{macrocode}
+%    Doing the reset the first time, we wait until we are out of
+%    math mode, so we use some recursive \cs{aftergroup} for this
+%    before we execute \cs{mv@\meta{version}@reset}.
+%    \begin{macrocode}
+  \aftergroup \__mv_init_freeze:N \exp_after:wN
```
PhelypeOleinik, line 596: `\group_insert_after:N` instead of `\aftergroup`

base/ltfssdcl.dtx
```diff
@@ -507,8 +569,138 @@
...
+%
+
+%  \begin{macro}{\freeze@math@version}
+% \changes{v3.0x}{2021/09/26}{Macro added for (gh/676)}
+%    This command stores the current state of the math version and sets
+%    things up to return to it after each formula from now on. We use
+%    L3 programming layer code to set it up.
+%    \begin{macrocode}
+\ExplSyntaxOn
+\cs_new:Npn\freeze@math@version #1 {
```
PhelypeOleinik, line 583: Maybe `\cs_new_protected:Npn`, to follow guidelines

PhelypeOleinik (changes requested):

car222222:
Useful, as ever, ideas from Phelype.

base/ltfssdcl.dtx
FrankMittelbach: oversight, thought it was still only `:D`

base/ltfssdcl.dtx
FrankMittelbach: I don't think  this can ever need protection, can you?  So technically I don't think it matters, though it doesn't hurt either I guess. Do we know if `\protect` introduces as (tiny) time penalty or not?

base/ltfssdcl.dtx
FrankMittelbach: I think you are right that "mv" is a bad module prefix, but in my opinion so is "latexe", just assume the whole kernel would be L3 based: if then each and every macro starts off with   `__latexe` that would be very unhelpful imo. Maybe the right prefix is "fss" (font selection scheme) or "nfss"  here, assuming that over time other parts of NFSS become L3 based.

base/ltfssdcl.dtx
Skillmon: There is a (very small) time penalty indeed. My benchmarks indicate typically around 6% slow down for a protected `\prg_do_nothing:` in pdfTeX on Linux.

protected: 2.6e-8 seconds (0.0829 ops)
normal: 2.59e-8 seconds (0.0765 ops)

Skillmon: If instead of `\prg_do_nothing:` we benchmark a single assignment `\def\mytmp{}` stored inside a protected and unprotected macro the difference becomes negligible:

protected:
9.47e-8 seconds (0.3 ops)
9.54e-8 seconds (0.305 ops)
9.48e-8 seconds (0.302 ops)

unprotected:
9.32e-8 seconds (0.298 ops)
9.47e-8 seconds (0.3 ops)
9.51e-8 seconds (0.302 ops)

FrankMittelbach: interesting here is the execution of a protected macro versus the execution of it unprotected, but in this particular case it doesn't really matter since that command is called at most once per mathversion in a document so ... :-) and not once per formula

base/ltfssdcl.dtx
FrankMittelbach: used nfss as this is a well known abeit not so "new" concept

PhelypeOleinik: Yeah, `__latexe` would not be great either, but for "one offs" it would be less worse than registering a new prefix for just two or three macros. But `__nfss` is much better indeed, if we end up needing more of those in the future.

PhelypeOleinik (approved):

base/ltfssdcl.dtx
Skillmon: I tend to do calculations on the ops, not the time. The time is in my experience far more inconsistent than the ops measured by `l3benchmark`. And I got the 6% pretty consistently for a macro that just expands to nothing.

PhelypeOleinik: Oh, the ops...  I don't know exactly what counts as an “operation” here, but I can see why that can be more stable than the raw time.  But in any case, the extra time it takes for `\protected` _vs._ otherwise makes practically zero difference, so I still favour following `l3styleguide`:

> The expandability of each function should be well-defined. Functions which cannot be fully expanded must be `protected`.

car222222: This is not at all the best place to discuss this, but )for a few reasons) I have never been happy with this sentence:
"Functions which cannot be fully expanded must be protected."

I am not sure that it is sufficiently precise; and it is definitely an over-simplified requirement.

PR #691, opened 2021-10-14
Gh687

FrankMittelbach:
# Internal housekeeping

## Status of pull request
- Ready to merge

- [x] Test file(s) added


FrankMittelbach:
> I supose we (in ltnews or somewhere?) should encourage class and package authors to do the same before it can be safely assumed that `\@currentcounter` is always set to match the current label.

ltnews might be best. Technically if it missing then hyperref will end up linking to the out entity so it is as good or bad as before I guess, but it would be good to say something about it. Care to to write a short para about it before I merge?

PR #707, opened 2021-11-15
Doc cleanup

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted (maybe something else needs dropping)
- Ready to merge

- [n/a] Test file(s) added

I would love to autogenerate manifest.md but the current result from `l3build manifest` is not yet really suitable imo.


u-fischer (approved):

josephwright:
> I would love to autogenerate manifest.md but the current result from `l3build manifest` is not yet really suitable imo.

Complain at @wspr  ;)

josephwright (approved):
Do we need the residual files?

FrankMittelbach:
> Do we need the residual files?

not really, but this way they aren't dropped totally while not ending up on ctan. I don't really mind keeping them on github but don't have strong feelings either way. It is just that my experience is that it is nice to keep a trace (and inside the gihub history they are essentially gone for good).

FrankMittelbach:
> > I would love to autogenerate manifest.md but the current result from `l3build manifest` is not yet really suitable imo.
> 
> Complain at @wspr ;)

less a question of complaints :-) more a question of, should we invest in that? We do have to keep a manifest and right now it is likely to be incomplete or out of date (as it is this time around). If we could auto-generate it is more likely to be correct.

FrankMittelbach:
@PhelypeOleinik could you have a look at the .md file? I think all of that is taken care of in which case this file could definitely vanish.

davidcarlisle (approved):

zauguin (commented):
> It is just that my experience is that it is nice to keep a trace (and inside the gihub history they are essentially gone for good).

I'm in favor of dropping these files completely. Especially with titles like "LaTeX installation TeX expert information" they are way too likely to be found by accident when someone is searching for keywords, and then they probably don't even look at some directory name. (That's mostly referring to `texpert.tex` and `tex2.tex`, IMO it makes sense to keep `latexchanges.tex` around for a while)

Beside that: `texpert.txt` and `tex2.txt` are still referenced in the README.md, we should probably drop them there too. `latexchanges.tex` should be removed from `.github/pull_request_template.md`.

PhelypeOleinik (approved):
In the beginning of `base/README.md`, it says:
> In addition to this bundle, a minimal LaTeX system also needs the files contained in the [list of required packages]

but it doesn't mention `l3kernel`.  I think we should add:

 - L3 programming layer (`l3kernel`)
 - L3 backend interfaces (`l3backend`)

and under `required`:

 - First aid for external packages (`firstaid`)

Other than that, changes seem fine.

zauguin (approved):

muzimuzhi:
> In the beginning of `base/README.md`, it says:
> 
> > In addition to this bundle, a minimal LaTeX system also needs the files contained in the [list of required packages]
> 
> but it doesn't mention `l3kernel`. I think we should add:
> 
> * L3 programming layer (`l3kernel`)
> * L3 backend interfaces (`l3backend`)
> 
> and under `required`:
> 
> * First aid for external packages (`firstaid`)
> 
> Other than that, changes seem fine.

Maybe also `unicode-data` (for unicode engines) and `l3package` (for `xparse`)?

PhelypeOleinik:
> Maybe also `unicode-data` (for unicode engines) and `l3package` (for `xparse`)?

Maybe `unicode-data`.  `l3packages` is not a dependency (anymore) because the core `xparse` is now part of the LaTeX2e kernel.

zauguin:
`unicode-data` is required (for all engines) by l3kernel, so I think it should be listed somewhere. I agree that it would make sense to add it here, but I think it should also be documented as a requirement in the latex3 repo since it's also needed when expl3 is loaded with other formats.

muzimuzhi:
> > Maybe also `unicode-data` (for unicode engines) and `l3package` (for `xparse`)?
> 
> Maybe `unicode-data`. `l3packages` is not a dependency (anymore) because the core `xparse` is now part of the LaTeX2e kernel.

But `xparse` is stored under `./texmf/tex/latex` repo path, while the mentioned `README.md` is in `./base` (hence readers may think it only has responsible for that subdir?)

FrankMittelbach:
> > 
> 
> But `xparse` is stored under `./texmf/tex/latex` repo path, while the mentioned `README.md` is in `./base` (hence readers may think it only has responsible for that subdir?)

what remains as xparse.sty is *not* required or used by the kernel, it is just another package. so I don't think this is something to list.

muzimuzhi:
> > > 
> > 
> > 
> > But `xparse` is stored under `./texmf/tex/latex` repo path, while the mentioned `README.md` is in `./base` (hence readers may think it only has responsible for that subdir?)
> 
> what remains as xparse.sty is _not_ required or used by the kernel, it is just another package. so I don't think this is something to list.

just found [`./base/ltcmd.dtx`](https://github.com/latex3/latex2e/blob/develop/base/ltcmd.dtx), sorry my bad

josephwright:
> `unicode-data` is required (for all engines) by l3kernel, so I think it should be listed somewhere. I agree that it would make sense to add it here, but I think it should also be documented as a requirement in the latex3 repo since it's also needed when expl3 is loaded with other formats.

Shouldn't that therefore be down as a dependency for `l3kernel` not for LaTeX2e? One normally lists _direct_ dependencies, as otherwise there is the potential to have conflicting information.

zauguin:
> Shouldn't that therefore be down as a dependency for `l3kernel` not for LaTeX2e? One normally lists _direct_ dependencies, as otherwise there is the potential to have conflicting information.

Normally I agree and that's why I suggested that it should be documented for `l3kernel`. But since it's (if I'm not mistaken) the only indirect dependency we currently have, I think that it would be useful to have on complete list here or at least to link to a list of requirements for the other packages. Especially since the current text talks about needing "the files of" `l3kernel` and not needing to install `l3kernel` and all it's dependencies.

I don't have any strong opinions about this though.

FrankMittelbach:
> I don't have any strong opinions about this though.

don't mind mentioning it but I think this is all a  storm in the water glass. nobody these days looks at such a readme to find out what he or she needs to install LaTeX, everybody uses a readymade installation, so all that is more something to drop rather than expand imho.

base/README.md
```diff
@@ -10,10 +10,15 @@ This bundle provides the core LaTeX kernel. In addition to this bundle,
 a minimal LaTeX system also needs the files contained in the
 
 - LaTeX team documentation (`doc`)
+- L3 programming layer (`l3kernel`)
+- L3 backend interfaces (`l3backend`)
+- L3 unicode support (`unicode-data`)
```
josephwright, line 15: Well it's 'core Unicode support' (used by plain XeTeX/LuaTeX), but I'll let you off ;)

FrankMittelbach:
Am 17.11.21 um 17:07 schrieb Joseph Wright:
> +- L3 unicode support (`unicode-data`)
> 
> Well it's 'core Unicode support' (used by plain XeTeX/LuaTeX), but I'll 
> let you off ;)
> 

you're invited to push a change :-)

Issue #712, opened 2021-11-27
Optional argument is not passed correctly to the end part


eg9:
## Brief outline of the bug

When trying `\NewDocumentEnvironment{O{x}}{#1}{#1}` we get `-NoValue-` in the end part.

This seems to be a regression, because compiling with TeX Live 2020 produces the expected result.

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}
\documentclass{article}

\NewDocumentEnvironment{test}{O{x}}{#1}{#1}

\begin{document}

\begin{test}
text
\end{test}

\end{document}
```

## Log file (required) and possibly PDF file  

```
This is pdfTeX, Version 3.141592653-2.6-1.40.23 (TeX Live 2022/dev) (preloaded format=pdflatex 2021.11.23)  27 NOV 2021 16:49
entering extended mode
 restricted \write18 enabled.
 %&-line parsing enabled.
**testopt
(./testopt.tex
LaTeX2e <2021-11-15>
L3 programming layer <2021-11-22>
(/usr/local/texlive/2021/texmf-dist/tex/latex/latexbug/latexbug.sty
Package: latexbug 2021/09/25 v1.0l Bug-classification
)
(/usr/local/texlive/2021/texmf-dist/tex/latex/base/article.cls
Document Class: article 2021/10/04 v1.4n Standard LaTeX document class
(/usr/local/texlive/2021/texmf-dist/tex/latex/base/size10.clo
size10.clo 2021/10/04 v1.4n Standard LaTeX file (size option)
)
\c@part=\count183
\c@section=\count184
\c@subsection=\count185
\c@subsubsection=\count186
\c@paragraph=\count187
\c@subparagraph=\count188
\c@figure=\count189
\c@table=\count190
\abovecaptionskip=\skip47
\belowcaptionskip=\skip48
\bibindent=\dimen138
)
(/usr/local/texlive/2021/texmf-dist/tex/latex/l3backend/l3backend-pdftex.def
l3backend-pdftex.def 2021-10-18 L3 backend support: PDF output (pdfTeX)
\l__color_backend_stack_int=\count191
\l__pdf_internal_box=\box50
)
No file testopt.aux.
\openout1 = `testopt.aux'.

LaTeX Font Info:    Checking defaults for OML/cmm/m/it on input line 6.
LaTeX Font Info:    ... okay on input line 6.
LaTeX Font Info:    Checking defaults for OMS/cmsy/m/n on input line 6.
[...]

PR #715, opened 2021-12-02
Insert correct defaults in end-of-environment code (see #712)

josephwright:
Fixes the bug: I imagine this one is going to be held for 2022-06-01 so the request is to go into dev.


eg9:
I'm believe this should be a hotfix. Code relying on optional arguments in the end part is currently broken.

PhelypeOleinik (approved):

FrankMittelbach (approved):
but I agree with Enrico, either firstaid or PL

eg9:
Here's a case:

https://tex.stackexchange.com/questions/624796/xparse-losing-default-values-of-optional-parameters-after-environment

PR #719, opened 2021-12-07
remove spurious spaces (xcolor issue 10)

u-fischer:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- Ready to merge

- [ ] Test file(s) added
- [x ] Version and date string updated in changed source files
- [x ] Relevant `\changes` entries in source included

### DIFF
--- a/required/graphics/changes.txt
+++ b/required/graphics/changes.txt
@@ -8,6 +8,10 @@ are not part of the distribution.
 All changes above are only part of the development branch for the next release.
 ================================================================================
 
+2021-12-07  Ulrike Fischer <Ulrike.Fischer@latex-project.org>
+
+	* color.dtx: Remove stray spaces after change for xcolor/gh/10}
+
 #########################
 # 2021-11-15 Release
 #########################
--- a/required/graphics/color.dtx
+++ b/required/graphics/color.dtx
@@ -21,7 +21,7 @@
 %<driver> \ProvidesFile{color.drv}
 % \fi
 %         \ProvidesFile{color.dtx}
-          [2021/10/11 v1.3b Standard LaTeX Color (DPC)]
+          [2021/12/07 v1.3c Standard LaTeX Color (DPC)]
 %
 % \iffalse
 %<*driver>
@@ -80,6 +80,8 @@
 %     {Save and restore catcodes}
 % \changes{v1.3a}{2021/09/27}
 %     {Use \cs{@backslashchar} in preference to \cs{string} gh/659}
+% \changes{v1.3c}{2021/12/07}
+%     {Remove stray spaces after change to \cs{@backslashchar} xcolor/gh/10}
 %
 %    \begin{macrocode}
 \edef\Gin@codes{%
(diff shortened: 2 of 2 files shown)


josephwright:
Needs to be in `changes.txt`

yihui:
I wonder if this PR could be merged asap. We have been waiting for this bug fix for a week. It has affected about 70 R packages on CRAN. Thank you very much!

davidcarlisle (approved):

FrankMittelbach:
it still needs a change log entry before it gets distributed

davidcarlisle:
@yihui  we'll do what we can although I must say I'm surprised it has a big impact: as far as I can tell this should only be broken for documents that load both color and xcolor which is never needed, simply loading xcolor should be enough.

yihui:
@davidcarlisle You were right---the broken R packages loaded both color and xcolor. The former was loaded through a template, and the latter was loaded by users. I haven't investigate why this happened yet, but I'll see if it's possible to replace `color` with `xcolor` in the template.

Thank you very much!

davidcarlisle:
@yihui  I just pushed the change from develop to main branch and tests seem to pass so we should be able to push this out shortly anyway but I couldn't help asking:-)

u-fischer:
> You were right---the broken R packages loaded both color and xcolor. 

Loading both is not really a problem, but loading color, defining a color and then loading xcolor will give you colors xcolor can not use and can lead to hard to spot problems. It is much safer if the template loads xcolor.

yihui:
> It is much safer if the template loads xcolor.

That's what I hope to do. My only concern is about possible option clashes, i.e., I `\usepackage{xcolor}` in the template but users load it again with options later in the document. I need to check these R packages before I make this change.

davidcarlisle:
@yihui  graphics bundle has been pushed to ctan, will take a couple of days to reach texlive normally

yihui:
Perfect. Many thanks again!

FrankMittelbach:
@yihui perhaps as a helpful hint for next time: errors happen even though we try to avoid them. But the upcoming release is always publicly available for about at least a month before it goes live and you can easily run your own test suite against it if you use pdflatex-dev or lualatex-dev on your system instead of the main pdflatex, lualate, xelatex. This way we can hopefully catch blunder like this one next time (if it is not caught by our own tests) before it is released as the main release. The more people do this the better the chances this works out perfectly.

yihui:
@FrankMittelbach Great suggestion! I'll add a daily test in my package to discover possible problems early. Thanks!

muzimuzhi:
Paying more attention to warnings may help as well, since

> loading color, defining a color and then loading xcolor

with older `color` package already gives warning

> Package xcolor Warning: Incompatible color definition on input line 6.

FrankMittelbach:
> @FrankMittelbach Great suggestion! I'll add a daily test in my package to discover possible problems early. Thanks!

small caveat: directly after a new release (especially after a patch release) the -dev release may be initially a bit behind, for example right now  main is 2011-11-15 PL1 but dev is still 2022-06-01 (which is equiv to 2011-11 PL0) because we haven't yet put out a new, just because we are in the middle of preparing a real dev release, so that hotfix is not yet in. But close to the end of the release cycle dev will be exactly what will show up in the next release (except for final fixes due to testing by you, us and others). Hope that description makes sense but you find more on the LaTeX project website.

PR #720, opened 2021-12-08
Document \DeclareEncodingSubset

dr-scsi:
As requested by FMi.

### DIFF
--- a/base/doc/fntguide.tex
+++ b/base/doc/fntguide.tex
@@ -40,9 +40,9 @@
 \author{\copyright~Copyright 1995--2021, \LaTeX\ Project
   Team.\thanks{Thanks to Arash Esbati for documenting the
     newer NFSS features of 2020}\\
-   All rights reserved.}
+  All rights reserved.}
 
-\date{March 2021}
+\date{December 2021}
 
 \begin{document}
 
@@ -2056,7 +2056,7 @@ Table~\vref{tab:textcomp} contains macros to access text symbols.  Again,
 the numbers are the slots in the encoding.
 
 \begin{table}
-\centering\footnotesize
+  \centering\footnotesize
   \begin{tabular}[t]{@{}lp{1.5em}l@{}}
     \verb|\textcapitalcompwordmark|  & \textcapitalcompwordmark  & 23 \\
     \verb|\textascendercompwordmark| & \textascendercompwordmark & 31 \\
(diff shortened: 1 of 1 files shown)


base/doc/fntguide.tex
```diff
@@ -2170,8 +2170,259 @@ \section{Additional text symbols -- \textsf{textcomp}}
...
+  \label{tab:ts1-subset-always}
+\end{table}
+
+\begin{table}[tbp]
+  \centering
+  \begin{tabular}[t]{@{}lll@{}}
+    \verb*|\textcircled{}|       & \textcircled{}          & acc
+  \end{tabular}
+  \caption{Symbol defined in \texttt{TS1} sub-encoding \texttt{1}}
+  \label{tab:ts1-subset-1}
```
FrankMittelbach, line 2256: The subset number is a bit strange (I know I'm to blame) but saying that something is defined in sub-encoding X is the wrong way around. \textcircled is only in 0 (at level 1 it is undefined). So the caption should probably say

   Symbol only available in sub-encoding 0

or

  Symbol not available in sub-encodings 1--9

base/doc/fntguide.tex
```diff
@@ -2170,8 +2170,259 @@ \section{Additional text symbols -- \textsf{textcomp}}
...
+which we declare the subset and finally the \m{subset number} between
+|0| (all of the encoding is supported) and |9| (many glyphs are
+missing).  Hence, it is assumed that some symbols are always available
+by all fonts and each sub-encoding defines macros which become
+unavailable (i.e., they are not provided in the subencoding with that
+number and all sub-encodings with higher numbers.)
+
+Thus the symbols that are available in sub-encoding $x$ are the symbols
+in table~\ref{tab:ts1-subset-always} (always available) and the symbols
+list in sub-encodings $x + 1$ and higher.  The tables
```
FrankMittelbach, line 2204: that's the wrong way around \textflorin has encoding 6 which means it is available in 0-5

base/doc/fntguide.tex
```diff
@@ -2170,8 +2170,259 @@ \section{Additional text symbols -- \textsf{textcomp}}
...
+    \verb|\textasciidieresis|    & \textasciidieresis      & 168 \\
+    \verb|\textcopyleft|         & \textcopyleft           & 171 \\
+    \verb|\textcircledP|         & \textcircledP           & 173 \\
+    \verb|\textasciimacron|      & \textasciimacron        & 175 \\
+    \verb|\textasciiacute|       & \textasciiacute         & 180 \\
+    \verb|\textreferencemark|    & \textreferencemark      & 184 \\
+    \verb|\textsurd|             & \textsurd               & 187
+  \end{tabular}
+  \caption{Symbols defined in \texttt{TS1} sub-encoding \texttt{2}}
+  \label{tab:ts1-subset-2}
```
FrankMittelbach, line 2331: same here and below: these are the symbols only in 0--1 but not in 2--9

FrankMittelbach (changes requested):
looks good overall; please also update the doc date at the top and take a look at my comments.

base/doc/fntguide.tex
dr-scsi: This text is actually a verbatim copy from [lttextcomp.dtx](https://github.com/latex3/latex2e/blob/c66a033aaa6efdc05203580ee7e395061d6b9ab1/base/lttextcomp.dtx#L478).  Did I get it wrong?

dr-scsi:
Thanks for your comments.  I admit that I had a hard time understanding the subset logic.  I will try to come up with a new suggestion tomorrow.  Meanwhile, please have a look at my comment above.  Maybe that is the source of my confusion.

base/doc/fntguide.tex
FrankMittelbach: @dr-scsi maybe more that I got it wrong in that file. The paragraph above the one you refer to says the opposite and that one is correct. I think I need to look at that text there more closely.

FrankMittelbach: @dr-scsi take a look at what it nows says in lttextcomp if that is more understandable

dr-scsi:
@FrankMittelbach - Thanks, I've updated my PR, please have a look now.  I changed the the table captions (acc. to your change in `lttextcomp.dtx` and re-phrased some other blocks.  I hope we have it together now.

FrankMittelbach (approved):

PR #722, opened 2021-12-09
Make a wide table narrower

dr-scsi:
* base/doc/fntguide.tex (subsection{Text font attributes}): Narrow
a wide table with a footnote.


FrankMittelbach:
I see your point but I'm not convinced it is an improvement to not immediately see to what it resolves. I'll see if it can be handled differently

dr-scsi:
Fine with me, I'm closing this PR.

Issue #723, opened 2021-12-10
Support for Croatian Unicode digraph letters within (at least) T1


ivankokan:
## Brief outline of the enhancement

It started here: https://tex.stackexchange.com/questions/625565/which-mappings-are-used-if-glyphtounicode-pdfgentounicode-are-not-used.

Croatian Latin alphabet (aka Gaj's Latin alphabet) contains three digraph letters, here listed in all Unicode variants:
```
Ǆ ǅ ǆ Ǉ ǈ ǉ Ǌ ǋ ǌ
```

**Truly the letters, not ligatures.**

Historically speaking, they are commonly typed as two consecutive characters they "consist of" (since Croatian Latin keyboard layout does not have special slots for them, compared to the Serbian Cyrillic keyboard layout which does have their Cyrillic equivalents which are never typed nor typeset as two consecutive characters) but this has few shortcomings.

Let me present by comparing `injekcija` (injection) and `iǌe` (frost):
- there are nine letters within `injekcija`, separate for `n` and `j`; but there are three letters within `iǌe`
- hyphenation: as mentioned before, we have three digraph **letters**, hence any word having e.g. `ǌ` must not be hyphenated in between those two glyphs.

Strictly speaking, these digraph letters must not be typeset as two consecutive characters - but even if they are, hyphenation must be suppressed. No special kerning is required. It seems that having the commented lines within the MWE in the distro would suffice.

Since T1 is commonly used font encoding when typesetting Croatian, it would be great to have support within the `t1enc.dfu` for the above-listed nine variants (just as for the other special Unicode cases listed there). Currently, using them yields an error.

I am not sure whether the proposal should be considered for other encodings as well. Why not, actually? Based on:
https://codepoints.net/U+01C4
https://codepoints.net/U+01C5
https://codepoints.net/U+01C6
https://codepoints.net/U+01C7
https://codepoints.net/U+01C8
https://codepoints.net/U+01C9
https://codepoints.net/U+01CA
https://codepoints.net/U+01CB
https://codepoints.net/U+01CC
...it seems that these Unicode code points are South-Slavic-specifics only.

## Minimal example showing the current behaviour

Does not compile.

```latex
\RequirePackage{latexbug}
\documentclass{article}

\usepackage[T1]{fontenc}
\usepackage[utf8]{inputenc}

% \DeclareUnicodeCharacter{01C4}{D\nobreak Ž}   % Ǆ
[...]

PR #724, opened 2021-12-10
extend UTF-8 support for issue #723

davidcarlisle:
Extend UTF-8 support as requested in issue #723

# Internal housekeeping

## Status of pull request

- Ready to merge

- [X] Test file(s) added


base/utf8ienc.dtx
```diff
@@ -1368,6 +1368,18 @@
 %<all,t1,ly1>\DeclareUnicodeCharacter{017E}{\v z}
 %<all,ts1,ly1>\DeclareUnicodeCharacter{0192}{\textflorin}
 %    \end{macrocode}
+% \changes{v1.2n}{2021/12/10}{add 01C4-01CC}
+%    \begin{macrocode}
+%<all,t1>\DeclareUnicodeCharacter{01C4}{D\v Z}
+%<all,t1>\DeclareUnicodeCharacter{01C5}{D\v z}
+%<all,t1>\DeclareUnicodeCharacter{01C6}{d\v z}
```
FrankMittelbach, line 1375: I kow that are are still people using LY1 and all such definitions would work for them too wouldn't they? so perhaps we should always set all,T1,LY1 if that is possible. (might also apply to other chars we added recently).

FrankMittelbach (approved):
one comment for consideration

base/utf8ienc.dtx
davidcarlisle: @FrankMittelbach  and OT1 but actually it doesn't make any difference these days as we pre-load OT1 and T1 into the format so any UTF-8 declaration in T1 is available whatever encodings you use.

FrankMittelbach: true, so forget my remark (which you probably already did :-))

PR #729, opened 2021-12-13
Metadata

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [x] Test file(s) added
- [n/a] Relevant `\changes` entries in source included -- is new


base/ltmeta.dtx
```diff
@@ -0,0 +1,189 @@
...
+%    tough.
+% 
+%    If the file can't be found we say so and carry on without it.
+%    \begin{macrocode}
+     {%
+       \@latex@warning@no@line{No support files for
+         \noexpand\DocumentMetadata found.\MessageBreak
+         Is the 'documentmetadata-support' bundle installed?%
+         \MessageBreak\MessageBreak
+         The declaration is ignored}%
```
zauguin, line 145: Not sure about this, but I think I would prefer an error here. `\DocumentMetadata`, especially since it loads PDF managment and other functionality, might easily change the behavior of the document and therefore ignoring it should not be hidden.

base/ltmeta.dtx
```diff
@@ -0,0 +1,189 @@
...
+%\ExplSyntaxOn
+%    \end{macrocode}
+%
+%    \begin{macrocode}
+%<latexrelease>\NewModuleRelease{2021/06/01}{ltmeta}
+%<latexrelease>                 {Document Metadata handling}
+%    \end{macrocode}
+%
+%    \begin{macrocode}
+\newcommand\DocumentMetadata{%
```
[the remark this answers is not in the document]
josephwright: If we are doing that, please make it protected!

base/ltmeta.dtx
FrankMittelbach: you are probably right. Changed.

base/ltmeta.dtx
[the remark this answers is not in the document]
FrankMittelbach: who whole setup is tmp which is why I didn't bother, but I now changed to `\protected\def`

eg9:
On line 169 of ltnews35.tex, `why` should be `while`.

zauguin (approved):

josephwright (approved):
Looks good to me

PhelypeOleinik (approved):

.github/tl_packages
```diff
@@ -38,6 +38,7 @@ varwidth
 #
 # special testing for latex-lab
 pdfmanagement-testphase
+l3experimental
```
josephwright, line 41: If we need this, shouldn't it be done manually using `texmf`?

FrankMittelbach: maybe, but on the other hand I think it is perfectly fine if this is not a "frozen" version but the one that is really on ctan because that is how it is used by others. So In my opinion this is fine.
It is not needed by the code in latex-lab only in Ulrike's code which is called in the test file.

FrankMittelbach:
another round of reviews please, from my perspective that is ready for now. Of course the lab code will change between now and then but that can and should be done after the first -dev release.

required/latex-lab/README.md
```diff
@@ -0,0 +1,34 @@
+# LaTeX laboratory
+
+This bundle holds optional files that are loaded in certain situations
+by kernel code (if available). For example, the new (as of 2021/12)
+`\DocumentMeta` command in the kernel loads a file from here holding
+the real payload. While this code is still in development and the use
+is experimental and mainly for the tagging project) the code is stored
```
mrpiggi, line 7: opening parentheses missing

davidcarlisle (approved):

required/latex-lab/README.md
```diff
@@ -0,0 +1,34 @@
+# LaTeX laboratory
+
+This bundle holds optional files that are loaded in certain situations
+by kernel code (if available). For example, the new (as of 2021/12)
+`\DocumentMeta` command in the kernel loads a file from here holding
```
PhelypeOleinik, line 5: `\DocumentMetadata`

required/latex-lab/documentmetadata-support.dtx
```diff
@@ -0,0 +1,376 @@
...
+%         More information can be found in the documentation
+%         of \pkg{l3pdfmeta}.
+%
+%       \item[\texttt{colorprofiles}] This allows to load icc-colorprofiles. Details
+%       are described in the documentation of \pkg{l3pdfmeta}.
+%
+%       \item[\texttt{pdfmanagement}] Boolean. This activates/deactivates
+%         the core management code. By default the value is true.
+%
+%       \item[\texttt{firstaidoff}] This accepts a comma lists of keysword and disable the patches
```
PhelypeOleinik, line 73: keywords and disables

required/latex-lab/documentmetadata-support.dtx
```diff
@@ -0,0 +1,376 @@
...
+%       related to them. More information can be found in the documentation of
+%       \pkg{pdfmanagement-firstaid}.
+%
+%       \item[\texttt{testphase}] This key is used to load testphase code. The values it accepts
+%       and their effect will change over time, when testphase packages are added or
+%       removed when the code is moved into the kernel. Currently the accepted values are
+%       \texttt{tagpdf}, this load the tagpdf package, \texttt{headings}, this loads
```
PhelypeOleinik, line 80: this loads

required/latex-lab/documentmetadata-support.dtx
```diff
@@ -0,0 +1,376 @@
...
+%    key and then force the loading by either loading l3pdf (+ expl3)
+%    or only expl3.  The backend contains now management command, so
+%    the boolean should true first.
+%    \begin{macrocode}
+        \keys_set_groups:nnn { document / metadata} {init}{ #1 }
+        %should be loaded after the backend is set, and only if not in the kernel
+        \bool_gset_true:N \g__pdfmanagement_active_bool
+        \cs_if_free:NTF \pdf_uncompress:
+           {\RequirePackage{l3pdf}}
+%FMi why load expl3? is there a better way to force the backend loading?
```
PhelypeOleinik, line 140: ` \sys_load_backend:n { <backend> }` should work as long as we know what `<backend>` is at this point.

PhelypeOleinik (approved):

required/latex-lab/documentmetadata-support.dtx
josephwright: `\sys_load_backend:n{}` will go with the default one.

base/doc/source2e.tex
```diff
@@ -264,6 +264,8 @@
 
  \DocInclude{ltpara}   % Paragraph hooks (L3 module)
 
+ \DocInclude{ltmeta}   % Document Metadara interface (L3 module)
```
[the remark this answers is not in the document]
FrankMittelbach: with this we are at Y with ltfinal, but that is not far off. I'll add an issue so that this doesn't get forgotten.

required/latex-lab/documentmetadata-support.dtx
u-fischer: I will remove the test and loading of `l3pdf` from ltdocinit, imho it is now long enough in the kernel.  

Instead of loading expl3 one could imho do now something like this
~~~~
\sys_load_backend:n { <value of DocumentMetadata backend key>}
\str_if_exist:NF \c_sys_backend_str
  { \sys_load_backend:n { } }
~~~~

Currently the backend key simply pass its value to the backend key of expl3, and so is automatically in sync and accept the same values. So if we change this one will have to adapt the value in both if needed.

FrankMittelbach: well, could you please to the change (now or at a later stage)? This code is only ever used if the format contains ltmeta which means it does have expl3 inside so we should never do the package loading (which gives option clashes if loaded later again to set options).

u-fischer: I pushed the change, I will have to adapt some of the tests, they fail as a luatex font number has changed, but this is unrelated.

FrankMittelbach: @u-fischer I had hoped you push the change here :-( ... ok reverse engineer time then

PR #731, opened 2021-12-13
Gh730

davidcarlisle:
## Status of pull request

- Feedback wanted 
- Ready to merge

- [X] Test file(s) added


base/testfiles-legacy/github-0730.lvt
```diff
@@ -0,0 +1,12 @@
...
+\input{test2e}
+\DeclareUnicodeCharacter{009C}{\fbox{U+009C}}
+
+
+\begin{document}
+
+\START
+% This should show the fbox (and not error)
+\setbox0\hbox{U+009C:}\showbox0
+\END
```
FrankMittelbach, line 12: I would be very surprised if this shows and error looking at it :-) but I guess the file actually contains the offending invisible characters. Difficult to tell in the github interface

FrankMittelbach: maybe be it would be better to show the character actually as the two bytes it is represented with (`^^c2^^9c`), because even looking with an editor at this file in the future will result in some head scratching I guess

FrankMittelbach (approved):
looks all fine to me (but see my comment on the test file)

PR #733, opened 2021-12-17
Only build released components for non dev-/release- tags

zauguin:

Let GitHub Actions only build and release a single module if the tag name does not start with `release-` or `dev-`.

This makes the release build for individual "required" packages faster and avoids unrelated issues and avoids attaching many unneeded files to the GitHub releases, but it also makes the latest released code a bit harder to find since the shown last release might often be a partial release then.

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- n/a Test file(s) added
- n/a Version and date string updated in changed source files
- n/a Relevant `\changes` entries in source included
- n/a Relevant `changes.txt` updated
- n/a `ltnewsX.tex` (and/or `latexchanges.tex`) updated

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -39,14 +39,35 @@ jobs:
           restore-keys: texlive-v0-
       # We need Ghostscript for dvips and XeTeX tests.
       - run: sudo apt-get update && sudo apt-get install ghostscript
+      # Determine the name of the tag we are working on. Sadly GH Actions only provides
+      # github.ref which has the name prefixed with github.ref and no way to remove a prefix
+      # in normal expressions, so we need an action for that.
+      - name: Get name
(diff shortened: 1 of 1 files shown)


FrankMittelbach:
> but it also makes the latest released code a bit harder to find since the shown last release might often be a partial release then.

would it be possible that such partial releases come out somewhat like this in the release list?
```
graphics-2021-12-14   (component only)
```
That should make it clear that it is not the full release  even though the "latest" label directs you to it

zauguin:
It's possible, but if we change the release name anyway, do we even want to include the literal tag name or use some more user friendly identifier? E.g. rewriting "graphics package version 2021-12-04" should also make it obvious that this is only one package.

FrankMittelbach:
one could be, I have no strong opinion on it but the word "package" may not be a good choice as in the Tex world that is really just a "package" while here it is the "bundle" consisting of many packages.

PR #754, opened 2022-01-16
Fix typo

kesslermaximilian:
# Internal housekeeping

## Status of pull request

- Ready to merge

I know that it says the merging / updating process is complicated and this is not intended for submitting regular changes,
but since it definitely doesn't affect behaviour / needs a version change / changelog ...  I'm submitting this anyways.

If this is not the right place: Where might I submit typos like these in the future (in case I find more?)

### DIFF
--- a/base/lthooks.dtx
+++ b/base/lthooks.dtx
@@ -1404,7 +1404,7 @@
 %    hook management (by adding various rules to ensure the
 %    appropriate ordering with respect to other code added by
 %    packages). However, this makes every document unnecessary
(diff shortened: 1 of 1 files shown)


FrankMittelbach:
Hi Maximilian, pull requests for typos if fine, but normally we like to see a file identification change even in such a case, more precisely a change in "date" (not version) in the `\Providesfile` (or package) line.   -- and there are many places where you can find typos if you look, so please avoid making pull requests for each individually :-)

PR #760, opened 2022-01-26
Typos and other corrections

### LINKED ISSUE #759: Typos
https://github.com/latex3/latex2e/blob/d025fbb0a1955783465c5f5e0011971ac74caab6/base/changes.txt#L77

Should be: `Add inputenc UTF-8 support for Digraphs U+01C4-U+01CC`

ivankokan:
Resolves #759.

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -74,7 +74,7 @@ are not part of the distribution.
 
 2021-12-10  David Carlisle  <David.Carlisle@latex-project.org>
 
-	* utf8ienc.dtx: Add inputenc UTF-8 support for Digraphs U+01C4-U+01CC
+	* utf8ienc.dtx: Add inputenc support for Digraphs U+01C4-U+01CC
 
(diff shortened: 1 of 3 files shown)


base/changes.txt
```diff
@@ -74,7 +74,7 @@ are not part of the distribution.
 
 2021-12-10  David Carlisle  <David.Carlisle@latex-project.org>
 
-	* utf8ienc.dtx: Add inputenc UTF-8 support for for Digraphs U_01C4-U+01CC
+	* utf8ienc.dtx: Add inputenc support for Digraphs U+01C4-U+01CC
```
FrankMittelbach, line 77: inputenc only handles UTF-8 (and calls it that) so I think it is correct to say UTF-8  here and below. Unicode is broader, if you give pdftex input in some other unicode flavor it will die.

FrankMittelbach:
we have done parts of this directly, and think that the other suggestions should not been accepted, so I'm closing that pull. Thanks for the effort.

Issue #762, opened 2022-01-27
utf8.def sets some commands globally


u-fischer:
## Brief outline of the bug

Unlike other encoding files utf8.def sets some commands globally. And this means that grouping `\inputencoding` (for example when using it the toc) doesn't work properly.

## Minimal example showing the bug

(the test file is ansinew encoded on my system, but one can also copy that into an utf8 encoded file, it only changes the actual message but shows the problem too). 

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{book}
\usepackage[latin9]{inputenc}

\begin{document}
\show ä
{\inputencoding{utf8}
 \inputencoding{latin9}
 \show ä
}
\show ä
\end{document}

```

The last `\show ä` then has still the utf8 definition:

~~~~
> ä=macro:
->\UTFviii@three@octets ä.
~~~~

## Log file (required) and possibly PDF file  

[test-ansi.log](https://github.com/latex3/latex2e/files/7950575/test-ansi.log)

FrankMittelbach:
this should be fixed by now but I guess we should make a test file from the original problem

PR #780, opened 2022-02-19
UTF-8 protected

davidcarlisle:
This PR uses `\protected` definitions for the UTF-8 buyes set up in `uf8.def` for UTF-8 support in 8bit TeX engines.

This greatly simplifies using these characters in expansion only contexts, keyval processing, but in general any `e` or `x` arguments from L3 code wheer `\protect` is not normally set up.

At the same time separate calls to `\string` on each byte are replaced by a singgle call to `\detokenize` and similarly separate calls to `\noexpand` by a single call to `\unexpanded`.  This change ensures that the inactive bytes are reconstitued as a valid UTF-8 string in a single expansion step rather than having `\string` commands between each byte (eg after f-expansion expands just teh first bye)

Finally, for issue #762, the group around the main loopis avoided, so that the definitions can be local as for other `inputenc` encodings, not global.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [X] Test file(s) added


davidcarlisle:
github-077 failing is

`\MakeLowercase{Ähh}`

which relies on `Ä` expanding to `\ddot{A}` in an `\edef`.

We may need to park this PR unless this can be fixed (using the L3 lowercase code would fix it but may be a step too far)

josephwright:
I wonder if we might update the code comments in `ltfinal`, otherwise looks good to me.

base/doc/ltnews35.tex
```diff
@@ -323,6 +323,22 @@ \subsection{???}
 
 \section{Code improvements}
 
+\subsection{\cs{protected} UTF-8 character definitions}
+The characters defined via \file{utf8.def} are now defined as \cs{protected}
+macros. This makes them safe to use in expansion contexts where the
+classic \cs{protect} mechanism is not enabled, notably L3 programming
+layer \texttt{e} and \texttt{x} arguments.
+
```
FrankMittelbach, line 331: and file name parsing

base/doc/ltnews35.tex
```diff
@@ -323,6 +323,22 @@ \subsection{???}
...
+Related to this change \cs{MakeUppercase} and \cs{MakeLowercase} have
+been updated to use the Unicode-aware case changing functions
+\cs{text\string_lowercase:n} in place of the \TeX-primitive \cs{lowercase}.
+A similar chaneg will be made in the \pkg{textcase} package.
+
```
FrankMittelbach, line 336: change

base/ltfinal.dtx
```diff
@@ -1051,28 +1055,35 @@
...
 %    long-term solution is to use all-caps fonts, but these aren't
 %    generally available.
+%
+% Two wrappers around the L3 case changing functions.
+% |\protected| to make them mostly safe as replacemenst for |luppercase|
+% and |\lowercase|.  
 %    \begin{macrocode}
+\ExplSyntaxOn
+\protected\def\@text@lowercase{\text_lowercase:n}
+\protected\def\@text@uppercase{\text_uppercase:n}
```
FrankMittelbach, line 1065: we have a different convention here `\@expl@...@@n` I suggest you follow that.

FrankMittelbach (changes requested):
perhaps also should add the firstaid for german.sty in firstaid (unless you have by now come up with a solution how to avoid it)

base/ltfinal.dtx
davidcarlisle: yes I wondered about that but thought the expl ones were mostly more internal  but happy to change.

davidcarlisle:
> perhaps also should add the firstaid for german.sty in firstaid (unless you have by now come up with a solution how to avoid it)

yes done (and german.sty removed from testfiles/helpers) although I seem to have broken some tests doing that, will look tomorrow.

base/doc/ltnews35.tex
davidcarlisle: actually filename parsing is already protcted via ifcsname but edef, e  and x have nothing without \protected

base/doc/ltnews35.tex
davidcarlisle: done

Issue #776, opened 2022-02-05
[graphics] .tex file is found instead of graphics file if both are in `\input@path`


PhelypeOleinik:
Reported [here](https://tex.stackexchange.com/q/630167/134574).

## Brief outline of the bug

If a call to `\includegraphics` asks for a file (say, `image`) without extension, and if both `folder-a/image.pdf` and `folder-b/image.tex` exist (both `folder-a` and `folder-b` in `\graphicspath`, but neither in a folder searched by kpse), then, because TeX implies `image.tex` from `image`, `folder-b/image.tex` is found at the time `\set@curr@path` is used to normalise the file name, and `folder-b` is fixed to the file name before `graphics` starts to search the known extensions, which is too early.

## Minimal example showing the bug

Given the following directory structure:

```none
.
├── folder-a
│   └── image.pdf
├── folder-b
│   └── image.tex
└── document.tex
```

and this `document.tex`:

```latex
\documentclass{article}
\usepackage{graphics}
\graphicspath{{./folder-a/}{./folder-b/}}
\begin{document}
\includegraphics{image}
\end{document}
```

results in the error

```none
! LaTeX Error: File `image' not found.
```

u-fischer:
@PhelypeOleinik  the fix breaks the search for graphics without extensions in a folder for me. E.g. this here:

~~~~
\documentclass{article}
\usepackage{graphicx}
\begin{document}
\includegraphics{figures/figure.pdf} %found

\includegraphics{figures/figure} % not found
\end{document}
~~~~

FrankMittelbach:
I wonder if it wouldn't be better to keep the search simple and call that a feature and revert. If a '.tex' is found when no extension is given and both exist then you have to give the extension --- this is the odd case. That none are found if just a file with a graphics extension exists is certainly counter productive and I'm not sure it is worth supporting with some complicated mechanism.

PhelypeOleinik:
Drat...

This could be fixed, for both cases, by using a non-searching version of `\set@curr@file` (which just clears `\input@path` before doing its thing).  I tested here and it seems to work.  Should I open a new issue, just push on top of this one, or completely revert?

FrankMittelbach:
either or; ok to push on this one, and then merge again later, but I think we should only do it if it is not complicating the normal case --- otherwise I'm for calling it a feature. The file handling got quite slow already and by the end of the day it is an odd case which should not make every file reading slower.

FrankMittelbach:
[...]

PR #784, opened 2022-03-02
Gh776 [graphics] Fix locating files with `\graphicspath` – take two

PhelypeOleinik:
Better fix for #776.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [n/a] `ltnewsX.tex` (and/or `latexchanges.tex`) updated&ndash;same ltnews entry as #777


base/ltfilehook.dtx
```diff
@@ -1117,6 +1128,32 @@
...
+%<latexrelease>      \@expl@@@filehook@resolve@file@subst@@w {#1}%
+%<latexrelease>    \@expl@@@filehook@if@file@replaced@@TF
+%<latexrelease>      {\@kernel@make@file@csname\@curr@file@reqd
+%<latexrelease>         \@expl@@@filehook@normalize@file@name@@w{#1}%
+%<latexrelease>       \if@tempswa \@expl@@@filehook@drop@extension@@N\@curr@file@reqd \fi}%
+%<latexrelease>      {\if@tempswa \@expl@@@filehook@drop@extension@@N\@curr@file \fi
+%<latexrelease>       \global\let\@curr@file@reqd\@curr@file}%
+%<latexrelease>    \@expl@@@filehook@clear@replacement@flag@@
+%<latexrelease>  \endgroup}
+%<latexrelease>\let\set@curr@file@nosearch\set@curr@file
```
FrankMittelbach, line 1152: I guess this is needed because if the kernel is rolled back graphics still uses that command. However, there should probably be a remark to that effect. Furthermore, what happens if the kernel is rolled back to some earlier state? Then the definition of `\setcurrfile`is altered back further (but  this guy stays at its definition and thus I'm failry sure chaos is the result ... or am I overlooking something?). In other words I guess that `\let` is also needed in all earlier parts.

FrankMittelbach (changes requested):
maybe I'm mistaken but my guess is some changes are still needed.
We should make a set of test files that do roll back to various earlier release dates to catch such issues (hopefully).

base/ltfilehook.dtx
PhelypeOleinik: > I guess this is needed because if the kernel is rolled back graphics still uses that command.

Bah, yes, `graphics.sty` doesn't roll back at all.

> Furthermore, what happens if the kernel is rolled back to some earlier state?

The kernel will use whatever definition of `\set@curr@file` was current at that time, and `graphics` should use that same `\set@curr@file` (but now named `@nosearch`).

> In other words I guess that `\let` is also needed in all earlier parts.

Yes, I think that should be enough.  I'll do that plus some tests in a bit.

FrankMittelbach: That's why I think we should add rollback tests both to base and to tools back to the last 4-5 main releases (and in tools load our packages and check that they actually remain functional  (as far as we can easily check). No of that will be perfect but better than nothing.

FrankMittelbach: and sorry to spoil it :-)

PhelypeOleinik: I ended up doing something else.  I think it is better, as far as possible, to have `latexrelease` roll back as close to _real_ rollback as possible, so instead of making `\set@curr@file@nosearch` a copy of `\set@curr@file` (which would also be a lie, because it _would_ search), I changed `graphics` to use `\Gin@set@curr@file`, and within `graphics` itself I pick up the proper definition for that internal macro, depending on what is available.  Indirectly this also supports plain, and whatever change `latexrelease` ends up doing to `\set@curr@file`, so it seems like a good solution to me.  Of course, I still have to manually undefine `\set@curr@file@nosearch` in every prior `\IncludeInRelease` block.

FrankMittelbach (approved):
looks fine to me as far as I can tell, and if it runes through the test suite I think it is good to go

Issue #714, opened 2021-12-01
Still some trouble with angle symbols copied from PDF and pasted


dbitouze:
#420 is partially fixed now (thanks!):
```latex
\RequirePackage{latexbug}
\documentclass{ltxdoc}
\pagestyle{empty}
\begin{document}
\meta{foo}
\end{document}
```
processed with `pdflatex` gives a PDF file which, if copied, gives `⟨foo⟩` when pasted.  Some remaining troubles nevertheless:

1. If the copied `⟨foo⟩` is inserted in the `.tex` file as verbatim:
    ```latex
    \RequirePackage{latexbug}
    \documentclass{ltxdoc}
    \pagestyle{empty}
    \begin{document}
    \meta{foo}
    
    \verb|⟨foo⟩|
    \end{document}
    ```
    the angle brackets of the 2<sup>nd</sup> `foo` are not the same in the resulting PDF and, if it is copied, it is pasted as `〈foo〉`, not as `⟨foo⟩`.
2. If the first MCE is processed with `lualatex`, what is pasted is `hfooi`, not `⟨foo⟩`.
3. From `interface3.pdf`, e.g. when `\seq_map_indexed_inline:Nn ⟨seq var⟩ {⟨inline function⟩}` (currently page 148) is copied, it is pasted  as `\seq_map_indexed_inline:Nn ⟨seq var ⟩ {⟨inline function ⟩}` (note the extra space before `⟩`).

zauguin:
These three points are rather different issues:

  1. That's a font issue. The font names these glyphs `angleleft` and `angleright` which according to the AGL marks them as glyphs corresponding to Unicode U+2329 and U+232A which are equivalent to the wider CJK glyphs you see. It might be usefull to addd an overwrite there, but these overwrites are maintained by [lcdf-typetools](https://github.com/kohler/lcdf-typetools). 
  2. This will be handled as part of other work on providing Unicode mapping for math fonts.
  3. The spaces there are inserted by your PDF viewer and can't be controlled by LaTeX.

dbitouze:
> These three points are rather different issues:

Many thanks for your insights!

>     1. That's a font issue.

So a Computer Modern one in the case of my example, isn't it?

>[...] It might be useful to add an overwrite there, but these overwrites are maintained by [lcdf-typetools](https://github.com/kohler/lcdf-typetools).

Well, beyond my scope, I'm afraid...

>     2. This will be handled as part of other work on providing Unicode mapping for math fonts.

[...]

PR #785, opened 2022-03-02
Define Unicode input for U+3008/U+3009

zauguin:
As mentioned in #714 we currently allow `U+2329` and `U+232A` to input `\textlangle` and `\textrangle`.
In order not to break documents when they are normalized we should therefore also allow their canonical equivalents `U+3008/U+3009`.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added


davidcarlisle:
I'm not sure that this is the right thing to do. U+3008/U+3009 are the CJK punctuation characters and if you map them to math angle brackets that breaks that use. The normalisation is from 2329 to 3008 which would mean that 2329 shouldn't be a math character rather than making 3008 a math character.  The whole thing is a mess which is why 2329 is essentially deprecated and 27E8 introduced which was seen as prefereable to breaking the stability rules and removing the normalisaion.

That said, the definitions can be over-written so if a CJK package redefined  U+3008 to be a suitable angle bracket having it default to a math angle bracket in the absense of such a package wouldn't be the end of the world.

A purer but perhaps less kind alternative would be to undeclare 2329 to match the undeclared 3008 so neither are usable by default.

zauguin:
> A purer but perhaps less kind alternative would be to undeclare 2329 to match the undeclared 3008 so neither are usable by default.

I do agree that that would be a nicer alternative *if* we are willing to break documents using U+2329/U+232A.

> U+3008/U+3009 are the CJK punctuation characters and if you map them to math angle brackets that breaks that use.

The question is IMO what the alternative looks like. Of course it's best to use actual CJK punctuation for these, but if there are alternative meanings for U+3008/U+3009 defined they should overwrite these anyway. If there are no alternatives defined, I would assume that using characters which are designed for mathematical use is still better than missing characters for CJK use. (Especially for CJK documents using non Unicode engines)
That leaves code which hooks into the undefined character handling, but even such code should be rather easy to adapt to add explicit cases for these.

> The normalisation is from 2329 to 3008 which would mean that 2329 shouldn't be a math character rather than making 3008 a math character.

But are we doing that? The suggestion is to map these to `\textlangle` and `\textrangle` which I would assume on a LICR to refer to a non-math angle bracket. For many font encodings these might actually be implemented as math characters since no non-math brackets are available, but I would see that more as a best effort fallback than as a classification of `\text[lr]angle` as a fundamentally mathematical character.

zauguin:
Generally I see four options here:

  1. Map U+3008/U+3009 to `\textlangle`/`\textrangle`.
  2. Unmap U+2329/U+232A.
  3. Map all four codepoint to something else™
  4. Keep the inconsistent handling of U+2329/U+232A and U+3008/U+3009.

Personally I think that we really should avoid option 4 since it leads to documents which break under mysterious circumstances and more generally is rather impractical in a modern Unicode workflow. I have no good suggestion what the "something else" in 3 would look like, so this leaves 1 or 2.

I would go for 1 since `\textlangle/\textrangle` already seems to express a fundamentally non mathematical nature (even though it might end up being implemented with mathematical characters) and to ensure as much compatibility as possible, but I wouldn't object to option 2 either. After all U+2329 and U+232A have been deprecated for more than a decade.

davidcarlisle:
ah I was thinking that we map U+2329 to \langle (math) I agree with
\textlangle it's a more or less plausible fallback for U+3008 as well so OK
let's do that.

PR #787, opened 2022-03-08
lab code for footnotes

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development

Still draft to look at concepts

- [x] Test file(s) added


u-fischer:
there is a stray closing brace in latex-lab-footnote.dtx at line 1316.

u-fischer:
The debug commands, e.g. `\__fnote_debug_footnotemark:` need some wrapper so that one can disable them. Currently every test stops in various places.

I could add a debug key for \DocumentMetadata, e.g. 

~~~~
,debug / footnote .bool_gset:N = some boolean 
~~~~

FrankMittelbach:
> The debug commands, e.g. `\__fnote_debug_footnotemark:` need some wrapper so that one can disable them. Currently every test stops in various places.
> 
> I could add a debug key for \DocumentMetadata, e.g.
> 
> ```
> ,debug / footnote .bool_gset:N = some boolean 
> ```

we could do that, but a) that coud should not use `\Showcommand`but `\LogCommand` (except that this is missing) and b) it was only as a tmp way to record stuff in the test files. Best would be to do both, a flag and also change from `\show` to `\typeout` basically

u-fischer:
If footmisc is not loaded the code errors:

~~~
\DocumentMetadata{uncompress,testphase=phase-II}
\documentclass{article}

\begin{document}
aaa
\end{document}
~~~

~~~
! Undefined control sequence.
\__hook begindocument ...in =\maxdimen \ifFN@para 
~~~

dbitouze:
(I'm sorry if this is not the right place to ask for a feature request about footnotes.) Something I'm often missing is a way to test whether a stuff is currently in a footnote or not.

A typical use case is for acronyms which, thanks to styles, can be automatically displayed at their first occurrences as ⟨short⟩\footnote{⟨long⟩}. That's what is expected for the first occurrences that happen to appear in "normal" text, but that's a non sense for the ones which happen to appear inside footnotes (where, typically, ⟨short⟩ (⟨long⟩) is expected).

FrankMittelbach:
> If footmisc is not loaded the code errors:

yeah, I keep forgetting that the code is split. Should be corrected now.

FrankMittelbach:
> (I'm sorry if this is not the right place to ask for a feature request about footnotes.)

@dbitouze Denis, you should  indeed know that this is a rather bad place for it. This is a code branch and it doesn't help  to pepper it with (somewhat) unrelated feature requests. Concerning your request I think it would be helpful if there is some general notion of context available that you can query (out of which  one would be  "footnote" another would  be float or more detailed which float area, side-margin and the like). So providing something hardwired just for footnotes  wouldn't be the right approach in my opinion. 

Your use case is clearly valid and I would welcome a clear documentation of what you described above as a normal issue, but please don't be upset if I immediately mark it as long -term.

dbitouze:
> @dbitouze Denis, you should indeed know that this is a rather bad place for it. This is a code branch and it doesn't help to pepper it with (somewhat) unrelated feature requests.

OK, sorry.

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+%
+%
+% \section{Tagging support}
+%
+%
+%
+%
+%
+%
+% \StopEventually{\setlength\IndexMin{200pt}  \PrintIndex  }
```
josephwright, line 383: `\MaybeStop`?

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+
+
+
+%-------------------------------------
+
+
+\ExplSyntaxOn
+
+
+\cs_new:Npn \fn_step_fnmark:nn #1#2 {
```
josephwright, line 435: protected please ;)

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+  \tl_if_novalue:nTF {#1}
+    {
+      \stepcounter {#2}
+      \protected@xdef \@thefnmark { \cs:w the#2 \cs_end: }
```
josephwright, line 439: Why `\cs:w` rather than `\use:c`?

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+    }
+    {
+     \group_begin:
+        \int_set:cn { c@#2 }{ #1 }
```
josephwright, line 443: Should counters be set globally?

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+        \unrestored@protected@xdef \@thefnmark { \cs:w the#2 \cs_end: }
```
josephwright, line 444: Again, `\use:c`?

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+     \group_end:
+    }
+}
+
+%-------------------------------------
+
+% the debugging code is just temp
+\bool_new:N \g_fnote_debug_bool
```
josephwright, line 465: These are `fnote` but above we have `fn`: what's the reasoning?

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+         \LogHook{fnmark}
+         \LogHook{fnmark/begin}
+         \LogHook{fnmark/end}
+         \LogHook{fnmark/after}
+         \cs_gset_eq:NN \__fnote_debug_footnotemark: \prg_do_nothing:
+       }
+}
+
+
+\cs_new:Npn \fnote_footnotemark: {
```
josephwright, line 491: Protected

josephwright: Isn't the name here repetitive? Just `\fnote_mark:`?

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+         \LogHook{fntext/end}
+         \LogHook{fntext/after}
+%    \end{macrocode}
+%    Show the info only once (if at all).
+%    \begin{macrocode}
+         \cs_gset_eq:NN \__fnote_debug_footnotetext: \prg_do_nothing:
+       }
+  }
+
+\cs_new:Npn \fnote_footnotetext:n #1 {
```
josephwright, line 599: Same comments: `\fnote_text:`? Protected

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+
+% default for config point (0 args)
+\def\@footnotetext@cfgpointiii { \rule\z@\footnotesep }
+
+% default for config point (0 args)
+\def\@footnotetext@cfgpointiv  { \@finalstrut\strutbox }
+
+
+% kernel hook for tagging (2 args)
+\cs_new:Npn \@kernel@process@footnotetext {}
```
josephwright, line 693: You are not exactly consistent in what to use: could we have `\cs_new_(protected):Npn` for all the new material? I fully understand sticking to LaTeX2e names for code you are retaining from existing implementation.

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+% don't want to do this like that in the kernel but for now I have
+% used this interface unchanged.
+
+
+\newdimen\footnotemargin
+\footnotemargin\maxdimen         % no value given
+
+\AtBeginDocument{
+  \ifdim \footnotemargin=\maxdimen
+    \setlength\footnotemargin{1.8em}
```
josephwright, line 721: Should this itself have a config point? Cf. what I've tried to do in `l3draw` to ensure that 'baked in' settings are still adjustable

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+% default for config point (1 arg)
+\long \def \@makefntext@cfgpointii #1 { #1 }
+
+
+
+%    If the definition for \cs{@makefntext} is that of the standard
+%    classes then replace it with \cs{fnote_makefntext:n}, otherwise
+%    try to patch the definition
+
+\newcommand\old@std@class@makefntext[1]{%
```
josephwright, line 763: I'm really not sure why `\newcommand` here but `\long \def` just before and `\cs_new:Npn` elsewhere!

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+    \parindent 1em%
```
josephwright, line 764: I'd like a space after the value here - I know the next token is non-expandable but it's a matter of good style :)

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+    \noindent
+    \hb@xt@1.8em{\hss\@makefnmark}#1}
+
+
+\tl_new:N \l__fnote_patch_tl
+\cs_new_protected:Npn \_fnote_patch:
```
josephwright, line 770: Could we have some comments here to explain what's going on?

josephwright (changes requested):

required/latex-lab/README.md
```diff
@@ -20,6 +20,8 @@ to documents already using the new functionality.
 
 ### Support for `\DocumentMetadata`
 
+### New output routine code (und development)
```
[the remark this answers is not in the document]
FrankMittelbach: und is the German word for 'and' but I meant to write 'under'.

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+%
+%    Historically, the footnote setup in \LaTeX{} was done by
+%    providing definitions for \cs{@makefnmark} (format the footnote
+%    mark in running text and in front of the footnote text) and
+%    \cs{@makefntext} (formatting the footnote text and placing a mark
+%    in front of it).
+%
+%    There was a default definition for \cs{@makefnmark} in the format
+%    that was used by most document classes,
+%    but \cs{makefntext} had to be defined in the class itself because
```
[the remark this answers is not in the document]
FrankMittelbach: yes

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+%    appears at the bottom of the page (default case), or possibly
+%    elsewhere, e.g. in the margin.
+%
+% \subsubsection{Configuration points}
+%
+%    To cater for different layout configurations there are four
+%    configuration points which can be set only by one package or
+%    class. These are:
+%    \begin{description}
+%    \item[\cs{@footnotetext@cfgpoint} (1 argument)]
```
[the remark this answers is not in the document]
FrankMittelbach: I consider all cfgpoint... names temporary until we have an interface for configuration points similar to what we have for hooks. For now I just "numbered" them but to call the first "i" looks odd to me (other roman numerals are easy to identify as such but cfgpointi looks weird in my eyes). Not sure what you mean with your last sentence.  The config points are "aux" they are "config points".

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+%
+%    This implements the formatting of the footnote text the way it
+%    appears at the bottom of the page (default case), or possibly
+%    elsewhere, e.g. in the margin.
+%
+% \subsubsection{Configuration points}
+%
+%    To cater for different layout configurations there are four
+%    configuration points which can be set only by one package or
+%    class. These are:
```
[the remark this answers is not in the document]
FrankMittelbach: What you say is correct, there should be a full-blown documentation on config points (but there isn't yet). And there should be more examples, I agree. However, we should keep in mind that in contrast to hooks, there are only very very few packages ever, that touch configuration points. Those here, for example would only be altered by footmisc (or its equivalent if it goes into the format) and "perhaps" KOMA and memoir if they intend to supply their own footnote handling for paragraph footnotes and that's it. Other packages and classes all just supply code for one of the hooks (or rather right now they overwrite `\makefntext`unconditionally instead of doing so) but not attempt to alter LaTeX's overall footnote handling. Even `manyfooot` or `bigfoot`really tried to do any of that but relied on footmisc to do that part.

So I have been a bit economical with the documentation for now, given priorities. Anybody, who want to see how they are used in detail can study the footmisc reimplementation using them.

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+%
+%      This configuration point receives the material to typeset the
+%      footnote mark. By default, all it does is running \cs{indent}
+%      to get a paragraph indentation (if one is set up---in most
+%      layouts it is 0~points)  and then typesets the mark, but in
+%      some designs it executes more elaborate code.
+%
+%      If tagging is produced this configuration point is also
+%      responsible for surrounding the mark with the appropriate tags
+%      marking the mark as an Lbl. It does this using the command \cs{tag@FELbl}.
```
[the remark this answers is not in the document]
FrankMittelbach: We have to think a bit further about the exact interfaces, but the cfg point code has to provide tagging code. That can'T be done from the outside depending on how the formatting is done. This is why for now I provide two macros that can be used for this (which needs more documentation but it is not yet 100% clear how that should look like). For this we do have to work with people who want to use that cfg point (which as I said will be few if anybody) and see what is needed and whether or not that is covered. For now everything that I can see in the 160 classes+packages that do something with footnotes all is taken care of without the need to do anything about the cfg points.

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+%      marking the mark as an MC of type FENote. It does this using
+%      the command \cs{tag@FENote}.
+%
+%    \end{description}
+%
+%
+%    The above configuration points are sufficient to implement all
+%    commonly used footnote layouts assuming L-R typesetting. For R-L
+%    typesetting they or may or may not need some extension (though
+%    that is not clear right now).
```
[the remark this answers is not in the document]
FrankMittelbach: yes, but one step at a time ... right now I'm happy enough to see it working well for L-R. In practical terms I think we don't have the bandwidth nor the skill right now to go beyond initially.

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1731 @@
...
+%      The configuration point \cs{@footnotetext@cfgpointii} normally
+%      runs the command \cs{@makefntext} or some code that eventually
+%      runs this command, and this then produces the footnote mark (in
+%      front of the footnote text) and the formatted footnote text. In
+%      front of both the mark and the footnote text the hook
+%      \hook{fntext/para2} is executed.  In fact it is not
+%      clear if it is really needed, as most likely all of its
+%      settings could probably go into \hook{fntext/para}, it is just
+%      that historically there have been settings in some classes at
+%      this point in \cs{@makefntext}.
```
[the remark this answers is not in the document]
FrankMittelbach: I guess you are right. Dropped.

required/latex-lab/latex-lab-footnotes.dtx
FrankMittelbach: upps, wonder where that came from ... not in footmisc.dtx

required/latex-lab/latex-lab-footnotes.dtx
FrankMittelbach: since you asked so nicely .. not that I think it matters for any of them ...

required/latex-lab/latex-lab-footnotes.dtx
FrankMittelbach: mechanical replacement of existing 2e code I guess :-)

required/latex-lab/latex-lab-footnotes.dtx
FrankMittelbach: well it should and it shouldn't :-) 2e  does it locally (what's what the group is for really). I just haven't altered the way the interface behaved there. But perhaps there should be made some remark to that effect.

josephwright: Ah, right: certainly worth a comment

required/latex-lab/latex-lab-footnotes.dtx
FrankMittelbach: do we? different times of development ...  not intentional.  I now normalized to fnote as prefix for now

required/latex-lab/latex-lab-footnotes.dtx
FrankMittelbach: it is but I thought I leave the trace to the two names which are `\@footnotetext` and `\@footnotemark`

required/latex-lab/latex-lab-footnotes.dtx
FrankMittelbach: I don't think so, this is setting the default from a format perspective if none is given by the class or by a package. Thus if a class wants a different default it would just give one for \footnotemargin and the user could then overwrite or not in the preamble and this code wouldn't touch it.  The reason it is only made in \document is because the historical defaul is in "em" and that is not established until after the preamble is over.

Anyway, that's the interface that footmisc has had the last 20 years and this is reimplementing that to stay compatible with the documents out there.

required/latex-lab/latex-lab-footnotes.dtx
FrankMittelbach: The reason is that this is a straight copy of what the std classes do. I could of course translate it to the equivalent and it would probably look better, but the point is to answer the question: "is the definition unchanged from what classes.dtx contains," so what I did was copying that definition 1-2-1

required/latex-lab/latex-lab-footnotes.dtx
FrankMittelbach: you may like it :-), but it would be wrong, because then the test would give false because that space is not there in the definition used in the std classes, that is the whole point. Okay, I add a note to the definition ...

required/latex-lab/latex-lab-footnotes.dtx
FrankMittelbach: if I would understand it? I'm I clever enough to describe Bruno's code? I'll try ...

josephwright (approved):

required/latex-lab/latex-lab-footnotes.dtx
blefloch: Btw, `\_fnote` should be `\__fnote` (two underscores).  Basically my code looks for `\hbox` and, if found, it grabs `\hbox#1#{`.  Otherwise it looks for `\hb@xt@` and if found it grabs it and places the tagging thing there.  It is a bit unsafe for instance if `\hbox` appears without any brace group afterwards for some reason.

FrankMittelbach: made stuff internal in various places and use `@@` notation

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1815 @@
...
+\cs_new_protected:Npn \tag_if_loaded:TF  { \cs_if_exist:NTF \tag_struct_begin:n }
+
+
+
+\cs_new_protected:Npn \tag@FEMark #1 {
+  \tag_if_loaded:TF
+     {
+       \tag_mc_end_push:
+       \exp_args:Nx
+       \tag_struct_begin:n{tag=Lbl,ref=fn.\the\c@absfootnote}
```
u-fischer, line 1032: `\the\c@absfootnote` should imho be replaced directly by some command with default meaning `\the\c@absfootnote` so that the actual reference can be changed from the outside.

required/latex-lab/latex-lab-footnotes.dtx
```diff
@@ -0,0 +1,1815 @@
...
+}
+
+
+
+
+\cs_new_protected:Npn \tag@FELbl #1 {
+  \tag_if_loaded:TF
+    {
+      % target should perhaps be raised ...
+      \IfPackageLoadedTF{hyperref}{ \hypertarget{fn.\the\c@absfootnote}{} }{}
```
u-fischer, line 1082: here too

u-fischer (approved):

required/latex-lab/latex-lab-footnotes.dtx
FrankMittelbach: I made a note in the source, but I suggest we do the next time around when we also integrate support for \footref etc.

PR #788, opened 2022-03-08
First conversion of xtemplate to l3keys internals

josephwright:

# Internal housekeeping

## Status of pull request

- [X] Feedback wanted 
- Under development
- Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/required/latex-lab/ltprototype.dtx
+++ b/required/latex-lab/ltprototype.dtx
@@ -1,0 +1,512 @@
+% \iffalse meta-comment
+%
+%% File: ltprototype.dtx
+%
+% Copyright (C) 1999 Frank Mittelbach, Chris Rowley, David Carlisle
+%           (C) 2004-2010 Frank Mittelbach, The LaTeX Project
+%           (C) 2011-2022 The LaTeX Project
+%
+% It may be distributed and/or modified under the conditions of the
+% LaTeX Project Public License (LPPL), either version 1.3c of this
+% license or (at your option) any later version.  The latest version
+% of this license is in the file
+%
+%    https://www.latex-project.org/lppl.txt
+%
+% This file is part of the "latex-lab bundle" (The Work in LPPL)
+% and all files in that bundle must be distributed together.
+%
+% -----------------------------------------------------------------------
+%
+% The development version of the bundle can be found at
+%
+%    https://github.com/latex3/latex2e
+%
+% for those people who are interested.
+%
+%<*driver>
+\documentclass[full]{l3doc}
+\begin{document}
+  \DocInput{\jobname.dtx}
+\end{document}
+%</driver>
+% \fi
+%
+% \title{^^A
+%   The \textsf{ltprototype} package\\ Prototype document functions^^A
+% }
+%
+% \author{^^A
+%  The \LaTeX{} Project\thanks
+%    {^^A
+%      E-mail:
+%        \href{mailto:latex-team@latex-project.org}
+%          {latex-team@latex-project.org}^^A
+%    }^^A
+% }
+%
+% \date{Released 2022-01-12}
+%
+% \maketitle
+%
+% \begin{documentation}
+%
+% There are three broad \enquote{layers} between putting down ideas into
+% a source file and ending up with a typeset document. These layers of
+% document writing are
+% \begin{enumerate}
+%   \item authoring of the text with mark-up;
+%   \item document layout design;
+%   \item implementation (with \TeX{} programming) of the design.
+% \end{enumerate}
+% We write the text as an author, and we see the visual output of the design
+% after the document is generated; the \TeX{} implementation in the middle is
+% the glue between the two.
+%
+% \LaTeX{}'s greatest success has been to standardise a system of mark-up that
+% balances the trade-off between ease of reading and ease of writing to suit
+% almost all forms of technical writing. It's
+% other original strength was a good background in typographical design; while
+% the standard \LaTeXe{} classes look somewhat dated now in terms of their
+% visual design, their typography is generally sound. (Barring the occasional
+% minor faults.)
+%
+% However, \LaTeXe{} has always lacked a standard approach to customising
+% the visual design of a document. Changing the looks of the standard classes
+% involved either:
+% \begin{itemize}
+%   \item Creating a new version of the implementation code of the class and
+%     editing it.
+%   \item Loading one of the many packages to customise certain elements of
+%     the standard classes.
+%   \item Loading a completely different document class, such as
+%     \textsf{KOMA-Script} or \textsf{memoir}, that allows easy customisation.
+% \end{itemize}
+% All three of these approaches have their drawbacks and learning curves.
+%
+% The idea behind \pkg{xtemplate} is to cleanly separate the three layers
+% introduced at the beginning of this section, so that document authors who
+% are not programmers can easily change the design of their documents.
+% \pkg{xtemplate} also makes it easier for \LaTeX{} programmers to provide
+% their own customisations on top of a pre-existing class.
+%
+% \section{What is a document?}
+%
+% Besides the textual content of the words themselves, the source file
+% of a document contains mark-up elements that add structure to the
+% document. These elements include sectional divisions, figure/table
+% captions, lists of various sorts, theorems/proofs, and so on.
+% The list will be different for every document that can be written.
+%
+% Each element can be represented logically without worrying about the
+% formatting, with mark-up such as \cs{section}, \cs{caption},
+% |\begin{enumerate}| and so on. The output of each one of these
+% document elements will be a typeset representation of the information
+% marked up, and the visual arrangement and design of these elements
+% can vary widely in producing a variety of desired outcomes.
+%
+% For each type of document element, there may be design variations that
+% contain the same sort of information but present it in slightly
+% different ways. For example, the difference between a numbered and an
+% unnumbered section, \cs{section} and |\section*|, or the difference
+% between an itemised list or an enumerated list.
+%
+% There are three distinct layers in the definition of
+% \enquote{a document} at this level
+% \begin{enumerate}
+%   \item semantic elements such as the ideas of sections and lists;
+%   \item a set of design solutions for representing these elements
+%     visually;
+%   \item specific variations for these designs that represent the
+%     elements in the document.
+% \end{enumerate}
+% In the parlance of the template system, these are called object types,
+% templates, and instances, and they are discussed below in sections
+% \ref{sec:objects}, \ref{sec:templates}, and~\ref{sec:instances},
+% respectively.
+%
+% By formally declaring documents to be composed of mark-up elements
+% grouped into objects, which are interpreted and typeset with a set of
+% templates, each of which has one or more instances with which to
+% compose each and every semantic unit of the text, we can cleanly
+% separate the components of document construction.
+%
+% \section{Object types}
+% \label{sec:objects}
+%
+% An \emph{object type} (sometimes just \enquote{object}) is an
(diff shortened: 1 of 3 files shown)


josephwright:
At the moment this is all draft as it is very much a sketch. Various notes

- You'll need the version of `expl3` with `\keys_precompile:Nnn` available for this to work: I've not included it in the PR at the moment since I imagine people will want to test locally
- I've deliberately not included design-level names, as those could prevent loading both this code and `xtemplate`. I think at the moment we want to be able to compare and contrast in one file
- I've named as 'prototype' as that's what the description says: using 'template' leads to some slightly awkward naming
- I wonder if 'element' would be better than 'object', as the latter might be misunderstood?
- I've not tried to do the same efficiency work as `xtemplate` does for combining defaults and instance values: I've simple set all of the defaults then added the settings for an instance. That will fail if keys are defined expecting to be set only once
- The interface ends up slightly repetitive as you have to set defaults after creating keys. `xtemplate` has a two-part structure too, just a different one. I've not tried (yet) to avoid having two passes over the keys as that means having to use more complexity in key definition, and we then can't use `\keys_define:nn` directly. My reasoning is that I think having `foo .tl_set:N = \l_some_tl : default` is overloading the interface too much
- In the same area, I did wonder if we want a `.preset:n` property similar to `xkeyval`, but to me the idea of resetting the keys for each setting really does link to prototyping/templates specifically
- I've not provided a dedicated interface for 'local override', nor have I reserved an argument for that: both are possible but need discussion
- I've not done anything about collections, etc.: I'm imagining something like LDB or CSS to handle contexts at some stage

required/latex-lab/ltprototype.dtx
```diff
@@ -0,0 +1,515 @@
...
+%          {latex-team@latex-project.org}^^A
+%    }^^A
+% }
+%
+% \date{Released 2022-01-12}
+%
+% \maketitle
+%
+% \tableofcontents
+%
```
FrankMittelbach, line 53: structure of the headings is right I would say

required/latex-lab/ltprototype.dtx
```diff
@@ -0,0 +1,515 @@
...
+% After the instances have been chosen, document commands must be
+% declared to use those instances in the document. \cs{UseInstance}
+% calls instances directly, and this command should be used internally
+% in document-level mark-up.
+%
+% \begin{function}{\prototype_use_instance:nn}
+%   \begin{syntax}
+%     \cs{prototype_use_instance:nn}
+%     ~~\Arg{object type} \Arg{instance} \meta{arguments}
+%   \end{syntax}
```
FrankMittelbach, line 277: needs to be discussed elsewhere but I think the instance (and template) use should accept one argument with one-time parameter settings, either always (in which case one would leave it empty if it shouldb't be used) or alternatively as two separate calls with different number of arguments

required/latex-lab/ltprototype.dtx
```diff
@@ -0,0 +1,515 @@
...
+%
+% \begin{implementation}
+%
+% \section{\pkg{ltprototype} Implementation}
+%
+% \subsection{File declaration}
+%    \begin{macrocode}
+%<*package>
+\ProvidesFile{ltprototype.sty}
+  [2022-03-25 v0.1a Experimental prototype document functions]
```
FrankMittelbach, line 316: what date is this?

FrankMittelbach (changes requested):
Overall looks all good. 
- Not keen on the name "prototype" but don't have a suggestion yet, so for no it is fine.
- Not keen on ltprototype.sty, make it latex-lab-prototype.sty perhaps; more to type but clear  what it is then (not for general \usepackage{...} use)
- I'm a big fan of test files first then code then update
- but even if done after the fact, I think it is about time now that all code interfaces get tests that exercise their  behavior --- helps a lot later imo

required/latex-lab/ltprototype.dtx
josephwright: That would be a copy-paste mess up

josephwright:
>     * Not keen on the name "prototype" but don't have a suggestion yet, so for no it is fine.

We could stick with `template`, of course.

>     * Not keen on ltprototype.sty, make it latex-lab-prototype.sty perhaps; more to type but clear  what it is then (not for general \usepackage{...} use)

I'll adjust.

>     * I'm a big fan of test files first then code then update
> 
>     * but even if done after the fact, I think it is about time now that all code interfaces get tests that exercise their  behavior --- helps a lot later imo

I can add some: I guess I will need some `show` functions to help with that.

FrankMittelbach:
> > ```
> > * Not keen on the name "prototype" but don't have a suggestion yet, so for no it is fine.
> > ```
> 
> We could stick with `template`, of course.

as I said (or wanted to say :-) ) for now it is fine and perhaps forever. Just wanted to record my stomach feeling on the name.

> > ```
> > * I'm a big fan of test files first then code then update
> > 
> > * but even if done after the fact, I think it is about time now that all code interfaces get tests that exercise their  behavior --- helps a lot later imo
> > ```
> 
> I can add some: I guess I will need some `show` functions to help with that.

Some `\ShowTemplate` and `\ShowInstance` would be useful anyway, I can't count how often I use \Showhook these days ..., so yes I think those should be added (as well as tests).

required/latex-lab/ltprototype.dtx
FrankMittelbach: :-)

josephwright:
Another thing to think about is the way keys are assigned. At present, I've not provided an equivalent to `\AssignTemplateKeys`: key assignment is automatic. If you look over TeX Live, the only place` that `\AssignTemplateKeys` is not the first thing in the code are in `enotez` and `xgalley`. In both cases, that's because they want to limit the scope of assignment. `enotez` uses
```latex
\group_begin:
   \AssignTemplateKeys
```
whereas `xgalley`  quite a long save list
```latex
    \tl_put_left:Nx \l__galley_justification_other_tl
      {
        \skip_set:Nn \exp_not:N \l_galley_par_end_skip
          { \skip_use:N \l_galley_par_end_skip }
        \bool_if:NTF \l_galley_fixed_spacing_bool
          { \bool_set_true:N  \exp_not:N \l_galley_fixed_spacing_bool }
          { \bool_set_false:N \exp_not:N \l_galley_fixed_spacing_bool }
        \galley_interword_spacing_set:N
          \exp_not:N \l_galley_fixed_spacing_bool
        \dim_set:Nn \exp_not:N \l_galley_par_indent_dim
          { \dim_use:N \l_galley_par_indent_dim  }
        \skip_set:Nn \l_galley_line_left_skip
          { \skip_use:N \l_galley_line_left_skip }
        \skip_set:Nn \exp_not:N \l_galley_line_right_skip
          { \skip_use:N \l_galley_line_right_skip }
        \skip_set:Nn \exp_not:N \l_galley_par_begin_skip
          { \skip_use:N \l_galley_par_begin_skip }
        \int_set:Nn \exp_not:N \l_galley_last_line_fit_int
          { \int_use:N \l_galley_last_line_fit_int }
        \skip_set:Nn \exp_not:N \@rightskip
          { \skip_use:N \l_galley_line_right_skip }
      }
    \tl_gput_right:Nn \g_galley_restore_running_tl
      { \l__galley_justification_other_tl }
    \AssignTemplateKeys
```
for restoring paragraph settings without a group.

Both of those use-cases can be covered in other ways: it's a question of setting up the template keys so they assign to an 'intermediate' variable, then assigning those as necessary to the 'live' ones for these cases.

I'd welcome thoughts from others: to me the balance is such that these rather unusual cases feel better handled on a per-case basis, leaving most templates not needing `\AssignTemplateKeys`. But I can see that this might feel awkward - there will be other save-and-restore situations. Perhaps an alternative is to have a dedicated function that deals with that?

josephwright:
Another thing to think about is `\KeyValue`. Implementing it using `l3keys` is non-trivial, but I'm not sure it's needed. Currently two packages use it: `xeCJK` and `diffcoeff`. My reading of both is that a similar outcome could be achieved using a meta key and setting one or more underlying keys that way. Again, thoughts welcome.

josephwright:
In particular, `\KeyValue` relies on keys being set in order of definition, not order they appear in the key settings. That's true for `xtemplate` but not for `l3keys`.

required/latex-lab/latex-lab-prototype.dtx
```diff
@@ -0,0 +1,813 @@
...
+%     ~~\Arg{settings} \meta{arguments}
+%   \end{syntax}
+%   Uses the \meta{template} of the specified \meta{object type},
+%   applying the \meta{settings} and absorbing \meta{arguments} as
+%   detailed by the \meta{object type} declaration. This in effect
+%   is the same as creating an instance using
+%   \cs{template_declare_instance:nnnn} and immediately using it with
+%   \cs{template_use_instance:nnn}, but without the instance having any
+%   further existence. It is therefore useful where a template needs to
+%   be used once.
```
FrankMittelbach, line 305: and only if there is no desire to allow alterations to key settings (eg via \EditInstance if that makes an appearence)

required/latex-lab/testfiles/latex-lab-prototype001.tlg
```diff
@@ -0,0 +1,46 @@
...
+<to be read again> 
+                   i
+l. ...}
+A number should have been here; I inserted `0'.
+(If you can't figure out why I needed to see a number,
+look up `weird error' in the index to The TeXbook.)
+! Use of \??? doesn't match its definition.
+<argument> \???  
+                 ! LaTeX3 Error: Relation 'i' not among =,<,>,==,!=,<=,>=.
+l. ...}
```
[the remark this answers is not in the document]
josephwright: I just felt like trying an imaginary number ;)

required/latex-lab/latex-lab-prototype.dtx
```diff
@@ -0,0 +1,813 @@
...
+%   \item Loading a completely different document class, such as
+%     \textsf{KOMA-Script} or \textsf{memoir}, that allows easy customisation.
+% \end{itemize}
+% All three of these approaches have their drawbacks and learning curves.
+%
+% The idea behind \pkg{xtemplate} is to cleanly separate the three layers
+% introduced at the beginning of this section, so that document authors who
+% are not programmers can easily change the design of their documents.
+% \pkg{xtemplate} also makes it easier for \LaTeX{} programmers to provide
+% their own customisations on top of a pre-existing class.
```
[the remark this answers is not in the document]
josephwright: Ah, yes: I just copy-pasted that. I'll think of something.

required/latex-lab/latex-lab-prototype.dtx
```diff
@@ -0,0 +1,813 @@
...
+%   \end{verbatim}
+%   Of course, the key names here are entirely imaginary, but illustrate
+%   the general idea of fixing some settings.
+% \end{function}
+%
+% \section{Document interface}
+%
+% After the instances have been chosen, document commands must be
+% declared to use those instances in the document. \cs{UseInstance}
+% calls instances directly, and this command should be used internally
```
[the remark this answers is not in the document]
josephwright: Another copy-paste: oops.

required/latex-lab/latex-lab-prototype.dtx
josephwright: This is one that Bruno suggests we could drop anyway (I think if we go with the approach here, he has a point as there really is no performance gain).

FrankMittelbach: true, but I'm not sure we should, but this is something to discuss over a pot of tea rather than here ... it is one of the listed open points so we'll get to it.

PR #796, opened 2022-03-18
Gh795 [ltcmdhooks] Adding hooks to simple macros cause them to expand

PhelypeOleinik:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


PhelypeOleinik:
I think this doesn't need rollback, because the old version is simply broken with some macros, so rolling back to that seems like something people wouldn't want.

FrankMittelbach:
> I think this doesn't need rollback, because the old version is simply broken with some macros, so rolling back to that seems like something people wouldn't want.

guess not, but should have some sort of roll forward in the sense, that if you start from current overleaf (which I think is early TL2021 you get it  if you use  a latexrelease for 2022-06

PhelypeOleinik:
Overleaf has LaTeX 2021-06-01, before I implemented `\ShowCommand` for `ltcmd` (and introduced this bug), so they don't have the bug.  They may be affected if they install TeX Live 2022 before the next release though...

PR #803, opened 2022-04-14
Define starred and protected versions of \ref, \Ref and \pageref

u-fischer:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge (apart from a conflict with a testfile changed in develop too)

- [x] Test file(s) added

Currently the definitions of reference commands are rather confused: the kernel defines (non robust) \ref and \pageref and a robust \Ref. If hyperref is loaded it makes \ref and \pageref robust, add links and  and a starred version.  nameref patches ifthenelse to enable it to use \pageref. \Ref is only adapted if varioref is detected as  hyperref still thinks that this is a varioref command. showkeys only handles the unstarred reference commands. 

The pull request tries to resolve some of these incompabilities: \ref, \pageref and \Ref  have now starred versions always. They are always defined with \NewDocumentCommand (similar changes in hyperref have been done there). Expandable commands to retrieve the labels have been added to the kernel and ifthen has been updated to use them. showkeys has been updated to handle the starred version too.  The texh4t maintainer has been contacted and has already adjusted the code.

As the commands are now protected the lof and toc no longer contain the "undefined label" code, this reduce the number of compilations needed if references are in captions or section commands and so changed two testfiles.


base/ltxref.dtx
```diff
@@ -63,17 +63,18 @@
...
 %
 % \section{Cross Referencing}
 %  The user writes  |\label|\marg{foo}  to define the following
 %  cross-references:
 %
-%   |\ref|\marg{foo}: value of most recently incremented referenceable
+%   |\ref|*\marg{foo}: value of most recently incremented referenceable
 %             counter. in the current environment. (Chapter, section,
 %             theorem and enumeration counters are
 %             referenceable, footnote counters are not.)
```
FrankMittelbach, line 75: unrelated but is that still true? they are referenceable, aren't they?

base/ltxref.dtx
```diff
@@ -87,6 +88,9 @@
 %  leaving a space on both sides of it is equivalent to leaving
 %  a space on either side.
 %
+%  Note: the starred version |\ref*| and |\pageref*| are provided
+%  to align with the of hyperref.
```
FrankMittelbach, line 92: Without \pkg{hyperref} (or some other package using the starred form) it is simply ignored.

base/ltxref.dtx
```diff
@@ -229,9 +235,43 @@
...
+%<latexrelease>                 {\ref}{Add starred reference commands}%
+%    \end{macrocode}
+%    \begin{macrocode}
+\def\@kernel@sref#1{\expandafter\@setref\csname r@#1\endcsname\@firstoftwo{#1}}
+\def\@kernel@spageref#1{\expandafter\@setref\csname r@#1\endcsname
+                 \@secondoftwo{#1}}
+\let\@kernel@ref\@kernel@sref
+\let\@kernel@pageref\@kernel@spageref
+\NewDocumentCommand\ref{s}
+   {\IfBooleanTF{#1}{\@kernel@sref}{\@kernel@ref}}
```
FrankMittelbach, line 252: Given that the kernel doesn't do anything special with the star and we don't want other packages to overwrite `\kernel@...` there doesn't seem to be a need to provided two different commands that are then made the same. I think one used by both branches is clearer and simpler.

base/doc/ltnews35.tex
```diff
@@ -455,6 +455,17 @@ \subsection{Better allocator for Lua command ids}
 %
 \githubissue{536}
 
+\subsection{Starred command version for \cs{ref}, \cs{Ref} and \cs{pageref}}
+
+Since a long time \pkg{hyperref} provided starred versions for the reference commands
+that don't create active links. The different syntax required from users and package authors
+to check for \pkg{hyperref} if both types should be supported. The commands have now been aligned with
+the \pkg{hyperref} usage and allow an optional star. The \pkg{showkeys} package
```
FrankMittelbach, line 463: and always allow an ...

FrankMittelbach (changes requested):

base/ltxref.dtx
u-fischer: I wondered about that. If we move at some time hyperlinking to the kernel then we would need them, wouldn't we? And then the switch would be easier (e.g. one wouldn't have to adapt showkeys again). But I don't mind to change it for now.

FrankMittelbach: That's an argument in favor of leaving it as you implemented it. Ok.

u-fischer:
It breaks with babel:

~~~~
\documentclass{article}
\usepackage[english]{babel}
\begin{document}

\ref{abc} \ref{abc}

\end{document}
~~~~

~~~~
! Missing \endcsname inserted.
<to be read again> 
                   \BooleanFalse 
l.34 \ref{abc} \ref{
                    abc}
~~~~

as babel redefines \ref in an incompatible way (and use \let on a now robust command)

~~~~
\ifin@
  \bbl@redefinerobust\ref#1{%
    \@safe@activestrue\org@ref{#1}\@safe@activesfalse}
  \bbl@redefinerobust\pageref#1{%
    \@safe@activestrue\org@pageref{#1}\@safe@activesfalse}
\else
  \let\org@ref\ref
  \let\org@pageref\pageref
\fi
~~~~

@jbezos could we discuss what babel needs here so that it no longer tries to patch \ref and \pageref ?

base/ltxref.dtx
```diff
@@ -87,6 +89,10 @@
 %  leaving a space on both sides of it is equivalent to leaving
 %  a space on either side.
 %
+%  Note: the starred version |\ref*| and |\pageref*| are provided
+%  to align with the use of hyperref. Without \pkg{hyperref} (or some
+%  other package using the starred form) it is simply ignored.
```
Skillmon, line 94: Imho, the wording here is a bit misleading, I'd change it to `the star is simply ignored` instead of "it".

base/ltxref.dtx
```diff
@@ -87,6 +89,10 @@
 %  leaving a space on both sides of it is equivalent to leaving
 %  a space on either side.
 %
+%  Note: the starred version |\ref*| and |\pageref*| are provided
```
Skillmon, line 92: "version" -> "versions"

base/ltxref.dtx
```diff
@@ -63,17 +63,19 @@
...
 % \section{Cross Referencing}
 %  The user writes  |\label|\marg{foo}  to define the following
 %  cross-references:
 %
-%   |\ref|\marg{foo}: value of most recently incremented referenceable
+%   |\ref|*\marg{foo}: value of most recently incremented referenceable
 %             counter. in the current environment. (Chapter, section,
-%             theorem and enumeration counters are
-%             referenceable, footnote counters are not.)
+%             theorem, footnote and enumeration counters and other counter
```
Skillmon, line 74: "counter" -> "counters"

PhelypeOleinik (approved):

base/ltxref.dtx
```diff
@@ -258,10 +258,20 @@
...
 %    \begin{macrocode}
-\def\@kernel@pageref@exp#1{\csname cs_if_exist:cTF\endcsname
-   {r@#1}{\csname tl_item:cn\endcsname{r@#1}{2}}{0}}
-\def\@kernel@ref@exp#1{\csname cs_if_exist:cTF\endcsname
-   {r@#1}{\csname tl_item:cn\endcsname{r@#1}{1}}{0}}
+\ExplSyntaxOn
+\cs_new:Npn \@kernel@pageref@exp #1
+  {
+    \cs_if_exist:cTF {r@#1}
+      { \exp_last_unbraced:Nv \use_ii:nn {r@#1} }
```
PhelypeOleinik, line 265: This is easier than `\tl_item:cn`, but assumes that `\r@#1` will always contain two items. I believe that's also true when `hyperref` is loaded, right?

u-fischer: No sorry this isn't correct. With hyperref there are five items and it will break there. This should be reverted. I don't mind the \ExplSyntaxOn/Off, but should one really switch that on and off only for two csnames?

PhelypeOleinik: Ah, that's what I feared.  I used `expl3` syntax because it's less work and looks cleaner than building the csname (and if gradually we're moving to `expl3`, I don't see why not).

But mainly I think `\tl_item:nn` is (maybe unnecessarily) slow compared to `\use_<item>:nn` for a often-used command like `\ref`.  Maybe (regardless if `expl3` syntax or otherwise):
```latex
\cs_new:Npn \@kernel@ref@exp #1
  {
    \cs_if_exist:cTF {r@#1}
      { \exp_last_unbraced:Nv \@kernel@ref@get {r@#1} }
      { 0 }
  }
\cs_new:Npn \@kernel@ref@get #1 #2 {#1}
```
then `hyperref` can simply redefine `\cs_gset:Npn \@kernel@ref@get #1 #2 #3 #4 #5 {#1}`.

FrankMittelbach: > then `hyperref` can simply redefine `\cs_gset:Npn \@kernel@ref@get #1 #2 #3 #4 #5 {#1}`.

we should not encourage changes to code marked kernel by non-kernel packages (even if our own), it looks like as we would be promoting that.

PhelypeOleinik: The name could be changed, of course, but yeah, there's that. `\tl_item:Nn` wouldn't have that problem, so I'm fine with that too.

u-fischer: beside not encouraging redefinition of `@kernel` commands: I think we should also try to define things so that there is no need for external packages to redefine interna. The more redefinitions I can take out of hyperref the better. Also once we have decided about how to extend the reference system  the number/type of arguments should be unified too and then we  can exchange that with a faster version.

Skillmon: Maybe use `\tl_head:e { \exp_args:Nv \tl_tail:n { r@#1 } }` instead?

Skillmon: The `\tl_head:e/tl_tail:n` combo should be considerably faster than `\tl_item:n`, but if you're really after maximum performance here, the following should be even faster:

```
\cs_new:Npn \use_item_ii:n #1
  {
    \__kernel_exp_not:w
    \tex_expanded:D
      {
        \if_false: { \fi: \__use_item_ii_aux:w #1 }
      }
  }
\cs_new:Npn \__use_item_ii_aux:w #1#2
  {
    { \__kernel_exp_not:w {#2} }
    \exp_after:wN \use_none:n \exp_after:wN { \if_false: } \fi:
  }
```

u-fischer: no I'm not after maximum performance, I like readable code more ;-). I'm okay with `\tl_head:v { r@#1 }` and `\tl_head:e { \tl_tail:v { r@#1 }}` for the first two (but imho it is already difficult see which item `\tl_head:e { \tl_tail:e { \tl_tail:e { \tl_tail:v { r@#1 } }}}` would extract.

Skillmon: I'd not use multiple nested `\tl_tail:(n|e)` either :) And I understand the point about readable code (though I think at the top-level the `\use_item_ii:n` would be readable, and the implementation is just as cryptic as the implementation of `\tl_head:n` or `\tl_tail:n`; but this would be an otherwise useless addition to the kernel).

jbezos:
@u-fischer  “if babel would make safe=none the default”. It seems sensible (`none` is in fact “strongly“ recommended), but even if the default setting is changed to `none` (testing right now), the former default should work somehow in case it has been set explicitly in documents or some code rely on it.

u-fischer:
@jbezos in my opinion the redefinitions of \ref and \pageref in babel should be removed as no longer needed and the option `safe=ref` should be a no-op. I don't think that a package will break because of this; but in the case that there is problem it will be easier to handle in the kernel if there is one package less to consider (beside hyperref, nameref, showkeys ...) that redefines these commands.

FrankMittelbach:
> ```
> 
> ```
> 
> 
>     
>   
> 
> as babel redefines \ref in an incompatible way (and use \let on a now robust command)
> 
> ```
> \ifin@
>   \bbl@redefinerobust\ref#1{%
>     \@safe@activestrue\org@ref{#1}\@safe@activesfalse}
>   \bbl@redefinerobust\pageref#1{%
>     \@safe@activestrue\org@pageref{#1}\@safe@activesfalse}
> \else
>   \let\org@ref\ref
>   \let\org@pageref\pageref
> \fi
> ```
> 
> 
>     
>   
> 
> @jbezos could we discuss what babel needs here so that it no longer tries to patch \ref and \pageref ?

Looking at this code, all that would be needed are suitable hooks for babel to add `\@safe@activestrue` and `\@safe@activefalse` in case the kernel is new enough. Probably using `cmd/ref/before` and `cmd/ref/after`is already fine, if not they could be explicitly added in the right places. This way there wouldn't be any special patching any more.

u-fischer:
@FrankMittelbach imho the redefinitions aren't needed at all nowadays as active chars and shorthands are safe inside ref even without them.  They are simply remains from older times.  E.g.

~~~~
\documentclass{article}
\usepackage[safe=none,french,ngerman]{babel}

\begin{document}

\section{a}\label{sec:grüß"=e}
\ref{sec:grüß"=e},\ref{sec2:grüß"=e}
sec:grüß"=e
\selectlanguage{french}

\section{a}\label{sec2:grüß"=e}

\ref{sec:grüß"=e},\ref{sec2:grüß"=e}
sec:grüß"=e
\end{document}
~~~~

jbezos:
`AddToHook` seems the right “somehow“. I was planning a version in a week or so. I’ll attempt to do it (and afer that I’ll analyze what other things I can handle with hooks, although there are already some).

u-fischer:
@jbezos please could you give an example why you think you need to change \ref at all?

FrankMittelbach:
> 

what about our famous spanish `>` ?

u-fischer:
@FrankMittelbach One get errors with spanish, (not with < but with "=). And not only in \ref, but also in `\@newl@bel` which babel redefines too. But I'm not quite sure why, the same shorthand definition works with german. Do you think we should add \@safe@activestrue to the kernel definitions?

jbezos:
Re `\@safe@activestrue`, I’ve always dreamed of removing it altogether.

u-fischer:
@jbezos  I'm quite willing to help with removing it, but why does spanish needs it?

jbezos:
@u-fischer Thanks, but it doesn't seem (technically) very difficult. I’m currently finishing the compatibility with `amsmath` (I _do_ need to finish it, for my own sanity, after struggling four months with the vertical spacing), but I think I can cope with both.

u-fischer:
@jbezos  does that mean that you will change spanish.ldf so that its shorthands no longer relies on `\@safe@activestrue`  but works like e.g. the german ones?

jbezos:
Yes, or something like that.

u-fischer:
That would be wonderful. Then I could perhaps start to remove all these `\@safe@activestrue` from hyperref and nameref. Do you know other language files which still rely on it?

jbezos:
Lots of test files failing with `safe=none`. Must investigate. Please, allow me a few weeks. This is not a trivial change at all, and there are many packages using `\@safe@actives`.

u-fischer:
@jbezos I changed in babel `\def\bbl@opt@safe{BR}` to `\def\bbl@opt@safe{B}` and then ran the testsuite with the latex-dev from the branch here and got no problematic failures. Some tests failed because of the new marks, some due to changes in \MakeUppercase, but nothing related to references. (If I disable also the bib/cite patches, one get more failures,  so I would leave that for now, but imho we should at least make the kernel \cite safe so that babel doesn't have to patch it). Do you have elsewhere other test files?

jbezos:
@u-fischer With the babel test files in GitHub, many of them failed because of `\cite`. In those failing with `\ref` they are apparently harmless changes in the messages. I’ve started working on it.

FrankMittelbach (approved):

PR #816, opened 2022-05-13
add no-op commands for the new linktarget interface of hyperref

u-fischer:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- Ready to merge

- [ ] Test file(s) added


base/doc/ltnews35.tex
```diff
@@ -183,6 +183,9 @@ \section{Document metadata interface}
 already existing keys are in \file{l3meta.pdf} and \file{documentmetadata-support.pdf}
 and also in the documentation of the \pkg{pdfmanagement-testphase} package.
 
+Package and class authors can test if a user has used \cs{DocumentMetadata}
+and so loaded the pdfmanagement package with \cs{IfPDFManagementActiveTF}. 
+
```
FrankMittelbach, line 188: My initial reaction would be that this is the wrong interface -- convince me otherwise :-) I think it makes more sense going forward to have a test that  would be called \IfDocumentMetadataTF because that is the deciding factor from an interface perspective

base/ltfinal.dtx
```diff
@@ -1325,7 +1325,25 @@
...
+%  \begin{macro}{\MakeLinkTarget}
+%  \begin{macro}{\LinkTargetOn}
+%  \begin{macro}{\LinkTargetOff}
+%  \begin{macro}{\NextLinkTarget}
+% \changes{v2.2u}{2022/05/10}{Default definition for targets added}
+%    \begin{macrocode}
+\NewDocumentCommand\MakeLinkTarget{sO{}m}{}
+\NewDocumentCommand\LinkTargetOn{}{}
+\NewDocumentCommand\LinkTargetOff{}{}
+\NewDocumentCommand\NextLinkTarget{m}{}
```
FrankMittelbach, line 1341: As mentioned by @zauguin  those should probably do more than nothing but insert a whatsit depending on the engine so that spacing is identical whether or not real actions happen.

base/ltfinal.dtx
```diff
@@ -1325,7 +1325,25 @@
...
 %  \end{macro}
 %
+%   To allow package and class author to support for document links
+%   we provide also the new interface commands of the hyperref package
+%   for the creation of targets.
 %
+%  \begin{macro}{\MakeLinkTarget}
+%  \begin{macro}{\LinkTargetOn}
+%  \begin{macro}{\LinkTargetOff}
+%  \begin{macro}{\NextLinkTarget}
```
FrankMittelbach, line 1335: there should probably be a rollback/rollforward as discussed.

FrankMittelbach (changes requested):

base/doc/ltnews35.tex
u-fischer: Well that command is not new code. It was only undocumented and so I added a sentence. But yes, I think in view of the development the focus has switched a bit and \IfDocumentMetadataTF is a better name.

base/ltfinal.dtx
u-fischer: Only the first needs imho a content. In vertical mode a special can be used, but in hmode hyperref inserts a box and surrounds it with `\Hy@SaveSpaceFactor`  and`\Hy@RestoreSpaceFactor`.  How would one  reproduce that at best?

base/ltfinal.dtx
u-fischer: i will try to figure it out ;-)

FrankMittelbach: I can do that for you if you like

base/ltfinal.dtx
FrankMittelbach: well you can mimic the mechanism if that's what it should do, because then this behavior would bessentially part of the command not part of what hyperref does with it (perhaps, dependes on where how it could/should be used in other places, but my guess is that this is *part of the command* in reality)

base/ltmeta.dtx
```diff
@@ -122,11 +122,12 @@
 %    \end{macrocode}
 %
 %    \begin{macrocode}
-%<latexrelease>\NewModuleRelease{2021/06/01}{ltmeta}
+%<latexrelease>\NewModuleRelease{2022/06/01}{ltmeta}
 %<latexrelease>                 {Document Metadata handling}
```
FrankMittelbach, line 126: guess that's wrong. This makes the  module appear only in 2022/06 but it was added in 2021.

u-fischer: the first commit was in dezember 2021, and \DocumentMetadata is currently defined only in latex-dev, so that makes it 2022 doesn't it?

base/ltfinal.dtx
FrankMittelbach: I guess that remark is obsolete (I thought we introduced \DocumentMetadata already in the previous release but since that is not the case it rolls back with the new module)

FrankMittelbach (approved):

base/ltfinal.dtx
u-fischer: Really? the hyperref commands are in ltfinal and not in ltmeta. (I could move them there if you think the location is better. I used ltfinal only because the IfPDFManagement-test was there).

FrankMittelbach: ah right, but they belong there don't they it is part of the ltmeta machinery by the end of the day.

u-fischer: ok I moved them to ltmeta and reverted the change in ltfinal.

PR #831, opened 2022-05-29
Set endlinechar earlier in v-type argument parser

user202729:
Originally mentioned in https://github.com/latex3/latex2e/issues/876.

Currently `\command%something%` does not absorb the content between the `%`. Moving the `dospecial` to before in addition will break it. (on the other hand setting endlinechar earlier should not break anything, as far as I can see.

# Internal housekeeping

## Status of pull request

- Feedback wanted (*is this desirable? Before I take a look at the items below*)

- [ ] Test file(s) added

### DIFF
--- a/base/ltcmd.dtx
+++ b/base/ltcmd.dtx
@@ -3132,6 +3132,7 @@
     \tl_set:Nn \l_@@_signature_tl {#1}
     \group_begin:
       \tex_escapechar:D = 92 \scan_stop:
+      \tex_endlinechar:D = `\^^M \scan_stop:
       \tl_clear:N \l_@@_v_arg_tl
       \peek_remove_spaces:n
         {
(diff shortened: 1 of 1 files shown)


josephwright:
I think this likely looks OK: lets see if I can get someone else to check too!

car222222:
And What is the probability of that? :-)

FrankMittelbach:
According to the LaTeX manual % can be used as char after \verb and for the kernel \verb that is in fact the case. Thus, making it also work for the v argument seems reasonable to me, but before merging it should get some news entry as well as perhaps a test file.

user202729:
I think something needs to be clarified here. This patch does not modify that behavior, these remains the same...

```
%! TEX program lualatex
\documentclass{article}
\begin{document}

\NewDocumentCommand\test{+v}{#1}

% current behavior: 123 is typeset verbatim
\verb%123%
456

% current behavior: 456 becomes the argument
\test%123%
|456|
```

it addresses this case.

```
\ExplSyntaxOn
% current behavior: 123<space> 456<newline> 789
% after the patch: 123<newline> 456<newline> 789
\test
|123
456
789|
\ExplSyntaxOff

\end{document}
```

The other one could be patched as well but that may need some deeper patches. (also the code to catch the opening `{` I think)

josephwright:
The code change looks fine but we do need the other 'housekeeping' here: please look over the tick list and make the required changes.

PhelypeOleinik:
Replaced by #1038.

PR #832, opened 2022-05-29
keyval.dtx: Fix a typo

dr-scsi:
* required/graphics/keyval.dtx (section{The Internal Interface}):
Fix a typo in the macro defined by \define@key.


davidcarlisle:
hmm I suppose so, `prefix`was the original term as shown in internal macros like `\def\KV@prefix{KV@#1@}%` the change to "family" was a documentation revision (possibly before the first relase:-) But the suggested change is just doc so OK...

Current policy would have us update (just) the package date for such a change.

dr-scsi:
Thanks for your response.  I updated the PR with a change of the package date (only).

PR #842, opened 2022-06-03
adjust ltxdoc to process existing dtx files

davidcarlisle:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [n/a] Test file(s) added - docs are the test

### DIFF
--- a/base/doc.dtx
+++ b/base/doc.dtx
@@ -148,8 +148,6 @@
 %
 % \newcommand\env{\texttt}
 % \newcommand\opt{\texttt}
-% \newcommand\cls{\texttt}
-% \newcommand\pkg{\texttt}
 % \newcommand\prg{\textsf}
 %
 % \newcommand\DOX{\env{DoX}\xspace}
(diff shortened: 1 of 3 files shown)


base/doc.dtx
```diff
@@ -148,8 +148,6 @@
 %
 % \newcommand\env{\texttt}
 % \newcommand\opt{\texttt}
-% \newcommand\cls{\texttt}
-% \newcommand\pkg{\texttt}
 % \newcommand\prg{\textsf}
```
FrankMittelbach, line 151: I guess I would probably still vote for doing \providecommand  (for all of them) rather than droppping some and keeping others.

There should be a at date change in the ProvidesPackage line (current version of doc is out in the wild)

FrankMittelbach (approved):
basically ok I guess, but date line change in doc please

base/doc.dtx
davidcarlisle: oh? I looked and it was only dated yeserday so I left it, I'll bump the date and add \providecommand

PR #865, opened 2022-06-19
Correct option handling logic

josephwright:


base/ltkeys.dtx
```diff
@@ -299,20 +300,20 @@
...
           }
       }
   }
-\cs_new_protected:Npn \@@_options_remove:nnn #1#2#3
+\cs_new_protected:Npn \@@_options_remove:nnnN #1#2#3#4
   {
     \keys_if_exist:nnTF {#3} {#1}
       { \clist_put_right:Nn \l_@@_options_clist {#2} }
-      { \clist_put_right:Nn \@unusedoptionlist {#1} }
+      { #4 \@unusedoptionlist {#1} }
```
u-fischer, line 314: I think the logic is wrong for a package. Nothing should happen if the key is *unknown*,  The "remove from \@unusedoptionlist" must be in the true branch. (the logic is different for a class if I got the code right).

josephwright: I'll try again a bit later on then :)

PR #867, opened 2022-06-20
Use raw options for keyval handing

josephwright:


josephwright:
@skillmon Could you take a look here too?

davidcarlisle (approved):
looks fine I noticed one typo in the comments, that could be picked up before merging, but not essential

davidcarlisle:
test github-0710 is showing a diff but I don't see that locally (general-option-C  = 333 being set or not)

base/testfiles/github-0862.lvt
```diff
@@ -0,0 +1,16 @@
+\documentclass{article}
+
+\begin{filecontents}[force]{\jobname.sty}
+  \ProvidesPackage{\jobname}
+  \DeclareKeys{foo .code = \typeout{Foo was gives: \detokenize{#1}}}
```
Skillmon, line 5: ~~gives~~ given

PR #873, opened 2022-06-23
Introduce \stock... dimens

josephwright:


base/classes.dtx
```diff
@@ -519,6 +519,14 @@
 %</article|report|book>
 %    \end{macrocode}
 %
+% \changes{1.4o}{2022/06/23}{Set \cs{stock\dots} registers}
+% No trim marks in the standard classes, so the stock and the paper are
+% the same size.
+%    \begin{macrocode}
+\stockheight=\paperheight
```
davidcarlisle, line 526: I think you need to initialize to 0pt not `\paperheight` otherwise you can not detect when `\stockheight` has been set , which is needed to implement

> set `\pdfpageheight`  to `\stockheight`  if set, else `\paperheight`

davidcarlisle (changes requested):

base/classes.dtx
josephwright: Updated version only creates the names, it never sets them.

josephwright:
Related plan for the l3 backend at https://github.com/latex3/latex3/tree/setpdfpagesize-alt

davidcarlisle (approved):
looks good, it also means the backend change at
https://github.com/latex3/latex3/compare/main...setpdfpagesize-alt
can be simplified to drop the is defined test  (possibly, depending on timings involved in using expl3 with other formats)

josephwright:
> looks good, it also means the backend change at [latex3/latex3@main...setpdfpagesize-alt](https://github.com/latex3/latex3/compare/main...setpdfpagesize-alt) can be simplified to drop the is defined test (possibly, depending on timings involved in using expl3 with other formats)

Older-format-newer-`expl3` ...

davidcarlisle:
@josephwright  
> Older-format-newer-expl3 ...
yes sure although that's a temporary change managment, but I was more wondering about other formats, as \stock... is allocated in 2e code, would we keep the defined test long term or drop it (and arrange the registers are allocated in other expl3 wrappers for plain etc (this PR doesn't depend on deciding either way)

base/ltoutput.dtx
```diff
@@ -30,7 +30,7 @@
 %%% From File: ltoutput.dtx
 %<flafter>\ProvidesPackage{flafter}
 %<fltrace>\ProvidesPackage{fltrace}
-%<flafter,fltrace>       [2021/07/31 v1.4e
+%<flafter,fltrace>       [2022/06/23 v1.4f
```
FrankMittelbach, line 33: this looks wrong. Why should those register affect the sty files? I think you update the wrong date here (not that it matters much)

PR #890, opened 2022-07-04
Introduce 2e names for new expl3 case functions

josephwright:


base/ltfinal.dtx
```diff
@@ -1064,8 +1064,11 @@
 %
 % \begin{macro}{\MakeUppercase}
 % \begin{macro}{\MakeUppercase}
```
davidcarlisle, line 1066: MakeLowercase (been wrong for years it seems, but could fix it)

base/ltfinal.dtx
```diff
@@ -1081,6 +1084,7 @@
 %                {Use \cs{@expl@text@uppercase@@n}, removing local redefinition of \cs{UTF@two@octets@noexpand}}%
 % \changes{v2.2t}{2022/04/21}{Support \cs{noexpand} in argument of \cs{@expl@text@uppercase@@n}}
 % \changes{v2.2v}{2022/06/30}{Just use \cs{text_lowercase:n} without \cs{protectd@edf} gh/881x}%
+% \changes{v2.2w}{2022/07/04}{Introduced \cs{CaseSwitch}, \cs{DeclareCaseChangeEquivalent} and \cs{MakeTitlecase} to support hooking into case changing gh/889}%
 % Two wrappers around the L3 case changing functions.
```
davidcarlisle, line 1088: not "Two" any more, may change again so

> Wrappers around...

davidcarlisle (approved):
approving, but a couple of cosmetic doc comments inline

FrankMittelbach (changes requested):
In my opinion this should have documentation in usrguide3 when it gets introduced (and just somewhere in interface3). I think we should add this now even if at some point we want to merge usrguide and usrguide3

josephwright:
@FrankMittelbach  Looking at the existing docs, `\MakeUppercase`/`\MakeLowercase` are in `clsguide`. I think at least `\MakeTitlecase` would go with them; to me, having all in the same place makes more sense. (That should include I suspect `\NoCaseChange`). Do you want me to add to `clsguide` or to make a new section in `usrguide3` and summarise there? If we do the latter, a proper tidy up can go into the Fall 2022 release.

base/ltfinal.dtx
josephwright: I don't follow

FrankMittelbach:
>  Do you want me to add to `clsguide` or to make a new section in `usrguide3` and summarise there? If we do the latter, a proper tidy up can go into the Fall 2022 release.

There are arguments both ways, some are clearly user facing other more likely to be used only by class code. My main point is that they should be described now not "at some point in the future", so perhaps, for now a section in usrguide3 is best and we resort when we combine old and new usr and cls guides.

josephwright:
I've added something to `usrguide3`. If it looks OK, after the merge I'll copy the text to my `usrguide` branch so it's set up for the change we plan, and probably then will look at a split between the part that is user facing and the part that belongs in `clsguide`. Although the latter is in much better shape, I think we do need to generally check that the docs are up to date - there are definitely a few things described as 'new' that are now 25+ years old.

FrankMittelbach:
> I think we do need to generally check that the docs are up to date - there are definitely a few things described as 'new' that are now 25+ years old.

closer to 30 by now, so yes, that should happen as part of combining/resorting the usr and cls guides.

PR #891, opened 2022-07-05
Support \CurrentOption for unknown keyval options

josephwright:


FrankMittelbach (commented):
Is \CurrentOption also supported for known keys and only the unknown key case was missing? (sorry no time to actually check myself)

josephwright:
> Is \CurrentOption also supported for known keys and only the unknown key case was missing? (sorry no time to actually check myself)

With the change, `\CurrentOption` is available to all keys set this way, yes: I think one would likely _not_ use this in general as it would go wrong if the keys are used outside of option processing, but that's a different matter!

PR #907, opened 2022-08-10
Add "={...}" modifier and suppress brace stripping

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [x] Test file(s) added


josephwright:
I think the obvious questions here are

- Do we want to apply to all argument types
- Are we happy with a non-expandable approach at present (it's easier than doing by expansion, and I think realistically not needed for expandable commands)
- Does the fuzzy logic for detecting keyvals match our expectations?

base/doc/usrguide3.tex
```diff
@@ -489,6 +493,79 @@ \subsection{Testing special values}
...
+\begin{verbatim}
+\DeclareDocumentCommand
+  \caption
+  {s ={short-text} +O{#3} +m}
+  {%
+    \showtokens{Grabbed arguments:^^J(#2)^^Jand^^J(#3)}%
+  }
+\end{verbatim}
+the default key name is \texttt{short-text}. When the command \cs{caption} is
+then used, if the mandatory argument is free text such as
```
Skillmon, line 524: ~~mandatory~~ optional

base/ltcmd.dtx
```diff
@@ -1224,12 +1251,12 @@
 % \end{macro}
 %
 % \begin{macro}{\@@_allowed_token_check:N}
-%   Some tokens are now allowed as delimiters for some argument types,
+%   Some tokens are not allowed as delimiters for some argument types,
 %   notably implicit begin/end-group tokens (|\bgroup|/|\egroup|).
 %   The major problem with these tokens is that for |\peek_...| functions,
 %   a literal~|{|$_1$. is virtually indistinguishable from a |\bgroup| or
```
Skillmon, line 1257: ~~literal~|{|$_1$.~~ literal~|{|$_1$ (no period)

base/ltcmd.dtx
```diff
@@ -1484,6 +1512,22 @@
...
 %
+% \begin{macro}{\@@_add_type_=:}
+% \changes{v1.1a}{2022/08/10}{Add support for \texttt{=} modifier}
+%   A mix of the ideas from above: set a flag and add a processor.
+%    \begin{macrocode}
+\cs_new_protected:cpn { @@_add_type_=:w } #1
+  {
+    \@@_flush_m_args:
+    \bool_set_true:N \l_@@_prefixed_bool
+    \bool_set_true:N \l_@@_suppress_sstrip_Bool
```
Skillmon, line 1523: Maybe the following is a stupid question, but why is `Bool` upper cased?

Skillmon: And why the double `s` on `sstrip`?

base/ltcmd.dtx
```diff
@@ -3873,6 +3954,166 @@
...
+%   |=,| then it is forced to be key--value. To avoid needing to worry about
+%   catcodes for this, and to allow spaces around the |=|, we use a
+%   series of steps rather than a delimited argument.
+%     \begin{macrocode}
+\cs_new_protected:Npn \@@_arg_to_keyvalue:nn #1#2
+  {
+    \str_if_eq:eeTF { \exp_not:n {#2} } { \exp_not:o { { \use:n #2 } } }
+      { \tl_set:Nn \ProcessedArgument { #1 = #2 } }
+      {
+        \exp_args:Ne \@@_arg_to_keyvalue_auxi:nnn
```
Skillmon, line 3989: `\tl_trim_spaces_apply:nN`?

base/ltcmd.dtx
```diff
@@ -3873,6 +3954,166 @@
...
+    \tl_if_head_is_N_type:nTF {#1}
+      { \@@_arg_to_keyvalue_auxii:nnNw {#2} {#3} #1 \q_@@_stop }
+      { \@@_arg_to_keyvalue_auxv:nn {#2} {#3} }
+  }
+\cs_new_protected:Npn \@@_arg_to_keyvalue_auxii:nnNw
+  #1#2#3#4 \q_@@_stop
+  {
+    \str_if_eq:nnTF {#3} { = }
+      {
+        \exp_args:Ne \@@_arg_to_keyvalue_auxiii:nnn
```
Skillmon, line 4004: Again `\tl_trim_spaces_apply:nN`.

base/ltcmd.dtx
```diff
@@ -3873,6 +3954,166 @@
...
+      }
+        {#1} {#2} #3 \q_recursion_stop
+  }
+\cs_new_protected:Npn \@@_arg_to_keyvalue_loop_group:nnn #1#2#3
+  { \@@_arg_to_keyvalue_loop:nnw {#1} {#2} }
+\use:x
+  {
+    \cs_new_protected:Npn
+      \exp_not:N \@@_arg_to_keyvalue_loop_space:nnw ##1##2 \c_space_tl
+  }
```
Skillmon, line 4052: Easier way to get a space delimited macro: `\use:n { \cs_new_protected:Npn \@@_arg_to_keyvalue_loop_space:nnw #1#2 } ~ {<stuff>}`

base/ltcmd.dtx
josephwright: Two things: would mean setting something to carry `#1`/`#2`, as we don't currently have an `:nn` version, and I have one eye on whether at some point I'll need to make the code expandable

base/ltcmd.dtx
josephwright: We tend to avoid having only part of the parameter spec being manipulated: using `\use:x` everything is inside the `{...}`, so doubling of `#` is consistent, etc. Using `\use:n` we have a mixed situation, which is not ideal.

base/ltcmd.dtx
Skillmon: That's not true, as you're only changing the first argument `#1`/`#2` are curried and don't need to be stored. `\tl_trim_spaces_apply:nN {#2} \@@_arg_to_keyvalue_auxi:nnn {#1} {#2}` behaves the same.

josephwright: That occurred to me after I'd posted my reply. I guess in general I'm wary of relying on arguments on the stack other than where required: at the `\tracingall` level it's much nicer to see `\foo:nnn` with `#1`, `#2` and `#3` than have to rely on having them noted down as 'still about'. Of course, there is an argument for performance, and I can see that one could argue the clarity situation both ways. Lets see what others think here.

base/doc/ltnews36.tex
```diff
@@ -180,6 +180,27 @@ \subsection{EC sans serif at small sizes}
...
+\end{verbatim}
+which would mean that if the optional argument does \emph{not}
+contain keyval data, it will be converted to a single keyval
+pair with the key name \texttt{short-text}.
+
+Arguments which begin with \texttt{=,} are always interpreted as
+keyvals even if they do not contain further \texttt{=} signs.
+Any \texttt{=} signs within math mode are ignored, meaning that
+only \texttt{=} outside of math mode will generally cause
+interpretation as keyval material.
```
FrankMittelbach, line 203: Math mode being determined how? Only `$...$` or also other constructs? Just saw that `\(`is also supported. If that is costly I'm not sure that it is really necessary but I must confess that I don't know how manypeople use this unreadable convention so maybe it is.

base/doc/usrguide3.tex
```diff
@@ -489,6 +493,79 @@ \subsection{Testing special values}
...
+\begin{verbatim}
+Grabbed arguments:
+(label = cap:demo)
+and
+(A much longer and more detailed text for demonstration purposes)
+\end{verbatim}
+
+Interpretation as keyval form is determinded by the presence of \texttt{=}
+characters within the argument. Those in math mode (inside \verb|$...$| or
+\verb|\(...\)|) are ignored. An argument can be forced to be read as keyvals by
```
FrankMittelbach, line 551: I think this can and should be a bit more repetitive with examples on what to do if math is not detected because it is caused by something other than `$`or `\(`.

FrankMittelbach (approved):
Only reviewed the overall docs not the details of the implementation

FrankMittelbach:
> I think the obvious questions here are
> 
>     * Do we want to apply to all argument types

Maybe we want to allow for it (as  the code does now unless I'm totally misreading it), but I would suggest to explicitly suggest to limit its use as "best practice" to only normal optionals and preferable the first argument
 
>     * Are we happy with a non-expandable approach at present (it's easier than doing by expansion, and I think realistically not needed for expandable commands)

expandable would mean that you could have an expandable command that in an expansion situation parses the arguments and turns them into key/vals, correct? I can't really see usecase for it, because processing those keyvals in the payload would then most likely be not expandable in places. So I think that restriction is fine

>     * Does the fuzzy logic for detecting keyvals match our expectations?

to me it looks good.

Skillmon:
> I can't really see usecase for it, because processing those keyvals in the payload would then most likely be not expandable in places.

Oh no, all my expandable key=value work goes to waste :) I can understand if the full rule set isn't implemented expandably (and maybe I'll find the time somewhen in the future to implement an expandable alternative)

josephwright:
> > I can't really see usecase for it, because processing those keyvals in the payload would then most likely be not expandable in places.
> 
> Oh no, all my expandable key=value work goes to waste :) I can understand if the full rule set isn't implemented expandably (and maybe I'll find the time somewhen in the future to implement an expandable alternative)

It's not impossible, of course, it's just that the combination of

- An existing document command with a free text (optional) argument
- A move of this argument to a keyval form
- A need to work by expansion in some context

seems extremely unlikely, so the effort isn't at this stage worth it. I've quite deliberately set up such that we could revisit this if there are use cases.

base/ltcmd.dtx
```diff
@@ -3873,6 +3954,166 @@
...
+% \begin{macro}{\@@_arg_to_keyvalue_math_space:nnw}
+%   If the entire argument is braced, we treat as free text and return as
+%   the value for the text key. Alternatively, if the start of the input is
+%   |=,| then it is forced to be key--value. To avoid needing to worry about
+%   catcodes for this, and to allow spaces around the |=|, we use a
+%   series of steps rather than a delimited argument.
+%     \begin{macrocode}
+\cs_new_protected:Npn \@@_arg_to_keyvalue:nn #1#2
+  {
+    \str_if_eq:eeTF { \exp_not:n {#2} } { \exp_not:o { { \use:n #2 } } }
```
PhelypeOleinik, line 3986: The `\exp_not:o` here does nothing useful.  I added a `\exp_after:wN` here, but the test output for:
```
    \DeclareDocumentCommand \foo { = { TOC-entry } m } { \TYPE { \detokenize { (#1) } } }
    \foo { { baz } }
```
changes from `(TOC-entry={{baz}})` to `(TOC-entry={baz})`.  I'm not sure which is correct...

PhelypeOleinik: If I understand correctly, this is testing if the grabbed argument is wrapped in a set of braces.  If so, this fails for the (likely unsupported) case of different bgroup and egroup characters.  For example:
```
\DeclareDocumentCommand \foo { = { TOC-entry } m } { \typeout { \detokenize { (#1) } } }
\foo { { baz } }
\catcode`\[=1
\catcode`\]=2
\foo [ [ baz ] ]
```
will print:
```
(TOC-entry={baz})
(TOC-entry={[baz]})
```

A version that protects against that would use (I think) `\tl_trim_spaces:n`, then `\tl_if_head_is_group:nTF`, and then test `\tl_if_blank:oTF { \use_none:n #2 }`.  Something like this:
```
\cs_new_protected:Npn \@@_arg_to_keyvalue:nn #1#2
  { \tl_trim_spaces_apply:nN {#2} \@@_arg_to_keyvalue_braces:nn {#1} }
\cs_new_protected:Npn \@@_arg_to_keyvalue_braces:nn #1 #2
  {
    \tl_if_head_is_group:nT {#1}
      {
        \tl_if_blank:oT { \use_none:n #1 }
          {
            \tl_set:Nn \ProcessedArgument { #2 = #1 }
            \use_none:nn
          }
      }
    \use:n
      {
        \exp_args:Ne \@@_arg_to_keyvalue_auxi:nnn
          { \tl_trim_spaces:n {#1} } {#2} {#1}
      }
  }
```

PhelypeOleinik (commented):
More to come tomorrow (sorry :)

base/ltcmd.dtx
josephwright: I've made the change also normalising the brace group

base/testfiles-ltcmd/ltcmd008.lvt
```diff
@@ -0,0 +1,66 @@
...
+    \foo [ bar ] { baz }
+  }
+
+\TEST { Collecting~keyvals }
+  {
+    \DeclareDocumentCommand \foo { = { TOC-entry } O{ } m } { \TYPE { \detokenize { (#1) (#2) } } }
+    \foo [ bong ] { baz }
+    \foo [ bong = bang ] { baz }
+    \foo [ = , bong ] { baz }
+    \foo [ { bong =  bang } ] { baz }
```
PhelypeOleinik, line 27: I'm not sure about the expected outcome of this test.  Are the braces supposed to hide the keyval as a free text argument with an `=`?  Currently it's seeing `\foo [ { bong =  bang } ] { baz }` exactly like `\foo [ bong = bang ] { baz }` because the `D` argument grabber removes a set of braces from within the delimiters.

josephwright: The brace stripping is supposed to be suppressed for these cases (which is why this is more than a processor), which means that the expected result is `TOC-entry={bong=bang}`. That _was_ the case, but I see it changed in 11cc8bd3706d8241065386ae10531f9d75437025 ...

josephwright: ... but as you adjusted other stuff after that, it will take me a while to track down what's gone wrong. If I have a chance I will do.

base/ltcmd.dtx
```diff
@@ -4037,68 +4039,70 @@
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_arg_to_keyvalue_auxv:nn #1#2
   {
-    \@@_arg_to_keyvalue_loop:nnw {#1} {#2} #2
-      \q_recursion_tail \q_recursion_stop
+    \@@_arg_to_keyvalue_loop:w #2
+      \q_recursion_tail \q_recursion_stop {#1} {#2}
```
josephwright, line 4049: I understand your thinking here, but this does make debugging quite a bit more 'interesting' as you have to remember what's on the stack all the time. (Not saying you should change it, just an observation.)

base/testfiles-ltcmd/ltcmd008.lvt
FrankMittelbach: `\foo [ { bong = bang } ] { baz }`  is supposed to turn the argument in a free text one, just like `\item[{foo [x] bar}]` in  classic LaTeX hides the  `]` in the item text so that it is not mistaken for the end of the optional argument. 
This is the std way people do this therefore has to be maintained. We can't go and say, fix your document and put the braces inside.

josephwright: I see the issue: in setting up the two aspects (brace strip suppression and post-processing to hunt for `=`), I've messed up and the brace stripping suppression 'gets lost' in the internals. I'll work out a fix shortly unless @PhelypeOleinik  beat me to it: the problem is in `\@@_prepare_signature:n` or perhaps what is done subsequently.

PhelypeOleinik: > That _was_ the case, but I see it changed in [11cc8bd](https://github.com/latex3/latex2e/commit/11cc8bd3706d8241065386ae10531f9d75437025) ...

Yeah, I had to change it there because it was comparing the wrong argument, thus the in-and-out of math mode wasn't working.  I see you already found the problem (not in the `D` grabber as I had thought).  Thanks!

base/ltcmd.dtx
PhelypeOleinik: I thought this one wasn't too convoluted (the final two arguments are just used at the very end), and it would spare a few cycles, since this is kind of “inner loop”.  I'll add something to the documentation to make it easier to follow in the future.

base/ltcmd.dtx
```diff
@@ -3996,19 +4008,15 @@
           }
       }
     \use:n
-      {
-        \tl_trim_spaces_apply:nN {#1} \@@_arg_to_keyvalue_auxi:nnn
-          {#2} {#1}
-      }
+      { \@@_arg_to_keyvalue_auxi:nnn {#1} {#2} {#3} }
```
Skillmon, line 4011: Why the `\use:n { \@@_arg_to_keyvalue_auxi:nnn {#1} {#2} {#3} }`? Just use `\@@_arg_to_keyvalue_auxi:nnn` directly and use `\use_none:nnnn` instead of `\use_none:nn`.

base/ltcmd.dtx
```diff
@@ -3996,19 +4008,15 @@
           }
```
Skillmon, line 4008: Instead of

```
            \tl_set:Nx \ProcessedArgument
              { #2 = { \exp_not:o { \use:n #1 } } }
```

you could use

```
            \tl_set:Nx \ProcessedArgument
              { #2 = { \exp_not:n #1 } }
```

(after all, you know that `#1` is a single `n`-type by now)

base/ltcmd.dtx
```diff
@@ -3996,19 +4008,15 @@
...
   {
     \tl_if_head_is_N_type:nTF {#1}
       { \@@_arg_to_keyvalue_auxii:nnNw {#2} {#3} #1 \q_@@_stop }
       { \@@_arg_to_keyvalue_auxv:nn {#2} {#3} }
   }
-\cs_new_protected:Npn \@@_arg_to_keyvalue_auxii:nnNw
-  #1#2#3#4 \q_@@_stop
+\cs_new_protected:Npn \@@_arg_to_keyvalue_auxii:nnNw #1#2#3#4 \q_@@_stop
   {
     \str_if_eq:nnTF {#3} { = }
```
Skillmon, line 4019: Though slower this should probably use something to split of the first token without imposing restrictions to the argument. I suggest the following code (if `\expanded` is required something equally fast can be built that also works for `n`-type heads if necessary):

```
\cs_new:Npn \@@_split_N_head_apply:Nn #1#2
  { \exp:w \if_false: { \fi: \@@_split_N_head_apply_aux:NNw #1#2 } }
\cs_new:Npn \@@_split_N_head_apply_aux:NNw #1#2
  {
    \exp_after:wN \exp_end:
    \exp_after:wN #1 \exp_after:wN #2 \exp_after:wN { \if_false: } \fi:
  }
```

Then change the definition of `\@@_arg_to_keyvalue_auxi:nnn` to `\@@_arg_to_keyvalue_auxiv:nnNw`:

```
\cs_new:Npn \@@_arg_to_keyvalue_auxi:nnn #1
  {
    \tl_if_head_is_N_type:nTF {#1}
      { \@@_split_N_head_apply:Nn \@@_arg_to_keyvalue_auxii:Nnnn {#1} }
      { \@@_arg_to_keyvalue_auxv:nn }
  }
\cs_new:Npn \@@_arg_to_keyvalue_auxii:Nnnn #1#2
  {
    \str_if_eq:eeTF { \exp_not:n {#1} } { = }
      { \tl_trim_spaces_apply:nN {#2} \@@_arg_to_keyvalue_auxiii:nnn }
      { \@@_arg_to_keyvalue_auxv:nn }
  }
\cs_new:Npn \@@_arg_to_keyvalue_auxiii:nnn #1
  {
    \tl_if_head_is_N_type:nTF {#1}
      { \@@_split_N_head_apply:Nn \@@_arg_to_keyvalue_auxiv:Nnnn {#1} }
      { \@@_arg_to_keyvalue_auxv:nn }
  }
\cs_new:Npn \@@_arg_to_keyvalue_auxiv:Nnnn #1#2
  {
    \str_if_eq:eeTF { \exp_not:n {#1} } { , }
      { \tl_set:Nn \ProcessedArgument {#2} \use_none:nn }
      { \@@_arg_to_keyvalue_auxv:nn }
  }
```

josephwright: I don't follow the need here: we have a standard `\tl_if_head_is_N_type:nTF` then split the `N`-type token safely. What case are you thinking fails?

PhelypeOleinik (approved):
Good to go from my end

base/ltcmd.dtx
Skillmon: `#4` might lose braces by the `w`-grabbing, `\q_@@_stop` can't be part of the argument.

Skillmon: (the latter is fine due to the private quarks rule)

PR #910, opened 2022-08-19
Gh888 [ltfilehook] Don't check provided version when a package is disabled

PhelypeOleinik:
This PR changes `\@disable@packageload@do` to prevent the version checking in `\@onefilewithoptions` (“You have requested...”) from happening when a package is prevented from loading with `\disable@package@load`, and adds an info to the log saying so.

# Internal housekeeping

## Status of pull request

- Under development

- [x] Test file(s) added


PhelypeOleinik:
I haven't added an ltnews entry yet (not sure if this needs one), and didn't add rollback for this (also not sure if it's worth it; the fix just removes a warning...).

base/testfiles-lthooks/filehook-011.tlg
```diff
@@ -18,6 +18,8 @@ File: size10.clo ....-..-.. v... Standard LaTeX file (size option)
 \bibindent=\dimen...
 )
 Your package is rubbish
+LaTeX Info: Package 'hbox.sty' has been disabled.
+            Load request ignored on input line ....
 (hbox.sty
 Package: hbox ....-..-.. v... The best package ever
 Option: opt unused
```
FrankMittelbach, line 25: why do I see hbox.sty being loaded after all?

PhelypeOleinik: In processing an `\usepackage`, LaTeX first checks if the _requested_ package was disabled, and only then it processes the file name and does file substitutions:
```latex
       \@disable@packageload@do{\@currname.\@currext}% <--- here we check if the package is disabled
         {\@expl@@@filehook@file@push@@
          \set@curr@file{\@currpath\@currname.\@currext}% <--- here we follow file substitutions
```
and the test file exercises that mechanism:
```latex
% Disable package 'hbox'
\disable@package@load{hbox}{\typeout{Your package is rubbish}}

% substitute file 'wrong-pkg.sty' by 'hbox.sty'
\declare@file@substitution{wrong-pkg.sty}{hbox.sty}

% try to load 'hbox' and fails because 'hbox' is disabled
\usepackage{hbox}

% try to load 'wrong-pkg', which is not disabled, but eventually is replaced by 'hbox.sty'
\usepackage{wrong-pkg}
```

FrankMittelbach (approved):

u-fischer (approved):

PR #933, opened 2022-10-10
protected@edef in option handler

davidcarlisle:
## Status of pull request

- [X] Feedback wanted 
  
  I have not added rollback which is tricky as parts of the option handler are pre-included
  in latexrelease, it could be considered  a bug-fix change which could have been done when
  raw option lists added, so I'm not sure rollig back just this part is worth the complication
  But rollback could be added if needed (I think:-)

- Ready to merge
 other than the above, and adding changes.txt and news.

- [X] Test file(s) added

### DIFF
--- a/base/ltclass.dtx
+++ b/base/ltclass.dtx
@@ -33,7 +33,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{ltclass.dtx}
-             [2022/06/20 v1.5c LaTeX Kernel (Class & Package Interface)]
+             [2022/10/10 v1.5d LaTeX Kernel (Class & Package Interface)]
 % \iffalse
 \documentclass{ltxdoc}
 \GetFileInfo{ltclass.dtx}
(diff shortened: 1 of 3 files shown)


FrankMittelbach (changes requested):
I'm ok with rollback to the earlier release (ie consider this a bug fix to that). If we find a good reason  to do otherwise we can still add that later.

But I think it should get a paragraph in ltnews describing the improvement .

base/doc/ltnews36.tex
```diff
@@ -335,7 +347,7 @@ \subsection{Rule based ordering for \LuaTeX\ callback handlers}
 
 \section{Bug fixes}
 
-\subsection{Prevent \TeX{} from loosing a \cs{smash}}
+\subsection{Prevent \TeX{} from losing a \cs{smash}}
```
FrankMittelbach, line 351: oh well :-)

base/changes.txt
```diff
@@ -6,6 +6,10 @@ completeness or accuracy and it contains some references to files that
 are not part of the distribution.
 ================================================================================
 
+2022-10-10  David Carlisle  <David.Carlisle@latex-project.org>
+
+	* ltclass.dtx: use \protected@edef rather than \edef wehe expanding option lists
```
FrankMittelbach, line 11: when

FrankMittelbach (approved):
with one minor typo

Issue #932, opened 2022-10-10
protecting option lists


davidcarlisle:
## Brief outline of the enhancement

Currently  `\usepackage[font=\bfseries]{...}` dies with internal errors from `\bfseries` in `\edef` even if the package defines a font key.

The edef is just used in `\opt@...` which could be (and is for the new kernel option handler) ignored in favour of `\@raw@opt@..`.  but only if the edef expands without error.

This suggested enhancement is to use `\protected@edef` rather than `\edef` so that latex constructs do not explode. This will make the classic option lists more usable, and more importantly allow them to be defined and ignored without error when using the new key-val option processing.

## Minimal example showing the current behaviour

```latex
% would  complain about edef1.sty \RequirePackage{latexbug}    
\documentclass{article}

\usepackage[font=\bfseries]{edef1}

\show\mainfont
\begin{document}

\end{document}
```

Using `edef1.sty`:

```

\ProvidesPackage{edef1}

\def\mainfont{\reset@font}

\DeclareKeys{
font .store = \mainfont
}

\ProcessKeyOptions
```

The above produces:

```
! Undefined control sequence.
\expand@font@defaults ->\edef \rmdef@ult 
                                         {\rmdefault }\edef \sfdef@ult {\sfd...
l.5
     
?
```

## Minimal example showing the desired new behaviour

The above should produce

```
Document Class: article 2022/07/02 v1.4n Standard LaTeX document class
(/home/davidc/texmf/tex/latex-dev/base/size10.clo)) (./edef1.sty)
> \mainfont=macro:
->\bfseries .
l.6 \show\mainfont
                  
? 
```

davidcarlisle:
see PR #933

davidcarlisle:
some more cases need protecting, a test case discussed in #939

```
\documentclass[foo,bar,baz=foo,ggg=\sqrt{4}]{article}
\usepackage{amsmath}
\begin{document}
\end{document}
```

PR #940, opened 2022-10-22
further guarding of options for #932

davidcarlisle:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [X] Test file(s) added

I added rollback in ltdefns but changes in ltclass treated as bug fix on recent changes so covered by existing rollbacks undoing the raw list handling


base/ltclass.dtx
```diff
@@ -1520,6 +1521,7 @@
    \ifx\@classoptionslist\relax\else
     \@for\CurrentOption:=\@classoptionslist\do{%
       \ifx\CurrentOption\@empty\else
+        \edef\CurrentOption{\detokenize\expandafter{\CurrentOption}}%
```
josephwright, line 1524: Using `\detokenize` here in re-defining `\CurrentOption` means that this 'shows up' in `\@use@ption`: if we imagine for example an option called `foo{=}bar`, the detokenization would now strip the braces of meaning, and we'd get 'unused option `foo`' not 'unused option `foo{=}bar`'. I can see that is somewhat contrived, but want to be sure that the effect is deliberate/expected.

base/ltclass.dtx
```diff
@@ -1557,13 +1559,15 @@
...
 % The common part of |\ProcessOptions| and |\ProcessOptions*|.
+% \changes{v1.5e}{2022/10/22}{Use \cs{detokenize}}
 %    \begin{macrocode}
 %</2ekernel>
 %<*2ekernel|latexrelease>
 %<latexrelease>\IncludeInRelease{2020/10/01}%
 %<latexrelease>                 {\@process@pti@ns}{Unused options issue}%
 \def\@process@pti@ns{%
   \@for\CurrentOption:=\@curroptions\do{%
+   \edef\CurrentOption{\detokenize\expandafter{\CurrentOption}}%
```
josephwright, line 1570: Same point about `\detokenize`.

josephwright (approved):
I'm happy for this to be merged, just want to be sure the `\detokenize` effect is by-design

base/ltclass.dtx
davidcarlisle: @josephwright  yes but this is the classic option handler and as existing behaviour is the low level csname error, there don't seem to be any compatibility issues, in practice only simple names are going to be declared, the only issue is to make sure foo=\whatever survives eough to get on the global class options list (without being detokenized)

davidcarlisle:
@josephwright  cyrillic tlb2888 fails due to

```
    \expandafter\in@\expandafter{\CurrentOption}%
                                {T2A,T2B,T2C,X2,LCY,OT2}%
```

now  \CurrentOption is detokenized, we could live that or replace

`\edef\CurrentOption{\detokenize\expandafter{\CurrentOption}}`

by detokenizing locally each time it is used in a csname

josephwright:
I suspect this might come up in other places: people expect `\CurrentOption` to be tokenized. So I think `\detokenize` will beed to be used on a need-to-apply basis.

base/ltclass.dtx
davidcarlisle: adjusted

Issue #820, opened 2022-05-17
Package "doc" assumes a fixed set of mathcodes


josephwright:
## Brief outline of the bug

As title: `\mod@math@codes` has a fixed set of codes, and this is not the case with Unicode engines/`unicode-math`.

## Minimal example showing the bug

```latex
\RequirePackage{latexbug} 
\documentclass{article}
\usepackage{doc}
\usepackage{unicode-math} %  To produce a \mathcode change
\begin{document}
    \begin{macrocode}
%<filea|fileb> test
%    \end{macrocode}
\end{document}
```

## Log file (required) and possibly PDF file  

Log attached: PDF shows
![test](https://user-images.githubusercontent.com/654518/168871991-de939ab9-142c-47f1-9a9d-e5be47bba859.png)

Details of the issue were first reported in https://tex.stackexchange.com/questions/644510/vertical-bar-in-docstrip-file-changes-to-j-with-unicode-math

FrankMittelbach:
Well, as the log clearly shows, this should be reported to unicode-math not to us :-), how far are we with moving unicode-math also under latex3 repository?

leaving that aside, the solution to the problem is to adjust the doc hook for it (which is from 1992), so an old fashioned hook (and it is really a configuration point anyway): but basically
```
\AddToHook{package/unicode-math/after}{%
  \AddToHook{cmd/mod@math@codes/after}{\mathcode`\|=28796 }}
```
The inner hook setting could go into unicode-math directly. Technically it might be better for unicode-math to fully (re)define `\mod@math@codes` instead of the above as that is a little bit more efficient, but the above works without needing to know if doc gets loaded

josephwright:
> Well, as the log clearly shows, this should be reported to unicode-math not to us :-), how far are we with moving unicode-math also under latex3 repository?

I did think of that, but the assumption is in `doc` not in `unicode-math` - it's reasonable in a 7-bit world, of course.

> leaving that aside, the solution to the problem is to adjust the doc hook for it (which is from 1992), so an old fashioned hook (and it is really a configuration point anyway): but basically
> 
> ```
> \AddToHook{package/unicode-math/after}{%
>   \AddToHook{cmd/mod@math@codes/after}{\mathcode`\|=28796 }}
> ```

Is `\mod@math@codes` a documented interface? Feels odd

FrankMittelbach:
> Is `\mod@math@codes` a documented interface? Feels odd

[...]

PR #947, opened 2022-11-04
first aid fix for #820

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


davidcarlisle (approved):

josephwright (approved):
All good but couldn't we just fix this in `unicode-math`? All that's needed is for @wspr to transfer it to the latex3 area then we are good-to-go.

FrankMittelbach:
> All good but couldn't we just fix this in `unicode-math`? All that's needed is for @wspr to transfer it to the latex3 area then we are good-to-go.

well we could, but it requires moving unicode-math under latex3 which will eventually happen (but when?) and until then ...

PR #950, opened 2022-11-06
Gh944

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development

- [x] Test file(s) added

The pr requires changes in other repos (not done):

 - update regression-test.tex so that it drops the release info if LaTeX is used
 - update expl3.ltx so that it adds the L3 layer info to the toks `\LaTeXReleaseInfo`and not to `\everyjob`

Perhaps there are better approaches, so open for suggestions.


support/regression-test.tex
```diff
@@ -293,6 +293,25 @@
   \fi
 \fi
 \reset@catcodes
-%% 
+%%
+
+%%% this needs to be moved to the source of regression-test.tex!
+
+\ifx\AddToHook\undefined
```
josephwright, line 300: I assume for the 'real thing' this should use the LaTeX release date?

support/regression-test.tex
```diff
@@ -293,6 +293,25 @@
...
+%
+% We don't want release information showing up in \end{document}
+% because that would change over and over again.
+% We therefore drop that information from the hook (should also
+% happen in the standard regression test when running a LaTeX
+% format). By using a voids relation we can reactivate the code in
+% a test by dropping "the testmode" label.
+%
+\AddToHook{enddocument/info}[kernel/testmode]{}
+\DeclareHookRule{enddocument/info}{kernel/testmode}{voids}{kernel/release}
```
josephwright, line 312: Easy enough to add - looks good to me

josephwright (approved):

support/regression-test.tex
FrankMittelbach: Maybe I misunderstand you, but I don't think so. Point is you can use regression-test.tex wiht formats other than LaTeX, don't you? So we need a test that does a) is it LaTeX and b) is it new enough? So I think the above is a simple way to do both.

PR #956, opened 2022-11-14
extend label to five data and add hook

u-fischer:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [ ] Test file(s) added 

## Summary
This pull request extends the `\label` command to store as with hyperref five 
data. It also adds a hook similar to the one added by nameref. This allows to 
remove some patches from hyperref, nameref and showkeys and to store 
structure data needed for example for the tagging of footnotes. 

## Open questions

* Name of the hook ok? (currently `label`)
* Name of the command which stores the current argument of `\label` ok?
  (currently `\label@name`)
* Are the names for the additional data ok? 
* Is the rollback code correct??
 
  
## Notes about the code

* As third argument it stores `\@currentlabelname`. An alternative would be 
to use a new name, e.g. `\@currenttitle`. But while this wouldn't be a problem for
nameref as it could do `\def\@currenttitle{\@currentlabelname}` or could be changed
to use the new command, it would be difficult for classes and other packages to support this, as they would have to fill 
either `\@currenttitle` or `\@currentlabelname` depending on the age of the format and
depending on if nameref is loaded or not.

* The fourth argument is `\@currentHref`. A better name would be
`\@currenttarget` but `\@currentHref` is set by various classes and packages
and so it can't be dropped for similar reasons. 

* The fifth argument should be reserved for the kernel, but currently no 
concrete use is planed or implemented. 
One option is to store an id for the current structure.

* The change of the \setref command avoids that documents compiled
with an older format and where the aux-file contains old 
„2-data-`\newlabel`'s“ error with "runaway argument". 

* The name of the label is saved into `\label@name`. This is the 
command name also used in nameref. But this could be changed, no package using
it (apart from hyperref and nameref) could be found. 

## Example of uses for the hook

* `showkeys` redefinition of `\label` can be replaced by

~~~~ 
\AddToHook{label}{\SK@\SK@@label{\label@name}} 
~~~~

* nameref uses its hook to strip a period from the title so would
have to do 

~~~~ 
\AddToHook{label}{\NR@sanitize@labelname} 
~~~~

* a zref user can do 

~~~~
\AddToHook{label}{\zlabel{\label@name}}
~~~~

## Compability with external packages.     

* The testsuite only showed test failures due to the changed write command. 
* cleveref redefines `\label` to take an optional argument but makes
use of the original command and seems to work fine. 
* No problems found with varioref, smartref, showlabels either.
* the `thref` option of ntheorem is incompatible. It doesn't error but
overwrites the changes again. (It is also incompatible with hyperref and cleveref
and should better not be used anyway.)
* xr-hyper/hyperref use currently the fifth argument and must be adapted if the 
kernel wants to use it too.


base/doc/ltnews37.tex
```diff
@@ -149,6 +149,29 @@ \section{New or improved commands}
 %
 \githubissue{xxx}
 
+\subsection{Extension of the \cs{label} command}
+
+Up to now the \cs{label} command stored in standard \LaTeX{} into the
+\cs{newlabel} command in the \texttt{.aux} file two data:
```
FrankMittelbach, line 155: would use "values" not "data" even if data is correct in its plural

base/doc/ltnews37.tex
```diff
@@ -149,6 +149,29 @@ \section{New or improved commands}
...
+\pkg{nameref} patch the \cs{label} command to store five data: 
+Additionally they save \cs{@currentlabelname}, which normally contains the 
+current title text and can be retrieved with \cs{nameref} and 
+\cs{@currentHref} which is the name of the destination needed to create an 
+active link. The fifth data container is currently only used if external 
+references are loaded with the \pkg{xr-hyper} package. 
+
+Starting with this release the number of arguments have be unified. 
+\cs{label} now writes a
+\cs{newlabel} command containing five data: \cs{@currentlabel},
```
FrankMittelbach, line 167: with five arguments

base/doc/ltnews37.tex
```diff
@@ -149,6 +149,29 @@ \section{New or improved commands}
...
+\cs{thepage}, \cs{@currentlabelname}, \cs{@currentHref}. The fifth 
+container is reserved for the kernel and contains data needed for the 
+tagged PDF project. 
+
+Additionally a hook with the name \texttt{label} has been added. It is 
+executed before the writing to the \texttt{.aux} file. Code using
+the hook can refer to the label argument with \cs{label@name}.
```
FrankMittelbach, line 174: if \@currentlabelname  etc why  \label@name  and not  \@labelargument, ie the @ always in the same position?

base/ltxref.dtx
```diff
@@ -108,7 +108,8 @@
 %  incremented by the command  \refstepcounter{CNT} , which sets
 %  \@currentlabel == {CNT}{eval(\p@cnt\theCNT)}.   The command
 %  \label{FOO} then writes the following on file \@auxout :
-%        \newlabel{FOO}{{eval(\@currentlabel)}{eval(\thepage)}}
+%        \newlabel{FOO}{{eval(\@currentlabel)}{eval(\thepage)}%
+%           {eval(\@currentlabelname)}{eval(\@currentHref)}{eval(\@kernel@currentdata)}}
```
FrankMittelbach, line 112: \@kernel@currentdata  is a very strange name ... for starters it is not at all linked to label/ref from the name

base/ltxref.dtx
```diff
@@ -343,16 +383,29 @@
...
+\providecommand\label@name{}
+\NewHook{label}
 \def\label#1{\@bsphack
+  \begingroup
+  \def\label@name{#1}%
+  \UseHook{label}%
   \protected@write\@auxout{}%
-         {\string\newlabel{#1}{{\@currentlabel}{\thepage}}}%
+         {\string\newlabel{#1}{{\@currentlabel}{\thepage}%
+          {\@currentlabelname}{\@currentHref}{\@kernel@currentdata}}}%
```
FrankMittelbach, line 401: is a full expansion (or any expansion for that matter) really desired for the fifth argument, or shouldn't tjat be passed unchanges (after one expansion through the .aux mechanism?

base/doc/ltnews37.tex
FrankMittelbach: from a general perspective I wonder if there shouldn't actually be 6 arguments (the last 2 key/value lists)
with 

 - arg5 reserved for packages to pass stuff around
 - arg6 reserved for the kernel to pass extra stuff around

shold probably be discussed elsewhere, not in this pull.

FrankMittelbach (commented):
Some comments added, I think this needs some separate discussions.

base/doc/ltnews37.tex
u-fischer: > if @currentlabelname etc why \label@name and not @labelargument, ie the @ always in the same position?

`\label@name` is the command currently used by nameref. But `\@labelargument would be fine with me. 

> 6 arguments (the last 2 key/value lists)

perhaps, but at first the patches from hyperref and nameref should disappear.

base/ltxref.dtx
u-fischer: > @kernel@currentdata is a very strange name ... for starters it is not at all linked to label/ref from the name

well currently it doesn't hold anything. I didn't want to leave it empty so that one can test in latex-lab what is needed or useful here and added some name that reserves that for us. Do you a have idea?

base/doc/ltnews37.tex
car222222: If changing to 6 args is too difficult, arg 5 could contain two groups, one for holding the user-information (in a well-defined and consistent format) and the other for the kernel.

FrankMittelbach: As discussed yesterday, we can do with one argument as long as we define that datastructure in a way that packages (or  users) can add to it in a controlled way. But if not we can indeed use 2 brace groups to separate them if that seems advisable in the end.

davidcarlisle (approved):
I reworded ltnews, pulling in Frank's comments at same time

base/doc/ltnews37.tex
car222222: or "data items"

base/doc/ltnews37.tex
car222222: No! They are not (and in ltxref never become) arguments of anything --

Suggestion: ... command that stores, in its second argument, the following five items of data:

base/doc/ltnews37.tex
car222222: argument ==> (data item)

base/ltxref.dtx
car222222: Maybe  \ref@kernel@data 
or   \ref@kernel@currentdata

base/ltxref.dtx
car222222: Agreed: *full* expansion may be very dangerous, but one-level seems to be necessary.

car222222:
I think that the ltnews37 entry will need further revision as I spotted a remaining, possibly wrong, use of "argument"; but not essential right now.

base/doc/ltnews37.tex
car222222: container ==>  (data) item

car222222:
Ooops!  It seems that I got misled about the data items never being used as arguments: the version of ltxref in this pull does seem to do so, for choosing one of them from the list.

I have seen a version of ltxref that avoids using them as arguments, by using some l3 commands for handling the data list as l3 items in an l3 token list (tl).  
Good stuff! But where is that version?

car222222:
No, I did see the no-arg versions, they are staring at me in the current ltxref.dtx.
But it seems they are not currently used!  It says: 
"As the commands are now protected we also need expandable versions for use in \ifthenelse:"
These 'expandable versions' do not use the data as arguments to anything.

Therefore my suggestion is to always use (something like) these  versions: at least, use their method for extracting the data, rather than using \@firstoffive etc., with arguments).  

I am unsure if it is worth making this change to ltxref at this stage.

u-fischer:
> Therefore my suggestion is to always use (something like) these versions: at least, use their method for extracting the data, rather than using @firstoffive etc., with arguments).

The expandable version is a bit slower than `\@setref`, so what would be the gain to use it always?

car222222:
Yes, it would be a bit slower. 

The main advantage is that it removes all of the dependence of the current code on the number of data items.  More adaptable code; packages do not need to know how many data items must be or have been been stored.
But maybe saving time is more important.

car222222:
I still believe that we should not use the term "argument" in the documentation as, even if the implementation does not change and thus they do sometimes get used as arguments, they should be thought of as data items!

FrankMittelbach:
> I still believe that we should not use the term "argument" in the documentation as, even if the implementation does not change and thus they do sometimes get used as arguments, they should be thought of as data items!

What is written to the aux file is `\newlabel{..}{..}{..}{..}{..}`  that is a command with X number of arguments and it is processed by picking up the arguments. I don't see why you want to call them anything else at this stage.

u-fischer:
> What is written to the aux file is `\newlabel{..}{..}{..}{..}{..}`

öh no. `\newlabel{key}{{1}{1}{}{}{}}` is written. So we have always two arguments, but with a varying numbers of values in the second argument. (I have locally changed the text to use "argument" only for the outer things, and "values" for the things in the second argument, will push when I checked and updated testfiles)

FrankMittelbach:
ok, looking instead of incorrectly remembering would help ... I see. Yes then value is better as arguments. I seem to always confuse that the way toc data is passed around.

car222222:
Yes, I too get confused every time I get back to looking at it again!

Thanks, @u-fischer, for sorting out the language. 

PS: I have now started looking at l3ref, which is even more brain-twisting!!

car222222:
Not sure if this the best place to put this comment?

AFAICT, the latest version of ltxref.dtx in branch newlabel still has this line (here, the date is definitely wrong, not sure about the version number?):

`\ProvidesFile{ltxref.dtx}
             [2022/11/12 v1.1q LaTeX Kernel (Cross Referencing)]`

car222222:
These three new slots/attributes(!) need some explanation 
in this file:
  {\@currentlabelname}{\@currentHref}{\@kernel@reserved@label@data}

For the first two:
How are these used, where/how do they get values?  
Where is their use explained (and they are defined)!?
Do they need, or have, accessor functions, like \ref \pageref: and where are these? 
For the third, . . . ??? but say something, please! 

Please also explain “something more about the hook”!   
Why is it there?
Maybe give some example contents!  
Or at least a reference to where its use is fully explained.

u-fischer:
> For the first two:
> How are these used, where/how do they get values?

the first has been introduced by nameref and contains typically the title. nameref patches sectioning commands to set it, but beamer and e.g. tabularray set it too. The second contains the destination/target name. It has been introduced by hyperref, but the latex-lab code sets it in various places too. 

Both is explained in the ltnews entry.

> Do they need, or have, accessor functions, like \ref \pageref: and where are these?

The first can be accessed by `\nameref`, the second has only internal accessor commands as it is typically used in links.

> For the third, . . . ??? but say something, please!

That was discussed on the team list. 

> Please also explain “something more about the hook”!
> Why is it there?

Because nameref added a similar hook, and hyperref needs it to change in certain cases the destination names. Beside this, if the PR is merged the showkeys package will use it. It also allows users to add more label commands, e.g. to record the position or the absolute page.

car222222:
These explanations all need to be in the file!

u-fischer:
> These explanations all need to be in the file!

I added a note in ltxref.

car222222:
Good, thanks!

FrankMittelbach (approved):

car222222:
These latest changes to ltnews37 by @FrankMittelbach are OK (and mostly good:-).  But much of this should at some stage also be added to ltxref.  
Explicit examples of current uses for the hook would als be useful in ltxref.

car222222:
This probably, at some stage, will need some test files.

FrankMittelbach:
> This probably, at some stage, will need some test files.

I was briefly wondering about that but given that it affects quite a number of test files arealdy I thought that extra ones aren't needed. What should the test in addition?

car222222:
Yes, I can understand that the new 5-slot version may well be covered by existing tests.  So, do some .tlg files need to be updated?

But the new hook is less likely to be covered. Also, and related, we anyway need examples of its use.

muzimuzhi (commented):
Whether the new hook `label` is used inside a group or not is inconsistent among `ltnews37` and `ltxref.dtx`.

It seems to me the functionality of `showkeys` package won't be influenced by the extra group since `\SK@@label` works globally.

base/ltxref.dtx
```diff
@@ -343,16 +393,27 @@
...
+\NewHookWithArguments{label}{1}
 \def\label#1{\@bsphack
+  %\begingroup
+  \UseHookWithArguments{label}{1}{#1}%
   \protected@write\@auxout{}%
-         {\string\newlabel{#1}{{\@currentlabel}{\thepage}}}%
+         {\string\newlabel{#1}{{\@currentlabel}{\thepage}%
+          {\@currentlabelname}{\@currentHref}{\@kernel@reserved@label@data}}}%
+  %\endgroup        
   \@esphack}
```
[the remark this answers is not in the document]
u-fischer: ups, good catch. I will add it (it is no longer strictly needed to protect the \@bshack value but we decided to keep it).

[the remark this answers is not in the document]
car222222: Also, it would be good to change the wording in ltnews37 as follows, to be more precise:

. . . thus its code can be made local, so as to affect only the write operation.

PR #959, opened 2022-11-16
Gh917+958 add nowarn option to filecontents

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -14,7 +14,7 @@ All changes above are only part of the development branch for the next release.
 2022-11-16  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
 
 	* ltclass.dtx:
-	Introduce key 'silent' on filecontents (gh/958)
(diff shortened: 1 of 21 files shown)


davidcarlisle (commented):
The code all looks OK as far as it goes, I'm not sure about the name `silent` as I'd expect that to not log at all. Perhaps `[nowarn]` as warning/info  is the distinction used elsewhere

With more modern interfaces in tex IDE or overleaf etc, not logging to the terminal doesn't make any difference as the terminal output is never exposed. TeX runs in the background and only the log is available. A real `silent`/`nolog` option would be required if we wanted to address the orginal use case in such an interface. Even if we don't add it now, it might be good to leave the name `silent` free for such an option.

FrankMittelbach:
just saw that fontspec calls it `quiet` instead of `nowarn` and the complete dropping `silent`, but I think "nowarn" is actually better.

davidcarlisle (approved):

PR #970, opened 2022-11-30
Set \oe/\OE equal during case changing

josephwright:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
[X] Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -10,6 +10,12 @@ are not part of the distribution.
 All changes above are only part of the development branch for the next release.
 ================================================================================	
 
+2022-11-30  Joseph Wright  <Joseph.Wright@latex-project.org>
+
+	* ltfinal.dtx:
+	Set \oe and \OE equal in case changing to allow detection
+	by babel
+
 #########################
 # 2022-11-01 Release
 #########################
--- a/base/ltfinal.dtx
+++ b/base/ltfinal.dtx
@@ -33,7 +33,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{ltfinal.dtx}
-             [2022/10/26 v2.2x LaTeX Kernel (Final Settings)]
+             [2022/11/30 v2.2y LaTeX Kernel (Final Settings)]
 % \iffalse
 \documentclass{ltxdoc}
 \GetFileInfo{ltfinal.dtx}
@@ -1144,6 +1144,7 @@
   }
 \exp_args_generate:n { cnx }
 %    \end{macrocode}
+% \changes{v2.2y}{2022/11/30}{Set \cs{oe}/\cs{OE} equal to act as a marker for \pkg{babel}}
 % The odd use of \emph{three} spaces here is needed as \pkg{ltcmd} uses the
 % name with one and two spaces to give a `friendly' error message for a runaway
 % argument: that means we can't use it here.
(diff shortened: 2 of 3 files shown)


davidcarlisle (commented):
This looks fine as far as it goes but can you say a bit why just oe needs this and not other entries of the old list, which are presumably handled OK by the underlying L3 code?

```
\def\@uclclist{\oe\OE\o\O\ae\AE
      \dh\DH\dj\DJ\l\L\ng\NG\ss\SS\ij\IJ\th\TH}
```

josephwright:
@davidcarlisle  It's because `\oe`/`\OE` are used as a marker by `babel`:
```
> \bbl@cased=macro:
->\ifx \oe \OE \expandafter \in@ \expandafter {\expandafter \OE \expandafter }\
expandafter {\oe }\ifin@ \bbl@afterelse \expandafter \MakeUppercase \else \bbl@
afterfi \expandafter \MakeLowercase \fi \else \expandafter \@firstofone \fi 
```

davidcarlisle:
>  It's because \oe/\OE are used as a marker by babel:

blrg OK  PR approved, but we should add this as a comment to the code

FrankMittelbach (commented):
given that this is (if you will) a temporary fix because of that somewhat strange method in babel right now, I think it would have been better placed into firstaid, as I think it is something that shold go out eventually again.

on the other hand we can certainly take it out of ltfinal, so given that it is already there ...

why is the test file getting an additional space? is that correct?

josephwright:
I'm happy adding to `firstaid` if that feels cleaner.

FrankMittelbach:
> I'm happy adding to `firstaid` if that feels cleaner.

well, firstaid differs in main and dev so it is probably not wirth the bother now that it is already in.

josephwright:
@FrankMittelbach  Space looks like an issue with having `\label` at the end of a text-cased string followed by a space vs not followed by a space: that's something I'll need to look at in the `expl3` code.

josephwright:
@FrankMittelbach  Thinking a bit more: it's because without the extra tokens, case changing gives ` \label{A} `, so the two spaces are combined, but now we get ` \label{A}<stuff> `, so they don't. To-be expected.

FrankMittelbach:
wait a second and let me understand that. why is a change in the outcome "to be expected"? All we did was changing the \MakeUppercase internal code so  documents should not change (unless they uppercased incorrectly in the past. (I must confess that I haven't looked what the test does ...)

josephwright:
> wait a second and let me understand that. why is a change in the outcome "to be expected"? All we did was changing the \MakeUppercase internal code so documents should not change (unless they uppercased incorrectly in the past. (I must confess that I haven't looked what the test does ...)

The test has `\MakeUppercase{a \label{b}} c`. If you set up case changing to be pure `f`-type expansion, this becomes, `A \label{b} c`, and then `\label` does it's thing and we see `a c`. However, if there are any tokens at the end of `\MakeUppercase` after the pure-expansion step, they mean we get `A \label{b}<tokens> c`, the two spaces are not compressed and we see them both in the output.

FrankMittelbach:
But that would be a clear change of functionality, wouldn't it?  and not unlikely to affect documents

davidcarlisle:
Until this year makeuppercase used a `{}` group so  a `\ignorespaces` in label would have no effect so it is reverting to that, but if we are thinking this is just temporary it might be good to avoid churn here at least in the kernel version. Perhaps we should leave this for babel to patch if needed?

u-fischer:
can't babel detect the context in a more natural way now or use \CaseSwitch?

josephwright:
@u-fischer  That will only work if it's reached by expansion of the input, not the way `babel` uses `\bbl@cased` (hidden inside protected macros).

FrankMittelbach:
> Until this year makeuppercase used a `{}` group so a `\ignorespaces` in label would have no effect so it is reverting to that, but if we are thinking this is just temporary it might be good to avoid churn here at least in the kernel version. Perhaps we should leave this for babel to patch if needed?

ah, so we overlooked that change back then and are now reverting to it. If that was always the case then I'm fine with it even if it was different in the last months. Otherwise it should have been documented (at least).

FrankMittelbach (approved):

jbezos:
My OK is late, but it's an OK.

PR #972, opened 2022-12-09
Initial proposal for math capture

josephwright:
# Internal housekeeping

## Status of pull request

- [x] Feedback wanted 
- Under development
- Ready to merge

- [x] Test file(s) added


required/latex-lab/latex-lab-everymath.dtx
```diff
@@ -0,0 +1,347 @@
...
+%
+% \subsection{Setup}
+%
+% Loading \pkg{amsmath} is an absolute requirement: this avoids needing to
+% have conditional definitions and deals with how to define \cs{[}/\cs{]}
+% neatly.
+%    \begin{macrocode}
+\AddToHook { begindocument / before }
+  { \RequirePackage { amsmath } }
+%    \end{macrocode}
```
FrankMittelbach, line 89: I think this would be better implemented by having a latex-lab-requirements.dtx file that hold everything we want to automatically load, possibly based on keys in \DocumentMetadata. That might need some thoughs/design, but I think it should be centralized.

required/latex-lab/latex-lab-everymath.dtx
```diff
@@ -0,0 +1,347 @@
...
+%
+% \begin{macro}{\@@_grab_inline_dollar:w}
+%   Grab up to a single |$|, for inline math mode.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_grab_inline_dollar:w % $
+  #1 $
+  {
+    \@@_process:n {#1} % $
+    #1 $
+  }
```
FrankMittelbach, line 155: what happens if the  closing dollar is hidden, e.g., an environment that has the dollars in begin+end?
We can't really avoid that, but perhaps we need to offer an interface (maybe a flag to set?) so that such cases can be handled if the code is updated. --- I realize I write comments while reading along, so maybe that is offered later ...

required/latex-lab/latex-lab-everymath.dtx
```diff
@@ -0,0 +1,347 @@
...
+%    \end{macrocode}
+%
+% \subsection{Setup}
+%
+% Loading \pkg{amsmath} is an absolute requirement: this avoids needing to
+% have conditional definitions and deals with how to define \cs{[}/\cs{]}
+% neatly.
+%    \begin{macrocode}
+\AddToHook { begindocument / before }
+  { \RequirePackage { amsmath } }
```
josephwright, line 88: If we simply force this here, then we don't need the delayed re-definition business later.

required/latex-lab/latex-lab-everymath.dtx
```diff
@@ -0,0 +1,347 @@
...
+%    \end{macrocode}
+% \end{macro}
+%
+% \begin{macro}
+%   {\equation, \@@_equation_begin:, \equation*, \@@_equation_star_begin:}
+% \begin{macro}
+%   {\endequation, \@@_equation_end:, \endequation*, \@@_equation_star_end:}
+%   These environments are not set up by \pkg{amsmath} to collect their body,
+%   so we do that here. This has to be done \emph{after} we can be sure
+%   \pkg{amsmath} is loaded.
```
FrankMittelbach, line 215: that would become simpler if we require amsmath early on. On the other hand perhaps the actual loading should even then only happen late, so not sure this is true.

required/latex-lab/latex-lab-everymath.dtx
```diff
@@ -0,0 +1,347 @@
...
+  }
+%    \end{macrocode}
+% \end{macro}
+% \end{macro}
+%
+% \begin{macro}{\[, \]}
+%   A slight oddity: we have to collect the body but want to pass on
+%   to the more formal structure. That means that \cs{]} is no longer
+%   a command, but rather a marker. Again, we need to watch for when
+%   \pkg{amsmath} is loaded after this code.
```
FrankMittelbach, line 246: Again it would be nice if that could handle `\[`and `\]` being hidden inside an environment

required/latex-lab/testfiles/everymath-001.tlg
```diff
@@ -0,0 +1,6 @@
+This is a generated file for the l3build validation system.
+Don't change this file in any respect.
+ y = mx + c
+y = mx + c
```
FrankMittelbach, line 4: what's the extra space doing here?

josephwright: `\begin{equation}` is followed by a newline in my test file, so that's a space ;) The aim is to capture the tokens exactly as TeX would see them: we likely want to do further processing, but I've not included that at present.

FrankMittelbach: isn't the amsmath grabbing cleaning those newlines out? if so we may want to do that for consistency too ... but it isn't really that important, just noticed the difference in the output.

required/latex-lab/latex-lab-everymath.dtx
```diff
@@ -0,0 +1,347 @@
...
+% \providecommand\class[1]{\texttt{#1.cls}}
+% \providecommand\pkg[1]{\texttt{#1}}
+% \providecommand\hook[1]{\texttt{#1}}
+%
+% \begin{abstract}
+% \end{abstract}
+%
+% \tableofcontents
+%
+% This file implements capture of all math mode material at the top level.
```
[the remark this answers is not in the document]
josephwright: Yes, I mean not dealing with recursion: would you prefer 'outer'?

required/latex-lab/changes.txt
```diff
@@ -1,3 +1,8 @@
+2022-12-09 Joseph Wright <Joseph.Wright@latex-project.org>
+
+	* latex-lab-everymath.dtx
```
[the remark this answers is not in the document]
josephwright: Happy to rename: I'll see what others feel then rename

required/latex-lab/latex-lab-everymath.dtx
```diff
@@ -0,0 +1,347 @@
...
+%
+% \begin{variable}{\l_@@_display_bool}
+%   Tracks display style (cases where \cs{everydisplay} is triggered).
+%    \begin{macrocode}
+\bool_new:N \l_@@_display_bool
+%    \end{macrocode}
+% \end{variable}
+%
+% \begin{variable}{\l_@@_env_bool}
+%   Tracks math environments which deal with grabbing themselves.
```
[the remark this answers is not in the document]
josephwright: The idea is that this is 'we are inside an environment': all of the `amsmath` ones are 'verb-like' and grab content, I've adjusted `equation(*)` so it does too, and I was imagining that any custom ones will do the same.

required/latex-lab/latex-lab-everymath.dtx
josephwright: I've favour forcing loading earlier but didn't want to do more than necessary

required/latex-lab/latex-lab-everymath.dtx
josephwright: Well really what we need is all math mode to be verb-like, as is already done by `amsmath`. I've not exposed the booleans at the moment, but we could. I think it's pretty unlikely to be required for inline maths.

required/latex-lab/latex-lab-everymath.dtx
josephwright: Does anyone ever do that? With `amsmath` loaded it's equivalent to `\begin{equation*}` in any case: I'd love to see a real use case here.

required/latex-lab/latex-lab-everymath.dtx
```diff
@@ -0,0 +1,347 @@
...
+%   And for the classical \TeX{} display structure.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_grab_dollardollar:w % $$
+  #1 $$
+  {
+    \@@_process:n {#1} % $$
+    #1 $$
+  }
+%    \end{macrocode}
+% \end{macro}
```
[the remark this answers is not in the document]
FrankMittelbach: suggesting to use `(+)b` is fine, but we should then document that with an example or two how to change an existing definition. amsmath envs can be hidden and are occasionally if you use the command form instead of the environment form.

required/latex-lab/latex-lab-everymath.dtx
FrankMittelbach: that's ok, I was just commenting on the fact that we may want to go into that direction (but it doesn't have to be initially like that.

required/latex-lab/latex-lab-everymath.dtx
FrankMittelbach: I can't claim I have a good example taken from the wild but I know I have misued math in the past , for example to get some separator, e.g.
```
\newenvironment{separator}{..some setup... \[}{\]}
```
used as
```
\begin{separator}  * \qquad * \qquad * \end{separator}
```
I'm not claiming such misused has to be supported :-)

but you can think of simply having  different "names" for displays for clarity in the source

required/latex-lab/latex-lab-everymath.dtx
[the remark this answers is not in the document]
josephwright: @FrankMittelbach  If we move the current hook-based approach to simply modifying `\<ams-thing>`, then there's no change compared with the normal `amsmath` situation here (as for `amsmath` stuff we don't actually need to do any collecting anyway). Want me to do that?

required/latex-lab/latex-lab-everymath.dtx
josephwright: OK, doable if we have `\[` check the 'environment' flag. If it's not set, then we do a collection step. If it's set, then we assume that a wrapper env. has already done the job.

FrankMittelbach: possibly. I'm not claiming any of this has to be done, more like we should at least add notes that this or that may need extending in one way or the other. The real test will be to see how much (if anything) is falling over if, say, the arXiv corpus is tried.

josephwright: I've adjusted the code to allow for this: as noted in the commit message, it currently needs the internal bool setting, so we'll need to decide how to handle that.

required/latex-lab/latex-lab-everymath.dtx
```diff
@@ -315,6 +315,16 @@
...
 % \end{macro}
 %
+% Places where math mode is (ab)used.
+%    \begin{macrocode}
+\clist_map_inline:nn
+  { tabular }
+  {
+    \AddToHook{ env / #1 / begin }
+      { \bool_set_true:N \l_@@_inner_bool }
+  }
```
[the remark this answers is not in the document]
josephwright: Probably: I suspect I can also combine the various booleans into one. I'll look over the weekend.

FrankMittelbach (changes requested):
some comments inline, but moere importantly, a number of things break, e.g., \textsuperscript and I think gather (but not figured out why), see mathcapture-002.lvt coming up soonish

car222222:
Re:`\textsuperscript` (amongst others): it would be good to remove all such hidden maths, now!

required/latex-lab/testfiles/mathcapture-001.lvt
```diff
@@ -0,0 +1,60 @@
...
+
+\makeatletter
+\input{latex-lab-mathcapture.ltx}
+\makeatother
+\ExplSyntaxOn
+\math_processor:n
+  {
+    \TIMO
+    \TYPE {#1}
+    \OMIT
```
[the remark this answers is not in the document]
josephwright: Will do

josephwright:
We can support `breqn`  in the same way as e.g. `eqnarray`: save the def. for the environment, collect the body, etc.:
```latex
\let\saveddmath\dmath
\let\savedenddmath\enddmath
\ExplSyntaxOn
\RenewDocumentEnvironment { dmath } { b }
  {
    \bool_set_true:N \l__math_collected_bool
    \__math_process:n {#1}
    \saveddmath
    #1
    \savedenddmath
  }
  { }
\ExplSyntaxOff
```

With current test still fails, but that's due to tagging: if I add the above, comment out loading the testphase, it all works. The question is then do we want to do that?

car222222:
Two points concerning test -011.  I hope this is a good place to report these, and similar points.

It would be good to add some explanation of why the `smallmatrix` environment is being tested.  I see that is no longer a "Registered" environment: good!

Since cases.sty is loaded here, why not test both its environments, maybe with some of the options? 

Noting that these, like the kernel/ams `cases`, have two columns, one for maths and one for text. But note that there is an important distinction as the `cases.sty` environments are complete displays (so need Registration) whereas the kernel/ams are only for use within mathmode, like smallmatrix etc.

FrankMittelbach:
Chris

> Two points concerning test -011. I hope this is a good place to report 
> these, and similar points.

probably not (I find comments in the middle of checkins are only useful 
if they can and need to lead to immediate actions) -- otherwise they 
tend to get lost there
> 
> It would be good to add some explanation of why the |smallmatrix| 
> environment is being tested. I see that is no longer a "Registered" 
> environment: good!

very simple ... I just copied example after example from TLC3 regardless 
of obvious or non obvious reasons, just with the notion that TLC3 tests 
a huge amount of math any typically in somewhat odd combinations to be 
concise.

doing that somewhat randomly discover a dozen issues now resolved.

If I would have though "what should I test" perhaps half of them would 
have not been discovered.

> Since cases.sty is loaded here, why not test both its environments, 
> maybe with some of the options?

It would be good to produce more tests and now perhaps also more 
structured ones testing various variants, perhaps some others could join 
in writing some (I spend already a day on those ... doing and checking 
them).

car222222 (changes requested):
Nothing very important, so far.  More to come.

required/latex-lab/latex-lab-mathpkg.dtx
```diff
@@ -0,0 +1,107 @@
...
+   \RegisterMathEnvironment{dgroup*}
+   \AddToHook{env/dmath/begin}{\tagpdfparaOff}
+   \AddToHook{env/dgroup*/begin}{\tagpdfparaOff}
+}
+%    \end{macrocode}
+%
+%    Not sure why that was set in a test file, mistake?
+%    \begin{macrocode}
+%\AddToHook{package/amsmath/after}{
+%  \RegisterMathEnvironment{smallmatrix}
```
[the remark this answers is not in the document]
josephwright: I've removed these lines: as you observe, not required.

required/latex-lab/latex-lab-mathcapture.dtx
```diff
@@ -0,0 +1,507 @@
...
+%\end{verbatim}
+% would only result in a single capture of the tokens
+% ``\verb*/a \in A \text{ for all $a<5$}/''.
+%
+% \subsection{Code level interfaces}
+%
+% \begin{function}{\math_register_env:n, \math_register_env:nn}
+%   \begin{syntax}
+%     \cs{math_register_env:n} \Arg{env}
+%     \cs{math_register_env:n} \Arg{env} \Arg{options}
```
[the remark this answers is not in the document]
josephwright: Sorted

required/latex-lab/latex-lab-mathpkg.dtx
```diff
@@ -0,0 +1,107 @@
...
+%}  
+%    \end{macrocode}
+%    
+%    \begin{macrocode}
+\AddToHook{package/cases/after}{
+  \RegisterMathEnvironment{subnumcases}
+}
```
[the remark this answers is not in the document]
josephwright: Good question. Before that, though, perhaps worth working out where is best to 'fix' the fact that `cases` can't be loaded after `amsmath`. I will move the load line to be with the one above, so the issue is collected in one place.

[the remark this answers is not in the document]
FrankMittelbach: I would say this is one of the topics we need a policy on once we see how much it involves either way. My expectation would be that for a set of "core" packages (where core needs some evaluation) we could do the registration using a scheme like the above initially, but long term I would expect to offer an interface and urge package owners to provide the registration themselves. So it is a bit like a first-aid for me: kick it off so that things work but get it midterm directly into the packages.

required/latex-lab/latex-lab-mathcapture.dtx
```diff
@@ -0,0 +1,505 @@
...
+\cs_new_protected:Npn \math_processor:n #1
+  { \cs_set_protected:Npn \@@_process_auxii:nn ##1##2 {#1} }
+%    \end{macrocode}
+% \end{macro}
+%
+% \subsection{Content grabbing}
+%
+% \begin{macro}{\@@_grab_dollar:w}
+%   Grab up to a single |$|, for inline math mode, suppressing
+%   any processing if the first token is \tn{m@th}.
```
car222222, line 178: Typo, should be:  
   . . . of the first . . .

required/latex-lab/latex-lab-mathcapture.dtx
```diff
@@ -0,0 +1,505 @@
...
+%   Collect inline math content and deal with the need to move to math mode.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_grab_inline:w % \(
+  #1 \)
+  {
+    \@@_process:nn { math } {#1}
+    $ #1 $
+    \bool_set_false:N \l_@@_collected_bool
+  }
+%    \end{macrocode}
```
car222222, line 211: Two+ items -

One (obscure) bug affecting an edge case:
   Here one should use ` $ \relax #1 $ ` ,
   to guard against an empty ` \(\) ` (as in \@ensuredmath).

Two, a possible enhancement (covers an unlikely but not 
impossible case): 
   Add look-ahead for ` \m@th ` at start of ` \( `.

Two+  Note, not currently within this file:
      this look-ahead for ` \m@th ` is definitely needed
      within ` \ensuremath `, since a common usage is: 
     ` \ensuremath { \m@th ...`.

car222222 (changes requested):
A few more suggestions, including bugs (obscure? :-) and typos.

required/latex-lab/latex-lab-mathcapture.dtx
josephwright: Pretty sure I mean `if` here :)

car222222: So you do!   Must have been somewhere else that needed this:-).

required/latex-lab/latex-lab-mathcapture.dtx
car222222: Amendment: `\ensuremath` is in this file.

required/latex-lab/latex-lab-mathcapture.dtx
```diff
@@ -0,0 +1,505 @@
...
+\cs_new_eq:NN \@@_m@th: \m@th
+\cs_gset_protected:Npn \m@th
+  {
+    \bool_set_true:N \l_@@_collected_bool
+    \@@_m@th:
+  }
+%    \end{macrocode}
+% \end{macro}
+%
+% \subsection{Modifying \pkg{amsmath}}
```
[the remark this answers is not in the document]
car222222: Correction, should be: alignat(*)

car222222:
Not sure if this should currently be classified as a bug, or listed as a deficiency? 

When "dollar grabbing", it looks as if the following type 
of input will, wrongly, collect only up to the `$`just 
before the `x` :

```Some text $ a + b + \begin{minipage}[t] with more non-math 
                    text $x \times  y$ \end{minipage} . . . ```

car222222 (changes requested):
A few more comments and suggestions.

required/latex-lab/latex-lab-mathcapture.dtx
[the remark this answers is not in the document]
FrankMittelbach: aren't all those automatically processed because they all use the internal amsmath grabbing method into which the code hooks? I thought I had made examples for those too (maybe I haven't).

FrankMittelbach:
> Not sure if this should currently be classified as a bug, or listed as a deficiency?
> 
> When "dollar grabbing", it looks as if the following type of input will, wrongly, collect only up to the `$`just before the `x` :
> 
> ```
>                     text $x \times  y$ \end{minipage} . . . ```
> ```

If that's the case I would consider this a bug not a misfeature to document. Where did you notice that? in one of the test file outputs?

car222222:
It just seemed to jump out at me when I was investigating the code: I find that such easy analysis works better than just throwing examples at it:-). 

More generally, the use within $-math of any environment that
has textmode content can produce such a bug. 
Noting that many environments that are primarily for use in textmode can also be used without problem within mathmode (since they simply produce a box). 

In addition, a $ within an array environment will lead to a similar problem. Maybe also in other environments that are mathmode only.

required/latex-lab/latex-lab-mathcapture.dtx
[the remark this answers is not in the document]
car222222: Very likely they are: thus adding them is probably just an "administrative matter": 
adding them to all necessary lists, including registration.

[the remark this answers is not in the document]
car222222: I did not find them in any of the current .lvt files.

josephwright:
> It just seemed to jump out at me when I was investigating the code: I find that such easy analysis works better than just throwing examples at it:-).
> 
> More generally, the use within $-math of any environment that has textmode content can produce such a bug. Noting that many environments that are primarily for use in textmode can also be used without problem within mathmode (since they simply produce a box).
> 
> In addition, a $ within an array environment will lead to a similar problem. Maybe also in other environments that are mathmode only.

I'm obviously missing something: could you add a small example?

josephwright:
> Not sure if this should currently be classified as a bug, or listed as a deficiency?
> 
> When "dollar grabbing", it looks as if the following type of input will, wrongly, collect only up to the `$`just before the `x` :
> 
> ```
>                     text $x \times  y$ \end{minipage} . . . ```
> ```

Tricky: could be handled by counting up `$` I suppose.

required/latex-lab/latex-lab-mathcapture.dtx
josephwright: I've covered both of these I hope.

josephwright:
> > Not sure if this should currently be classified as a bug, or listed as a deficiency?
> > When "dollar grabbing", it looks as if the following type of input will, wrongly, collect only up to the `$`just before the `x` :
> > ```
> >                     text $x \times  y$ \end{minipage} . . . ```
> > ```
> 
> Tricky: could be handled by counting up `$` I suppose.

Or likely better by counting `\begin`/`\end` pairs inside math mode and 'continuing grabbing' if there is an issue.

required/latex-lab/latex-lab-mathcapture.dtx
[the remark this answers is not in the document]
car222222: Yes, they are all intended for use as displayed math environments (and some, maybe all, can also be used within a row in gather):

`flalign` is simply a left-aligned, rather than centred, align.

`alignat` and the, not well-documented) "x-aligns" are other variants on the details of the horizontal layout.  They all use `\@startalign` (unlike the "aligned" family that are for use only within mathmode in certain displays).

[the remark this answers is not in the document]
car222222: The "-ed" variants all check that they are used only within mathmode.

car222222:
What is the effect of adding x-aligns to this list?

car222222:
Ah, I see now: my fault for not being explicit.

x-aligns was just my shorthand for: `xalignat` and `xxalignat`
and it is these two that need to be added. 

There is in fact no `xxalignat*` but I guess there is no harm in adding to its hook.

car222222:
I think that Joseph has now dealt with all my current concerns with the code.

But now I cannot see a button to press in order to complete my review: should there be one within this PR?

FrankMittelbach:
> I think that Joseph has now dealt with all my current concerns with the code.
> 
> But now I cannot see a button to press in order to complete my review: should there be one within this PR?

Chris, you can click on "Files changed" at the top of this page. That opens another page showing all the diffs and a green buttom on the top right called "Review Changes". That's the one.

There are probably other ways to get at it, but this is what I normally do.

FrankMittelbach (approved):

car222222 (approved):
Permission granted to move to develop for (a lot of) further testing in latex-lab.

car222222:
Ah yes, of course: senior moment!  Thanks.

Note that the some of the code documentation is now wrong and will need updating (and much is missing:-).  But for another day/year I guess.

PR #978, opened 2023-01-02
Remove unused local variable `tex_setattribute` in `ltluatex.lua`

Josef-Friedrich:


josephwright:
Change is logical, but we need the various bits of admin doing:

- Change the file date and version for the `.dtx`
- Add a `\changes` line
- Include in `changes.txt`

Josef-Friedrich:
Thank you very much for your review. Hopefully I have documented the change correctly.

josephwright:
Close enough ;) I'll fix the order issue in `changes` myself.

josephwright:
Based on the test failure, are you _sure_ this is unused (@zauguin - thoughts)

Josef-Friedrich:
When I opened the file ltluatex.lua in the vscode, it gave me an warning about an unused local variable:

Screenshot from 2023-01-20 16-59-42

When I search for the `tex_setattribute` in the file https://github.com/latex3/latex2e/blob/develop/base/ltluatex.dtx it finds only one occurrence: the definition of the local variable itself.

zauguin (approved):

davidcarlisle (approved):
2 tests fail due to a change in the line numbers in the log.
I'll approve/merge then adjust here.

PR #985, opened 2023-01-18
Switch to PDF-based setup for pdfTeX and LuaTeX

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
[x] Ready to merge

- [ ] Test file(s) added


u-fischer (approved):

base/testfiles-lthooks2/ltpara-008.tlg
```diff
@@ -25,14 +25,20 @@ PARA: 1 end   (document level on input line ...)
 Completed box being shipped out [1]
 \vbox(633.0+0.0)x407.0
 .\hbox(0.0+0.0)x0.0
+..\kern -72.26999
+..\vbox(0.0+0.0)x0.0, glue set 72.26999fil
+...\kern -72.26999
+...\hbox(0.0+0.0)x0.0
+....\glue 0.0 plus 1.0fil minus 1.0fil
+...\glue 0.0 plus 1.0fil minus 1.0fil
 .\glue 16.0
```
FrankMittelbach, line 34: does anybody understands that difference? It doesn't hurt but it is a construction that looks very unnecessary, ie no content but boxes and glue put together. Happens in a couple of places.

base/testfiles-lthooks2/shipout2-010.tlg
```diff
@@ -119,9 +117,9 @@ Absolute page = 4 (target: 4)
 Completed box being shipped out [2]
 \vbox(633.0+0.0)x407.0
 .\hbox(0.0+0.0)x0.0
-..\kern -72.27
-..\vbox(0.0+0.0)x0.0, glue set 72.27fil
-...\kern -72.27
+..\kern -72.26999
+..\vbox(0.0+0.0)x0.0, glue set 72.26999fil
+...\kern -72.26999
```
FrankMittelbach, line 122: is that a calculated 1in? interesting that it comes out differently in the two approaches (not that it matters)

base/testfiles/tlb-hfloat-01.luatex.tlg
```diff
@@ -284,6 +284,6 @@ Underfull \vbox (badness 10000) has occurred while \output is active
 .\penalty 10000
 .\glue 0.0
 [1
-Caption1: 32634387
-Caption2: 35492222
+Caption1: 34426431
+Caption2: 37284266
```
FrankMittelbach, line 288: what kind of change is this?

base/testfiles/tlb0250.tlg
```diff
@@ -39,4 +39,6 @@ Completed box being shipped out [1]
 ....\glue 0.0 plus 1.0fil
 ....\OT1/cmr/m/n/12 1
 ....\glue 0.0 plus 1.0fil
+Non-PDF special ignored!
+<special> header=gnudict.pro
```
FrankMittelbach, line 43: was there a "test" reason for a specific header file? ie is this still testing something?

FrankMittelbach (approved):
only minor comments/querries mainly being curious

base/testfiles/tlb-hfloat-01.luatex.tlg
josephwright: Because the offset changes when you set `\pdfpagesize`, etc.

base/testfiles/tlb0250.tlg
josephwright: The test is that specials were lost in the preamble: clearly we are not _loosing_ this one, it's just it does nothingin PDF mode

base/testfiles-lthooks2/ltpara-008.tlg
josephwright: It's the page colour code in `pdftex.def`, but I'm not sure why it shows as `\ifGPT@pagecolor` should be false.

FrankMittelbach: then perhaps that is worth investigating rather than accepting just because it doesn't hurt (in this case), no?

u-fischer: You get that as soon something is in the shipout hooks. E.g. 
~~~~
\documentclass{article}
\showoutput
\AddToHook{shipout/background}{\relax}
\begin{document}
aaa
\end{document}
~~~~

car222222: Yes, it looks about right for the "background box".

FrankMittelbach: so it does, but why is there an (empty) background box if pdftex is used but not if etex was used?

u-fischer: because pdftex.def puts code into `\@kernel@before@shipout@background` (for the pagecolor), and any non empty content triggers the box too: 

~~~~
\documentclass{article}

\showoutput 
\makeatletter 
\def\@kernel@before@shipout@background{\iffalse\fi}
\makeatother
\begin{document}
blub
\end{document}
~~~~

u-fischer: well the code in pdftex.def is a bit more complicated, but it is the same principle: as soon as you add some code to the hook the box for the shipout/background is created, even if the code at the end expands to nothingness.

PR #990, opened 2023-01-30
Gh989 spacing at top of minipage

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [x] Test file(s) added


base/ltpara.dtx
```diff
@@ -800,6 +823,43 @@
...
+%    \end{macrocode}
+%    We also have to add the \cs{everypar} toks register at the
+%    end. In case of rollback this is already allocated and we have to
+%    find out the correct number (hope this is correctly done)
+%    \begin{macrocode}
+%<latexrelease>\cs_set:Npn \@@_tmp:w #1#2#3#4#5 { }
+%<latexrelease>\tl_gput_right:Nx \g_@@_standard_everypar_tl {
+%<latexrelease>    \exp_not:N \the
+%<latexrelease>    \exp_not:N \toks
+%<latexrelease>    \exp_after:wN \@@_tmp:w \token_to_meaning:N \everypar
```
PhelypeOleinik, line 859: I'm not at home, so I haven't tried this code, but `\token_to_meaning:N` doesn't seem to be the right thing here... Shouldn't this be `\exp_after:wN \@@_tmp:w \the \everypar` (or `\exp_last_unbraced:NV \@@_tmp:w \everypar`, to be more `expl3`-compliant)?

FrankMittelbach: no :-) the intention is to get the token allocation number back, so \meaning\everypar results in \toks11 and grabbing 5 tokens from that gives you 11

PhelypeOleinik: Ah, indeed, sorry about the noise.

In the unlikely scenario of a non-printable (or space) `\escapechar`, this will grab one token too much.  Defining `\@@_tmp:w` with `%<latexrelease>\use:e { \cs_set:Npn \exp_not:N \@@_tmp:w #1 \tl_to_str:n { toks } { } }` should solve that.

Skillmon: Or change the `\escapechar` in a group with

```latex
\group_begin:
  \escapechar=-1
  \exp_args:NNNx
\group_end:
\tl_gput_right:Nn \g_@@_standard_everypar_tl
  {
    \exp_not:N \the
    \exp_not:N \toks
    \exp_after:wN \use_none:nnnn \token_to_meaning:N \everypar
    \c_space_tl
  }
```

Skillmon: Also perhaps `\exp_stop_f:` instead of `\c_space_tl`?

FrankMittelbach: As it is now we end with a real space in the definition. Don't really like `\exp_stop_f` which seems out of place.
As to the escapechar being weird: the definition is done once in the kernel at a weldefined place (or in a rollback situation). In both situations we know that it is not odd, so I guess this may be overthinking the situation.

PR #1012, opened 2023-03-09
Math tagging experiments fmi1

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


FrankMittelbach:
> Quite a bit of cleanup needed before widespread us, but for feedback at this stage looks good to me.

true, but it is (and documented as such) just a first prototype for experimentation. And as such I think (like you) it is good enough to get some feedback while we then refactor an update the prototype.

u-fischer (approved):
I agree with Joseph. Perhaps one should add a warning to the l3news text, that the math code can clash with special uses of the dollar (e.g. mathescape in the listings package) and that hidden math can lead to unwanted tagging as Formula?

Issue #1028, opened 2023-03-31
bm does not handle localmathalphabets


davidcarlisle:
## Brief outline of the bug

If the math alphabet allocations are undone, bm gets confused and the mapping of math alphabets gets corrupted.
`\mathsf{Q2}` is made bold by use of `\bm` elsewhere in the same expression.

Should be fixable, but a workaround is to disable localmathalphabets although of course you may then run out of math alphabets, which is particularly likely with bm.

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}
\documentclass{article}


\DeclareSymbolFont{symbols1}     {OMS}{cmsy}{m}{n}
\DeclareSymbolFont{symbols2}     {OMS}{cmsy}{m}{n}
\DeclareSymbolFont{symbols3}     {OMS}{cmsy}{m}{n}
\DeclareSymbolFont{symbols4}     {OMS}{cmsy}{m}{n}
\DeclareSymbolFont{symbols5}     {OMS}{cmsy}{m}{n}
\DeclareSymbolFont{symbols6}     {OMS}{cmsy}{m}{n}
\DeclareSymbolFont{symbols7}     {OMS}{cmsy}{m}{n}

\usepackage{bm}

%\setcounter{localmathalphabets}{0}

\begin{document}

\showoutput

$\mathsf{Q1}$

$\mathsf{Q2}+\bm{\mathsf{Q3}}$

$\mathsf{Q4}$

\end{document}
```

## Log file (required) and possibly PDF file  

[bmlocal.log](https://github.com/latex3/latex2e/files/11119903/bmlocal.log)

davidcarlisle:
@FrankMittelbach 

I can reproduce  this without `bm` . for math alphabets, bm does essentially `\boldsymbol` ie a mathchoice around `\mbox{\boldmath$...` 

```
\RequirePackage{latexbug}
\documentclass{article}


\DeclareSymbolFont{symbols1}     {OMS}{cmsy}{m}{n}
\DeclareSymbolFont{symbols2}     {OMS}{cmsy}{m}{n}
\DeclareSymbolFont{symbols3}     {OMS}{cmsy}{m}{n}
\DeclareSymbolFont{symbols4}     {OMS}{cmsy}{m}{n}
\DeclareSymbolFont{symbols5}     {OMS}{cmsy}{m}{n}
\DeclareSymbolFont{symbols6}     {OMS}{cmsy}{m}{n}
[...]

PR #1030, opened 2023-04-01
reset frozen mathversion for #1028

davidcarlisle:
Possible fix for #1028

# Internal housekeeping

## Status of pull request

- Feedback wanted 
I didn't add rollback, if it is needed I guess it should have the original from 0000/00/00 and this from 2022/11/01 (when freeze mechanism added)  but this version should work with old formats anyway so no rollback seems ok?

- [X] Test file(s) added

### DIFF
--- a/base/ltfssbas.dtx
+++ b/base/ltfssbas.dtx
@@ -35,7 +35,7 @@
 %
 %
 \ProvidesFile{ltfssbas.dtx}
-             [2022/07/04 v3.2k LaTeX Kernel (NFSS Basic Macros)]
+             [2023/04/01 v3.2l LaTeX Kernel (NFSS Basic Macros)]
 % \iffalse
 \documentclass{ltxdoc}
 \begin{document}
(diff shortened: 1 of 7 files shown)


base/ltfssbas.dtx
```diff
@@ -1050,11 +1050,22 @@
 %      {\cs{def} $\to$ \cs{edef} for \cs{math@version}.}
 % \changes{v1.2g}{1990/02/16}{\cs{@nomath} added.}
 % \changes{v2.1a}{1994/01/17}{New math font setup}
+% \changes{v3.2l}{2023/04/01}{Reset frozen mathversion gh/1028}
 %    \begin{macrocode}
 \DeclareRobustCommand\mathversion[1]
          {\@nomath\mathversion
           \expandafter\ifx\csname mv@#1\endcsname\relax
           \@latex@error{Math version `#1' is not defined}\@eha\else
+%    \end{macrocodpe}
```
FrankMittelbach, line 1059: why didn't that fail?

FrankMittelbach (approved):
probably doesn't need  an entry in ltnews but it wouldn't hurt either I guess

and the \end... should be fixed, otherwise fine I think

base/ltfssbas.dtx
davidcarlisle: my typing doesn't improve... It doesn't generate an error, just matches the next one.

FrankMittelbach (approved):
looks good now

Issue #1035, opened 2023-04-09
Add and document \BCPdata to kernel


josephwright:
## Brief outline of the enhancement

As discussed in https://github.com/reutenauer/polyglossia/issues/581 and elsewhere, a common data format for BCP-47 data is needed. This really needs to be initiated and documented at the kernel level, with packages such as `polyglossia` and `babel` then re-defining to provide the actual data.

This issue is to allow us to track the design of the interface: currently `\BCPdata{...}` where `...` needs to be refined.

jspitz:
Just to summarize the proposal so far. The arguments that should be supported are:

1. `language`  (e.g., `de`)
2. `region` (e.g., `AT`)
3. `script` (e.g., `Latn`)
4. `variant` (e.g., `1901`)
5. `extension.t` (transformation, e.g., `en-t-ja`)
6. `extension.u` (additional locale information, e.g., `ar-u-nu-latn`)
7. `extension.x` (e.g., `la-x-classic`)

and possibly
- `tag` (for the "full" CBP-47 tag as set by babel/polyglossia which however might omit presupposed values)

To access the main language, it was proposed to have a `main.*` argument as a prefix to the above tags.

@u-fischer furthermore proposed some semantic links such as `casing` which returns the appropriate tag/subtag for the casing commands.

josephwright:
@jbezos I assume you are also happy with that list? If so, I'll add to `ltfinal` and document there and in `ltnews` for the June release, and adjust the case code so it picks up `\BCPdata` and thus solves the `polyglossia` aspect. 

(Thanks all for the contributions here: I think we are getting to a very good situation for locale data.)

u-fischer:
@josephwright  will you use a semantic wrapper like `casing`  for `\MakeUppercase`? Imho it would be sensible as the discussion showed that none of the "primitive" values is the correct one in all cases.  I will then adapt hyperref in a similar way.

josephwright:
> @josephwright will you use a semantic wrapper like `casing` for `\MakeUppercase`? Imho it would be sensible as the discussion showed that none of the "primitive" values is the correct one in all cases. I will then adapt hyperref in a similar way.

Yes that is the plan I think: makes a lot of sense

jbezos:
👍

josephwright:
OK, I will go with `\BCPdata{casing}`: that I hope should be 'available' from both `babel` and `polyglossia` by the time we do the kernel release.

jbezos:
[...]

PR #1036, opened 2023-04-11
Add \BCPdata and adapt \Make<thing>case (#1035)

josephwright:


josephwright:
The main question here is are we happy with the docs I've provided

car222222:
Where is this documentation that needs checking?

car222222:
It would be useful to provide a link to where there us a full description of the basics of this method of data provision, i.e. the syntax and how it is accessed by LaTeX (presumably by expansion of something?).  I am unsure whether the examples given here are useful without referring to some such explanation? 

Is this general type of syntax for data used elsewhere in LaTeX?  

I have seen similar ideas in other contexts.  Should its use be set-up generically in expl3 kernel or some other l3 file?

josephwright:
> It would be useful to provide a link to a full description of the basics of this method of data provision, i.e. the syntax and how it is accessed by LaTeX (presumably by expansion of something?). I am unsure whether the examples given here are useful without referring to some such explanation?

The idea is that `\BCPdata{X}` will expand to some BCP-47 subtag, so e.g.
```latex
\expanded{\noexpand\foo{\BCPdata{language}}}
```
becomes
```latex
\foo{en}
``` 
(or whatever) and then `\foo` has the data to do 'whatever'.

> Is this general type of syntax for data used elsewhere in LaTeX?

Not sure about that, but we need a way to do this by expansion (as we need the info in expansion-only contexts, e.g. bookmark creation). It felt like a reasonable plan in that sense.

> I have seen similar ideas in other contexts. Should its use be set-up generically in expl3 kernel or some other l3 file?

This needs to work 'directly' from `babel` and `polyglossia`, so it's a question of what those package authors are happy-enough with. This seemed like a workable compromise.

base/doc/clsguide.tex
```diff
@@ -1138,6 +1138,40 @@ \subsection{Normalising spacing}
...
+
+Localisation information is needed to customise a range of outputs. The
+\LaTeX{} kernel does not itself manage localisation, which is well-served
+by the bundles \pkg{babel} and \pkg{polyglossia}. To allow the kernel and
+other packages to access the current localisation information provided
+by \pkg{babel} or \pkg{polyglossia}, the command \cs{BCPdata} is defined
+by the kernel. The initial kernel definition absorbs one argument and
+expands to an empty value, as the kernel does not track localisation.
+However, if \pkg{babel} or \pkg{polyglossia} are loaded, it is redefined
+to takes one argument and expands to the BCP-47 information matching this
```
davidcarlisle, line 1151: to take one argument and expand

base/doc/clsguide.tex
```diff
@@ -1138,6 +1138,40 @@ \subsection{Normalising spacing}
...
+expand to the main language even if another language is currently active.
+
+In addition to the tag breakdown, the following semantic arguments are
+supported
+\begin{itemize}
+  \item \texttt{casing} The tag for case changing, e.g.~\texttt{el-x-iota}
+    could be selected rather than \texttt{el} to select a capital adscript
+    iota on uppercasing an \emph{ypogegrammeni}
+\end{itemize}
+
```
davidcarlisle, line 1174: this section seems a bit abstract, would probably help to give some example of a package-level command definition that uses this, assuming some minimal  babel setup.

davidcarlisle (approved):

car222222:
All of @josephwright's explanation here is fine, but it would be better added to the documentation somewhere. 

Also, it still needs a reference to a full explanation of this method of storing and accessing data.

josephwright:
> Also, it still needs a reference to a full explanation of this method of storing and accessing data.

The addition to the kernel is a stub: it's (currently) down to `babel` and `polyglossia` to handle data storage. Both of them do it their own way, and what's important here is a consistent way of getting the info _out_.

josephwright:
> All of @josephwright's explanation here is fine, but it would be better added to the documentation somewhere.

@car222222  Did you see the adjustment I made to the LaTeX News entry? I've added an example which I hope might help here.

car222222:
Understood, thus it seems that a short example may therefore be better here, along the lines of this one: 

"The idea is that \BCPdata{X} will expand to some BCP-47 subtag, so e.g.

\expanded{\noexpand\foo{\BCPdata{language}}}
becomes

\foo{en}
(or whatever) and then \foo has the data to do 'whatever'."

And refer to polyglossia/babel for more complex examples and the details of how each of these stores the data.

car222222:
. . . and removing all the current examples.

josephwright:
> . . . and removing all the current examples.

You mean the breakdown of parts? That won't really work, as the idea is that _we_ define what is supposed to be returned, then `babel`/`polyglossia` 'make it so'.

car222222:
Aha!  That is not made at all clear.  

In that case, this whole set-up and reasoning needs more careful explanation somewhere, with a full list of what 'we define' (and why?) and how this list can be extended (if it can be!).

car222222:
On the ltnews entry: 

1.  The new example is useful but it leaves a mystery: 
What is the relationship between "\BCPdata{casing}" and "locale"? 
Is "casing" the name of a data point whose value must be a locale? 

2.  See also relevant further comments on the original text.

base/doc/ltnews37.tex
```diff
@@ -289,6 +289,20 @@ \subsection{\cs{IfFileAtLeastTF}}
 %
 \githubissue{1015}
 
+\subsection{\cs{BCPdata}}
+
+Improvements in the Unicode handling for case changing have highlight that
```
FrankMittelbach, line 294: have highlighted ?

FrankMittelbach (approved):
minor comment/query

FrankMittelbach:
> > . . . and removing all the current examples.
> 
> You mean the breakdown of parts? That won't really work, as the idea is that _we_ define what is supposed to be returned, then `babel`/`polyglossia` 'make it so'.

given that, wouldn't it be more correct if the default (without babel or polyglossia loaded) returns the "english" in full glory, ie all value populated?

davidcarlisle:
> And by what is it returned?

it's just tex expansion.

> Where do "we" define "what is supposed to be returned"?

The list added to clsguide in this pr defines the list of possible arguments. The possible returned values would be a rather long list as that's BCP-47

jspitz:
I'm not sure `extension.u` will be used in the foreseeable future. But we have requests for transliteration in polyglossia, and in this case this extension might be useful. I've implemented it now for the sake of completeness.

As for `tag`, I think we talked about it. It might have been overlooked (or I might have misunderstood), but I find it handy.

jbezos:
Here is what `babel` provides with `\localeinfo` (copypasted from the manual):

> tag.ini is the tag of the ini file (the way this file is identified in its name).
>
> tag.bcp47 is the full BCP 47 tag [...]. This is the value to be used for the ‘real’ provided tag (babel may fill other fields if they are considered necessary).
>
>language.tag.bcp47 is the BCP 47 language tag.

[...]

>script.tag.bcp47 is the BCP 47 tag of the script used by this locale. This is a required field for the fonts to be correctly set up, and therefore it should be always defined.

[...]

>region.tag.bcp47 is the BCP 47 tag of the region or territory. Defined only if the locale loaded actually contains it (eg, es-MX does, but es doesn’t), which is how locales behave in the CLDR.
>
> variant.tag.bcp47 is the BCP 47 tag of the variant (in the BCP 47 sense, like 1901 for German). 
>
> extension.*s*.tag.bcp47 is the BCP 47 value of the extension whose singleton is *s* (currently the recognized singletons are x, t and u). The internal syntax can be somewhat complex, and this feature is still somewhat tentative. An example is classiclatin which sets extension.x.tag.bcp47 to classic. 


What `\BCPdata` does in `babel` is just to call #1.tag.bcp47, which means the full tag is not currectly accesible with `\BCPdata`, but it can be added more or less straightforwardly. 

Currently no locale uses `extension.u` or `extension.t`.

jspitz:
> > > I'm not sure `extension.u` will be used in the foreseeable future.
> > 
> > 
> > Sorry, I am talking about `extension.t` here, of course
> 
> Even if it not used now, in case it is used later, the documentation should be correct. So I wonder if the example Joseph added to clsguide is the right one. That means should it say
> 
> ```
> \item \texttt{extension.t} (transformation, e.g.,~\texttt{en-t-ja})
> ```
> 
> or
> 
> ```
> \item \texttt{extension.t} (transformation, e.g.,~\texttt{t} if the full tag is \texttt{en-t-ja})
> ```
> 
> or something completely different?

Both are correct, but not very clear. I think it should become clear that what follows the tag is the language, or script, to which transformation (translation, transliteration, or transcription) has happened.

`polyglossia` documents it as

```
extension.t The subtag indicating transformation between languages or
scripts (e.g., ja as in en-t-ja: English transformed to Japanese)
```

u-fischer:
> `polyglossia` documents it as
> 
> ```
> extension.t The subtag indicating transformation between languages or
> scripts (e.g., ja as in en-t-ja: English transformed to Japanese)
> ```

That sounds good to me.

jspitz:
>     * why is does the script example use a titlecase `Latn` and  the extension.u  lowercase `latn`?

By definition. The [specification](https://www.rfc-editor.org/bcp/bcp47.txt) holds:

```
All subtags, including extension and private
   use subtags, use lowercase letters with two exceptions: two-letter
   and four-letter subtags that neither appear at the start of the tag
   nor occur after singletons.  Such two-letter subtags are all
   uppercase (as in the tags "en-CA-x-ca" or "sgn-BE-FR") and four-
   letter subtags are titlecase (as in the tag "az-Latn-x-latn").
```


For the `u` extension, see also https://www.rfc-editor.org/rfc/rfc6067:

```
As required by [[BCP47](https://www.rfc-editor.org/rfc/rfc6067#ref-BCP47)], the use of uppercase or lowercase
letters is not significant in the subtags used in this extension.  The canonical
form for all subtags in the extension is lowercase.
```

jbezos:
I‘ve just read `clsguide` (I've been very busy these last three days), and I have nothing to add to what Jürgen has said.

car222222:
@davidcarlisle : judging by the subsequent lengthy discussions, it does not seem to be as straightforward as you claim!

But many thanks for your description of what the text is intended to convey; I shall now reread it and see if I can adjust it to make sense along the lines you suggest are intended.  For example, I did not pick up that, with luck on odd Tuesdays at least, only the listed (as examples) values of its argument might cause its expansion to be non-empty and be a BCP-42 tag. My understanding of the term BCP data is that it is the value of some key in the data for a given BCP-42 tag, rather than only the tag itself, which I would have expected to be the argument. 

There needs to be a reference to where this is all explained, probably in the babel or polyglossia documentation? Thanks to Javier for helping me to understand this.

I still have no clear understanding of why more than one such tag might be needed, and hence produced (by supplying different argument values).

car222222:
42 ==> 47  (I never was very good with numbers, so I became, in a previous life, a mathematician!)

car222222:
@davidcarlisle wrote, somewhat strangely:
 The possible returned values would be a rather long list as that's BCP-47.

But no one is asking for, or needs a list, just the fact that the return value is always such a tag.  Even better would be an explanation of why different arguments are needed and exactly which tag they each expand to (also, exactly when this expansion is not void).  Also, what each of these tag values is used for.

jbezos:
After re-reading the thread, a couple of quick thoughts:
* I think we are *too* biased towards `babel` and `polyglossia`. There are other ways to localize documents, like `xgreek`, `arabluatex`, `xepersian`, the CJK frameworks, etc. I think they must be taken into account somehow.
* What `tag` means is somewhat fuzzy, because the requested locale can be different from the supplied one. For example, if I request `es-Latn` or `es-AR`, the supplied locale by `babel` is `es` (it performs a simple, or rather simplistic, lookup), but if I request `es-MX` or `es-Latn-MX`, the supplied locale is `es-MX`.

car222222:
Agreed: @jbezos makes many good and pertinent points here.

I had also been wondering what should be returned for non-default (i.e. US-English only) documents that do not use babel or polyglossia (or even those that use no package at all to handle distinct languages, such as my favo_u_rite "en-uk":-).

base/doc/clsguide.tex
```diff
@@ -1138,6 +1138,48 @@ \subsection{Normalising spacing}
...
+to takes one argument and expand to the BCP-47 information matching this
+argument. The supported arguments are the BCP-47 tag breakdowns:
+\begin{itemize}
+  \item \texttt{language} (e.g.,~\texttt{de})
+  \item \texttt{region} (e.g.,~\texttt{AT})
+  \item \texttt{script} (e.g.,~\texttt{Latn})
+  \item \texttt{variant} (e.g.,~\texttt{1901})
+  \item \texttt{extension.t} (transformation, e.g.,~\texttt{en-t-ja})
+  \item \texttt{extension.u} (additional locale
+    information, e.g.,~\texttt{ar-u-nu-latn})
```
car222222, line 1160: This description is too vague, and there is no indication of why the tag in the example is l additional information.  Also, suppose that there is more than one piece of such information?

base/doc/clsguide.tex
```diff
@@ -1138,6 +1138,48 @@ \subsection{Normalising spacing}
...
+  \item \texttt{extension.x} (private use area, e.g.,~\texttt{la-x-classic})
+\end{itemize}
+The information for the \emph{main} language for a document is be provided
+if these are prefixed by \texttt{main.}, e.g.~\texttt{main.language} will
+expand to the main language even if another language is currently active.
+
+In addition to the tag breakdown, the following semantic arguments are
```
car222222, line 1167: What is a "tag breakdown"? 
What tag is broken down, and how/why?

car222222 (commented):
A lot more queries on the text in clsguide, following on from @jbezos and @jspitz and @u-fischer adding lots of useful information.

jbezos:
The *internal* syntax in extensions should be defined, too. For example, with `lg-Scrt-RG-x-first-second-third`, is `extension.x` `first-second-third`, or `first, second, third`, or `first second third`, or something else? What about `el-x-iota-xxxx`, where `xxxx` is another private extension? With the `u` extension, with its internal structure, things have more fun. Either I wouldn’t document them or they must be tagged as ‘tentative’, ‘experimental’...

jspitz:
I agree we should define on that. A comma separated list makes most sense to me. The order is arbitrary, right?

base/doc/clsguide.tex
```diff
@@ -1138,6 +1138,40 @@ \subsection{Normalising spacing}
...
+by \pkg{babel} or \pkg{polyglossia}, the command \cs{BCPdata} is defined
+by the kernel. The initial kernel definition absorbs one argument and
+expands to an empty value, as the kernel does not track localisation.
+However, if \pkg{babel} or \pkg{polyglossia} are loaded, it is redefined
+to takes one argument and expands to the BCP-47 information matching this
+argument. The supported arguments are the BCP-47 tag breakdowns:
+\begin{itemize}
+  \item \texttt{language} (e.g.,~\texttt{de})
+  \item \texttt{region} (e.g.,~\texttt{AT})
+  \item \texttt{script} (e.g.,~\texttt{Latn})
```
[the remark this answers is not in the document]
jspitz: > Why Latn/latn, not "Latin"?

Script subtags always have four letters, hence `Latn`, not `Latin`, and `Laoo`, not `Lao`

jspitz:
> given that, wouldn't it be more correct if the default (without babel or polyglossia loaded) returns the "english" in full glory, ie all value populated?

I am not sure the value without any language package loaded should be English. Sure it tends towards English, but isn't it rather some more (language-)"abstract" thing, similar to the `C` locale?

I am wondering whether we shouldn't use the private use area here and return something such as `x-latex` or even `x-tex`.

For polyglossia, the question also comes up, as we have a special "language" (we call it "latex"), with the default (English) captions, which is loaded when language settings have to be "reset" to the extra-polyglossia state. Currently we do not return any BCP47 information for this internal "language" which is in fact always immediately overwritten with a real language anyway. But I feel like it would be good to tag this as well.

FrankMittelbach:
> > given that, wouldn't it be more correct if the default (without babel or polyglossia loaded) returns the "english" in full glory, ie all value populated?
> 
> I am not sure the value without any language package loaded should be English. Sure it tends towards English, but isn't it rather some more (language-)"abstract" thing, similar to the `C` locale?
> 
> I am wondering whether we shouldn't use the private use area here and return something such as `x-latex` or even `x-tex`.
> 
> For polyglossia, the question also comes up, as we have a special "language" (we call it "latex"), with the default (English) captions, which is loaded when language settings have to be "reset" to the extra-polyglossia state. Currently we do not return any BCP47 information for this internal "language" which is in fact always immediately overwritten with a real language anyway. But I feel like it would be good to tag this as well.

my rationale was that LaTeX sets up a default language which is *English* as far has hyphenation and captions are concerned and in that respect is haves as if english was requested. Thus if anything is querying the context it seems to me sensible to return the same values. Given that this default is hard-wired I see not much point in requring to interpret it differently from an explicit setting.

car222222:
@FrankMittelbach Surely the default is US-English?  Or am I outdated? 
Note also that the expansion should be a BCP-47 tag, rather than a language name.

FrankMittelbach:
> @FrankMittelbach Surely the default is US-English? Or am I outdated? Note also that the expansion should be a BCP-47 tag, rather than a language name.

yes, on both accounts. I could have been more precise. But this is what I meant.

u-fischer:
> I am not sure the value without any language package loaded should be English. Sure it tends towards English, but isn't it rather some more (language-)"abstract" thing, similar to the `C` locale?

We do have to set a language in the PDF catalog (in the /Lang) entry for standards and also if the PDF is tagged and the only sensible default is some sort of english. Once the dust has settled down a bit I plan to retrieve \BCPdata{main.language} (and perhaps also \BCPdata{main.region}) and so the first should have a default value of `en`  imho.

base/doc/clsguide.tex
```diff
@@ -1138,6 +1138,48 @@ \subsection{Normalising spacing}
...
+  \item \texttt{variant} (e.g.,~\texttt{1901})
+  \item \texttt{extension.t} (transformation, e.g.,~\texttt{en-t-ja})
+  \item \texttt{extension.u} (additional locale
+    information, e.g.,~\texttt{ar-u-nu-latn})
+  \item \texttt{extension.x} (private use area, e.g.,~\texttt{la-x-classic})
+\end{itemize}
+The information for the \emph{main} language for a document is be provided
+if these are prefixed by \texttt{main.}, e.g.~\texttt{main.language} will
```
car222222, line 1164: Why is there "information for any language", and is the "main language" always defined?

Better: . . . for the main language of the document.
And start with this clause: If any of these is prefixed by . . . then . . . 

Not sure about "provided"? 

Without this prefix, which language is used?  
Is it possible to prefix them with an explicit language name?

car222222 (commented):
One more thing!

car222222:
@jspitz   What, please, is the "C locale"?  Is this C as in the programming language (an oldie but goodie!??)

@u-fischer  Are you sure that this default should not be en-us?
My reasoning is that if you do not specify anything more precise, then the only hyphenation patterns you get are (most unfortunately) for US English.

jspitz:
> @jspitz What, please, is the "C locale"? Is this C as in the programming language (an oldie but goodie!??)

Yes. But I am fine with setting it to en-US given the arguments here.

u-fischer:
> Are you sure that this default should not be en-us?

it is currently, but imho a more vague default ("some sort of english") is better.  We don't know if the document writes color or colour. (From the PDF side, using only `en` is fine.)

jbezos:
Note `es-US` in the CLDR is the ‘default content locale’ for `en`, which basically means they are synonymous. So, it doesn’t matter. See https://cldr.unicode.org/translation/translation-guide-general/default-content

car222222:
@jbezos  But @u-fischer thinks we should not commit ourselves to US-English!  So how does one indicate that "the language is en, but no assumption should be made about which variant is used for things such as units, dates, hyphenation and spelling(??)".

Is spelling really a concern of BCP or locale??  that seems very odd and it would imply that, for older documents within the U.K., each printer/publisher needs to have a code assigned.

u-fischer:
> But @u-fischer thinks we should not commit ourselves to US-English!

That is not what I meant. We can not avoid that *implicitly* the default is en-US, I only think it looks better if we don't set it *explicitly* (even more as that is not required and makes no difference).

jbezos:
> So how does one indicate that "the language is en, but no assumption should be made about which variant is used for things such as units, dates, hyphenation and spelling(??)".

@car222222 I think what you want is the ‘global’ English, with tag `en-001`.

davidcarlisle:
@car222222  see for example https://www.ibm.com/docs/en/aix/7.1?topic=locale-c-posix

 `LANG=C`   or `LC_ALL=C`  sets up minimal locale defaults in `*ix` systems

car222222:
@davidcarlisle Fascinating stuff, but is it relevant to language-related data for use in document processing?

car222222:
More fun facts:

001 and 150 are UN M.49 codes for "the world" and "Europe" respectively.

en_US_POSIX was invented to refer to the C (or 'null') locale code used in POSIX libraries. It's basically American English, with some peculiarities.

Note that this last one is under heavy discussion at this instant.

car222222:
And UN M.49 is, of course, from the United Nations publication "Standard Country or Area Codes for Statistical Use".

davidcarlisle:
> @davidcarlisle Fascinating stuff, but is it relevant to language-related data for use in document processing?

Well LANG is supposed to set up human languge related options but LANG=C is  "unset all options and set up a minimal ascii environment enough to use printf in C"

Jürgen was floating a similar base `x-latex` language here that provided minimal defaults without necessarily claiming to be English. 

But as Ulrike says, we need a "real" language anyway for some pdf cases, so defaulting to en(-us) is probably better

car222222:
@davidcarlisle OK, I see the relevance now, but as you suggest, the idea of a non-language or an empty locale is not really useful for documents (and maybe not for code either, in the real world!). 

It is not clear whether hyphenation and other line-breaking conventions ate ever part of locales: this needs investigation. 

Noting that, for tex and friends at the engine level, hyphenation is the only language thing of which it is aware. But there may be other locale-related concepts that are built-in to the engine.

jbezos:
A technical point. While `x-subtag` is allowed by the BCP 47, the CLDR requires a base language. It can be `und` (‘undetermined’), which is what `babel` sets when no language is explicitly loaded (in addition, the script is set to `Latn`).

But imo the default in LaTeX should be some English, because after all it’s the actual default. Just write a document with the single command `\chaptername`. Concealing this very fact is of no help, I think.

josephwright:
>     * I think we are _too_ biased towards `babel` and `polyglossia`. There are other ways to localize documents, like `xgreek`, `arabluatex`, `xepersian`, the CJK frameworks, etc. I think they must be taken into account somehow.

In an ideal world, the locale data would be handled by a single mechanism, presumably in the kernel, even though _outcomes_ would be delegated to packages. But we are not in an ideal world :) So the idea is to find a mechanism that is 'good enough': if someone wants to modify `xgreek` or whatever to define `\BCPdata` appropriately, that's great, but IO think most users will load either `babel` or `polyglossia` anyway.

FrankMittelbach:
> But imo the default in LaTeX should be some English, because after all it’s the actual default. Just write a document with the single command `\chaptername`. Concealing this very fact is of no help, I think.

precisely, that was my starting point. I would expect that \BCPdata in the kernel should produce exactly the same as when babel or polyglossia or ... is loaded in a way that the LaTeX defaults all remain the same, ie, when you ask for English with US hyphenation

josephwright:
I can arrange that the 'out of the box' `\BCPdata` gives the right parts of `en-US`: does that sound OK?

car222222:
I did not know that en-US has "parts". Please explain?

josephwright:
> I did not know that en-US has "parts". Please explain?

Well, only two parts: `en` and `US`, but they need to come out from the right requests.

car222222:
@josephwright Sure!  OK.

jspitz:
> > I did not know that en-US has "parts". Please explain?
> 
> Well, only two parts: `en` and `US`, but they need to come out from the right requests.

I think you should also provide
```
{ script }  { Latn }
```

jbezos:
I’ve just uploaded to CTAN `babel`, but it doesn’t include the `main` data yet — to be implemented in the next release (by mid May). But a little question, @u-fischer: do you need these values in the preamble or at `\begin{document}` is enough?

u-fischer:
> But a little question, @u-fischer: do you need these values in the preamble or at `\begin{document}` is enough?

I can set the value at every time, or you can do it. Basically only something like this is needed (the parentheses around the `de-DE` are important!):

~~~~
\ExplSyntaxOn
\IfDocumentMetadataTF
 {\pdfmanagement_add:xxx{Catalog}{Lang}{(de-De)}}{}
\ExplSyntaxOff
~~~~

It doesn't harm if this is done more than once, the last value will win.

jbezos:
> It doesn't harm if this is done more than once, the last value will win.

👍 Perfect, because the main language can be reset.

josephwright:
@FrankMittelbach Do we need rollback support here? I'm wondering mainly about roll-forward ...

FrankMittelbach:
> @FrankMittelbach Do we need rollback support here? I'm wondering mainly about roll-forward ...

good question. it is a rather fundamental change, but it only takes effect really if babel or polyglossia is new as well (and they don't have roll forward, so most likely people get it only when they  (or overleaf) update the whole system). So I guess, probably not really necessary and if it is rather complicated then I wouldn't do it.

jspitz:
From `polyglossia` perspective, what this change does is to "silently" fix casing issues (e.g., with Greek) that have been "silently" broken with the advent of the new `\Make<thing>case` implementation in 2022/11. So it's merely a bug fix, and probably if so at all, the rollback/forward would have to be done in 2022/11. From my perspective, though, it's not necessary (but I might overlook implications).

josephwright:
@jspitz The new case changer was brought in because the updates to UTF-8 handling didn't work with the classical approach. We can't really rollback the UTF-8 change, so as a result there was no sensible way to rollback the adjustment to `\MakeUppercase`. So the only thing I was wondering about was the uses of `\BCPdata`, which as @FrankMittelbach says doesn't really make sense to rollback.

Issue #1033, opened 2023-04-05
MakeUppercase fails with "latin letter j with caron" under pdflatex


gmilde:
## Brief outline of the bug
Upcasing the letter [j with caron](https://en.wikipedia.org/wiki/J%CC%8C) (U+01F0) leads to
`LaTeX Error: Unicode character ̌ (U+030C) not set up for use with LaTeX.`
and there is no caron on the captital J.
This is a regression from the TL21 behaviour, where there is no error and a caron is placed
on the upcased letter.

This seems related to "ǰ" not having a pre-composed capital variant.
It is similar to the behaviour for some Greek letters with diacritics, that also do not have a capital pre-composed variant (cf. issue #987; however, in contrast to the affected Greek letters, the capital J with caron seems to have a use case).

## Minimal example showing the bug

```latex
\RequirePackage{latexbug} 
\documentclass[a4paper]{article}
\usepackage[T1]{fontenc}
\begin{document}
ǰ → \MakeUppercase{ǰ}
\end{document}
```
## Log file (required) and possibly PDF file  
This is pdfTeX, Version 3.141592653-2.6-1.40.25 (TeX Live 2023) (preloaded format=pdflatex 2023.4.3)  5 APR 2023 02:16
entering extended mode
 restricted \write18 enabled.
 %&-line parsing enabled.
**case-changing-diacritics.tex
(./case-changing-diacritics.tex
LaTeX2e <2022-11-01> patch level 1
L3 programming layer <2023-03-30>
(/usr/local/texlive/2023/texmf-dist/tex/latex/latexbug/latexbug.sty
Package: latexbug 2022/06/14 v1.0n Bug-classification
)
(/usr/local/texlive/2023/texmf-dist/tex/latex/base/article.cls
Document Class: article 2022/07/02 v1.4n Standard LaTeX document class
(/usr/local/texlive/2023/texmf-dist/tex/latex/base/size10.clo
size10.clo 2022/07/02 v1.4n Standard LaTeX file (size option)
)
\c@part=\count185
\c@section=\count186
\c@subsection=\count187
\c@subsubsection=\count188
\c@paragraph=\count189
\c@subparagraph=\count190
\c@figure=\count191
\c@table=\count192
\abovecaptionskip=\skip48
\belowcaptionskip=\skip49
\bibindent=\dimen140
)
(/usr/local/texlive/2023/texmf-dist/tex/latex/base/fontenc.sty
[...]

PR #1054, opened 2023-04-19
Add \DeclareUppercaseMapping, etc.

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
-[x] Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/doc/ltnews37.tex
+++ b/base/doc/ltnews37.tex
@@ -358,9 +358,9 @@ generally or to a specific locale (see also the next section). A small number
 of pre-defined customisations have been set up in the kernel where the outcomes
 for \pdfTeX{} should be different for those from Unicode engines. For
 example
-\begin{quotation}
-  |\DeclareUppercaseMapping{"01F0}{\v{J}|
-\end{quotation}
+\begin{verbatim}
+  \DeclareUppercaseMapping{"01F0}{\v{J}
+\end{verbatim}
 allows \v{J} to be produced in 8-bit engines: without this customisation,
 an error would occur as there is no pre-composed \v{J} in Unicode. More detail
 is given in \texttt{usrguide}.
--- a/base/doc/ltnews37.tex
+++ b/base/doc/ltnews37.tex
@@ -359,7 +359,7 @@ of pre-defined customisations have been set up in the kernel where the outcomes
 for \pdfTeX{} should be different for those from Unicode engines. For
 example
 \begin{verbatim}
-  \DeclareUppercaseMapping{"01F0}{\v{J}
+  \DeclareUppercaseMapping{"01F0}{\v{J}}
 \end{verbatim}
 allows \v{J} to be produced in 8-bit engines: without this customisation,
 an error would occur as there is no pre-composed \v{J} in Unicode. More detail
--- a/base/doc/ltnews37.tex
+++ b/base/doc/ltnews37.tex
@@ -341,6 +341,25 @@ line.
 %
 \githubissue{1015}
 
+\subsection{\cs{DeclareLowercaseMapping}, \cs{DeclareTitlecaseMapping}
+  and \cs{DeclareUppercaseMapping}}
+
+The move from a case-changing approach using \cs{lccode} and \cs{uccode} data
+to one where information is stored by a kernel-managed structure left a gap in
+the ability of the user to \emph{tune} the case changing outcomes. This has now
+been addressed by the addition of three commands
+\begin{itemize}
+  \item \cs{DeclareLowercaseMapping}
+  \item \cs{DeclareTitlecaseMapping}
+  \item \cs{DeclareUppercaseMapping}
+\end{itemize}
+which can be used to customise the outcome for codepoints. This can be apply
+generally or to a specific locale (see also the next section). A small number
+of pre-defined customisations have been set up in the kernel where the outcomes
+for \pdfTeX{} should be different for those from Unicode engines.
+%
+\githubissue{1033}
+
 \subsection{\cs{BCPdata}}
 
 Improvements in the Unicode handling for case changing have highlighted that
--- a/base/doc/ltnews37.tex
+++ b/base/doc/ltnews37.tex
@@ -353,7 +353,7 @@ been addressed by the addition of three commands
   \item \cs{DeclareTitlecaseMapping}
   \item \cs{DeclareUppercaseMapping}
 \end{itemize}
-which can be used to customise the outcome for codepoints. This can be apply
+which can be used to customise the outcome for codepoints. This can be applied
 generally or to a specific locale (see also the next section). A small number
 of pre-defined customisations have been set up in the kernel where the outcomes
 for \pdfTeX{} should be different for those from Unicode engines. For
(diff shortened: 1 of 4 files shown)


josephwright:
I've gone here for a solution to #1033 that uses an optional argument for the locale. The alternative approach with two commands is a possible: I'm not sure which is clearer.

base/doc/usrguide.tex
```diff
@@ -1216,8 +1216,21 @@ \section{Case changing}
...
+\begin{decl}
+  |\DeclareLowercaseMapping| \oarg{locale} \arg{codepoint} \arg{output} \\
+  |\DeclareTitlecaseMapping| \oarg{locale} \arg{codepoint} \arg{output} \\
+  |\DeclareUppercaseMapping| \oarg{locale} \arg{codepoint} \arg{output}
+\end{decl}
+All three take a \meta{codepoint} (as an integer expression) and will
+result in the \meta{output} being produced under the appropriate case changing
+operation. The optional \meta{locale} can be given if the mapping should only
+apply to a specific one: this is given in BCP-47 format.
+
```
FrankMittelbach, line 1233: I suggest to add a useful link for BCP-47 maybe https://en.wikipedia.org/wiki/IETF_language_tag and also give two examples, one with and one without the optional locale

base/doc/ltnews37.tex
```diff
@@ -341,6 +341,25 @@ \subsection{\cs{IfFileAtLeastTF}}
...
+  \item \cs{DeclareLowercaseMapping}
+  \item \cs{DeclareTitlecaseMapping}
+  \item \cs{DeclareUppercaseMapping}
+\end{itemize}
+which can be used to customise the outcome for codepoints. This can be apply
+generally or to a specific locale (see also the next section). A small number
+of pre-defined customisations have been set up in the kernel where the outcomes
+for \pdfTeX{} should be different for those from Unicode engines.
+%
+\githubissue{1033}
```
FrankMittelbach, line 361: maybe one example + a reference to usrguide (or at least a reference to usrguide for more details).

base/ltfinal.dtx
```diff
@@ -1209,6 +1214,24 @@
 \cs_new_eq:NN \CaseSwitch \text_case_switch:nnnn
 \cs_new_eq:NN \DeclareCaseChangeEquivalent
   \text_declare_case_equivalent:Nn
+\NewDocumentCommand \DeclareLowercaseMapping { o m m }
```
FrankMittelbach, line 1217: I'm fine with use of an optional argument

base/ltfinal.dtx
```diff
@@ -1209,6 +1214,24 @@
...
+  {
+    \IFNoValueTF {#1}
+      { \text_declare_lowercase_mapping:nn {#2} {#3} }
+      { \text_declare_lowercase_mapping:nnn {#2} {#1} {#3} }
```
FrankMittelbach, line 1221: ... but I'm less sure about the order of arguments here: it might seems natural to put the locale in the middle in this case, but there may be other declarations that have a dependency on the locale and they many not naturally have "locale" in second place. My guess it is better to put it first in all such commands so that the logic is: "do XYZ"  or "for locale=abc do XYZ"  with XYZ having as many arguments as needed but in both cases in the same order.

FrankMittelbach (approved):
some comments that might need addressing or a decision before merging

base/doc/usrguide.tex
josephwright: OK will do

base/doc/ltnews37.tex
josephwright: Cool again will do

base/ltfinal.dtx
josephwright: I only finalised the `expl3` layer yesterday so I think we are free to change - it does happen. I did wonder about the ordering, which would comes currently as
```latex
\text_declare_lowercase_mapping:nnn { "01F0 } { xx } { \v { j } }
```
but would if changed be
```latex
\text_declare_lowercase_mapping:nnn { xx } { "01F0 } { \v { j } }
```
I can live with either: the latter is closer to what `\text_lowercase:nn` does, but at the same time here with three arguments I could see a logic to having the codepoint first.

FrankMittelbach: as I said, my argument is that later on you may have other declarations that are locale specific and they may have more or less number of additional arguments so it seems more consistent to have the locale always first or for the global case "missing" rather than somewhere in the middle (or missing there). Or saying it differently are there a lot of things related to a code point and then subdivided by locale each or a lot of things related to a locale? I think the latter.

FrankMittelbach: another argument is that this matches order in the optional arg version

josephwright: OK, I'll adjust in `expl3`, do a release and adjust in the 2e PR

base/doc/ltnews37.tex
```diff
@@ -341,6 +341,32 @@ \subsection{\cs{IfFileAtLeastTF}}
...
+The move from a case-changing approach using \cs{lccode} and \cs{uccode} data
+to one where information is stored by a kernel-managed structure left a gap in
+the ability of the user to \emph{tune} the case changing outcomes. This has now
+been addressed by the addition of three commands
+\begin{itemize}
+  \item \cs{DeclareLowercaseMapping}
+  \item \cs{DeclareTitlecaseMapping}
+  \item \cs{DeclareUppercaseMapping}
+\end{itemize}
+which can be used to customise the outcome for codepoints. This can be apply
```
muzimuzhi, line 356: "be apply" -> "be applied"
```suggestion
which can be used to customise the outcome for codepoints. This can be applied
```

base/doc/ltnews37.tex
```diff
@@ -341,6 +341,32 @@ \subsection{\cs{IfFileAtLeastTF}}
...
+generally or to a specific locale (see also the next section). A small number
+of pre-defined customisations have been set up in the kernel where the outcomes
+for \pdfTeX{} should be different for those from Unicode engines. For
+example
+\begin{verbatim}
+  \DeclareUppercaseMapping{"01F0}{\v{J}
```
muzimuzhi, line 362: Unpaired braces?
```suggestion
  \DeclareUppercaseMapping{"01F0}{\v{J}}
```

base/doc/usrguide.tex
```diff
@@ -1216,8 +1216,35 @@ \section{Case changing}
...
+\end{decl}
+All three take a \meta{codepoint} (as an integer expression) and will
+result in the \meta{output} being produced under the appropriate case changing
+operation. The optional \meta{locale} can be given if the mapping should only
+apply to a specific one: this is given in BCP-47 format
+(\url{https://en.wikipedia.org/wiki/IETF_language_tag}). For example,
+the kernel customises the mapping for U+01F0 (\v{j}) when uppercasing in
+8-bit engines:
+\begin{quotation}
+  |\DeclareUppercaseMapping{"01F0}{\v{J}|
```
muzimuzhi, line 1237:
```suggestion
  |\DeclareUppercaseMapping{"01F0}{\v{J}}|
```

muzimuzhi (commented):
found some typos

Diegiwg (approved):

davidcarlisle (changes requested):
code change looks good `\v{J` as reported by @muzimuzhi  in ltnews stlll outstanding

davidcarlisle (approved):

PR #1056, opened 2023-04-24
Use 'mac' instead of 'function'

dr-scsi:
* base/doc/usrguide.tex (subsection{Access to the argument specification}): Talk about 'mac' instead of 'function' like in the rest of the document.

### DIFF
--- a/base/doc/usrguide.tex
+++ b/base/doc/usrguide.tex
@@ -844,23 +844,23 @@ The argument specifications for document commands and environments are
 available for examination and use.
 
 \begin{decl}
-  |\GetDocumentCommandArgSpec| \arg{function}         \\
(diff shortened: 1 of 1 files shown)


muzimuzhi:
From the diff, it seems the issue title should be "Use '**cmd**' instead of 'function'", not "Use '**mac**' ...".

FrankMittelbach:
@dr-scsi preparing a new pr seems overkill, why not stick with this one and update? but it is only the title so all it needs is changing the title.  --- there is no competition for increasing the pr numbers :-)

dr-scsi:
@FrankMittelbach - Thanks, I wasn't sure how you deal with 'amended' commits; and to be honest, I had some bad experience with Github and amending commits to PR's.  I'll try not to increase the PR numbers in future `;-)`

PR #1068, opened 2023-05-22
Cmd hooks with arguments

PhelypeOleinik:
This PR adds the ability to access a command's arguments from generic `cmd` hooks using `\AddToHookWithArguments`. 

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


base/ltcmdhooks.dtx
```diff
@@ -76,8 +88,8 @@
 %
 %    The hooks are not physically present before
 %    \verb=\begin{document}= (i.e., using a command in the preamble
```
u-fischer, line 90: I would suggest to add here that commands in the aux-file are not seeing the hooks either.

base/ltcmdhooks.dtx
```diff
@@ -271,6 +290,55 @@
 % If you do that then both hooks automatically work and are patched
 % into the right places.
 %
+% \subsection{Arguments and redefining commands}
+% \label{sec:redef-warn}
+%
+% The code in \pkg{ltcmdhooks} does its best to find out how many
+% arguments a given command has, an to insert the appropriate call to
```
u-fischer, line 297: and to insert
    ^

base/ltcmdhooks.dtx
```diff
@@ -271,6 +290,55 @@
...
+% \cs{UseHookWithArguments}, so that the arguments seen by the hook are
+% exactly those grabbed by the command (the hook, after all, is a macro
+% call, so the arguments have to be placed in the right order, or they
+% won't match).
+%
+% When using the package writer interface, as dicussed in
```
u-fischer, line 303: as discussed

base/ltcmdhooks.dtx
```diff
@@ -271,6 +290,55 @@
...
+% section~\ref{sec:pkg-author}, to change the position of the hooks in
+% your commands, you are also free to change how the hook code in your
+% command sees its arguments.  When a \hook{cmd} hook is declared with
+% \cs{NewHook} (or \cs{NewHookWithArguments} or other variations of
+% that), it loses its \enquote{generic} nature and works as a regular
+% hook.  This means that if you may choose to declare it without
```
u-fischer, line 309: This means that you may

base/ltcmdhooks.dtx
PhelypeOleinik: Done, thanks.

This begs the question: is the timing right, or should the commands be patched earlier?  Or should we have a command `\hook_cmd_add_even_if_before_begindocument_because_I_need_it_in_the_aux:n`?

base/ltcmdhooks.dtx
```diff
@@ -75,9 +87,12 @@
 %    \end{description}
 %
 %    The hooks are not physically present before
-%    \verb=\begin{document}= (i.e., using a command in the preamble
-%    will never execute them) and if nobody has declared any code for
-%    them, then they are not added to the command code ever.  For
+%    \verb=\begin{document}=\footnote{More specifically, they are
+%    inserted in the commands after the \hook{begindocument} hook, so
+%    they are also not present while \LaTeX{} is reading the
+%    \texttt{.aux} file.} (i.e., using a command in the preamble
```
FrankMittelbach, line 93: haven't I read a discussion that this was changed so that they are available during .aux reading? If so some of the later documentation would need to change too.

FrankMittelbach (commented):
not done with the review but I have to stop for now ...

base/ltcmdhooks.dtx
PhelypeOleinik: I think you skipped a word while reading: they are _not_ present when reading the `.aux`. Thus my comment to Ulrike: https://github.com/latex3/latex2e/pull/1068#discussion_r1200456488

FrankMittelbach: Right I misread.

FrankMittelbach (approved):
some minor comments

davidcarlisle (approved):
I read over ltnews diff and comments in dtx and look good to me, I can't review detailed code this evening

base/ltcmdhooks.dtx
FrankMittelbach: Interesting question, but I think for now it might be better to leave it as that and just document the fact (which you did and I misread). If we really see a compelling need we might revisit but on the whole I think it make more confusion that it could gain. After all if we change it would either need to be a special "request" which is odd or we would need to move all code earlier and that doesn't quite work as package may add hooks in begin document

PhelypeOleinik: @FrankMittelbach Yes, not for now, definitely.  The problem with `cmd` hooks and `\begin{document}` is not the hooks added `\AtBeginDocument` (they are patched in, as long as the command exists), but the commands defined there:  that's why we delayed back then, and it's likely hard to do it any other way.

base/testfiles-lthooks/ltcmdhooks-002.tlg
```diff
@@ -14,7 +14,7 @@ Code labels for sorting:
 Data structure for label rules:
  x = 0 -> 
 Handled code for x
--> The generic hook 'cmd/foo/before':
+-> The generic hook 'cmd/foo/before' (unknown arguments):
 > Code chunks:
```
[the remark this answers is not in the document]
PhelypeOleinik: No(t yet).  I'll add a paragraph somewhere.  I'm thinking maybe `lthooks.dtx` should have a section explaning the output of `\ShowWook`.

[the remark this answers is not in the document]
PhelypeOleinik: I suggest we merge this, so we can do the dev release, then I add the documentation for the final release.  I likely won't be able to work on this this week.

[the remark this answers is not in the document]
FrankMittelbach: ok by me

PR #1071, opened 2023-05-23
fix fntguide typos

mbertucci47:
Small typo fixes in fntguide.tex :

- artifically -> artificially
- |\DeclareFontsubstitution| -> |\DeclareFontSubstitution|

### DIFF
--- a/base/doc/fntguide.tex
+++ b/base/doc/fntguide.tex
@@ -323,7 +323,7 @@ A less common value for font shape is:
   \end{minipage}
 \end{center}
 and there is also |ui| for upright italic, i.e., an italic shape but
-artifically turned upright. This is sometimes useful and available in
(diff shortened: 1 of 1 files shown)


FrankMittelbach (changes requested):
could you please in such pull's also change the date line of the file, which is either something inside \ProvidesFile or some "date" near the title of the thing (depends on the file) -- thanks. That helps me to merge without the need to go into the file and update the date afterwards

mbertucci47:
Of course, done. The only place for a date I saw was in `\date{...}`

FrankMittelbach:
quite likely in that particular file (as I said it depends on the file), but the convention we use is that we always change the date but a version number only if we change the code, ie in \ProvidesFile or \ProvidesPackage we only change the date part if it is updated for typos

PR #1074, opened 2023-05-25
Generate simpler variant for \cs_generate_from_arg_count:NNnn

muzimuzhi:
It's first n-arg already accepts an `<int expr>`.

Only date string is updated, not sure if it's enough.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,6 +6,12 @@ completeness or accuracy and it contains some references to files that
 are not part of the distribution.
 ================================================================================
 
+2023-05-26  Yukai Chou <muzimuzhi@gmail.com>
+
+    * ltcmd.dtx (subsection{Structure of \pkg{xparse} commands}):
+    Use simpler 'NNno' variant instead of 'NNVo' for
+    \cs_generate_from_arg_count:NNnn
+
 2023-05-22  Joseph Wright  <Joseph.Wright@latex-project.org>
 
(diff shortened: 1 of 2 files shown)


josephwright:
Sure but aren't we passing by value deliberately here?

josephwright:
As there is a code change, we'd need the version number to be stepped; also, it needs to go into `changes.txt`.

FrankMittelbach:
But given that we want to get rid of generating variants in 2e and move them to L3 kernel shouldn't that be done in your upcoming cleanup @josephwright ?

josephwright:
> But given that we want to get rid of generating variants in 2e and move them to L3 kernel shouldn't that be done in your upcoming cleanup @josephwright ?

I've added all of the variants from LaTeX2e to `expl3`, but I've not removed from the 2e sources as we need to allow for the `expl3` version to be released. I think we should give that 'a while' to make sure people don't get odd failures.

josephwright:
> Sure but aren't we passing by value deliberately here?

Ah, no, we are not: I see what you are getting at :)

If you can sort the changes and version string, we can merge.

josephwright:
> Please add `changes.txt` details and step the version number.

Also add a `\changes{}{}` line

josephwright (approved):

PR #1081, opened 2023-06-07
Lab review 01

u-fischer:
This adapts a few of the latex-lab files to the new hooks with arguments and corrects a number of small bugs. The bib-module handles now biblatex if hyperref is loaded too. The tagpdf version has been updated (which corrects a bug with empty structures). A new firstaid module collects some changes to external packages and classes , it is not loaded automatically by phase-III.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [ x] Test file(s) added (tests for changes in external packages must be called locally)
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated  (not sure if the firstaid should be mentioned?)


FrankMittelbach (commented):
looked it over only up firstaid, but have something coming up for the next hours

required/latex-lab/latex-lab-testphase.dtx
```diff
@@ -83,12 +83,18 @@
...
 \input{phase-II-latex-lab-testphase.ltx}
-\RequirePackage{latex-lab-testphase-block-tagging}
-\RequirePackage{latex-lab-testphase-sec}
-\RequirePackage{latex-lab-testphase-toc}             
-\RequirePackage{latex-lab-testphase-minipage}             
-\RequirePackage{latex-lab-testphase-graphic}             
-\RequirePackage{latex-lab-testphase-float}             
+\IfFormatAtLeastTF{2023-06-01}
+ {
+  \RequirePackage{latex-lab-testphase-block-tagging}
```
FrankMittelbach, line 88: thought we wanted to drop -tagging (not important though)

required/latex-lab/latex-lab-toc-kernel-changes.dtx
```diff
@@ -181,15 +179,12 @@
...
 %    \begin{macrocode}
-\NewHook{cmd/addcontentsline/before}
-\def\addcontentsline#1#2#3{% toc extension, type, tag
-    \def\addcontentsline@arg@one{#1}% name??
-    \def\addcontentsline@arg@two{#2}%
-    \def\addcontentsline@arg@three{#3}%
-    \UseHook{cmd/addcontentsline/before}%
+\NewHookWithArguments{cmd/addcontentsline/before}{3}
+\def\addcontentsline#1#2#3{% #1=toc extension, #2= heading type, tag
+    \UseHookWithArguments{cmd/addcontentsline/before}{3}{#1}{#2}{#3}%
```
FrankMittelbach, line 187: we should review if we really want to turn individual generic hooks into hooks with arguments. I think we should not and instead use \AddTohookWithArguments{addcontentsline} here and leave the generic one for code that goes around the whole command for one or the other reason. It is not less efficient (ie when the generic hook is not filled it is not patched in)

required/latex-lab/latex-lab-toc-kernel-changes.dtx
```diff
@@ -227,17 +222,27 @@
...
+%  \item[\#3] target  (the fourth argument of a \cs{contentsline}
+%  \end{itemize}
 %    \begin{macrocode}
-\NewMirroredHookPair{contentsline/number/before}{contentsline/number/after}%
-\NewMirroredHookPair{contentsline/text/before}{contentsline/text/after}%
-\NewMirroredHookPair{contentsline/page/before}{contentsline/page/after}%
+\NewMirroredHookPairWithArguments{contentsline/number/before}{contentsline/number/after}{1}%
+\NewMirroredHookPairWithArguments{contentsline/text/before}{contentsline/text/after}{4}%
+\NewMirroredHookPairWithArguments{contentsline/page/before}{contentsline/page/after}{4}%
 %    \end{macrocode}
```
FrankMittelbach, line 238: TODO: convince myself if we really need all of these hooks, I doubt it. In particular I doubt that the /after hooks need to have arguments (or need to be there).

FrankMittelbach (approved):
I may be wrong with some of the remarks, but I think a few of the should be addressed now  (or refuted) others could be addressed later after a merge.

required/latex-lab/latex-lab-block-tagging.dtx
```diff
@@ -1346,23 +1346,26 @@
 %    will help with the transion. But instead of using a
 %    \env{trivlist} we use a blockenv and some tagging for the title
 %    (as a Caption).
-%
+%    We do not want the \texttt{Strong} tagging from \cs{textbf} here,
+%    so we use \cs{bfseries} to set the font.
```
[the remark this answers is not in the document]
u-fischer: yes it isn't anymore (the text was added before). But imho `\bfseries` is better nevertheless, in case some document adds tagging to `\textbf`. I changed the text a bit.

required/latex-lab/latex-lab-testphase.dtx
u-fischer: yes, next round. I did run out of time  ;-)

required/latex-lab/latex-lab-toc-kernel-changes.dtx
u-fischer: well I need to inject link + tagging code around the number, the text and the page. And this code must be mirrored: the link code must be either inside or outside the tagging code. I don't quite see how  do it differently or how to produce a pair where the begin takes arguments and the end hook not.

FrankMittelbach: make the number an argument to the hook maybe, really comes down to configuration points by the end of the day ... but it isn't something that need to be solved or changed now, but perhaps add a TODO to review the current approach

required/latex-lab/config-firstaid.lua
```diff
@@ -0,0 +1,11 @@
+-- Tests for firstaid
+--
+-- Tests here only function locally due to the class not loaded in github actions!
+
```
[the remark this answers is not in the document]
FrankMittelbach: my position is that the tests for firstaid help to notice that something got corrected and the firstaid is no longer nevessary (or even wrong). And once, the firstaid is no longer required the corresponding test can go again. But yes these are judgement calls

required/latex-lab/latex-lab-bib.dtx
```diff
@@ -110,36 +152,38 @@
...
+%    \end{macrocode} 
+% we store the target name for the label.
+%    \begin{macrocode} 
+   \tl_set:Nn\l_@@_bib_citekey_tl{#2}
+ }  
 %    \end{macrocode}
+% The target is added at the begin of the paragraph. We give that a label as
+% it perhaps need to be removed by packages
+%    \begin{macrocode}  
+\AddToHookWithArguments{cmd/@lbibitem/before}[latex-lab-testphase-bib/target]
```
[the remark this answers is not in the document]
FrankMittelbach: ok

required/latex-lab/latex-lab-block-tagging.dtx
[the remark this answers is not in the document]
FrankMittelbach: yes \bfseries is better (in fact \textbf is probably wrong as the user input may have or wants \par inside the caption)

required/latex-lab/latex-lab-toc-kernel-changes.dtx
FrankMittelbach: as I said: I think we should not  change generic hooks to become hooks with arguments. I think this leads to bad style and difficulties. Why is the generic hook filled? It is only filled by the tagging code so if that goes into a hook by itself then the generic hook would normally not be there and would be free to be used on the outside for somehting else if needed.

josephwright:
@FrankMittelbach  Are we good to go here?

FrankMittelbach:
> @FrankMittelbach Are we good to go here?

I would suggest not to make generic hooks hooks with arguments (I think just one case), otherwise yes.

required/latex-lab/latex-lab-toc-kernel-changes.dtx
u-fischer: I already removed the declaration and the UseHook command hook.  (But I don't quite understand the distinction, we just made generic hooks into hooks with argument).

FrankMittelbach:
> > > @FrankMittelbach Are we good to go here?
> > 
> > 
> > I would suggest not to make generic hooks hooks with arguments (I think just one case), otherwise yes.
> 
> OK, I'll wait for that to be changed

I meant use them in that case, but as I said that was partly nonsense

josephwright:
Merged, letting this build then will start on the release (will be tomorrow now)

Issue #1085, opened 2023-06-13
Support tabs in \verb*


josephwright:
## Brief outline of the enhancement

Currently, LaTeX2e treats tabs as spaces generally, but in `\verb*` they remain as normal spaces not visible ones.

## Minimal example showing the current behaviour

```latex
\RequirePackage{latexbug}
\documentclass{article}
\begin{document}
M\verb*|	|M\verb*| |M * % First is tab, second is space
\end{document}
```

![image](https://github.com/latex3/latex2e/assets/4047173/34d35a63-2c94-4e2f-a259-79d1cb87b0cd)

## Minimal example showing the desired new behaviour

Both should be shown.

car222222:
The behaviour of `\verb*` is weird, but I guess that no one cares about it.  

If this needs to change (at this late stage) then a new, good, representation of "visible tab" will be needed. 

A 'true verbatim' treatment of tabs may be useful for some cases, but has anyone ever asked for it?

josephwright:
> The behaviour of `\verb*` is weird, but I guess that no one cares about it.

Yes, it's odd

> If this needs to change (at this late stage) then a new, good, representation of "visible tab" will be needed.

I'm going to push a suggestion shortly: as tabs are normally treated as spaces, I think they should do the same inside `\verb*` so give the 'visible space' *but with a mechanism to alter this*.

> A 'true verbatim' treatment of tabs may be useful for some cases, but has anyone ever asked for it?

Not in the kernel, but it is available for example in `listings`. Moreover, the idea of a change here is to have something consistent between `\verb` and `+v` from `ltcmd`, and I'd like to start with 'something that I feel I can logically document'.

FrankMittelbach:
> The behaviour of `\verb*` is weird, but I guess that no one cares about it.

I guess yes, nobody cares about it (not if it comes to core LaTeX `\verb*` or `verbatim*`). The situation is different though if you look at `listings` or `fancyvrb`. There people do and there is code to handle TABs properly.

> If this needs to change (at this late stage) then a new, good, representation of "visible tab" will be needed.

if you take the position that core LaTeX just treats TAB = space then we don't need a "visible tab" because
TAB --> you meant space, right? --> make a visible space if you are in `\verb*`
 
> A 'true verbatim' treatment of tabs may be useful for some cases, but has anyone ever asked for it?

as Joseph already said, yes a few packages provide proper  support for this for a long time.

--------

[...]

PR #1087, opened 2023-06-14
Adjustments to \verb and +v (ltcmd)

josephwright:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [x] Test file(s) added


base/doc/ltnews38.tex
```diff
@@ -149,6 +149,43 @@ \section{New or improved commands}
 
 \section{Code improvements}
 
+\subsection{Support for tabs in \cs{verb*}}
+
+\LaTeX{} converts tabs to spaces when typsetting. The same has been true to
```
FrankMittelbach, line 154: converts as TAB to a single space when typesetting.

(as tabs are often represented as several spaces it is not what what we mean if the plural form is used here.)

base/ltmiscen.dtx
```diff
@@ -1188,11 +1188,27 @@
...
+% \changes{v1.2b}{2023/06/14}{Macro added}
 %    \begin{macrocode}
 {\catcode`\ =\active%
 \gdef\@vobeyspaces{\catcode`\ \active\let \@xobeysp}}
+%</2ekernel>
+%<latexrelease>\IncludeInRelease{2023/11/01}%
+%<latexrelease>                 {\@vobeytabs}{Obeyed tabs}%
+%<*2ekernel|latexrelease>
+{\catcode`\^^I=\active
+\gdef\@vobeytabs{\catcode`\^^I\active\let^^I\@xobeytab}}
```
FrankMittelbach, line 1201: are we sure that it is better to have a separate command or to stick the TAB handling into \@vobeyspaces? I'm just wondering, but there are several packages out there that mimic verbatim setup and maybe they would not have to change and nevertheless pick up the  ^Î handling this way. Just as thought ... not thought through

FrankMittelbach (commented):
I'm not sure that it is good to provide \@vobeytabs. It might complicate the situation in the field --- needs some thought

base/doc/ltnews38.tex
josephwright: I was trying to get at that multiple tabs might convert to multiple spaces, but TeX then collapses the result to one space. So perhaps 'converts a single tab to a single space, which is then treated like 'real' spaces in typesetting.'?

base/ltmiscen.dtx
josephwright: I can see this both ways. I agree that if you want to say 'basically tabs are spaces' then yes, it makes sense not to have a separate macro. But the moment you handle tabs there's the possibility of diverging from space handling, and it was that I was thinking of: in particular, people could allow a tab to convert to multiple spaces in verb material.

base/doc/ltnews38.tex
FrankMittelbach: something like this sounds good to me

base/ltmiscen.dtx
josephwright: Ah, you mean include `\@vobeytabs` in the def. for `\@vobeyspaces`?

FrankMittelbach: I didn't mean this originally, but yes, your argument about  being able to adjust it is valid, so I now thought moving it in there gives you the best of both worlds

PhelypeOleinik (approved):

base/ltmiscen.dtx
```diff
@@ -1306,11 +1325,23 @@
...
+%<latexrelease>                 {\verbatim}{Obeyed tabs}%
+%<*2ekernel|latexrelease>
+\def\verbatim{\@verbatim \frenchspacing\@vobeyspaces\@vobeytabs \@xverbatim}
+%</2ekernel|latexrelease>
+%<latexrelease>\EndIncludeInRelease
+%<latexrelease>\IncludeInRelease{0000/00/00}%
+%<latexrelease>                 {\verbatim}{Obeyed tabs}%
+%<latexrelease>\def\verbatim{\@verbatim \frenchspacing\@vobeyspaces \@xverbatim}
+%<latexrelease>\EndIncludeInRelease
+%<*2ekernel>
```
FrankMittelbach, line 1344: if tabs are handled inside \@vobeyspaces then this hsouldn't be needed, should it?

base/ltmiscen.dtx
```diff
@@ -1423,10 +1479,13 @@
...
 %     {Provide visible space in \cs{verb*} also for XeTeX and LuaTeX (github/69)}
+% \changes{v1.2b}{2023/06/14}
+%     {Support visible tabs in \cs{verb*}}
 %    \begin{macrocode}
 \@namedef{verbatim*}{\@verbatim
   \@setupverbvisiblespace
-  \frenchspacing\@vobeyspaces\@sxverbatim}
+  \@setupverbvisibletab
+  \frenchspacing\@vobeyspaces\@vobeytabs\@sxverbatim}
 \expandafter\let\csname endverbatim*\endcsname =\endverbatim
```
FrankMittelbach, line 1489: ditto

base/ltmiscen.dtx
```diff
@@ -1436,20 +1495,48 @@
 %    \begin{macrocode}
 %</2ekernel|latexrelease>
 %<latexrelease>\EndIncludeInRelease
+%<latexrelease>\IncludeInRelease{2018/12/01}%
+%<latexrelease>                 {\verbatim*}{Setup visible space for \verb}%
+%<latexrelease>\@namedef{verbatim*}{\@verbatim
+%<latexrelease>  \@setupverbvisiblespace
+%<latexrelease>  \frenchspacing\@vobeyspaces\@sxverbatim}
+%<latexrelease>\EndIncludeInRelease
```
FrankMittelbach, line 1503: ditto

base/ltmiscen.dtx
josephwright: I'd wondered about that. However, that leaves you vulnerable -  to a re-definition of `\@vobeyspaces`, as that would also kill tab support. That *might* of course be the position we want to take. So I've backed out my change and will look again at this - there's also something not quite right and I want to understand that too.

FrankMittelbach: the way I see it is this: we interpret `\@vobeyspaces` as "handle all space like chars (space TAB, maybe others one day) for verbatim" and  `\@vobeytabs` as an indirection to do something special to tabs alone on top of it. To my knowledge there is no redefinition of `\vobeyspaces` right now and if there is one in the future it has to take care of the assumption.

josephwright: See if you like the new version

base/ltmiscen.dtx
```diff
@@ -1493,15 +1580,32 @@
 %    it the definition for the active space character.
 % \changes{v1.1o}{2018/10/11}
 %     {Provide visible space in \cs{verb*} also for XeTeX and LuaTeX (github/69)}
+% \changes{v1.2b}{2023/06/14}
+%     {Support visible tabs}
 %    \begin{macrocode}
   \ifnum\catcode`\ =\active
   \else  \@setupverbvisiblespace \@vobeyspaces \fi
+  \ifnum\catcode`\^^I=\active
+  \else  \@setupverbvisibletab \@vobeytabs \fi
```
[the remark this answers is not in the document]
josephwright: Indeed: take a look at the modified one

car222222:
There seem to be a few comments etc that are still pending.

car222222 (commented):
A typo and some replies/comments.

car222222:
When the code is finalised I shall review the text in usrguide:
I cannot follow some parts of the current version.

base/doc/usrguide.tex
```diff
@@ -794,6 +794,42 @@ \subsection{Commands at the start of tabular cells}
 \end{tabular}
 \end{verbatim}
 
+\subsection{Using the verbatim argument types}
+
+As described about, the \texttt{v}-type argument may be viewed as similar to
```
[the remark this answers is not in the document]
josephwright: Fixed

josephwright:
> When the code is finalised I shall review the text in usrguide: I cannot follow some parts of the current version.

Removed: Frank suggested activating tabs was done as part of activating spaces, so it's all in one place.

car222222:
What was "Removed:"? 

I assume that it was not all the text in usrguide about verbatim and friends!

josephwright:
> What was "Removed:"?
> 
> I assume that it was not all the text in usrguide about verbatim and friends!

I meant the lines you'd been talking about: Frank had various suggestions which I addressed.

car222222:
Aha! But I am now, at least, referring to the whole of the entry in usrguide.

josephwright:
> Aha! But I am now, at least, referring to the whole of the entry in usrguide.

That's all still there :)

base/doc/usrguide.tex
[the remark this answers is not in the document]
car222222: Are you sure: the diff suggests that it now reads: aboveß  ??

base/doc/usrguide.tex
```diff
@@ -796,7 +796,7 @@ \subsection{Commands at the start of tabular cells}
 
 \subsection{Using the verbatim argument types}
 
-As described about, the \texttt{v}-type argument may be viewed as similar to
+As described aboveß, the \texttt{v}-type argument may be viewed as similar to
```
mrpiggi, line 799: that's a typo as well ;)

car222222:
Is all this code now finalised? So that I can attempt to sort out the documentation.

josephwright:
> Is all this code now finalised? So that I can attempt to sort out the documentation.

Yes no changes planned: awaiting thoughts from review

car222222:
@josephwright Reviews by whom?  Four people have already made comments! Did you request further reviews?

josephwright:
@car222222  At today's team meeting, I checked - there are no expected changes to the code, so please do review the docs.

josephwright:
@car222222 Have you had a chance to look at this?

car222222:
It would probably be useful to provide a command (somewhere) that sets up what is needed in order to typeset the contents of a 
v-type argument in a similar way to the typesetting done by `\verb` and `verbatim`.

car222222:
Thanks for the updates.  It would help me a lot if you also changed the date so that I can easily check that I am downloading the correct version.  I hope that is OK with you.

josephwright:
> Thanks for the updates. It would help me a lot if you also changed the date so that I can easily check that I am downloading the correct version. I hope that is OK with you.

Done. I assume you will re-review the updated text shortly.

car222222:
A short review, or in a short time? :-)

Not that I am guaranteeing either!
But yes, I will continue work on the latest version.

josephwright:
> A short review, or in a short time? :-)

:)

> Not that I am guaranteeing either! But yes, I will continue work on the latest version.

Thanks - provided we can consider this for the Fall'23 release, I'm happy.

car222222:
Sounds good!

car222222:
Are you also considering any of the following good ideas 
for the Autumn of this year?  
Or for any time?
 
1.  Documenting, somewhere, how to use the content of a 
v-type (or just +v) argument in various ways:

—  the set up needed for typesetting;

—  the set up needed for writing it out to a file;

—  maybe for other common uses? 

2.  Adding some commands+code to help with any of the cases in 1. 

3.  Some resolution of the treatment of empty lines in +v args? 

Also, what about maybe adding, somewhere, some similar 
documentation of how \verb and verbatim work?

josephwright:
>     1. Documenting, somewhere, how to use the content of a
>        v-type (or just +v) argument in various ways:
> 
> 
> — the set up needed for typesetting;
> 
> — the set up needed for writing it out to a file;
> 
> — maybe for other common uses?

I take it you want more than I've already added to `usrguide`?

>     2. Adding some commands+code to help with any of the
>        cases in 1.

I assume you mean you'd like to see the `usrguide` stuff expanded with demos?

>     3. Some resolution of the treatment of empty lines
>        in +v args?

I think one for a team meeting!

> Also, what about maybe adding, somewhere, some similar documentation of how \verb and verbatim work?

I'm not sure there is, but then they are much more 'fixed'.

car222222:
> I take it you want more than I've already added to usrguide?

I had not noticed very much!  
I shall check again, and then make some suggestions.

car222222:
> Also, what about maybe adding, somewhere, some 
> similar documentation of how \verb and verbatim work?

> I'm not sure there is, but then they are much more 'fixed'.

We already established that this does not exist.  So I guess I am asking "everyone"! if this should be added somewhere, at some stage.

car222222:
Do not forget to close #1085 when this is finished (if ever).

tobiasBora:
Wait, I’m very confused by:
```
We have now revised this behavior, such that end-of-line characters
are converted to the \cs{par} command when parsed by \texttt{+v}-type argument.
```
In my case, I’m using `+v` to parse non-LaTeX code, like python code for instance… in that case I really don’t want to have new lines replaced with `\par`! Wouldn’t it be better to create a new letter for the argument, and keep `+v` as it is?

josephwright:
@tobiasBora  What's your _exact_ use case? We suspect that the vast majority of `+v`-type grabbing is then being used in typesetting. The parallel plan is to have something filecontents-like, but that's not yet implemented.

tobiasBora:
@josephwright  I am doing a nearly complete rewrite of my newborn library [robust-externalize](https://raw.githubusercontent.com/leo-colisson/robust-externalize/master/doc/robust-externalize.pdf). The goal of this library is to cache any kind of object (pdf generated by tikz, image or even video generated by some python code…). These objects are created by:
1. creating a template (e.g. a latex file with a preambule, and some placeholders to add text, or a python code generating some images)
2. replacing the placeholders with some values (e.g. insert the content of a tikz picture in the template)
3. create the file containing this source file, whose name contains the hash of all the dependencies for maximum purity
4. generate a compilation command like `pdflatex \mysourcefile`. Note that the source file is different everytime
5. run this command from latex

Step 1 needs code that might look like:
```
\placeholderFromCode{__PYTHON_CODE__}{
import PIL
def create_picture():
    im = PIL.Image.new(mode="RGB", size=(__IMAGE_SIZE__, __IMAGE_SIZE__))
    return im

im = create_picture()
im.save(__OUTPUT_FILE__)
}
```

If the verbatim environment turns newlines into `\par`, I will save a python file like:
```
import PIL\pardef create_picture():\par    im = PIL.Image.new(mode="RGB", size=(__IMAGE_SIZE__, __IMAGE_SIZE__))\par    return im\par\parim = create_picture()\parim.save(__OUTPUT_FILE__)\par
```
and it will never compile correctly.

I also sometimes want to print also in the pdf the content of the template I’m using (e.g. for debugging purpose), but this is another question https://tex.stackexchange.com/questions/694716/print-latex3-string-verbatim/694717.

davidcarlisle:
@tobiasBora  your use case is actually simplified by the use of `\par` here. If the string has newlines marked with catcode 12 `^^J` (character 10) it is hard to print anything, you need to iterate through teh string and replace `^^J` by `\par` or `\newline` or some such. Conversely if each end of line is marked by a control token you simply define it to whatever you need, so locally define `\par` as `^^J` while writing a file for example, while it will work naturally for typesetting.

tobiasBora:
Wait, do you mean that a string can still contain some macros like `\par` which are not decomposed into a `\`, a `p`, a `a` and a `r`, and that are evaluated when being written?? I think I will never understand the complicated point of view of LaTeX ^^'

To me, replacing while writing to a file does not look much simpler than replacing when printing… (e.g. the tt family font could insert a newline instead of a weird \Omega by default no?) and the user needs to be aware of all the edge cases (like `n` is treated differently, `\t` is treated differently…)… but you certainly know better than me what you are doing. But at least, please document all the characters that you handle in a special way, and how to replace the characters with their original definition.

josephwright:
@tobiasBora The aim with the new behaviour is to tighten up exactly as you suggest: the _only_ macro that can appear in the grabbed material is `\par`. Other stuff retained is active chars: spaces and bytes >127 for pdfTeX. That is all covered in the updated docs.

FrankMittelbach:
> @tobiasBora The aim with the new behaviour is to tighten up exactly as you suggest: the _only_ macro that can appear in the grabbed material is `\par`. Other stuff retained is active chars: spaces and bytes >127 for pdfTeX. That is all covered in the updated docs.

correct me if I'm wrong @josephwright but I would expect that a `\par` as part of the input is converted to 4 tokens, ie the only `\par` tokens that remain (or rather are added) are those that denote line endings in the original input.

josephwright:
Are we good to go here? This is one that I'd like to see in a pre-release with plenty of time for testing.

car222222:
There is just one thing outstanding on the code:

"3. Some resolution of the treatment of empty lines
   in +v args?"

To which, one @josephwright stated, somewhat ungrammatically:

"I think one for a team meeting"

davidcarlisle:
@josephwright I wonder if `\par` is right and it would be better to use `\obeyedline` like verbatim (from `\obeylines`) then you can define it to be `\mbox{}\par` if you want consecutive newlines to make blank lines or more generaly define some soecial line behaviour without so explicitly locally redefining `\par`

tobiasBora:
@josephwright  @FrankMittelbach Ok now it is clearer, thanks a lot. I’m curious though, how would you make the difference between `\par` (the token) and `\par` (the string), for instance in the string replacement command?

josephwright:
> @josephwright @FrankMittelbach Ok now it is clearer, thanks a lot. I’m curious though, how would you make the difference between `\par` (the token) and `\par` (the string), for instance in the string replacement command?

(Aside: TeX strings are not like strings in other languages: you mean a token list.)

Using `expl3`, is we do something like
```latex
\tl_set:Nn \l_tmpa_tl {#1} % The grabbed tokens
\tl_replace_all:Nnn \l_tmpa_tl { \par } { [PAR] }
```
we would  replace the `\par` _tokens_ as TeX will tokenize `\par` as normal. However, if we switch to
```latex
\tl_set:Nn \l_tmpa_tl {#1} % The grabbed tokens
\tl_replace_all:Nxn \l_tmpa_tl { \token_to_str:N \par } { [PAR] }
```
then the `\par` is converted by `\token_to_str:N` to four 'other' tokens, ``\``, `p`, `a` and `r`, and we would replace that string.

tobiasBora:
Oh, so `\tl_replace` would be usable on strings as well. Thanks!

car222222:
One further typo, in ltnews38:
obeydline  ==>  obeyedline

I shall deal with the (slightly improved, hooray!) documentation later.

Skillmon:
One could further generalise and leave one end-of-line control sequence and one begin-of-line control sequence. Would make processing line wise easier and not hurt if the default would expand to empty. If one has to insert control sequences one could as well make processing really easier and not just hard in another way (and it wouldn't hurt typesetting with an empty control sequence -- though I doubt that direct typesetting is really the most used functionality for `+v`).

FrankMittelbach:
> One could further generalise and leave one end-of-line control sequence and one begin-of-line control sequence. Would make processing line wise easier and not hurt if the default would expand to empty. If one has to insert control sequences one could as well make processing really easier and not just hard in another way (and it wouldn't hurt typesetting with an empty control sequence -- though I doubt that direct typesetting is really the most used functionality for `+v`).

quite an interesting though, worth exploring.

car222222:
+1 (re +v)from me on both:

@FrankMittelbach certainly any such is worth investigating.

@Skillmon on the usage of +v
But with 
Questions/Requests 

Please can you provide some examples of real-world usages of +v, with associated necessary/possible settings for the two commands. 

These examples will be useful both for the consideration stage and as a basis for some badly needed documentation of such uses. 

TIA !

josephwright:
Packages/classes I can find using `+v`

- `cprotectinside`
- `ducksay`
- `l3doc`
- `letgut`
- `nodetree-embed`
- `scontents`

Most of the use cases here seem to be typesetting the content rather than writing to files, etc.

Skillmon:
I use `+v` in some documentations of mine to typeset examples and show the results next to each other. An example for this is the macro `\codeexample` in `pxpic.dtx` (using `\scantokens`, but writing to a temporary file would be essentially the same).

Other than that `ducksay` uses `+v` for its `\AddAnimal` (directly typeset with EOL turned into `\tabularnewline\null` via `\tl_replace_all:Nnn`) and for one of its argument-reading modes (which uses `\scantokens` as well, and replaces two consecutive EOLs to `\par` and a single EOL to a single space).

daleif:
> I use `+v` in some documentations of mine to typeset examples and show the results next to each other. An example for this is the macro `\codeexample` in `pxpic.dtx` (using `\scantokens`, but writing to a temporary file would be essentially the same).
> 
> Other than that `ducksay` uses `+v` for its `\AddAnimal` (directly typeset with EOL turned into `\tabularnewline\null` via `\tl_replace_all:Nnn`) and for one of its argument-reading modes (which uses `\scantokens` as well, and replaces two consecutive EOLs to `\par` and a single EOL to a single space).

Nice example (the `\codeexample`). I'm wondering, can this sort of thing also be done via envs? I have a latex book with many examples and there I use the method of writing it to disk and doing a dual input (input + lstlisting) and I've always wondered if it can be done via code alone (aka no writing to disk)

josephwright:
@car222222 I've pulled out the parts of the PR that only impacted on `\verb`, and merged those - so we now cover tabs. That will be in the pre-release, but the adjustments to `v`-type (both tab support and `^^M` tokens) are left here to be resolved.

josephwright:
Marking beginning and end of each line is easy enough when grabbing, but does it really offer a significant step up over simply marking boundaries? I'm not sure I see a use case where there is a non-trivial impact.

Skillmon:
If you want to process the contents line wise you can simply do `\def\myBOL#1\myEOL{<stuff>}` and are done if you place a BOL and EOL marker. If you only place an EOL marker you're back to checking whether your loop is done after each line, even though the reading process already knew that.

car222222:
@Skillmon Understood, thanks!

PR #1095, opened 2023-06-16
Add theHcounter represention 

u-fischer:
hyperref and also the tagging code need for links and structure references an additional counter represention `\theH<counter>`.  This moves the relevant hyperref and latex-lab patches  of `\@definecounter`, `\@addtoreset` and `\refstepcounter` into the kernel.

From the tagging project if will resolve the issues
https://github.com/latex3/tagging-project/issues/656
https://github.com/latex3/tagging-project/issues/414
as the code no longer will use cmd hooks. 

 
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


base/ltcounts.dtx
```diff
@@ -305,22 +307,56 @@
...
 % \begin{macro}{\@addtoreset}
+%  \changes{v1.1n}{2023/06/15}{add the parent theHfoo if a counter is reset}
 %    \begin{macrocode}
-\def\@addtoreset#1#2{\expandafter\@cons\csname cl@#2\endcsname {{#1}}}
+%</2ekernel>
+%<latexrelease>\IncludeInRelease{2023/11/01}{\@addtoreset}
+%<latexrelease>                             {provide theHfoo commands}%
+%<*2ekernel|latexrelease>
+\def\@addtoreset#1#2{\expandafter\@cons\csname cl@#2\endcsname {{#1}}%
+  \expandafter\gdef\csname theH#1\endcsname{\csname theH#2\endcsname.\the\value{#1}}%
```
car222222, line 351: Does this work as intended in the following cases? 
--  Counter A is reset by Counter B and 
    Counter A is also reset by Counter C ;
--  Counter A is reset by Counter B and 
    Counter B is reset by Counter D .

car222222 (changes requested):
See two pending comments: 
one being a possible problem with the code.

car222222:
One more comment, but just a typo!

car222222:
When the code is finalised I shall review the text in usrguide:
I cannot follow some parts of the current version.

base/ltcounts.dtx
u-fischer: The exact definition of `\theHA` depends naturally on the order of the \@addtoreset calls. It "works" if one get unique `\theHA` value, which isn't 100% garantied -- users can always find a way to build something odd, -- but it is the definition used by hyperref since many years, so it is compatible with the current behaviour.

car222222:
Looking further into this suggests that it is not necessary for any of this to be in the kernel, at least not in this very specialised form.  Is this move considered essential?  If so, why?

car222222:
It seems likely that it would be safer to only add a hook here to the kernel version.  This enables packages, including the various options in hyperref, to independently add whatever code they need. And it avoids building in particular, and not totally robust, code that is specific to some uses (options) of hyperref.

u-fischer:
The code is not specific to hyperref. As said above the counter representation is also used for structure references and for this the H-representation is needed also for counters created before hyperref is loaded and also if hyperref is not loaded at all. The latex-lab code contained therefore a copy of the hyperref code, and this PR mainly moves this code from latex-lab into the kernel. 

Generally: one goal of the tagging project is to move support for hyperlinks into the kernel. To quote one of the subtasks:

> Add the hyperref document interfaces (as appropriate) to the LaTeX kernel and then retire the package itself.

And this requires that hyperref stops to patch kernel commands, both directly and through hooks.

car222222:
OK, so the ability to adapt is needed more generally, including maybe for applications not yet imagined.

In that case there are even stronger reasons to only add hooks or generic commands to the kernel, rather than very specific, and in some cases suboptimal, chunks of code. Or not?

u-fischer:
> In that case there are even stronger reasons to only add hooks or generic commands to the kernel, rather than very specific, and in some cases suboptimal, chunks of code. Or not?

Sorry I'm loosing you. `\@definecounter{blub}` defines currently a count `\c@blub`, and the commands `\theblub`, `\cl@blub`, and `\p@blub`, and with this PR it will also define `\theHblub`.  Why is the last a specific, suboptimal chunk of code? And do you really suggest that I add to the kernel `\AddToHooksWithArguments{cmd/@definecounter/after}{\expandafter\gdef\csname theH#1\endcsname{\the\value{#1}}` only to get the H-representation defined all the time? Hooks are fine, but they have an overload, why use them if I can do the definition directly?? And do not forget that we need the representation also for counters like enumi and equation which are defined by the kernel.

> This is one of many ideas in hyperref that is strangely implemented

What is strange? It is simply another command associated with a counter and not stranger than `\p@blub` or `\theblub`.

base/changes.txt
```diff
@@ -6,14 +6,525 @@ to completeness or accuracy and it contains some references to files that are
...
+	* ltluatex.dtx modify new_luafunction to ensure
+	\e@alloc@luafunction@count always matches the size of the Lua
+	table of function allocations (gh100)
+
+2023-06-12  Joseph Wright  <Joseph.Wright@latex-project.org>
+
+	* usrguide.tex
+	Correct description of font encodings supported for case changing
+
+>>>>>>> develop
```
mrpiggi, line 527: probably remnant from merge conflict

base/ltcounts.dtx
```diff
@@ -328,12 +335,50 @@
...
+%<*2ekernel>
 %    \end{macrocode}
 %  \end{macro}
 %
 % \begin{macro}{\@addtoreset}
+%  \changes{v1.1o}{2024/09/20}{add the parent theHfoo if a counter is reset}
+%  If a counter is reset when a parent counter changes it no longer has an unique value
+%  across the document. As |\theH<counter>| should be unique 
+%  this representation is changed to include also the 
+%  representation of the parent.
```
FrankMittelbach, line 366: Maybe add something like:

This is not 100\% guaranteed to work but has been used this way by hyperref for many years without causing problems.

base/lttagging.dtx
```diff
@@ -254,6 +253,34 @@
 % TODO: move this into lttagging.
```
FrankMittelbach, line 253: when is that move happening?

base/ltxref.dtx
```diff
@@ -413,18 +416,55 @@
...
+% the plug hyperref can add a conditional and e.g.
+% suppress the processing in a PDF context. 
+%    \begin{macrocode}
+\NewSocket{refstepcounter}{1}
+%    \end{macrocode}
+% \end{socketdecl}
+% \begin{socketdecl}{refstepcounter/target}
+% This socket takes an argument, the counter name, and
+% should at least set from it the target name |\@currentHref|.
+% With hyperref it will also sets the actual target.
```
FrankMittelbach, line 447: will also set   or   also sets

base/ltxref.dtx
```diff
@@ -413,18 +416,55 @@
...
+%    \begin{macrocode}
+\NewSocket{refstepcounter}{1}
+%    \end{macrocode}
+% \end{socketdecl}
+% \begin{socketdecl}{refstepcounter/target}
+% This socket takes an argument, the counter name, and
+% should at least set from it the target name |\@currentHref|.
+% With hyperref it will also sets the actual target. 
+% This is done with a socket so that the target name 
+% is not set more than once to perhaps different names.
```
FrankMittelbach, line 449: to (possibly) different

FrankMittelbach (approved):
ok I guess, with a few comments or questions

base/doc/ltnews40.tex
```diff
@@ -295,6 +297,16 @@ \subsection{Improved error raised by empty hook}
 %
 \githubissue{1423}
 
+\subsection{Provide counter representations for link targets}
+
+To create unique link targets the package 
+\pkg{hyperref} needs special counter representations
+\verb+\theH+\meta{counter}. For this it patches the 
+commands \verb+\@definecounter+, \verb+\@addtoreset+
```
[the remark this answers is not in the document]
u-fischer: I reworded that. Better so? And just to be sure that the implications are known: the change (that \theHcounter gets defined) means that document or classes doing `\newcounter{blub}\newcommand\theHblub{...}` will error. This won't affect classes like memoir (it changed its code last year at my request) but ...

base/doc/ltnews40.tex
```diff
@@ -300,6 +302,25 @@ \subsection{Improved error raised by empty hook}
 %
 \githubissue{1423}
 
+\subsection{Provide counter representations for link targets}
+
+To create unique target names for links the package 
+\pkg{hyperref} uses a special counter representation
+\verb+\theH+\meta{counter}. To ensure that this 
+counter representation exist it redefined the
```
FrankMittelbach, line 310: ... representation exists, hyperref redefined ...

base/doc/ltnews40.tex
```diff
@@ -300,6 +302,25 @@ \subsection{Improved error raised by empty hook}
...
 
+\subsection{Provide counter representations for link targets}
+
+To create unique target names for links the package 
+\pkg{hyperref} uses a special counter representation
+\verb+\theH+\meta{counter}. To ensure that this 
+counter representation exist it redefined the 
+commands \verb+\@definecounter+, \verb+\@addtoreset+ 
+and \verb+\refstepcounter+. This counter representation is also
+needed for the Tagged PDF project and so the augmented commands
```
FrankMittelbach, line 313: and so these augmented command definitions

base/doc/ltnews40.tex
```diff
@@ -300,6 +302,25 @@ \subsection{Improved error raised by empty hook}
...
+
+To create unique target names for links the package 
+\pkg{hyperref} uses a special counter representation
+\verb+\theH+\meta{counter}. To ensure that this 
+counter representation exist it redefined the 
+commands \verb+\@definecounter+, \verb+\@addtoreset+ 
+and \verb+\refstepcounter+. This counter representation is also
+needed for the Tagged PDF project and so the augmented commands 
+have now been incorporated
+into the kernel. So from now on every \verb+\newcounter{+\meta{counter}\verb+}+ will not
```
FrankMittelbach, line 315: Thus from now on ...

base/doc/ltnews40.tex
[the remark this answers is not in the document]
FrankMittelbach: but they would have errored before the moment hyperref got loaded or not?

u-fischer:
> but they would have errored before the moment hyperref got loaded or not?

only if hyperref is loaded first. `hyperref, \newcounter, \newcommand\theH` naturally errored  always, but `\newcounter, \newcommand\theH, hyperref ` was fine.

base/ltcounts.dtx
```diff
@@ -306,19 +308,24 @@
...
      \@addtoreset{#1}{@ckpt}%
      \global\expandafter\let\csname p@#1\endcsname\@empty
+     \expandafter\gdef\csname theH#1\endcsname{\the\value{#1}}%
 %    \end{macrocode}
 %    If \cs{the\#1} is undefined or \cs{relax} we define it with the
 %    standard definition for counters, otherwise we warn. This will
 %    catch, for example, that somebody defines a counter named
 %    ``index'' conflicting with the \env{theindex} environment.
-%    \begin{macrocode}
+%    \begin{macrocode}
```
josephwright, line 328: Stray whitespace?

davidcarlisle (approved):

josephwright (approved):
All good but perhaps zap those stray spaces?

base/ltcounts.dtx
josephwright: (We can fix post-merge)

Issue #1100, opened 2023-06-25
luadef/luafunction allocations in ltluatex


davidcarlisle:
## Brief outline of the bug

ltluatex allocates `\luadef` tokens in the classic style using  a tex count `\e@alloc@luafunction@count` to count the allocations. the luatex manual and generic code shows an allocation using the first free slot in the Lua table,

```
  local lft = lua.get_functions_table()
  lft[\string#lft+1] = function()
 ...
```

The end result is that (for example) pgf and luapstricks over-write the same slots

see

https://tex.stackexchange.com/a/689524/1090

I don't think it is feasible to force generic code to use `\e@alloc` so I think the allocators should ensure the count being allocated is bigger than the table, something like

```
local function new_luafunction(name)
-- new
  if tex_count[luafunction_count_name] < table.getn(lua.get_functions_table()) then
  tex_setcount("global", luafunction_count_name,
                         table.getn(lua.get_functions_table())
-- end addition			 
  tex_setcount("global", luafunction_count_name,
                         tex_count[luafunction_count_name] + 1)
  if tex_count[luafunction_count_name] > 65534 then
    luatexbase_error("No room for a new luafunction register")
  end
  luatexbase_log("Lua function " .. (name or "") .. " = " ..
                 tex_count[luafunction_count_name])
  return tex_count[luafunction_count_name]
end
```

it would probably have been better not to introduce `\e@alloc@luafunction@count` and always use the table size but it may be difficult to remove, so keeping them in sync seems the best compromise

muzimuzhi:
Is this issue already "fixed-in-dev"?

PR #1104, opened 2023-07-02
Guard against overwiting function table for #1100

davidcarlisle:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [X] Test file(s) added


base/testfiles-TU/github-1100.tlg
```diff
@@ -0,0 +1,2 @@
+This is a generated file for the LaTeX2e validation system.
+Don't change this file in any respect.
```
FrankMittelbach, line 2: is the test not showing what it is supposed to show, ie some typeout data?

FrankMittelbach: forget it that is the pdftex one ... got it

davidcarlisle: actually xetex, I put this in the -TU config to cut down on non-luatex runs, I was surprised we don't have a luatex-only config but so far we have avoided that and just had empty tests for non lua engines eg testfiles/tlb4445.lvt... Could add a `config-lua` and remove the `\ifx\directlua\undefined` test as we have a few lua-only tests and likely to get more I suspect

```
$ grep -l ifx.directlua testfiles*/*lvt
testfiles-TU/github-0282.lvt
testfiles-TU/github-1100.lvt
testfiles-disabled/tlb-early-expl3.lvt
testfiles/github-0591.lvt
testfiles/tlb-callbacks-001.lvt
testfiles/tlb-callbacks-002.lvt
testfiles/tlb-extraprimitives-001.lvt
testfiles/tlb-latexrelease-rollback-003-often.lvt
testfiles/tlb-ltluatex-001.lvt
testfiles/tlb-luacmd-001.lvt
testfiles/tlb-mathcallbacks-001.lvt
testfiles/tlb-mathcallbacks-002.lvt
testfiles/tlb4445.lvt
```

josephwright:
On 03/07/2023 06:09, David Carlisle wrote:
> @davidcarlisle commented on this pull request.
> 
> 
> 
>> +This is a generated file for the LaTeX2e validation system.
> +Don't change this file in any respect.
> 
> actually xetex, I put this in the -TU config to cut down on non-luatex runs, I was surprised we don't have a luatex-only config but so far we have avoided that and just had empty tests for non lua engines eg testfiles/tlb4445.lvt... Could add a `config-lua` and remove the `\ifx\directlua\undefined` test as we have a few lua-only tests and likely to get more I suspect

I do wonder if we should review generally what tests run with what engines.

We have a *lot* of tests that really should have no engine dependence at 
all (true of 2e and l3). Now, I see the 'things can crop up' argument, 
but I also see that with the engine situation more-or-less stable, and 
the fact we are not supposed to be testing engines ourselves, one could 
cut down quite a bit.

I have wondered if we could take most of the tests which don't need any 
engine-specific .tlg files and simply test them once (pdfTeX?). We could 
do much the same for Lua-specific tests, and then would be able to 
target engine tests where they are needed.

One for the team meeting I suspect.

Joseph

base/ltluatex.dtx
```diff
@@ -1284,14 +1284,18 @@ luatexbase.new_chunkname = new_chunkname
...
 %    \begin{macrocode}
 local luafunction_count_name =
                          luafunction_count_name or "e@alloc@luafunction@count"
 local function new_luafunction(name)
   tex_setcount("global", luafunction_count_name,
-                         tex_count[luafunction_count_name] + 1)
+                         math.max(
+                           #(lua.get_functions_table()),
+                           tex_count[luafunction_count_name])
+                          + 1)
```
zauguin, line 1298: We should additionally do something like

```
lua.get_functions_table()[tex_count[luafunction_count_name]] = false
```

to ensure that code relying on the table length doesn't consider a slot to be free after it has been returned by `new_luafunction` in case it's not immediately used.

davidcarlisle: @zauguin sounds good (also highlights we shoudl probably never had had a count on the tex side at all, but since we do, ensuring it really is the same as the Lua table size sounds like a good plan), I'll add to this branch

davidcarlisle: resolved by fbac7f05

josephwright (approved):

josephwright:
Looks good but `changes.txt` entry?

davidcarlisle:
> Looks good but `changes.txt` entry?

oh I wrote one, but looks like it was never checked in...

davidcarlisle:
> Looks good but `changes.txt` entry?

https://github.com/latex3/latex2e/pull/1104/commits/acda3951ffd9aac5dd7d4b7262ff375a35b54bc7

PR #1108, opened 2023-07-18
Cleanup firstaid

u-fischer:
This removes three unneeded firstaids (everysel, bidi, memoir) from the firstaid file.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added (not needed)
- [ ] Relevant `\changes` entries in source included (imho unneeded in firstaid)
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated (imho unneeded)

### DIFF
--- a/base/testfiles-lthooks/ltcmdhooks-001.tlg
+++ b/base/testfiles-lthooks/ltcmdhooks-001.tlg
@@ -178,18 +178,6 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid = 0 -> 
 Handled code for firstaid
-Update code for hook 'file/memoir.cls/before' on input line ...:
-Code labels for sorting:
- firstaid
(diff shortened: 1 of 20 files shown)


FrankMittelbach (approved):
looks fine to me. The only thing I wonder is whether we want to keep the test (such as the one for bidi) or if we want to drop them too. Not sure.

u-fischer:
I thought we needed tests that the removal doesn't break something and the original firstaid tests would cover that.

FrankMittelbach:
> I thought we needed tests that the removal doesn't break something and the original firstaid tests would cover that.

agreed, but those where I had commented out the code last time did already served their purpose :-) but I agree it might be better to keep them around ... I was just wondering if we want to keep them forever.

PR #1109, opened 2023-07-21
Lab review 02

u-fischer:
This add a number of small fixes to latex-lab (\part, verse, floats), it imports the newest tagpdf, cleans up the naming of the various module and documents them properly.

# Internal housekeeping

## Status of pull request
- Ready to merge

- [x] Test file(s) added
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated (not relevant)


required/latex-lab/latex-lab-math.dtx
```diff
@@ -198,15 +198,15 @@
 %
 %    \car{Change description here?}
 %    \begin{macrocode}
-\ProvidesFile{latex-lab-mathtagging.ltx}
-             [\ltmathtaggingdate\space
-              \ltmathtaggingversion\space
+\ProvidesFile{latex-lab-math.ltx}
+             [\ltmathdate\space
+              \ltmathversion\space
```
FrankMittelbach, line 203: in some sense this shows that we went the wrong way around ie going from XXX-tagging to just XXX. We should have probably done it the other way around, but I don't think we want to change that now. However, I think these 2 commands should be changed back to ltmathtagging... as ltXXXdate is plain wrong and might conflict with ltXXXdate used elsewhere (e.g. there is an ltmath.dtx file even though it currently doesn't  use the commands.

required/latex-lab/latex-lab-sec.dtx
```diff
@@ -437,45 +437,48 @@
...
-              \int_compare_p:nNn {\c@secnumdepth}<{-1} 
-            } 
-          }  
-          {
-            \MakeLinkTarget[part]{}
-          }
+           \bool_lazy_any:nT  
+            {
+              { #1 } 
+              {
```
FrankMittelbach, line 462: the file seems to have tabs not spaces here and elsewhere, maybe better to change that (I can do this easily in emacs)

required/latex-lab/latex-lab-sec.dtx
```diff
@@ -31,7 +31,7 @@
 % \title{The \textsf{latex-lab-sec} package\\
 % Changes related to the tagging of sectioning commands}
 % \author{\LaTeX{} Project\thanks{Initial implementation done by Ulrike Fischer}}
-% \date{v0.83 2023-06-07}
+% \date{v0.84 2023-06-21}
 %
```
FrankMittelbach, line 35: is this file missing a \Provides... that the version is in the doc only?

required/latex-lab/testfiles-sec/test-stop-sect-pdf.pvt
FrankMittelbach, line 1: for consistency that should be "sec" :-)   Actually I guess I would have always used sect but that is a different matter and again not worth altering

required/latex-lab/testfiles-sec/test-suppress-sect.pvt
FrankMittelbach, line 1: again sect -> sec

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1173,8 +1171,32 @@
 % \end{environment}
 %
 %
+% \subsubsection{verse environment}
 %
-%
+%  \begin{environment}{verse}
+%  The verse environment has not special tagging currently. It is
+%  defined as a simple standard list and takes the tagging from there.
+%  But it must be redefined so that \cs{itemindent} is correctly set.
```
FrankMittelbach, line 1179: maybe we should explain why this has to be defined inside that hook (because it is class defined)

FrankMittelbach (approved):
looks ok to me with some minor comments in a few places

required/latex-lab/latex-lab-math.dtx
u-fischer: What about say `\ltlabmathdate`  everywhere? That would never clash. I'm not so keen about "tagging" as while the tagging project triggers the changes, the code will have an impact on untagged documents too.

required/latex-lab/latex-lab-sec.dtx
u-fischer: spaces and tabs are the bane of my life ;-). I deactivated the "space cleaning" facility of my editor as people complained that changes created too large diffs ... but perhaps I should simply reactivate it. (I have no idea where the tabs are from as I never use tab.)

required/latex-lab/latex-lab-sec.dtx
u-fischer: I will check

required/latex-lab/latex-lab-math.dtx
FrankMittelbach: fine

PR #1110, opened 2023-07-26
latex-lab sockets

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development

- [x] Test file(s) added

At this point in time this is just a request for feedback in particular on `latex-lab-socket` as the base functionality and its use in `latex-lab-new-or-2` and `latex-lab-footnotes`.  In particular socket und plug names are likely to need changes ... I'm really just experimenting with that to see how it feels.


required/latex-lab/socket-sample.tex
```diff
@@ -0,0 +1,411 @@
...
+
+By the way:
+\begin{itemize}
+\item
+  should it be \texttt{gassign} ? or some other verb?
+\item
+  And what about \cs{socket\_use:n} ? Should that be \texttt{:nw}
+  as wedon't really control the arguments or should we have
+  \cs{socket\_use:nn}, \cs{socket\_use:nnn}, \ldots, just to make it
+  clearer how many arguments are (supposed to be) picked up?
```
PhelypeOleinik, line 292: In lthooks we used `\hook_use:nw` for the version with arguments, so I think `\socket_use:nw` for consistency (also because the code is the same, regardless of the actual number of arguments).

required/latex-lab/socket-sample.tex
```diff
@@ -0,0 +1,411 @@
...
+  \NewSocket{foo}{2}    % Error (already declared)
+  \NewSocket{foo}{1}    % Error (already declared)
+\end{verbatim}
+\NewSocket{foo}{2}
+\NewSocket{foo}{1}
+
+You also get an error if you attempt to declare some socket code and the
+socket name is not yet defined, e.g.,
+\begin{verbatim}
+  \DeclareSocketCode{baz}{undeclared}{} % Error (Socket not declared)
```
PhelypeOleinik, line 325: Should this error, or just store the code?  For example if, say, `multicols` defines a socket `outputcolumn`, and another package `fancycols`, which is loaded before, adds some code to that socket?  Of course this could be solved by doing `\DeclareSocketCode` in the hook `package/multicols/after`, but it's kinda workaroundish, I think.

PhelypeOleinik (commented):
Just a few comments. I Haven't looked at the code yet

required/latex-lab/socket-sample.tex
car222222: This suggestion @PhelypeOleinik to use `:nw` sounds sensible.

FrankMittelbach:
sorry my bad, the file scockets-sample.tex is obsolete, I just forgot to take it out. All code and its documentation is in latex-lab-socket.dtx

required/latex-lab/socket-sample.tex
FrankMittelbach: This is what I have done in the real code, that file should have been deleted it was an early draft.

car222222: OK, good.  Sorry about the noise.

FrankMittelbach: and it is already the case in the .dtx

required/latex-lab/socket-sample.tex
FrankMittelbach: Good point, probably it should show the same behavior even though the number of use cases will be far less compared to hooks, but I guess they aren't zero. So perhaps I should change that.

FrankMittelbach: Thinking about this a bit more I'm not so sure that it is a good idea after all. You would need to do the same then to `\AssignSocketPlug` and then the  question is what happens you have  two such assignments before the socket is declared? Who is then going to win? The one assigned to the socket when it is declared "noop" ie the last assignment that was made? That would be logical but not helpful. And how would you adjust the situation?

If you go via `package/multicol/after` then there is a clear intend  that you overwrite and if two packages do this you could order the decision with a rule if necessary. So for now I'm inclined to leave it as it is and see what use case appear in reality.

FrankMittelbach:
By now the request for checking and feedback no longer involves the socket code but only its *usage* in the other two files (as the sockets code moves to the kernel and has its own PR). I have now updated the pr to already including the pending pull for ltsockets.dtx, so the tests now run with the kernel version of sockets.

car222222:
!!:-) Loved the "pins" idea!

car222222:
No way can things like this be described as “identity” :
#1#2#3#4#5#6#7#8#9 

Rather, this is called “concatenation” ;  
and it can be very dangerous for some 
combinations of “inputs/pins”!! --
Beware of electric shocks, fires, etc!

josephwright (approved):

u-fischer (commented):
the rcb-code has been merged with the tagging-9 code, so this PR can be closed.

PR #1121, opened 2023-08-05
Sockets in 2e

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added

Basically done, smaller updates and improvements could be added at a later stage.


u-fischer:
* The output of `\ShowSocket` looks imho a bit odd:

~~~~
Socket foo0:
    number of arguments = 0
    available plugs = noop
    current plug = noop
    definition = \protected\long macro:->

! .
\ShowSocket #1->\socket_log:n {#1}\errmessage {}
~~~~

Where does the `!.` and the next line come from?

* Why is the default plug for a socket with one argument `identity`, but for a socket with two or more arguments `no-op`? (The behaviour imho also contradicts the documentation of `\socket_new:nn`.

FrankMittelbach:
> 
> Where does the `!.` and the next line come from?
> 
> 
>     * Why is the default plug for a socket with one argument `identity`, but for a socket with two or more arguments `no-op`? (The behaviour imho also contradicts the documentation of `\socket_new:nn`.

I can't see that it contradicts, but perhaps the word argument is misleading and I should better use "inputs". The second argument of `\socket_new:nn` defines how many "inputs" this socket has (the first its name).  So a socket with one argument (input) gets identity and you call it as `\socket_use:nn {name}{ input }`.
For two or more inputs identify makes little sense and as there has to be a default noop is better (and for those without inputs "noop" make sense too.

As to the `!.`: it says in the documentation the debugging code this is a rough draft and it uses simply `\errmessage` and I guess that 's where it comes come. Certainly need a better design and a better implementation but I still have hope somebody else is going to do that between now and then.

u-fischer:
> So a socket with one argument (input) gets identity and you call it as `\socket_use:nn {name}{ input }`.

But `\UseSocket` is `\socket_use:nw` and not `\socket_use:nn`. 

> For two or more inputs identify makes little sense

why? Isn't `#1#2->#1#2` an identity too? 
I at least I would have expected that in the following example either both `\UseSocket` swallow arguments, or that both leave them alone, so either no output, or `aaabbbccc`:

~~~~
\documentclass{article}
\NewSocket{foo1}{1}
\NewSocket{foo2}{2}
\begin{document}
\UseSocket{foo1}{aaa}\UseSocket{foo2}{bbb}{ccc}
\end{document}
~~~~

FrankMittelbach:
> > So a socket with one argument (input) gets identity and you call it as `\socket_use:nn {name}{ input }`.
> 
> But `\UseSocket` is `\socket_use:nw` and not `\socket_use:nn`.

that is because with 2e names you just don't know how many arguments follow (not from the command name syntax).
in fact \socket_use:nnn etc is also just \socket_use:nw unless you have checking enabled (but that is only a "future feature and not implemented)

> 
> > For two or more inputs identify makes little sense
> 
> why? Isn't `#1#2->#1#2` an identity too? I at least I would have expected that in the following example either both `\UseSocket` swallow arguments, or that both leave them alone, so either no output, or `aaabbbccc`:

Then you have to read documentation :-) yes `#1#2`can be thought of as identity, but in practical terms you would nearly ever have any use for it, more likely with noop ... but in any case I would expect we will end up seeing only plugs with 0 or 1 input.

FrankMittelbach:
one more comment: identity is useful as a default for 1 input (it is its typical use case) but as I claim nearly never for sockets with more inputs. But defining it (unnecessarily) would be a fair amount of work, because it differs for each and every socket whereas noop is trivial to define (and a sensible default in some cases).

car222222:
Moved to the right place!

u-fischer (approved):
The last lines of the dtx seems to contain forgetten rollback settings and should probably be removed. Apart from this it looks imho fine.

josephwright (approved):
Looks good to me

PR #1129, opened 2023-08-25
Gh1128

u-fischer:
This puts longtable in sync with tabular/array but it can change some output, so the question is if a rollback version is needed.
 
# Internal housekeeping

## Status of pull request

X Feedback wanted 
- Under development
- Ready to merge

- [X] Test file(s) added


required/tools/testfiles/github-1128.lvt
```diff
@@ -0,0 +1,16 @@
...
+\hline 
+some text & \medskip xxx
+\end{tabular}
+
+\begin{longtable}{lp{1cm}}
+\hline 
+some text & \medskip xxx
+\end{longtable}
+
+\end{document}
```
davidcarlisle, line 16: missing end of line here

davidcarlisle (changes requested):
looks good apart from one e0l missing from end of a test file.

In https://www.latex-project.org/cgi-bin/ltxbugs2html?pr=tools/3785

I queried whether we could do this after 15 years and now it's more like like 25 but apart from being more important to fix these things, the user now has the option of using `[=v4.13]` to get legacy behaviour, so I think we can be braver.

base/doc/ltnews38.tex
```diff
@@ -299,8 +299,10 @@ \section{Changes to packages in the \pkg{amsmath} category}
 \section{Changes to packages in the \pkg{graphics} category}
 
 \section{Changes to packages in the \pkg{tools} category}
+\subsection{\pkg{longtable}: correct p-column definition}
 
-
+In general the \pkg{longtable} implementation follows 
+the \pkg{array} usage but the package didn't take over a change made 1992 in \pkg{array} which adjusted the handling of the strut inserted at the begin of p-columns. As a consequence there are a number of inconsistencies in the output of p-columns between \pkg{tabular} and \pkg{longtable}. This has been corrected, \pkg{longtable} uses now for the strut the same definition as \pkg{array}. \githubissue{1128}
```
FrankMittelbach, line 305: may add a few newlines :-) but more importantly, maybe add a note that the old behavior can be obtained using an explicit rollback version as @davidcarlisle mentioned

FrankMittelbach (approved):
other than the possible addition to ltnews I think this can go

required/tools/longtable.dtx
```diff
@@ -2206,12 +2206,16 @@
...
 %    \begin{macrocode}
     \color@begingroup
     \let\@footnotetext\LT@p@ftntext
     \setlength\hsize{#1}%
     \@arrayparboxrestore
-    \vrule \@height \ht\@arstrutbox \@width \z@}
+    \everypar{%
+      \vrule \@height \ht\@arstrutbox \@width \z@
+      \everypar{}}%    
+    }
```
muzimuzhi, line 2218: Loosely related: If using paragraph hook `\AddToHook{para/begin}{...}`  instead, since `\AddToHook` acts globally, a `\RemoveFromHook` is needed in `\LT@endpbox`. Is it?

FrankMittelbach: > Loosely related: If using paragraph hook `\AddToHook{para/begin}{...}` instead, since `\AddToHook` acts globally, a `\RemoveFromHook` is needed in `\LT@endpbox`. Is it?

without having looked at the context, I think you would want to use \AddToHookNext instead (because this is what it does, but perhaps that wouldn't quite work, not sure)

PR #1131, opened 2023-09-01
tagging-project issue #9

FrankMittelbach:
# Internal housekeeping

- Ready to merge

- [x] Test file(s) added


u-fischer:
It looks fine. I found two problem. 

* At first it errors if tagging is not active as part of the code is hidden behind \tag_if_active:

~~~~
\DocumentMetadata{testphase={block}}
\documentclass{article}
\newtheorem{theorem}{Theorem} 
\begin{document}%
\begin{theorem}
\begin{enumerate}
\item blub
\end{enumerate}
\end{theorem}
\end{document}
~~~~

~~~~
! Undefined control sequence.
\@begintheorem ...ara_structure_unconditionally:n 
                                                  {\PARALABEL }\itshape 
~~~~

* The second problem is with hyperref, or more precisely with the development version. hyperref tries to avoid problems with anchors in theorems (spacing problems and lost anchors) by delaying the setting of the anchor to the next `\item` and as the new implementation doesn't use `\item` this doesn't work ;-(. 

I would suggest following changes here. 
The most logical place to add an anchor would be imho in `\@begintheorem` after the switch to H-mode. 

~~~~
\def\@begintheorem#1#2{
  \UseInstance{blockenv}{theorem}{}
  \tagpdfparaOff
  \mode_leave_vertical:
  \MakeLinkTarget{} %<--- 
  \tag_struct_begin:n{tag=Caption}
 .....

\def\@opargbegintheorem#1#2#3{
  \UseInstance{blockenv}{theorem}{}
  \tagpdfparaOff
  \mode_leave_vertical:
  \MakeLinkTarget{} %<-
  \tag_struct_begin:n{tag=Caption}
 ...
~~~~

Sadly `\@begintheorem`  doesn't get the counter name as argument, so one can't simply use ` \MakeLinkTarget{#1}` above, but one has to change \@thm anyway to replace the `\refstepcounter`  and can set the target name there:

~~~~
\def\@thm#1#2{%
  \@kernel@refstepcounter{#1}% don't set anchor
  \NextLinkTarget{#1.\csname theH#1\endcsname}% set target name (test if \theHXX exist?)
  \@ifnextchar[{\@ythm{#1}{#2}}{\@xthm{#1}{#2}}}
~~~~

At last one has to prevent hyperref to overwrite everything again (alternatively hyperref could test if the block code is loaded). 

~~~~
\def\hyper@nopatch@thm{}
~~~~

FrankMittelbach:
Feel free to update/ improve. I won't be able to do anything before the end of my vacation.

u-fischer (approved):

josephwright (approved):

PR #1145, opened 2023-10-10
add properties and MakeLinkTarget to clsguide

u-fischer:

# Internal housekeeping

## Status of pull request
- Feedback wanted: Someone should find the typos etc ...

- [ ] Test file(s) added


base/doc/clsguide.tex
```diff
@@ -1181,6 +1181,113 @@ \subsection{Querying localisation}
...
+\begin{description}
+ \item[\texttt{abspage}] The absolute value of the current page: starts at $1$ and increases
+   monotonically at each shipout.
+ \item[page] The current page as given by \cs{thepage}: this may or may not
+   be a numerical value, depending on the current style. Contrast with
+   \texttt{abspage}. You get this value also  with the standard \cs{label}/\cs{pageref}.
+ \item[\texttt{pagenum}] The current page as arabic number. This is suitable for integer operations and
+   comparisions.
+ \item[\texttt{currentlabel}] The content of \cs{@currentlabel}. This is the value that
+   you get also with the standard \cs{label}/\cs{ref}.
```
FrankMittelbach, line 1229: Reading over the list of predefined properties, this strikes me as inconsistent. Either we should have currentlabel, currentcounter and currenttarget (because all three are produced by explicit commands in the document like headings or \refstepcounter or the like) or we should always drop the "current" and just call it label. 

I think there is something in making the predefined properties concise so my voe goes for call this one "label".

base/doc/clsguide.tex
```diff
@@ -1181,6 +1181,113 @@ \subsection{Querying localisation}
...
+ \item[\texttt{counter}] The content of \cs{@currentcounter}. 
+   This command contains after a \cs{refstepcounter} the name of the counter.
+ \item[\texttt{xpos}, \texttt{ypos}] These properties records the $x$~and $y$ coordinates of a point previously 
+   stored with \cs{pdfsavepos}/\cs{savepos}. 
+   E.g.~(if \pkg{bidi} is used it can be necessary to save the position
+   before and after the label):
+   \begin{verbatim}
+     \pdfsavepos 
+     \RefProperty{myposition}{xpos,ypos}%
+     \pdfsavepos
```
FrankMittelbach, line 1244: not sure I understand this example it seems a little incomplete.

base/doc/clsguide.tex
```diff
@@ -1181,6 +1181,113 @@ \subsection{Querying localisation}
...
+   \end{verbatim}  
+\end{description} 
+ 
+Class and package authors can define more properties to store other values they are interested in.
+\begin{decl}
+  |\NewProperty|\arg{name}\arg{setpoint}\arg{default}\arg{code}\\
+  |\SetProperty|\arg{name}\arg{setpoint}\arg{default}\arg{code}
+\end{decl}
+These commands declare or change a property \meta{name}\footnote{The standard set of properties and properties of other package should not be changed!}. If a new property
```
FrankMittelbach, line 1253: of other packages should \emph{never}

base/doc/clsguide.tex
```diff
@@ -1181,6 +1181,113 @@ \subsection{Querying localisation}
...
+ 
+Class and package authors can define more properties to store other values they are interested in.
+\begin{decl}
+  |\NewProperty|\arg{name}\arg{setpoint}\arg{default}\arg{code}\\
+  |\SetProperty|\arg{name}\arg{setpoint}\arg{default}\arg{code}
+\end{decl}
+These commands declare or change a property \meta{name}\footnote{The standard set of properties and properties of other package should not be changed!}. If a new property 
+is declared within a package it is suggested that its name is always structured as follows:
+\meta{package-name}\texttt{/}\meta{property-name}. \meta{setpoint} is either |now| or |shipout| and decides if the 
+value is written directly or at the next shipout. \meta{default} is used if the property is referenced but not yet known.
```
FrankMittelbach, line 1256: not yet known, e.g., in the first run.

base/doc/clsguide.tex
```diff
@@ -1181,6 +1181,113 @@ \subsection{Querying localisation}
...
+
+The commands related to properties are offered as a set of CamelCase
+commands for traditional \LaTeXe{} packages (and for use in the
+document preamble if needed) as well as \texttt{expl3} commands
+for modern packages, that use the L3 programming layer of
+\LaTeX{}. The \texttt{expl3} commands and more details can by found in \texttt{ltproperties-doc.pdf}. 
+
+\subsection{Preparing link targets}
+
+Active links in a document need targets where they can jump to. Such targets are often created automatically (if the package \pkg{hyperref} is loaded) by the \cs{refstepcounter} command but there are also cases where class or package authors need to add a target manually for example in unnumbered sectioning commands or in environments. For this \LaTeX{} provides the following commands. \emph{Without} \pkg{hyperref} they do nothing or insert only a whatsits, \emph{with} \pkg{hyperref} they add the necessary targets. Details about the behaviour and the arguments of the following commands can by found in the \pkg{hyperref} package in \texttt{hyperref-linktarget.pdf}.
```
FrankMittelbach, line 1270: appreciate if you put newlines into your paragraphs :-) here and elsewhere ... much better for diffing

FrankMittelbach: targets where -> target to where ??? not sure

FrankMittelbach: only a whatsit -> only a whatsit (to ensure that spacing doesn't change when hyperref is loaded),

base/doc/clsguide.tex
```diff
@@ -1181,6 +1181,113 @@ \subsection{Querying localisation}
...
+A property is something that \LaTeX{} can track, such as a page number, a section number or a name. Such properties
+can be labeled, their current value written to the \texttt{aux}-file and referenced in the next compilation, similar to the standard \cs{label}/\cs{ref} commands. 
+
+\begin{decl}
+  |\RecordProperties|\arg{label}\arg{list of properties} 
+\end{decl}
+This command writes the current or shipout value (depending on the declaration of the property) 
+of the properties in the list into the \texttt{aux}-file labeled by \meta{label}. The arguments can contain commands which are expanded. \meta{label} can expand to an arbitrary string (as long as it can safely be written to the \texttt{aux}-file) but note that the label names of \cs{label} and \cs{RecordProperties} share a singe namespace. That means that you get a \texttt{Label `A' multiply defined} warning with the following code:
+\begin{verbatim}
+\label{A}\RecordProperties{A}{page}
```
FrankMittelbach, line 1195: I would use a different property to avoid the impression that it is because  \label remembers the "page".

base/doc/clsguide.tex
```diff
@@ -1181,6 +1181,113 @@ \subsection{Querying localisation}
...
+\RefProperty{mylabel}{counter} % outputs section
+\setcounter{mycounter}{\RefProperty{mylabel}{pagenum}} 
+\end{verbatim}
+
+
+
+As \cs{RefProperty} is expandable it can not issue a rerun warning if a label is not found. If needed such a warning can be forced by the following command:
+\begin{decl}
+ |\RefUndefinedWarn|\arg{label}\arg{property}
+\end{decl}
```
FrankMittelbach, line 1216: This check can be carried out anywhere after the aux file has been read, e.g., in the begindocument hook or any time later.

base/doc/clsguide.tex
```diff
@@ -1181,6 +1181,113 @@ \subsection{Querying localisation}
 where |#1| is the user input and the first argument to
 \cs{MakeUppercaseAux} takes two arguments, the locale and input text.
 
+\subsection{Extended and expandable references of properties}
+
+A property is something that \LaTeX{} can track, such as a page number, a section number or a name. Such properties
+can be labeled, their current value written to the \texttt{aux}-file and referenced in the next compilation, similar to the standard \cs{label}/\cs{ref} commands.
```
FrankMittelbach, line 1187: and then referenced in the next compilation, similar to the standard \cs{label}/\cs{ref} commands (that record/reference a fixed set of properties: currentlabel,page,title, and target).

FrankMittelbach (changes requested):
some suggestions for change (one involves a property name change).

base/doc/clsguide.tex
FrankMittelbach: or even more forcefully: \footnote{Only change properties that you have defined. The standard ...  }

base/doc/clsguide.tex
car222222: Strong agreement with these not being in amy sense "current" when the item is retrieved! 

"label" has other problems, since it also used for the "logical label", with which this "label" has zero (or even negative:-) connection!

base/doc/clsguide.tex
car222222: I have a lot of comments and ideas about storing/recording the "current position".  Not sure if I should add them to this review?

base/doc/clsguide.tex
car222222: Does this mean that packages cannot, for example, redefine \@currentlabel ??

Or just that this property must always have code \@currentlabel ?

base/doc/clsguide.tex
car222222: "to which they can jump"

Not that I approve of using the verb "jump" here!  But will leave that to my full review.

base/doc/clsguide.tex
```diff
@@ -1181,6 +1181,113 @@ \subsection{Querying localisation}
...
+\begin{decl}
+ |\LinkTargetOn|\\
+ |\LinkTargetOff|
+\end{decl}
+These commands allow to enable and disable locally the creation of targets. This can be useful to suppress targets created
+automatically by \cs{refstepcounter}.
+
+\begin{decl}
+ |\NextLinkTarget|\arg{manual target}
+\end{decl}
```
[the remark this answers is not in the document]
car222222: I have even more to write about this one!

This will involve the whole area of targets and finding "names" for "named (structure) destinations".
There are many possible applications of this idea beyond the (somewhat strange) possible set-ups in the current hyperref. 

Maybe we should keep all this private (within latex-lab), pending further understanding of more generally what is really needed here.
NOTE: this is a comment on the .dtx file itself, not just on this documentation. 

But I shall wait until this stage is finished.

base/doc/clsguide.tex
```diff
@@ -1181,6 +1181,113 @@ \subsection{Querying localisation}
 where |#1| is the user input and the first argument to
 \cs{MakeUppercaseAux} takes two arguments, the locale and input text.
 
+\subsection{Extended and expandable references of properties}
+
+A property is something that \LaTeX{} can track, such as a page number, a section number or a name. Such properties
```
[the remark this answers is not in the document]
car222222: "references of properties" ???  
Pure invention!! == nonsense!! 

Sorry, I just spotted this in passing.

car222222:
Mostly, so far, I have just commented further on those things Frank picked up on.

I shall postpone my proper review until at least the less contentious of these are resolved, and we get our line-ends back:-).

car222222:
Overall, reading both these comments and the documentation in ltproperties.dtx (including the far too short appendix:-) has convinced me that this "package" is not yet sufficiently mature and complete for "prime-time" release.  

Very sorry!

base/doc/clsguide.tex
```diff
@@ -1181,12 +1181,162 @@ \subsection{Querying localisation}
...
+   suitable for integer operations and comparisions.
+
+ \item[\texttt{currentlabel}] The content of \cs{@currentlabel}. This
+   is the value that you get also with the standard
+   \cs{label}/\cs{ref}.
+
+ \item[\texttt{title}] The content of \cs{@currentlabelname}.  This
+   command is filled beside others by the \pkg{nameref} package and
+   some classes (e.g.~\pkg{memoir}) and typically gives the title of
+   some sectioning command.
```
wspr, line 1253: Maybe I’m just tired but I first read this as meaning it would contain a string like “Chapter”. Perhaps “typically gives the title defined in the document by some sectioning command”

base/doc/clsguide.tex
FrankMittelbach: They are "current" in my opinion: code in LaTeX sets them up and here is a "current" value for them at any time in the document and that is what is retrieved. And on top of it this is what LaTeX called them for the last 40 years (and nobody wondered or objected). My point is only that a shorter name would do as well and for consistency it should be either "current" on all of them or on none. And given that "on all of them" means a lot of  extra tokens for no good reasing in the code I'm for "none" ie also here. 

As to "label" being overloaded, I agree, but that doesn't mean it couldn't be used. I think it can and it will not cause practical problems, on the contrary inventing a really new name for somehting that is linked to `\@currentlabel` would cause more problems and removing the `\@currentlabel` interface on the code-level would be a non-starter.

base/doc/clsguide.tex
```diff
@@ -1181,12 +1181,162 @@ \subsection{Querying localisation}
...
+These commands declare or change a property \meta{name}\footnote{The
+standard set of properties and properties of other package should not
+be changed!}. If a new property is declared within a package it is
+suggested that its name is always structured as follows:
+\meta{package-name}\texttt{/}\meta{property-name}. \meta{setpoint} is
+either |now| or |shipout| and decides if the value is written directly
+or at the next shipout. \meta{default} is used if the property is
+referenced but not yet known.  \meta{code} is the code executed when
+storing the value. For example the \texttt{pagenum} property is
+declare as
```
wspr, line 1289: “declared as”

base/doc/clsguide.tex
FrankMittelbach: The latter. The property retrieves `\@currentlabel` and that should not change. Setting `\@currentlabel` in code is what the interface to this property is (if you want to call that an interface :-) ) and so sure code can and has to set `\@currentlabel`so that the `\RecordProperties` can slap a name onto it at some point in the document and then retrieve the value associated with the name elsewhere.

base/doc/clsguide.tex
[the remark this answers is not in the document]
wspr: I’m not close enough to this material to have a strong opinion — I presume that hyperref provides these at this level so that package authors have the tools they need to ensure targets are created in the right places if they have complex document structures. If that is the case (i.e., a foreseeable use case beyond being used internally within hyperref) then I think they justify being kept.

base/doc/clsguide.tex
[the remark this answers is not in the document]
wspr: > "references of properties" ??? Pure invention!! == nonsense!!

I don't understand what you mean, it makes sense to me.

base/doc/clsguide.tex
```diff
@@ -1181,12 +1181,162 @@ \subsection{Querying localisation}
...
+
+Active links in a document need targets where they can jump to. Such
+targets are often created automatically (if the package \pkg{hyperref}
+is loaded) by the \cs{refstepcounter} command but there are also cases
+where class or package authors need to add a target manually for
+example in unnumbered sectioning commands or in environments. For this
+\LaTeX{} provides the following commands. \emph{Without}
+\pkg{hyperref} they do nothing or insert only a whatsits, \emph{with}
+\pkg{hyperref} they add the necessary targets. Details about the
+behaviour and the arguments of the following commands can by found in
```
josephwright, line 1311: 'behavior' (sigh, but that's the general position)

base/doc/clsguide.tex
u-fischer: I changed the text and will change the actual code in the ltproperties-activechars branch if that is ok.

base/doc/clsguide.tex
u-fischer: it is not incomplete but wrong ;-), corrected.

base/doc/clsguide.tex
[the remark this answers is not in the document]
u-fischer: I changed from "manual target" to "target name". Apart from this: The code here is not new, it is in the kernel since about a year, I only thought it would be good if it is mentioned in clsguide. The general idea is to allow package authors to provide support for active links without having to add a \@ifpackageloaded{hyperref} everywhere. hyperref redefines the commands to create real targets.

base/doc/clsguide.tex
u-fischer: done

base/doc/clsguide.tex
u-fischer: hm, the rest of the file uses behaviour everywhere (with one exception).  So what is the general position ???

davidcarlisle: Generally use American english. In the color package  I used American English in the code and English English in the documentation (but I'm not sure that was the best idea ever).

FrankMittelbach: as David said, in general we try to use American English (with a few historical exceptions) and with the exception that if David or Joseph or Chris write documentation their fingers tend to write British English (sometimes).
"behavior" versus "behaviour" is more borderline as far as I'm concerned (in contrast to writing "you change the colour to red by using `\color{red}`" which indeed I think wasn't the best idea ever :-)

So if Joseph spotted "behaviour" change it thoughout, I would say.

FrankMittelbach (approved):

Issue #1004, opened 2023-02-24
Provide a test for l3kernel date


muzimuzhi:
## Brief outline of the enhancement

**LaTeX2e generally cannot add new features without an extreme amount of care to accommodate backwards compatibility. Please do not be offended if your request is closed for being infeasible.**

Although the LaTeX3 programming layer (expl3) is bundled into the LaTeX2e format since Feb, 2020, the LaTeX3 (or, l3kernel) has its own version, is upgraded independently and more frequently, hence there's the need to test for its date.

## Minimal example showing the current behaviour

As of LaTeX2e 2022-11-01, there's no public command for this.

## Minimal example showing the desired new behaviour

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{article}

\makeatletter
\def\IfExplAtLeastTF{%
  \@ifl@t@r\ExplLoaderFileDate}
\makeatother

\begin{document}
\IfExplAtLeastTF{2023-02-02}TF % T

\IfExplAtLeastTF{2100-01-01}TF % F
\end{document}
```

One use case is to make example files testing against latex3 bugs/new features work for both old and new latex3 releases.
```tex
\IfExplAtLeastTF{<date>}{<patch code or empty>}{<empty or patch code>}
```

It seems l3kernel need a similar test as well, especially for non-LaTeX users. But since in LaTeX there's already a helper `\@ifl@t@r`.

muzimuzhi:
Perhaps also provide a test/tester for `l3backend` date, see https://github.com/latex3/latex3/issues/1177#issuecomment-1457584599.

Currently there's no specific macro holding that date, but in `latex2e` once the backend is loaded its date is stored at the beginning of `\cs{ver@l3backend- \c_sys_backend_str .def}`.

```tex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{article}

\ExplSyntaxOn
\msg_new:nnn { kernel } { backend-not-loaded }
  { l3backend~ not~ yet~ loaded }

\cs_new:Npn \IfExplBackendAtLeastTF
  {
    \bool_lazy_and:nnTF
      { \cs_if_exist_p:N \c_sys_backend_str }
      { \cs_if_exist_p:c { ver@l3backend- \c_sys_backend_str .def }}
      {
        \exp_args:cc { @ifl@t@r } { ver@l3backend- \c_sys_backend_str .def }
      }{
[...]

PR #1146, opened 2023-10-12
Add \IfExplAtLeastTF (see #1004)

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added


base/ltexpl.dtx
```diff
@@ -516,4 +516,27 @@
...
 %    \end{macrocode}
 %
+% \DescribeMacro\IfExplAtLeastTF
+%   A pretty simple wrapper.
+%
+% \begin{macro}{\IfExplAtLeastTF}
+% \changes{v1.3g}{2023/10/12}{Provide a test for \pkg{expl3} date (gh/1004)}
+%    \begin{macrocode}
+%<*2ekernel|latexrelease>
+%<latexrelease>\IncludeInRelease{2022/06/01}%
```
muzimuzhi, line 526: Is the date 2022/06/01 specifically chosen? The next release should be 2023/11/01.

muzimuzhi:
Some checks failed because tests rolling back to 2022-06-01, 2022-11-01, and 2023-06-01 need to be updated.
```
l3build save -e pdftex,xetex,luatex \
  tlb-latexrelease-rollback-2022-06-01 \
  tlb-latexrelease-rollback-2022-11-01 \
  tlb-latexrelease-rollback-2023-06-01
```

josephwright:
@muzimuzhi  Yeah, I was just fixing that

base/ltexpl.dtx
```diff
@@ -516,4 +516,27 @@
...
+% \end{macro}
+%
+%    \begin{macrocode}
+%</2ekernel|latexrelease>
+%<latexrelease>\EndIncludeInRelease
+%<latexrelease>\IncludeInRelease{0000/00/00}%
+%<latexrelease>                 {\IfExplAtLeastTF}{Test for expl3 date}%
+%<latexrelease>
+%<latexrelease>\let\IfExplAtLeastTF\@undefined
+%<latexrelease>\EndIncludeInRelease
```
FrankMittelbach, line 539: I don't know  what we did concerning rollback in case of \IfFormatAtlLeastTF (would need to check) but is it wise to undefine this command in rollback? I  mean it is intended to be used by packages and those package would then automatically fail in rollback. But we don't roll back the L3 prog layer, so in fact it will in rollback still offer the same  as before.

Thus, I think in this particular case it should either not have a rollback at all or it should have the same definition as line 528 (allowing for a roll forward).

FrankMittelbach (changes requested):
maybe not really request changes but suggest changes or think about changes :-)

base/ltexpl.dtx
muzimuzhi: FYI, it's pretended that `\IfFormatAtLeastTF`, `\IfPackageAtLeastTF` and similar commands were "available since the beginning of dawn".
https://github.com/latex3/latex2e/blob/6e3a3e37cf74dc101b445ba799aabb9de9cafc82/base/ltclass.dtx#L912-L923

muzimuzhi: The `latexrelease` package was introduced in 2015-01-01, while expl3 was not part of the format untill 2020-02-02. Thus `\ExplLoaderFileDate` may not exist in the format when rolling forward.

FrankMittelbach: Well, roll-forward was always rather fragile and only intended to help with situations like Overleaf being 1 or 2 releases behind. Integrating expl3 was a kind of major cut  (as it is not roll back or rolled forward) so I doubt that you can roll forward at all from a format before 2020-02-02 at all, I guess it would be in other places first.

base/doc/ltnews38.tex
```diff
@@ -207,8 +207,18 @@ \section{Extended cross-referencing of document properties}
...
 
+\section{Testing for \pkg{expl3} version: \cs{IfExplAtLeastTF}}
+
+The integration of \pkg{expl3} into the kernel means that programmers can use
+all of the features available without needing to load explicitly. However, as
+\pkg{expl3} is not a separate package, the version can't be tested using
+\cs{IfPackageAtLeastTF}. To date, low-level methods have therefore been needed
+to check for the availablity of features in \pkg{expl3}. We have now added
+\cs{IfExplAtLeastTF} as a test working in the same way as
+\cs{IfPackageAtLeastTF} but focussed on the pre-loaded programming layer.
```
FrankMittelbach, line 218: one could perhaps mention that this date shows up when LaTeX starts and also in the .log file near the end as in

LaTeX2e <2023-06-01> patch level 1
L3 programming layer <2023-10-10>

might help people to find out which date they need.

base/ltexpl.dtx
```diff
@@ -516,4 +516,27 @@
...
+% \DescribeMacro\IfExplAtLeastTF
+%   A pretty simple wrapper.
+%
+% \begin{macro}{\IfExplAtLeastTF}
+% \changes{v1.3g}{2023/10/13}{Provide a test for \pkg{expl3} date (gh/1004)}
+%    \begin{macrocode}
+%<*2ekernel|latexrelease>
+%<latexrelease>\IncludeInRelease{2023/11/01}%
+%<latexrelease>                 {\IfExplAtLeastTF}{Test for expl3 date}%
+\def\IfExplAtLeastTF{\@ifl@t@r\ExplLoaderFileDate}
```
FrankMittelbach, line 528: not sure, but to address the comment of \ExplLoderFileDate not available in case of a roll forward on could have an additional line here that sets it to something (not sure what) if it is undefined.

However, my guess is that such a roll forward would not work anyway, so probably not really helpful

FrankMittelbach (approved):
with 2 comments

base/ltexpl.dtx
josephwright: `\ExplLoaderFileDate` has been defined since 2017 (https://github.com/latex3/latex3/commit/d060bf686b5982ef5ee20927369a70fc17df0240#diff-f1498259ee1774283bc5d8fea536a113cac11edca429dddab2695291d24a0444) - I think if it doesn't exist there are other issues to worry about!

FrankMittelbach: Then I think we are good if it is that old.

PhelypeOleinik (approved):

PR #1153, opened 2023-10-26
Use a protected def for tilde which is safe in csnames

josephwright:


base/ltspace.dtx
```diff
@@ -1217,7 +1218,23 @@
 \DeclareRobustCommand{\nobreakspace}{%
    \leavevmode\nobreak\ }
 \catcode `\~=13
-\def~{\nobreakspace{}}
+%</2ekernel>
+%<latexrelease>\IncludeInRelease{2023/11/01}%
+%<latexrelease>                 {[tilde]}{Protected tilde}%
```
davidcarlisle, line 1223: I thought we documented this argument had to be a single token, but we don't document that explicitly (although do rely on a single `\string` making the whole argument safe in a csname)

base/ltspace.dtx
```diff
@@ -1217,7 +1218,23 @@
...
+%<*2ekernel|latexrelease>
+\protected\edef~{%
+  \noexpand\ifincsname\noexpand\expandafter\string~%
+  \noexpand\else
+    \noexpand\expandafter\noexpand\nobreakspace
+  \noexpand\fi
+}
```
davidcarlisle, line 1230: OK I guess it's safe to drop the `{}` as `\protected` will mean will write as `~` to a file anyway. {perhaps a code level comment saying that would be useful) The expandafters are not really needed here, doing a nobreakspace inside an if doesn't hurt, but I guess it's good practice anyway.

davidcarlisle (commented):
approve apart from query over using `[tilde]` rather than (say) `~`  or `\nobreakspace`

base/ltspace.dtx
josephwright: I was worried we'd simply loose a `~` here, hence not using it

base/ltspace.dtx
josephwright: Yes, I originally had more `\expandafter` and kept the `{}`, but I think there were only there for the case where `~` expands to give `\nobreakspace .... [space] ...`, and we now don't have that issue. Happy to document.

I guess I'm used to having the tokens available to want to clean up conditionals all the time 'just in case'.

base/ltspace.dtx
josephwright: Changed

base/ltspace.dtx
josephwright: Code comment added

davidcarlisle (approved):

u-fischer (approved):

PR #1158, opened 2023-10-29
CAR rewrite of section with changes by FMi

FrankMittelbach:
# Internal housekeeping

rewrite of one section (mainly CAR + my updates changes)

### DIFF
--- a/base/doc/ltnews38.tex
+++ b/base/doc/ltnews38.tex
@@ -280,29 +280,27 @@ documentation see \texttt{texdoc ltsockets-doc}.
 
 \section{Document properties and cross-referencing}
 
-Traditional \LaTeX{} uses \cs{label} to record the values of
-two \enquote{local}
-properties of the document: the textual representations of the
-\emph{current
-page number} and
-the \emph{current \textup{\cs{ref}} value} set by
-\cs{refstepcounter} declarations~\cite[p.~209]{38:Lamport}.
-(These declarations are issued, for example, by
-sectioning commands, by numbered environments like \env{equation}, and by \cs{item} in an
-\env{enumerate} or a similar environment.)
+Traditional \LaTeX{} uses \cs{label} to record the values of two
+\enquote{local} properties of the document: the textual
+representations of the \emph{current page number} and the
+\emph{current \textup{\cs{ref}} value} set by \cs{refstepcounter}
+declarations~\cite[p.~209]{38:Lamport}.  (These declarations are
+issued, for example, by sectioning commands, by numbered environments
+like \env{equation}, and by \cs{item} in an \env{enumerate} or a
+similar environment.)
 
-These two recorded values can then be accessed and typeset
-(from anywhere in the next run of the document) by use of the
-(non-expandable) commands \cs{ref} and \cs{pageref}
-using the \emph{key} that was specified as the argument to \cs{label} when recording these values.
-This supported basic cross-referencing (within a document) using
-these recorded values to provide both page-related and counter-related
-information (such as the page xvii or the subsection number 4.5.2).\footnote{In the Spring
-2023 release of \LaTeX{}, the \cs{label} command was extended
-to record, in addition, both a title (such as the text used in a
-section head) and the \emph{logical name} used
-for an associated link target
-if these have been set by packages such as  \pkg{nameref} or \pkg{hyperref}.}
+These two recorded values can then be accessed and typeset (from
+anywhere in the next run of the document) by use of the
+(non-expandable) commands \cs{ref} and \cs{pageref} using the
+\emph{key} that was specified as the argument to \cs{label} when
+recording these values.  This supported basic cross-referencing
+(within a document) using these recorded values to provide both
+page-related and counter-related information (such as the page xvii or
+the subsection number 4.5.2).\footnote{In the Spring 2023 release of
+\LaTeX{}, the \cs{label} command was extended to record, in addition,
+both a title (such as the text used in a section head) and the
+\emph{logical name} used for an associated link target provided these
+have been set by packages such as \pkg{nameref} or \pkg{hyperref}.}
 
 
 Over the years \LaTeX{} packages have appeared that extend
@@ -325,14 +323,15 @@ idea of properties and lists of properties, with methods to record, and
 subsequently access, the value of any declared property.
 
 
-Starting with this release, the \LaTeX{} kernel provides handling of general document
-properties as a core functionality with standard interfaces.
+Starting with this release, the \LaTeX{} kernel provides handling of
+general document properties as a core functionality with standard
+interfaces.
 %
-This is based on concepts introduced by the \pkg{zref}
-package but with some differences in detail, particularly
-in the implementation.
+This is based on concepts introduced by the \pkg{zref} package but
+with some differences in detail, particularly in the implementation.
 It supports the declaration of new properties, and the recording of
-the values of any list of properties. These values are retrieved expandably.
+the values of any list of properties. These values are retrieved
+expandably.
 
 To set up a new property that is the current chapter
 number, for example, here is the declaration to use.
@@ -340,11 +339,13 @@ number, for example, here is the declaration to use.
   \NewProperty{chapter}{now}{?}{\thechapter}
 \end{verbatim}
 The second argument means that the property value will be recorded
-immediately (\enquote{now}), and not \enquote{during the next \cs{shipout}}.
-The third argument sets a default
-to be used when, for example, an unknown label is supplied.
+immediately (\enquote{now}), and not \enquote{during the next
+  \cs{shipout}}.  The third argument sets a default to be used when,
+for example, an unknown label is supplied.
 %
-The final argument contains the code that will, as part of the recording process, be expanded to obtain the value to record for this property.
+The final argument contains the code that will, as part of the
+recording process, be expanded to obtain the value to record for this
+property.
 
 Then, to record the value of this new property,
 together with others, use this command.
@@ -353,12 +354,12 @@ together with others, use this command.
                    {chapter,page,label}
 \end{verbatim}
 This records the current values for the properties \texttt{chapter},
-\texttt{page}, and \texttt{label}, using \texttt{mylabel} as the label, or \emph{key}
-for the record.
+\texttt{page}, and \texttt{label}, using \texttt{mylabel} as the
+label, or \emph{key}, for the record.
 
-To reference (i.e., to retrieve)
-this recorded value for use in a cross-reference to this chapter, use
-the \cs{RefProperty} command with two arguments: the label, \emph{key}, and the property.
+To reference (i.e., to retrieve) this recorded value for use in a
+cross-reference to this chapter, use the \cs{RefProperty} command with
+two arguments: the label, or \emph{key}, and the property.
 \begin{verbatim}
   \RefProperty{mylabel}{chapter}
 \end{verbatim}
@@ -368,27 +369,43 @@ The
 declarations for some generally useful properties,
 including these:
 \begin{description}
-\item[\texttt{label}] the textual representation of the \emph{current \textup{\cs{ref}} value}, see above;
-\item[\texttt{page}] the textual representation of the page number for the page currently under construction;
+\item[\texttt{label}] the textual representation of the \emph{current
+\textup{\cs{ref}} value}, see above;
+
+\item[\texttt{page}] the textual representation of the page number for
+  the page currently under construction;
+
 \item[\texttt{title}] the title, if set by, e.g.,~\pkg{nameref};
-\item[\texttt{target}] the logical name of the associated link target, if set by,
-e.g.,~\pkg{hyperref};
-\item[\texttt{pagetarget}] the logical name of the target added
-by \pkg{hyperref} at the origin of each shipped out page;
-\item[\texttt{pagenum}] the value of the \LaTeX{} counter \texttt{page} in Arabic numerals;
-\item[\texttt{abspage}] the absolute page number of the page under construction, i.e.,
-  one more than the number of pages shipped out so far (thus it starts at 1 and is increased by 1 whenever a page is shipped out);
-\item[\texttt{counter}] the name of the counter that produced the \emph{current \textup{\cs{ref}} value}, i.e.,
-                  the counter that was stepped in the most recent \cs{refstepcounter} within the current scope;
-\item[\texttt{xpos}, \texttt{ypos}] the position on the shipped out page as
-set by the most recent \cs{pdfsavepos}: recording these properties should
-be done as soon as possible after saving the position.
+
+\item[\texttt{target}] the logical name of the associated link target,
+  if set by, e.g.,~\pkg{hyperref};
+
+\item[\texttt{pagetarget}] the logical name of the target added by
+  \pkg{hyperref} at the origin of each shipped out page;
+
+\item[\texttt{pagenum}] the value of the \LaTeX{} counter
+  \texttt{page} in Arabic numerals;
+
+\item[\texttt{abspage}] the absolute page number of the page under
+  construction, i.e., one more than the number of pages shipped out so
+  far (thus it starts at 1 and is increased by 1 whenever a page is
+  shipped out);
+
+\item[\texttt{counter}] the name of the counter that produced the
+  \emph{current \textup{\cs{ref}} value}, i.e., the counter that was
+  stepped in the most recent \cs{refstepcounter} within the current
+  scope;
+
+\item[\texttt{xpos}, \texttt{ypos}] the position on the shipped out
+  page as set by the most recent \cs{pdfsavepos}: recording these
+  properties should be done as soon as possible after saving the
+  position.
 \end{description}
 
 Both \LaTeXe{} commands (using camel-case names) and
 \LaTeX3 programming layer commands are provided.
-
-For the full documentation, see \texttt{texdoc ltproperties-doc}.
+%
+For a more complete documentation, see \texttt{texdoc ltproperties-doc}.
 
 
 
--- a/base/doc/ltnews38.tex
+++ b/base/doc/ltnews38.tex
@@ -357,7 +357,7 @@ This records the current values for the properties \texttt{chapter},
 \texttt{page}, and \texttt{label}, using \texttt{mylabel} as the
 label, or \emph{key}, for the record.
 
-To reference (i.e., to retrieve) this recorded value for use in a
+To \emph{reference} (i.e., retrieve) this recorded value for use in a
 cross-reference to this chapter, use the \cs{RefProperty} command with
 two arguments: the label, or \emph{key}, and the property.
 \begin{verbatim}
--- a/base/doc/ltnews38.tex
+++ b/base/doc/ltnews38.tex
@@ -335,15 +335,15 @@ The \pkg{cleveref} package stores (by means of a second, internal
 The \pkg{hyperref} package adds the \cs{autoref} command, which
 tries to retrieve the name of a counter from the  \emph{logical name}
 used for a link target.
-The \pkg{tikzmarks} package records information about
- \emph{labelled positions} in pictures when using \pkg{tikz}.
+The \pkg{tikzmarks} library records information about
+ \emph{labelled positions} on the page when using \pkg{tikz}.
 Finally, the \pkg{zref} package implements many related ideas, including a general
 idea of properties and lists of properties, with methods to record, and
 subsequently access, the value of any declared property.
 
 
 %%FMi some rewrite:
-Starting with this release, the \LaTeX{} kernel now offers the handling general document
+Starting with this release, the \LaTeX{} kernel now offers the handling of general document
 properties as a core functionality with standard interfaces.
 %
 This is based on concepts introduced by the \pkg{zref}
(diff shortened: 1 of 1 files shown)


josephwright (approved):

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+And the \pkg{smart-ref} package supports the storage of a larger
+collection of counter values so that, for example, a cross-reference
+can refer to the relevant chapter together with an equation tag.
+The \pkg{cleveref} package stores (by means of a second, internal
+\enquote{logical label}) extra information such as the name of the counter.
+The \pkg{hyperref} package adds the \cs{autoref} command, which
+tries to retrieve the name of a counter from the  \emph{logical name}
+used for a link target.
+The \pkg{tikzmarks} package records information about
+ \emph{labelled positions} in pictures when using \pkg{tikz}.
```
u-fischer, line 339: tikzmark is not restricted to pictures. It can be used everywhere in the text. I wouldn't try to be too exact here and replace "in pictures when using tikz" by "on the page"-

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
-(\enquote{now}) and not at \cs{shipout}. The third argument sets a default if a 
-label is unknown. It is then possible to record a list of properties with 
-\cs{RecordProperties}: 
+The second argument means that the property value will be recorded
+immediately (\enquote{now}), and not \enquote{during the next \cs{shipout}}.
+The third argument sets a default
+to be used when, for example, an unknown label is supplied.
+%%FMi added
+The final argument contains the code that produces the current
+property value when when recording the value is requested.
```
u-fischer, line 366: when when -> when

u-fischer (approved):

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+used for a link target.
+The \pkg{tikzmarks} package records information about
+ \emph{labelled positions} in pictures when using \pkg{tikz}.
+Finally, the \pkg{zref} package implements many related ideas, including a general
+idea of properties and lists of properties, with methods to record, and
+subsequently access, the value of any declared property.
+
+
+%%FMi some rewrite:
+Starting with this release, the \LaTeX{} kernel now offers the handling general document
```
davidcarlisle, line 346: the handling   of  general document

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+\item[\texttt{title}] the title, if set by, e.g.,~\pkg{nameref};
+\item[\texttt{target}] the logical name of the associated link target, if set by,
+e.g.,~\pkg{hyperref};
+\item[\texttt{pagetarget}] the logical name of the target added
+by \pkg{hyperref} at the origin of each shipped out page;
+\item[\texttt{pagenum}] the value of the \LaTeX{} counter \texttt{page} in Arabic numerals;
+%% NOT as an Arabic number -- Arabic numbers are "something totally different"!!
+%%FMi changed:
+%\texttt{abspage} (the value of the \texttt{shipout count});
+\item[\texttt{abspage}] the current number of shipped out pages;
```
davidcarlisle, line 407: past  tense "shipped out" suggests that it starts at 0 but I think it starts at 1 so the value is 1 if you access this on the first page, even though It's really only set at shipout so perhaps.

`\item[\texttt{abspage}] the absolute page number of the page on which the label was recorded`

or something, not sure about that wording

davidcarlisle (approved):
two minor comments (especially the second one where I'm not sure my suggested wording is better)  but approve anyway

base/doc/ltnews38.tex
FrankMittelbach: good point. but what is also wrong I just noticed is that this is not package but a tikzlibrary

base/doc/ltnews38.tex
FrankMittelbach: good point, but the property is independent of being recorded or not, so should not be describle in that way I think.

   the absolute page number of the page currently produced

maybe

FrankMittelbach: or just: the current number of shipped out pages plus 1;

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+properties of the document: the textual representations of the
+%% FMi added back in, be cause it is the "current", see Lamport:
+\emph{current
+page number} and the
+%%FMi from Lamport
+the \emph{current \textup{\cs{ref}} value} set by
+\cs{refstepcounter} declarations~\cite[p.~209]{38:Lamport}.
+(These declarations are issued by
+sectioning commands, by numbered environments, and by \cs{item} in an
+\env{enumerate} or a similar environment.)
```
muzimuzhi, line 294: Adding back "(by numbered environments) like \env{equation}" existing in Lamport's book?

I found saying solely "numbered environments" a bit hard to understand. At first I thought of numbered list env, then numbered theorem-like envs...

```suggestion
(These declarations are issued by
sectioning commands, by numbered environments like \env{equation},
and by \cs{item} in an \env{enumerate} or a similar environment.)
```

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+
+These two recorded values can then be accessed and typeset
+(from anywhere in the next run of the document) by use of the
+(non-expandable) commands \cs{ref} and \cs{pageref}
+%%FMi new
+using the \emph{key} that was assigned to the values with the \cs{label} command.
+This supported basic cross-referencing (within a document) using
+these recorded values to provide both page-related and counter-related
```
muzimuzhi, line 302: not quite sure
```suggestion
these recorded values to provide both page- and counter-related
```

base/doc/ltnews38.tex
u-fischer: as it records `\g_shipout_readonly_int` you could perhaps point to ltshipout-doc.pdf?

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+(These declarations are issued by
+sectioning commands, by numbered environments, and by \cs{item} in an
+\env{enumerate} or a similar environment.)
+
+These two recorded values can then be accessed and typeset
+(from anywhere in the next run of the document) by use of the
+(non-expandable) commands \cs{ref} and \cs{pageref}
+%%FMi new
+using the \emph{key} that was assigned to the values with the \cs{label} command.
+This supported basic cross-referencing (within a document) using
```
muzimuzhi, line 301: Otherwise this sentence doesn't have no predicate (usually a verb).
```suggestion
This supported basic cross-referencing (within a document) uses
```

base/doc/ltnews38.tex
FrankMittelbach: it is pretty hard to be accurate here (without writing a book). It is also true for theorems and for floats (even though there it happens through the \caption), and ...  this is  why I made it a bit more vague than Leslie.

Perhaps I should add  "issued, for example, by" instead

base/doc/ltnews38.tex
FrankMittelbach: I wouldn't. The reference to that doc wouldn't help in this situation (the property is recorded at shipout) and the technical documentation there about \g_shipout_readonly_int would be rather confusing and we don't say that this is what apspage us looking at (and we shouldn't in ltnews). I think the above statement is fine.

base/doc/ltnews38.tex
FrankMittelbach: Think you are mistaken in this case.

base/doc/ltnews38.tex
FrankMittelbach: would do that either but leave it as written by Chris

base/doc/ltnews38.tex
muzimuzhi: You're right, "supported" is the predicate 😄

base/doc/ltnews38.tex
FrankMittelbach: I meant would *not* do ...

base/doc/ltnews38.tex
FrankMittelbach: I rewrote it slightly

base/doc/ltnews38.tex
car222222: Both are OK: ask the copy-editor about “house style”?

base/doc/ltnews38.tex
car222222: Not sure that I like the “plus one”, better? longer! :

one more than the number of pages shipped out so far (thus it starts at 1 and is increased by 1 whenever a page us shipped out).

base/doc/ltnews38.tex
car222222: Yes, better to add the “for example”

base/doc/ltnews38.tex
car222222: Typo: us ==> is (near to end)

base/doc/ltnews38.tex
car222222: OK: the important point is that it is a tikz module rather than a standalone package.

base/doc/ltnews38.tex
car222222: why not simply:
  . . . kernel provides general . . .   
??

base/doc/ltnews38.tex
car222222: Thanks for the correction re “only pictures” :  I
I could not properly understand the documentation since (naturally) it assumed great familiarity with the tikz universe.

base/doc/ltnews38.tex
car222222: Suggestions:

. . . the code that will, as part of the recording process, be expanded to obtain the value to record for this property.

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+\section{Document properties and cross-referencing}
+
+Traditional \LaTeX{} uses \cs{label} to record the values of
+two \enquote{local}
+properties of the document: the textual representations of the
+%% FMi added back in, be cause it is the "current", see Lamport:
+\emph{current
+page number} and the
+%%FMi from Lamport
+the \emph{current \textup{\cs{ref}} value} set by
```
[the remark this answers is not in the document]
car222222: Agreed.

base/doc/ltnews38.tex
car222222: Also, minor nit-pick on the grammar, delete the “a” here:
“or a similar environment”
Maybe also add commas, thus:
“, or similar, environment”

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+\cs{refstepcounter} declarations~\cite[p.~209]{38:Lamport}.
+(These declarations are issued, for example, by
+sectioning commands, by numbered environments like \env{equation}, and by \cs{item} in an
+\env{enumerate} or a similar environment.)
+
+These two recorded values can then be accessed and typeset
+(from anywhere in the next run of the document) by use of the
+(non-expandable) commands \cs{ref} and \cs{pageref}
+%%FMi new
+using the \emph{key} that was assigned to the values with the \cs{label} command.
```
car222222, line 300: But I am not so sure about this bit:
"assigned to these values".

Does Lamport help us here, with a good phrase? 

Here is one ideas:

. . . that was specified as the argument to \cs{label} when recording these values.

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+%if these have been set by packages such as  \pkg{nameref} or \pkg{hyperref}.
+
+%%FMi: meanwhile sounds as if it just happened after 2023 Spring ...
+%%     So moved the above text into a footnote (is anyway old news) and
+%%    did some rewrites below.
+%%
+%Meanwhile, other packages had been extending
+%this \enquote{label-ref system} in various ways.
+
+Over the years \LaTeX{} packages appeared that extended
```
car222222, line 325: Or (reads better to me:-): 

. . . packages have appeared that extend . . .

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+used for a link target.
+The \pkg{tikzmarks} library records information about
+ \emph{labelled positions} on the page when using \pkg{tikz}.
+Finally, the \pkg{zref} package implements many related ideas, including a general
+idea of properties and lists of properties, with methods to record, and
+subsequently access, the value of any declared property.
+
+
+%%FMi some rewrite:
+Starting with this release, the \LaTeX{} kernel now offers the handling of general document
```
car222222, line 346: Better (I think:-) 😀 definitely shorter! :

. . . kernel provides general . . .

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
-\RecordProperties{mylabel}
-                 {chapter,page,currentlabel}
+  \RecordProperties{mylabel}
+                   {chapter,page,label}
 \end{verbatim}
-To reference the value the \cs{RefProperty} command is provided which takes 
-two arguments, the label and the property: 
+%%FMi added:
+This records the current values for the properties \texttt{chapter},
+\texttt{page}, and \texttt{label}, using \texttt{mylabel} as a label
```
car222222, line 376: addition of key:
. . . as a label, or \emph{key}, . . .

car222222: Should be: "the label" not "a label"!

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+for the record.
+
+%%FMi changed: again following Lamport terminology
+To reference (i.e., to retrieve)
+this recorded value for use in a cross-reference to this chapter, use
+the \cs{RefProperty} command with two arguments: the label and the property.
```
car222222, line 382: add
the label, or \emph{key},

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
-  
-The module provides both \LaTeXe{} camel case commands and
-\LaTeX3 programming layer commands.    
-
-For documentation see \texttt{texdoc ltproperties-doc}. 
+
+The
+%%FMi changed
+%extension itself contains
+\LaTeX{} kernel itself already contains
```
car222222, line 390: Ok, but remove the "already" as it indicates that this happened 'some time before this release'!

base/doc/ltnews38.tex
```diff
@@ -278,67 +278,149 @@ \section{Hooks, sockets and plugs}
...
+by \pkg{hyperref} at the origin of each shipped out page;
+\item[\texttt{pagenum}] the value of the \LaTeX{} counter \texttt{page} in Arabic numerals;
+%% NOT as an Arabic number -- Arabic numbers are "something totally different"!!
+%%FMi changed:
+%\texttt{abspage} (the value of the \texttt{shipout count});
+\item[\texttt{abspage}] the absolute page number of the page under construction, i.e., number of already shipped out pages plus 1;
+%%FMi changed:
+%\texttt{counter}] (the name of the counter stepped in the (local) most recent \cs{refstepcounter});
+\item[\texttt{counter}] the name of the counter that produced the \emph{current \textup{\cs{ref}} value}, i.e.,
+                  from the most recent \cs{refstepcounter} within the current scope;
```
car222222, line 411: More precisely:
the counter that was stepped in the most . .  .

car222222 (changes requested):
I hope I have covered all the changes and comments.

I still need to reread the whole thing again. 

Big thanks to everyone who has made suggestions and changes: all very helpful.

base/doc/ltnews38.tex
FrankMittelbach: I know that both work (in German we would usually use the shorter) but I think it gives a slightly different emphasis and I would keep it like that -- bb might have a different opinion :-)

FrankMittelbach:
> I still need to reread the whole thing again.

please do, but preferably today, so they can get integrated
 
> Big thanks to everyone who has made suggestions and changes: all very helpful.

getting there I think :-)

base/doc/ltnews38.tex
car222222: Also, for extra clarity, start this as follows (note that only one "to"  is needed 😀 

To \emph{reference} (i.e., retrieve) . . .

car222222:
One more thing to fix (or leave for bnb?):
Right at the end there is very bad break, after "texdoc".

car222222:
And one more (will this ever be finished??😃) less important idea:

Use "provided" rather than "if" here (approx line 310?): 

". . . if these have been set by packages such as  \pkg{nameref} or \pkg{hyperref}"

FrankMittelbach:
From experience I'm not bothering with pagination before the bb/kb edits are in. Yes, saw that one too.

car222222:
I think that is me finished (for now:-) . . . 

if you are not careful I shall go on to edit "sockets" too.

car222222:
> From experience I'm not bothering with pagination before the bb/kb edits are in.

Actually the pagination looks pretty good at present: e.g., exactly two cols for this section! 

But this one is a line-break:
 Yes, saw that one too.
Not easy to miss, is it?

car222222:
I do not seem to be able to do "a review" at this stage.

@FrankMittelbach Are you currently working on it, or are you waiting on me for anything?

car222222:
I did find a typo in your latest changes.
And of course I have some questions!

davidcarlisle (approved):

FrankMittelbach:
> I do not seem to be able to do "a review" at this stage.

you should be, starting from this page:

https://github.com/latex3/latex2e/pull/1158/files

> 
> @FrankMittelbach Are you currently working on it, or are you waiting on me for anything?

was not at home, now back

car222222:
@FrankMittelbach 

It claims that I already reviewed it today, and I see no button to "review", or to change my review. Probably just my system.  Maybe if you requested a review from me (yet again!) something  will change. 

Otherwise I shall simply send another comment on two small things.

FrankMittelbach:
> @FrankMittelbach
> 
> It claims that I already reviewed it today, and I see no button to "review", or to change my review. Probably just my system. Maybe if you requested a review from me (yet again!) something will change.

hmm, I did rerequest, reload the page perhaps?

> Otherwise I shall simply send another comment on two small things.

or that, but you said you wanted to look at the whole text and that is much easer to work with (from my end) if you do a full review with all comments are somewhat bundled together. but if it doesn't work individual comments  it is ...

car222222:
The typo (I think) is here, approx line 356-7:

add final comma  here:  ", or \emph{key},"

And one more, related, on line 361:

add "or" to get ", or \emph{key},"

And re line 328? "handling of" 
Comment: is it really better to emphasize the "handling" here, when the major new thing is the properties themselves!  (One cannot handle something until it  exists:-)! 

I do believe that is all!  Can I get a beer now?

car222222:
And, of course, there is always "one more thing" --

Did you think about the need for some introduction to the (good!) use of the term "key" for "logical label"??

FrankMittelbach:
> The typo (I think) is here, approx line 356-7:
> 
> add final comma here: ", or \emph{key},"
> 
> And one more, related, on line 361:
> 
> add "or" to get ", or \emph{key},"

done


> And re line 328? "handling of" Comment: is it really better to emphasize the "handling" here, when the major new thing is the properties themselves! (One cannot handle something until it exists:-)!

the importance is "handling ...as a core functionality of the kernel".  In the past they have been around but with individual (partially incompatible) solutions that not necessarily knew about each other.

> I do believe that is all! Can I get a beer now?

you can :-)

car222222:
I would argue that "the functionality" is the existence of the basic "concept of properties", and not just the "handling" of them.  Maybe just a "meaning of words" difference thing, again!?

FrankMittelbach:
> I would argue that "the functionality" is the existence of the basic "concept of properties", and not just the "handling" of them. Maybe just a "meaning of words" difference thing, again!?

maybe :-) but as you say there is always another thing and we need to stop at some point. So I suggest we go with what we have now (else you never get to your beer and we never get this finished :-) )

car222222:
Yes, sir!!  I agree that it is not spending more time on this now.

On to my second beer already!

car222222:
Did you ever respond to my query concerning the introduction of the term "key"?

Does this need some explanation?

FrankMittelbach:
> Did you ever respond to my query concerning the introduction of the term "key"?
> 
> Does this need some explanation?

no I did not I guess. I don't think it needs an extra explanation or introduction. But I'm sure opinions can vary here.

car222222:
I did some rechecking!

This idea seems to have got lost in the depths of a review:

It applies to line ~360 --
The suggestion was to start this as follows (note that only one "to" is needed 😀)

  To \emph{reference} (i.e., retrieve) . . .

car222222:
Ok!  . . . If I promise not to look at it again.

PR #1164, opened 2023-11-07
Gh1102

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [n/a] Test file(s) added

I decided to use the simpler way and extend the error message with some general text.


josephwright (approved):

base/ltoutenc.dtx
```diff
@@ -3685,7 +3686,11 @@
           {}{\PackageError{fontenc}%
            {Encoding file `\reserved@f' not found.%
             \MessageBreak
-             You might have misspelt the name of the encoding}%
+            You might have misspelled the name of the encoding
+            \MessageBreak
+            or a required support package (e.g., cyrillic) is
+            \MessageBreak
+            missing in your installation}%
```
davidcarlisle, line 3693: I think for packages it's not that the package is not installed, but rather not used in the document (unlike fd files that get found automatically if installed and the encoding name is right)

so rather than `missing in your installation`  perhaps `not loaded`

davidcarlisle (changes requested):

base/ltoutenc.dtx
FrankMittelbach: I think the point is that people do install TL minimal (or some other scheme that does not contain the cyrillic or greek support (package/bundle whatever you call it) and thus the .fd files and the fonts are simply physically not on the machine. So is a question of something like babel is trying to use T2A or LGR and simply doesn't find it. E.g. for cyrillic you don't have to load a LaTeX package but still that's what caused the original request for a better message.

FrankMittelbach: more to the point , this error message is only triggered if the .fd file is not on the system. Otherwise it would find it and you would only get an error if the output fonts are not there but this would happen much later.

FrankMittelbach: @davidcarlisle do you still stay by your suggestion (or a variant one)? Or is it ok as is?

davidcarlisle: leave it as it is (I'm persuaded by argument that if cyrillic was installed but not loaded the fd files would have been found) I'll adjust the approve flag

davidcarlisle (approved):

PR #1165, opened 2023-11-07
Gh823

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge  (more or less)

- [x] Test file(s) added

I doubt it needs a rollback, but we could add one, and of course it should eventually get a news entry if we are good with the warning text.


base/ltcounts.dtx
```diff
@@ -313,8 +314,20 @@
      \@addtoreset{#1}{@ckpt}%
      \global\expandafter\let\csname p@#1\endcsname\@empty
      \expandafter
```
davidcarlisle, line 316: You could move this `\expandafter` after the added comment as  the block starting ` \ifx\csname the#1\endcsname\relax`  looks like a missing `\expandafter`

davidcarlisle (approved):
approve although I left one suggested change

u-fischer:
Assuming that that is for the next release: could/should we combine that with the open PR about \theHcounter (and so moving the hyperref patches)? The main open question there is if `\@definecounter` should directly define also `\theHcounter` (which could five errors if a package contains a `\newcommand\theHcounter`)  or only define at begin document the missing ones.

base/ltcounts.dtx
car222222: Make that "could move" into a "should move"!

car222222:
It probably makes sense to look again at this branch (add-theHcount) soon.

But only as part of a larger area of concern, involving:

details of how, in general, to manufacture "names for named destinations";
whether PDF structural/other "named destinations" are useful, or even needed hyperref (current or future);
probably also more general questions around names and namespaces for many use-cases on the LaTeX side (I can explain this further);
and the relationships between "names", "keys", "mappings", "look-up tables", "properties for programming use-cases", etc. 

Maybe some of this needs new structures at the expl3 level.

car222222:
@u-fischer What, please, is the number of that PR?

car222222:
Found it now: #1095

base/ltcounts.dtx
FrankMittelbach: done

base/ltcounts.dtx
```diff
@@ -305,16 +305,29 @@
...
+%    \end{macrocode}
+%    If \cs{the\#1} is undefined or \cs{relax} we define it with the
+%    standard definition for counters, otherwise we warn. This will
+%    catch, for example, that somebody defines a counter named
+%    ``index'' conflicting with the \env{theindex} environment.
+%    \begin{macrocode}
      \expandafter
-     \gdef\csname the#1\expandafter\endcsname\expandafter
-          {\expandafter\@arabic\csname c@#1\endcsname}}
+     \ifx\csname the#1\endcsname\relax
```
josephwright, line 323: Are we really worried about `\relax` here? Why not just `\ifcsname`?

josephwright (changes requested):

base/ltcounts.dtx
FrankMittelbach: because it would be wrong. Packages, e.g., bigfoot, touch the counter name and as a result it is defined but as \relax. The above is just the normal way of LaTeX to determine if a command is free, ie either undefined or \relax and given that in this case we define it, it doesn't matter that we test with \ifx

josephwright: I was thinking that if the counter has ended up as `\relax` its not really 'in use' ... and I guess with my `expl3` view of the word, it's really not defined at all ;)

Assuming you want to stick this way, why are you not using `\@ifundefined`?

FrankMittelbach: it is not the counter that is \relax, it is `\the<counter>` that is \relax and that is what technically happens, because packages do (for one or the other reason) `\csname the<counter>\endcsname` before the a `\newcounter{<counter>}` was issued. I started with using  `\ifcsname` and several tests blow into my face, especially those around footnotes and floats.  In the past touching `\the<counter>` didn't matter because afterwards it was unconditionally defined by the code above.

as for using `\@ifundefined` I could have used that, it is just a little bit slower as it unnecessarily first tests with \ifcsname. But it is perhaps cleaner.

davidcarlisle: I think `\expandafter\ifx\csname` is fine here and matches the style of the rest of the block (` \expandafter
     \gdef\csname...`  etc)

FrankMittelbach: that was roughly my thinking (despite the fact that it is slightly faster)

josephwright (approved):

PR #1177, opened 2023-11-10
enhance mark extraction to account for forced breaks within the material

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


josephwright (changes requested):
Changes.txt entry?

FrankMittelbach:
> Changes.txt entry?

sure, but too early for that in my opinion.  That's why it says under development and nothing is checkmarked above (I'm one of those folks who are really using Will's checkmark list :-) ). So right now all I'm really interested in is whether the logical seems to be right.

josephwright:
> > Changes.txt entry?
> 
> sure, but too early for that in my opinion. 

Obviously we view this a bit differently: I thought the `changes.txt` _had_ to be updated when you edit the source, even if that means we have a series of edits to `changes.txt`during a PR. I am though happy if you want to leave until the end.

FrankMittelbach:
@josephwright I agree that doing this in parallel is useful most of the time. However, in this case I expect many more changes  (and possibly alternative implementation), so I'm going lightweight on programmer-level docs and do them wehn things are a bit more settled.

FrankMittelbach:
I altered and extended the implementation in anticipation what is necessary for multicols and/or in general for extracting marks out of box material to reinsert them afterwards. This is now closer to what I think is needed so change entries will come up soon @josephwright :-)

muzimuzhi:
To reduce the ambiguity, perhaps a new item can be added to section "Status of pull request", in pull request template, to indicate that the PR author is aware of the missing `changes.txt` entries and friends.

For example,
```
- In early stage (`changes.txt` entries and friends might be missing)
```

FrankMittelbach:
> To reduce the ambiguity, perhaps a new item can be added to section "Status of pull request", in pull request template, to indicate that the PR author is aware of the missing `changes.txt` entries and friends.

sorry don't get it. Of course, I'm aware (and so should everybody else) since I didn't checkmark those item nor did I put n/a (not applicable) into the checkmark box. So it is all there at the very top. Why should a different string from 

  -  Feedback wanted
  - Under development

plus the checkmark list, provide any additional info? It is more that often people do not use the checkmark list, but that is not helped I think by asking for providing a different status string.

josephwright (approved):

base/ltmarks.dtx
```diff
@@ -912,23 +966,100 @@
...
+%    \end{macrocode}
+%    After splitting we check if there is anything left in
+%    \cs{l_@@_box}. If not then the above split has set some split marks
+%    that we can then use to finish the extraction:
+%    \begin{macrocode}
+          \box_if_empty:NTF \l_@@_box
+              { #1 }
+%    \end{macrocode}
+%    If we have a remainder after the split then this means that there was some forced
+%    break in the material. We get rid of that be combining the
```
davidcarlisle, line 989: by

davidcarlisle (approved):

base/ltmarks.dtx
```diff
@@ -784,40 +807,47 @@
...
+\cs_new_protected:Npn \@@_extract_and_handle_marks:nnn #1#2#3 {
 %    \end{macrocode}
-%    Getting the first and last marks out of the material in \verb=#2=
+%    This macro expects a region name as its third argument and
+%    vertical material (not boxed) as its fourth. It attempts to
+%    extract mark information from \verb/#4/ and if it succeeds It
+%    executes \verb/#1{#3}/ to do something with the marks, e.g., to
+%    update the mark structure for the region.
+%  
+%    If it finds infinite negative glue it generates an error message
```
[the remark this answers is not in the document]
FrankMittelbach: Interesting thought, but I don't think there is a way to detect then that the error has happened and it is an error, ie the user is losing the marks and so I think they should learn that this is the case. But perhaps if it is enough if that is something you can find in the .log? It would certainly has the advantage that then the processing doesn't stop if they can't avoid the infinite negative glue. Maybe.

base/ltmarks.dtx
```diff
@@ -987,29 +1118,160 @@
...
+}
+%    \end{macrocode}
+%  \end{macro}
+%
+%
+%
+%  \begin{macro}{\@@_get_marks_for_reinsertion:nNN}
+%
+%    This function extracts the marks from the material in the first
+%    argument but it does not update any the mark structure. Instead,
```
car222222, line 1151: . . . update any mark structures. ?? 
OR:
. . . update the mark structure(s). ??

base/ltmarks.dtx
```diff
@@ -987,29 +1118,160 @@
...
+%    it collects the marks in the token lists given as the second and
+%    third argument, in a way that they can be reinserted by just
```
car222222, line 1153: . . . , in such a way that they . . .

base/ltmarks.dtx
```diff
@@ -987,29 +1118,160 @@
...
+%    executing the token lists.\footnote{It is probably enough to
+%    collect everything in a single token list as long as we put the
+%    first marks first and the last marks last). But for extra
+%    flexibility, I currently use 2 tls. This might change when it is
+%    really clear that this is never needed.}
+%
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_get_marks_for_reinsertion:nNN #1#2#3 {
+%    \end{macrocode}
+%    First we clear the tmp token lists as we haven't seen any marks yet.
```
car222222, line 1163: tmp ==> temporary  ??

base/ltmarks.dtx
```diff
@@ -987,29 +1118,160 @@
...
+  \tl_clear:N \g_@@_last_marks_tl
+%    \end{macrocode}
+%    The we try to extract all marks, thereby filling the token lists
+%    with suitable \cs{mark_insert:nn} calls.
+%    \begin{macrocode}
+  \@@_extract_and_handle_marks:nnn
+%    \end{macrocode}
+%    If we can extract them (i.e., if we don't detect negative infinite
+%    glue) then this is done with \cs{@@_get_from_splitmarks:} otherwise
+%    we issue warning that mark extraction was not possible.
```
car222222, line 1175: issue a warning

base/ltmarks.dtx
```diff
@@ -987,29 +1118,160 @@
...
+%    \begin{macrocode}
+     { \@@_get_from_splitmarks: }
+     { \msg_warning:nn { mark } { infinite-shrinkage-box } }
+     { #1 }
+%    \end{macrocode}
+%    finally we copy the updated (or not updated) tmp token lists to
```
car222222, line 1181: Finally we . . .

car222222: and "tmp" again!

base/ltmarks.dtx
```diff
@@ -987,29 +1118,160 @@
...
+%    \begin{macrocode}
+\cs_new_protected:Npn  \@@_get_from_splitmarks:  {
+  \seq_map_inline:Nn \g_@@_classes_seq
             {
-              \tl_gset_eq:cc { g_@@_#1_top_  ##1_tl }
-                             { g_@@_#1_last_ ##1_tl }
-              \tl_gset_eq:cc { g_@@_#1_first_##1_tl }
-                             { g_@@_#1_last_ ##1_tl }
+%    \end{macrocode}
+%    First we to get the last last mark for the current class from the
```
car222222, line 1204: last last  ==>  last ??

base/ltmarks.dtx
```diff
@@ -987,29 +1118,160 @@
...
+                { \tex_splitbotmarks:D \use:c { c_@@_class_##1_mark } }
+%    \end{macrocode}
+%    
+%    If this mark doesn't exist then obviously first mark doesn't
+%    either, so we do nothing (other than issuing some debugging
+%    info). We have to be a little careful here: something like
+%    \verb=\mark_insert:nn{foo}{}= adds an \enquote{empty} mark that
+%    should not be confused with no mark at all. But no mark in our
+%    material will result in \cs{g_@@_tmp_tl} being fully empty. This
+%    is why we have to make sure that \enquote{empty} from
```
car222222, line 1217: I got lost here!  Try again?

base/ltmarks.dtx
```diff
@@ -987,29 +1118,160 @@
...
+                {
+%<*trace>
+                  \@@_debug:n { \iow_term:x { Marks:~ extract~ last~
+                       mark~ for~ class~ '##1'~ =~  \g_@@_tmp_tl } }
+%</trace>
+                  \tl_gput_right:Ne \g_@@_last_marks_tl
+                     { \mark_insert:nn {##1} { \g_@@_tmp_tl } }
+%    \end{macrocode}
+%    Because we had a last mark we also have a first mark (which might
+%    be the same, but might be not), so we pick that up and add it to
```
car222222, line 1244: be not  ==>  not be

base/ltmarks.dtx
```diff
@@ -987,29 +1118,160 @@
...
+}
 %    \end{macrocode}
-%    Once all mark classes have been processed the data structures are
-%    updated and we can close the group which undoes our local
-%    changes and retains only the global ones.
+%  \end{macro}
+%
+%
+%  \begin{macro}{\g_@@_first_marks_tl,\g_@@_last_marks_tl}
+%    The two global variables used in the code above.
```
car222222, line 1271: These are the two . .

base/ltmarks.dtx
```diff
@@ -1158,7 +1420,7 @@
 %    To retrieve the first, last or top region mark, we grab the
 %    appropriate value stored in the corresponding token list variable
 %    and pass its contents back. These functions should be used only
-%    in output routines after \cs{@@_update_structure:nn} has acted,
+%    in output routines after \cs{@@_update_structure_from_material:nn} has acted,
```
car222222, line 1423: ??

in an output routine and only after . . . 

OR?? 

. . . only immediately after . . .

base/ltmarks.dtx
```diff
@@ -1248,17 +1510,28 @@
...
     \c__msg_coding_error_text_tl
     The~mark~region~'#1'~contains~some~infinite~negative~glue~
-    allowing~it~to~shrink~to~an~arbitrary~size.~
+    allowing~it~to~shrink~to~an~arbitrarily~small~size.~
     This~makes~it~impossible~to~split~the~region~apart~to~
     get~at~its~marks.~They~are~lost.
   }
 %    \end{macrocode}
 %
+%    If box content checked for marks contains some infinite negative glue
```
car222222, line 1524: If some box content to be checked for marks contains any . . .

base/ltmarks.dtx
```diff
@@ -1248,17 +1510,28 @@
...
     The~mark~region~'#1'~contains~some~infinite~negative~glue~
-    allowing~it~to~shrink~to~an~arbitrary~size.~
+    allowing~it~to~shrink~to~an~arbitrarily~small~size.~
     This~makes~it~impossible~to~split~the~region~apart~to~
     get~at~its~marks.~They~are~lost.
   }
 %    \end{macrocode}
 %
+%    If box content checked for marks contains some infinite negative glue 
+%    allowing it to shrink to an arbitrarily small size we can't extract any marks.
```
car222222, line 1525: glue, allowing . . . size, then we . . .

base/ltmarks.dtx
```diff
@@ -1248,17 +1510,28 @@
...
+    allowing~it~to~shrink~to~an~arbitrarily~small~size.~
     This~makes~it~impossible~to~split~the~region~apart~to~
     get~at~its~marks.~They~are~lost.
   }
 %    \end{macrocode}
 %
+%    If box content checked for marks contains some infinite negative glue 
+%    allowing it to shrink to an arbitrarily small size we can't extract any marks. 
+%    They are lost, but this should not generate an error because in
+%    certain places, e.g., in some \env{minipage}s such glue might be legitimate.
```
car222222, line 1527: . . .  , such . . .

base/ltmarks.dtx
```diff
@@ -1370,6 +1643,24 @@
 %  \end{macro}
 %
 %
+%  \begin{macro}{\ShowMarksAt}
+%    Debugging helper that displays a snapshot of all known mark
+%    structures. The argument is a text string that this is
```
car222222, line 1648: . . . that is . . .

car222222 (approved):
Lots of commas (and a few "uncommas"), etc.

base/ltmarks.dtx
```diff
@@ -784,40 +807,47 @@
...
 %    is done by putting the material in a box and then doing a
 %    split operation to the maximum size possible (which hopefully
-%    means all of the content).\footnote{We could verify this, maybe we
-%    should.} Because this is an action only for the sake of getting
+%    means all of the content).\footnote{With normal column material cut
+%    from the main galley we should always get all material in one go, but
+%    in certain situations, for example,
+%    in a \pkg{multicols} environment that contains
+%    some \cs{columnbreak}s a single split operation will not be
+%    enough. Thus, this this is something we need to handle.}
```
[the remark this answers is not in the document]
FrankMittelbach: yeah, stuttering me ... there is also last/the/may elsewhere. I should really use Nelson's "dw" program regularly.

base/ltmarks.dtx
```diff
@@ -842,11 +872,34 @@
...
+%
+%
+%  \begin{macro}{@@_prepare_and_extract:nnn}
+%
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_prepare_and_extract:nnn #1#2#3 {
+%    \end{macrocode}
+%
+%    To handle the \cs{enlargethispage} case we do an \tn{unskip} to
+%    get rid of that glue if present and also check if we have then a
```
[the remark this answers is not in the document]
FrankMittelbach: "that" glue from \enlargethispage. But we are in fact removing "any" glue if present your wording is better

base/ltmarks.dtx
```diff
@@ -842,11 +872,34 @@
...
+%    \tn{vbox} as the last item and if so unpack that too, but only
+%    under certain conditions, see below. All this is temporary, just
+%    for getting the marks out, so it doesn't affect the final page
+%    production.
 %
 %    In fact, we go one step further and set the box to a large
```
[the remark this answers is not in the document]
FrankMittelbach: it is not really the largest (becaus then TeX has a problem as explained later, so I guess just drop the word "possible"

base/ltmarks.dtx
```diff
@@ -889,10 +942,11 @@
                                               % is not empty
                     }
 %    \end{macrocode}
-%    If not, then we unpack it, if yes we still ignore it for the process of
-%    mark extraction. We do not generate an error though, because in all
-%    likelihood this is an ordinary box like a marginal that does
-%    contain something like \tn{vss}.
+%    If we don't detect infinite negative glue we unpack the box. If
+%    yes we ignore it for the process of mark extraction. However, we
```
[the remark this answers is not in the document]
FrankMittelbach: it = the box
"If yes" should have been "If we do,"

base/ltmarks.dtx
FrankMittelbach: nothing "immediate" here.  The update_structure_from_material is usually called once in the OR (near the beginning) and then for the whole time tha OR is active you can retrieve the marks, e.g., while attaching headers and footers.

base/ltmarks.dtx
FrankMittelbach: how about this?
```
%    We have to be a little careful here: something like
%    \verb=\mark_insert:nn{foo}{}= adds an \enquote{empty} mark that
%    we should not confuse with the case where there is no mark at
%    all.
%
%    When there is no mark at all we get a truly empty
%    \cs{g_@@_tmp_tl} as a result. This is why we have to make sure
%    that an \enquote{empty} mark generated with \cs{mark_insert:nn}
%    only appears to be empty when it is typeset, but fails the next
%    test (see below how this is done).
```

car222222: Much better, thanks!

base/ltmarks.dtx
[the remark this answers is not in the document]
FrankMittelbach: hmm, back to the drawing board perhaps. If we run with \interactionmode 0, the error only appears in the log and we actually get to the marks even if there is real infinite negative glue (because TeX makes it finite).

In other words, if we accept that we may see
```
! Infinite glue shrinkage found in box being split.
\__box_dim_eval:n ...box_dim_eval:w #1\scan_stop: 
l. ...   }
The box you are \vsplitting contains some infinitely
shrinkable glue, e.g., `\vss' or `\vskip 0pt minus 1fil'.
Such glue doesn't belong there; but you can safely proceed,
since the offensive shrinkability has been made finite.
! Infinite glue shrinkage found in box being split.
```
in the log then we can get at the marks in all cases, which seems better then issuing an error or a warning that it doesn't work. 

If we want to we could additionally do
```
\wlog{Any immediately following 'Infinite glue shrinkage' error is harmless and can be ignored}
```
but I'm not sure this is better, because most of the time there will be none and we only end up with these extra lines. Thoughts?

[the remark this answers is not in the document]
davidcarlisle: You could use `\ignore_infinite_shrinkage_error  defined as `\let\ignore_infinite_shrinkage_error\__box_dim_eval:n ` perhaps....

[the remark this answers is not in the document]
davidcarlisle:
```

\expandafter\def\csname Ignore any infinite glue shrinkage error on this line\endcsname{
\setbox0\vbox{aaa \vss bbb}%
\setbox2=\vsplit0 to \maxdimen
}
\expandafter\def\expandafter\foo\expandafter{%
  \csname Ignore any infinite glue shrinkage error on this line\endcsname}
{
\interactionmode=0
\foo
}
```

makes a log

```
! Infinite glue shrinkage found in box being split.
\Ignore any infinite glue shrinkage error on this line ...
                                                  
l.13 \foo
```

FrankMittelbach:
> The test file `xmarks-009` fails showing this diff, I didn't update it, to make it easier to back out this commit if you don't like it, and to highlight the diff. It requires a bit of balancing to get the string at the right expansion level but it seems to work here without breaking the other ltmarks tests.

I like the idea and it makes the code much simpler --- and we get correct results even if there is shrinkage in the material which is great. I altered the txt a bit (not sure if it is for the better). Right now the test suite fails in a very strange way though and I don't understand why.

davidcarlisle:
> > The test file `xmarks-009` fails showing this diff, I didn't update it, to make it easier to back out this commit if you don't like it, and to highlight the diff. It requires a bit of balancing to get the string at the right expansion level but it seems to work here without breaking the other ltmarks tests.
> 
> I like the idea and it makes the code much simpler --- and we get correct results even if there is shrinkage in the material which is great. I altered the txt a bit (not sure if it is for the better). Right now the test suite fails in a very strange way though and I don't understand why.

yes lots of newlines in the raw log before the `[4`  page marker which looks like some extra empty `\write` but I don't see how that can be.  eg the raw log for tlb-dox002 has

```
[3


]) (./tlb-dox002.gls



[4



])
```

On the terminal it looks like

```
[3]) (./tlb-dox002.gls



[4]) (./tlb-dox002.aux)

```

so the extra newlines are only going to the log

FrankMittelbach:
> but I don't see how that can be. eg the raw log for tlb-dox002 has

neither can I but I really would like to understand before I update hundreds of test files

davidcarlisle:
@FrankMittelbach  turns out `\nonstopmode`  causes a newline in the log, if you keep all the code apart from this , the breaks go.

possibly from

```
@<Change the interaction...@>=
begin error_count:=0; interaction:=batch_mode+c-"Q";
print("OK, entering ");
case c of
"Q":begin print_esc("batchmode"); decr(selector);
  end;
"R":print_esc("nonstopmode");
"S":print_esc("scrollmode");
end; {there are no other cases}
print("..."); print_ln; update_terminal; return;
end

```

but my web is a bit rusty:-)

davidcarlisle:
@FrankMittelbach 

```

\wlog{===}
\wlog{===}
\nonstopmode
\interactionmode=1
\interactionmode=2
\wlog{===}

\bye
```

produces

```
\wlog{===}
\wlog{===}
\nonstopmode
\interactionmode=1
\interactionmode=2
\wlog{===}

\bye
```

but since the test suite  uses \scrollmode anyway an alternative plan to avoid changing the tests and adding newlines to end user log files  would be to skip saving and resting \interactionmode unless its current value is 3 (errorstopmode) I think any of the values 0,1,2  would work for this so we don't need to switch?

FrankMittelbach:
I guess I'm more for fixing the tlgs instead of doing something tailored to our test suite. But I have another issue. When I locally run l3build check it seems too loop after processing (or while?) github-0094 and I don't find what happens there. When I run 0094 by itself (or the one that I think comes after it all seems fine

davidcarlisle:
@FrankMittelbach 

If I run

` l3build check github-0094`

It stops in error stop mode at the terminal with

```
! Undefined control sequence.
\pdfstartlink ->\pdfextension 
                              startlink 
l.25 \leavevmode\pdfstartlink
                             
? 
```

FrankMittelbach:
and the same happens for  l3build check tltc001  with xetex (as if the interaction mode it not properly restored), now why is that?

davidcarlisle:
> and the same happens for l3build check tltc001 with xetex (as if the interaction mode it not properly restored), now why is that?

well why is it using luatex back end syntax with xetex

davidcarlisle:
me> well why is it using luatex back end syntax with xetex

because it's defined that way in the test file, this ought to be in a pdftex-luatex config, probably

FrankMittelbach:
All true but why is it not running scrollmode?

FrankMittelbach:
After the test suite issues have been identified and corrected, I think this is now ready to be merged. Open questions
- Should there be a rollback (would be tricky and I tend to think no)
- Is the error message ok, the way it is now? (there is not much space availabel, ie one has to stay with more or less that many chars)
- Is the news entry ok?

josephwright (approved):

davidcarlisle (approved):
looks good to me, I don't think it needs rollback,  I made a couple of minor tweaks to the news text. Error text is OK, I played around in my initial draft but getting the text just to appear in the right place doesn't leave a lot of room, as you noted Your version is fine.

Issue #1155, opened 2023-10-28
lthooks documentation


jlaurens:
I find the hooking support implemented in LaTeX extremely useful, much more than the documentation suggests.
In fact it was very hard for me to really understand this documentation.
I found typos and minor errors here and there and wanted to post a simple issue.
But the more I understood the concept, the more I found that the documentation does not play its role very well:
many terms are not defined explicitly, same terms are used with different meanings, information on the same matter is spread all over the text, some information is of limited interest.
For that reasons, I decided to do a bigger clean up and will post a pull request when done.

Two questions

1. Issues and PRs are not that practical to work on pdf documents. Exchanging by mail a pdf with annotations seems more efficient to me. Who would review such a doc?
2. I am facing a more important problem:
When \NewHook{foo} has not been executed yet, the documentation and the L3 messages say interchangeably that the hook

- does not exist
- is not created
- is not declared
- is undeclared

The problem is that "not declared" and "undeclared" are not synonyms, and the latter does not apply here because one
can declare hooks but cannot undeclare them. So I decided to consistently use the term "not declared" everywhere.
Unfortunately, "undeclared" also appears in the code implementation in various functions and variables that are not documented.
It makes sense to change _undeclared_ into_not_declared_.
If someone disagrees with that change, I will add instead a note in the implementation section to avoid a misunderstanding.
Any objection?

josephwright:
@FrankMittelbach and @PhelypeOleinik are the experts on the hook code and docs, but in terms of workflow the PDFs are just 'output': if you want to log comments or make a PR, this needs to be done at the level of the `.dtx` sources.

Issue #410, opened 2020-10-12
Please provide details/examples about hooks


dbitouze:
It would be nice if additional details/examples would be provided about hooks.

For instance, the `everypage`'s documentation says:

>Functionality similar to that provided by this package is directly implemented in LaTeX since its 2020 Fall release. Do not use `everypage` in new documents and do not rely on it in new packages or classes of yours. 

but section 2.4 "Emulating everypage" of `ltshipout`'s documentation doesn't help me to know how to do a simple:

```latex
\AddEverypageHook{
  \tikz[remember~picture,overlay] { ... ; }
}
```

I have read, in `ltfilehook`'s documentation, section 1.5 "High-level interfaces for LaTeX":
>We do not provide any high-level LaTeX commands (like `filehook` or `scrlfile` do) but think that for package writers the commands from for hook management are sufficient.

and that makes me sad :wink:

u-fischer:
> but section 2.4 "Emulating everypage" of ltshipout's documentation doesn't help me to know how to do a simple:

I would use the shipout/background or foreground hook to add pictures, for example 

```
\usepackage{tikzducks}
\AddToHook{shipout/background}{\put(0.5\paperwidth,-0.5\paperheight){\tikz[scale=3]\duck;}}
```

I don't really see a need for `remember picture` and `overlay`. 

> and that makes me sad

Why?

FrankMittelbach:
> I have read, in `ltfilehook`'s documentation, section 1.5 "High-level interfaces for LaTeX":
> 
> > We do not provide any high-level LaTeX commands (like `filehook` or `scrlfile` do) but think that for package writers the commands from for hook management are sufficient.
> 
> and that makes me sad 😉

I think that sentence is a little misleading. The new hook implementation provides a single and consistent set of high-level commands, e.g., `\AddToHook`and there is therefore no need to invent additional wrappers around each and every hook with names that often are rather arbitrary. What is the problem with writing in your package or class (or preamble)
```
\AddToHook{package/after/array}{ ... do something after array package was loaded ...}
```
This works consistently across all hooks and whether you have to remember the hook names (with consitent naming conventions) or a set of wrapper commands with sometimes strange names doesn't make much difference and I don't think it is anything to be sad about.

As to "please provide more examples" we agree with you that there is some potential for improvement and we plan at some point to write up an article for TUB on that subject. Anyway, please remember that we try to keep the issue list for latex2e as a list of "bugs" not as a more general discussion forum - thanks.

[...]

PR #1179, opened 2023-11-12
Update lthooks documentation

jlaurens:
This is a Draft PR open for discussion after a suggestion from Joseph Wright, this Draft PR should not be merged now.

Initially, I wanted to post issues but it rapidly appeared that there were faaaar too many of them.
Compared to interface3.pdf that just needs fine tuning, the actual documentation of lthooks really needs serious rework, at least.
Open issues are #1155, #410.

Typesetting `lthooks.dtx` with `pdflatex` gives a `pdf` file with many differences from the original, from pages 1 to 24.
This new documentation should really be read with clean new eyes starting with the TOC.

Improvements and corrections are welcome, the best way is to annotate and share a pdf.
It is possible to [download](https://filesender.renater.fr/?s=download&token=a26821f2-1ec0-46e9-a9aa-8a69c05c9179) the pdf until 2023/11/27

This documentation needs only one thing: support for RE instead of raw labels in rules.

The implementation has not been changed, this is for the next step, if any.


josephwright:
We notice your putting in quite a bit of effort here, but we are concerned that this might not be productive. As it stands, it's not really possible to review the PR: picking out what has changed from the original, and why you'd like to change it, isn't workable. Do you plan to take your changes and split into several commits each with a defined 'target'?

FrankMittelbach:
If I may add: any individual suggested change/improved  should nevertheless stick to the documentation conventions used across all of the LaTeX kernel, i.e., please do not invent your own documentation syntax. It might well be  the better one, no doubt, but it will be unmanageable for us if a single files differ in their conventions.

josephwright:
We seem to now have no changes at all here: I'm closing as we could to do keep on top of PRs.

PR #1181, opened 2023-11-15
Add optional argument to \listfiles

josephwright:

# Internal housekeeping

## Status of pull request

- [x] Feedback wanted 
- Under development
- Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/doc/ltnews39.tex
+++ b/base/doc/ltnews39.tex
@@ -194,7 +194,13 @@ related to variation in package versions. However, this has to date relied on
 the information in the \cs{ProvidesPackage} line, or similar: that can be
 misleading if for example a file has been edited locally. We have now extended
 \cs{listfiles} to take an optional argument which will then include the MD5
-hash of each file (and the size of each file) in the \texttt{.log}.
+hash of each file (and the size of each file) in the \texttt{.log}. Thus for
+example you can use
+\begin{verbatim}
+\listfiles[hashes,sizes]
+\end{verbatim}
+to get both the file sizes and file hashes in the \texttt{.log} as well as
+the standard release information.
 %
 \githubissue{945}
 
--- a/base/doc/ltnews39.tex
+++ b/base/doc/ltnews39.tex
@@ -187,7 +187,16 @@ defined it is not altered, but instead a warning message is displayed.
 %
 \githubissue{823}
 
+\subsection{Extended information in \cs{listfiles}}
 
+The \cs{listfiles} command provides useful information when finding issues
+related to variation in package versions. However, this has to date relied on
+the information in the \cs{ProvidesPackage} line, or similar: that can be
+misleading if for example a file has been edited locally. We have now extended
+\cs{listfiles} to take an optional argument which will then include the MD5
+hash of each file (and the size of each file) in the \texttt{.log}.
+%
+\githubissue{945}
 
 %\section{Bug fixes}
 
--- a/base/doc/usrguide.tex
+++ b/base/doc/usrguide.tex
@@ -1258,4 +1258,39 @@ between dotted- and dotless-i, one could use for example
   |\DeclareUppercaseMapping[xx]{"0131}{I}|
 \end{quote}
 
+\section{Support for problem solving}
+
+\begin{decl}
+  |\listfiles| \oarg{options}
+\end{decl}
+
+If this command is placed in the preamble then a list of the files
+read in (as a result of processing the document) will be displayed
+on the terminal (and in the log file) at the end of the run. Where
+possible, a short description will also be produced. These descriptions
+will (hopefully) include the descriptions, dates and version numbers
+for package and class files.
+
+Sometimes, if may be that a local edit has been made to a package or
+class file (or rather a copy of such a file). To allow these cases to
+be identified, \cs{listfiles} takes an optional argument which allows
+adjustment of the information printed using a key--value approach
+\begin{description}
+  \item[\texttt{hashes}] Adds the MD5 hash for each file to the
+    information printed
+  \item[\texttt{sizes}] Adds the file size for each file to the
+    information printed
+\end{description}
+Note that the difference in line ends used by Windows and Unix
+systems \emph{does} show up here: as such, you should compare
+results from computers using the same operating system.
+
+\emph{Warning}: this command will list only files which were read
+using \LaTeX{} commands such as |\input|\arg{file} or
+|\include|\arg{file}.  If the file was read using the primitive \TeX{}
+syntax |\input |\emph{file} (without |{ }| braces around the file name)
+then it will not be listed; failure to use the \LaTeX{} form with the
+braces can cause more severe problems, possibly leading to overwriting
+important files, so \textbf{always put in the braces}.
+
 \end{document}
--- a/base/doc/usrguide.tex
+++ b/base/doc/usrguide.tex
@@ -1281,9 +1281,10 @@ adjustment of the information printed using a key--value approach
   \item[\texttt{sizes}] Adds the file size for each file to the
     information printed
 \end{description}
-Note that the difference in line ends used by Windows and Unix
-systems \emph{does} show up here: as such, you should compare
-results from computers using the same operating system.
+Note that as Windows and Unix use different line endings (LF \emph{versus} LF
+CR), the hashes and file sizes from the two systems will not be the same. As
+such, you should compare these values between operating systems of the same
+type.
 
 \emph{Warning}: this command will list only files which were read
 using \LaTeX{} commands such as |\input|\arg{file} or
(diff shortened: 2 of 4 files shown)


josephwright:
Making a start here: would welcome idea on what @FrankMittelbach was aiming for with `names` ...

base/doc/usrguide.tex
```diff
@@ -1258,4 +1258,39 @@ \section{Case changing}
...
+\end{description}
+Note that the difference in line ends used by Windows and Unix
+systems \emph{does} show up here: as such, you should compare
+results from computers using the same operating system.
+
+\emph{Warning}: this command will list only files which were read
+using \LaTeX{} commands such as |\input|\arg{file} or
+|\include|\arg{file}.  If the file was read using the primitive \TeX{}
+syntax |\input |\emph{file} (without |{ }| braces around the file name)
+then it will not be listed; failure to use the \LaTeX{} form with the
```
muzimuzhi, line 1292: Starting from TeX Live 2020, `\input` primitive "also accepts a group-delimited filename argument".
https://tug.org/texlive/doc/texlive-en/texlive-en.html#x1-890009.1.17

josephwright: Sure, but in LaTeX a braced argument will always go _via_ the LaTeX mech: the point is that anything which doesn't use braces is a problem.

base/doc/ltnews39.tex
```diff
@@ -190,7 +190,16 @@ \subsection{Warn if counter names are problematic}
...
 
-
+\subsection{Extended information in \cs{listfiles}}
+
+The \cs{listfiles} command provides useful information when finding issues
+related to variation in package versions. However, this has to date relied on
+the information in the \cs{ProvidesPackage} line, or similar: that can be
+misleading if for example a file has been edited locally. We have now extended
+\cs{listfiles} to take an optional argument which will then include the MD5
+hash of each file (and the size of each file) in the \texttt{.log}.
```
FrankMittelbach, line 200: I think this should give an example, such as "So instead of \listfiles you can say \listfiles[hashes,sizes] to get the full information."

base/doc/usrguide.tex
```diff
@@ -1258,4 +1258,39 @@ \section{Case changing}
...
+
+Sometimes, if may be that a local edit has been made to a package or
+class file (or rather a copy of such a file). To allow these cases to
+be identified, \cs{listfiles} takes an optional argument which allows
+adjustment of the information printed using a key--value approach
+\begin{description}
+  \item[\texttt{hashes}] Adds the MD5 hash for each file to the
+    information printed
+  \item[\texttt{sizes}] Adds the file size for each file to the
+    information printed
```
FrankMittelbach, line 1282: There was the suggestion to also offer "paths". I have my doubts that it is that helpful, because if somebody sends me his output that includes full paths of filename then I can't do a simple diff to what I get. So for that use case it would be counterproductive I think. But I can see that locally it might be helpful, so I'm not sure if that should be included.

I suggested "titles=false" as a further possibility, but thinking about it I feel this only make things more complex without any real gain, so I guess I take that back.

base/doc/usrguide.tex
```diff
@@ -1258,4 +1258,39 @@ \section{Case changing}
...
+\end{description}
+Note that the difference in line ends used by Windows and Unix
+systems \emph{does} show up here: as such, you should compare
+results from computers using the same operating system.
```
FrankMittelbach, line 1286: is that really an issue, I mean if somebody sends me such a listing from a Windows machine, would I have problems to compare it with my OSX output after copy-paste?

I mean if I get the full log then this will not be something one can compare across OSes anyway, because all the paths are likely to be different, but for something put into a github issue or send by mail ... ?

So I think the "show up here" needs some qualifying or perhaps the whole para is not needed because all of this was true also for the original \listfiles

FrankMittelbach (approved):
with a few comments/suggestions

base/doc/usrguide.tex
josephwright: I've revised the text: the impact is that you have a different number of bytes and thus a different MD5 hash based on LF vs. LF CR.

base/doc/ltnews39.tex
josephwright: Adjusted

base/doc/usrguide.tex
josephwright: The alternative would be to read the files to a token list then hash that: TeX then deals with the system-dependence. But that will be slower ..

FrankMittelbach: ah, that is more understandable that the sizes and the hashes differ in that case, so "here" is in the hash and size result :-)

base/doc/usrguide.tex
FrankMittelbach: I think it is fine as it is given what you want to convey

base/doc/usrguide.tex
muzimuzhi: To indicate the source of hash and size differences, maybe the current platform (`\c_sys_platform_str`) can be logged too, when one or more of "hashes" and "sizes" are specified.

Or even a full sentence,
```tex
 *File List*
    (Current platform is "unix". Note the sizes and hashed may differ between platforms due to different default line ends.)
 article.cls    2023/05/17 v1.4n Standard LaTeX document class
    (size: 20144, hash: FBCF51733730D8F1A62143512FF5B1FE)
  size10.clo    2023/05/17 v1.4n Standard LaTeX file (size option)
    (size: 8448, hash: 74078C4A887F2CDCD57C3EC111622740)
l3backend-luatex.def    2023-11-04 L3 backend support: PDF output (LuaTeX)
    (size: 30409, hash: 4F844A9024E5A9B01FC5871F268DC768)
  ts1cmr.fd    2023/04/13 v2.5m Standard LaTeX font definitions
    (size: 2430, hash: A2E8A746C2C6DDF1DF1FB6AD609D5F6A)
 ***********
```

base/doc/usrguide.tex
```diff
@@ -1258,4 +1258,40 @@ \section{Case changing}
...
+\end{decl}
+
+If this command is placed in the preamble then a list of the files
+read in (as a result of processing the document) will be displayed
+on the terminal (and in the log file) at the end of the run. Where
+possible, a short description will also be produced. These descriptions
+will (hopefully) include the descriptions, dates and version numbers
+for package and class files.
+
+Sometimes, if may be that a local edit has been made to a package or
```
muzimuzhi, line 1274: typo? "if may be that" -> "it may be that"

```suggestion
Sometimes, it may be that a local edit has been made to a package or
```

base/ltfiles.dtx
```diff
@@ -1693,8 +1709,10 @@
 %         {Stop \cs{listfiles} being run twice}
 % \changes{v1.0i}{1994/10/18}
 %         {code moved here from ltclass}
+% \changes{v1.2v}{2023/11/15}{Extend file list information}
 %    \begin{macrocode}
-\def\listfiles{%
+\NewDocumentCommand\listfiles{O{}}{%
+  \SetKeys[__kernel/listfiles]{#1}%
```
muzimuzhi, line 1715: Do the two booleans need to be initialized in `\listfiles`?

This only makes sense if multiple `\listfiles` with different options were used, which is too rare a real case.

base/doc/usrguide.tex
josephwright: We can revisit this if there is a clear benefit, I guess: that's why I went keyval rather than a simple `\listfilehashes`

muzimuzhi (commented):
A possible typo and an overthinking suggestion about option initialization.

base/doc/usrguide.tex
josephwright: You have to watch it a bit: you _can_ have files with different line endings to the system default and they will work (as TeX normalises), so if someone has e.g. something by email then it will still work ...

base/ltfiles.dtx
josephwright: Suggested alternative?

base/doc/usrguide.tex
muzimuzhi: Oh, you're right.

base/ltfiles.dtx
muzimuzhi:
```suggestion
  \@listfiles@sizesfalse
  \@listfiles@hashesfalse
  \SetKeys[__kernel/listfiles]{#1}%
```
or
```suggestion
  \SetKeys[__kernel/listfiles]{sizes=false,hashes=false,#1}%
```

josephwright: Switches start off false, and these are internal - I really don't think we need to worry about resetting them.

josephwright:
I think this one looks safe - will merge

PR #1182, opened 2023-11-15
Fix error message for para/begin hook

viccie30:
The error message is correctly issued if the hook switches out of horizontal mode, but the message itself says that the hook switched out of vertical mode instead.

I can send the one-word patch by e-mail as well if that's easier.

### DIFF
--- a/base/ltpara.dtx
+++ b/base/ltpara.dtx
@@ -813,7 +813,7 @@
 %    If we aren't in horizontal mode any longer the hooks above misbehaved.
 %    \begin{macrocode}
   \if_mode_horizontal: \else:
-    \msg_error:nnnn { hooks }{ para-mode }{begin}{vertical} \fi:
+    \msg_error:nnnn { hooks }{ para-mode }{begin}{horizontal} \fi:
 %    \end{macrocode}
 %    Finally we reinsert the indentation box (unless suppressed) and
 %    then call \cs{everypar} the way legacy \LaTeX\ code expects it.


FrankMittelbach:
thanks for the pr.  There is unfortunately more to do even for one line fixes, e.g., making a test file for the issue (usually, not this time), updating the change log, updating the file date, updating the file version (if code is changed), etc. It is therefore usually easier if you just make a small test file reporting the issue (maybe together with a suggestion for a fix) and let us do the rest.

For example, your change fails the test suite in ltpara-002.lvt (because that "incorrectly" claims that the curent code is correct --- in other words I tested but overlooked the problem in the results).

It is possible to do all this in a pr (and some people do) but you would then need to be familiar with the test suite LaTeX is using etc. This is why we say pull requests are normally for the team --- anyway take a look at branch [gh1182](https://github.com/latex3/latex2e/tree/gh1182) to see what I did (so far) to integrate your correction. I close here and fix in the other branch.

viccie30:
Thanks for fixing the issue so quickly. I'll send an e-mail if I ever find a bug in the future.

FrankMittelbach:
> Thanks for fixing the issue so quickly. I'll send an e-mail if I ever find a bug in the future.

no please don't!  If you find an issue, make an issue at https://github.com/latex3/latex2e/issues instead of a pull and follow the guidelines there. It may not be me who is going to fix it after all and emails tend to get burried and lost. Thanks - Frank

Issue #1183, opened 2023-11-19
unknown-keyhandler disables removing from unused options


Skillmon:
## Brief outline of the bug

If one activates an `unknown` handler in an `l3keys`-set the new option handler does no longer remove known keys from the list of unused options if used inside a class (in a package the handling seems fine).

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\begin{filecontents}[overwrite]{\jobname.cls}
\NeedsTeXFormat{LaTeX2e}
\ProvidesClass{\jobname}
\LoadClassWithOptions{article}

\DeclareKeys[testwork]{%
  test .code  = \newcommand{\foo}{#1},
  test .usage = load,
  % the following three lines give an equivalent definition of your bertha
  % option as defined with `\DeclareOption{bertha}{}`
  bertha .code = {},
  bertha .value_forbidden:n = true,
  bertha .usage = load,
  % if you use this unknown handler, you get "test" and "bertha" as unknown
  % options, if you omit it you get "a4paper" as unknown option
  unknown .code = {}
}
\ProcessKeyOptions[testwork]

\endinput
\end{filecontents}
\documentclass[a4paper,test=wtf,bertha]{\jobname}

\begin{document}
test \foo
\end{document}
```

## Log file (required) and possibly PDF file  

[mwe_unknownbug.log](https://github.com/latex3/latex2e/files/13404793/mwe_unknownbug.log)

muzimuzhi:
Caused by this logic, which was copied from `l3keys2e`. See the [similar logic in `l3keys2e.dtx`](https://github.com/latex3/latex3/blob/e1fa5022262f9936de44352b047dfb17f891ff43/l3packages/l3keys2e/l3keys2e.dtx#L247-L274).
https://github.com/latex3/latex2e/blob/5064cfc98a1b3a2dd01bfb45709974e08a8a9972/base/ltkeys.dtx#L301-L326

muzimuzhi:
It seems the cause is, compared to `\__keys_options_class:nnn`, currently in `\__keys_options_class:n`, when `unknown` key exists, raw options are only appended to `\l_@@_options_clist`, but not removed from `\@unusedoptionlist`.

https://github.com/latex3/latex2e/blob/5064cfc98a1b3a2dd01bfb45709974e08a8a9972/base/ltkeys.dtx#L308-L338

```tex
[...]

PR #1184, opened 2023-11-20
Add options to unused list when unknown key is active

josephwright:


base/testfiles/github-1183.lvt
```diff
@@ -0,0 +1,28 @@
...
+  bertha .usage = load,
+  % if you use this unknown handler, you get "test" and "bertha" as unknown
+  % options, if you omit it you get "a4paper" as unknown option
+  unknown .code = {}
+}
+\ProcessKeyOptions[testwork]
+
+\endinput
+\end{filecontents}
+\documentclass[a4paper,test=wtf,bertha]{github-1183}
```
muzimuzhi, line 24: Use an extra option which is unknown to both `article` and `githu-1183` classes, to trigger the "Unused global option(s)" waning?

FrankMittelbach: From a conceptual point I think that global options should be handled as follows:
- if a package uses a global option it is marked as used (but still offered to other packages as a global option)
- if no package uses a global option there should be an "unused global option warning"

The big question is what does this mean if a a package defines an unknown option handler? 
- Given that for the default unknown option handler (raising an unknown option error) global options are ignored one can argue that the same should happen if a different unknown option handler is implemented 
- Another possibility is to pass global options to an unknown option handler (if it is not the default) and consider them used if that happens (that means that with one such package we would automatically use all global options and there is no warning ever)
- Finally one could implement some sort of a mix of the two by offering an interface to be used in the unknown option handler to explicitly mark a global option as used if it is passed through and is handled (not sure how that should look like exactly)

muzimuzhi: > We still shouldn't see a warning?

@josephwright I'm not sure which behavior is expected (#1183 has some discussions on it), but with current proposed change, if `aaapaper` is used as class option, it will remain in unused options list. See https://github.com/muzimuzhi/latex2e/commit/587588348942bd38a77fdcf20eb4e836cda38bed.

Issue #1185, opened 2023-11-21
error if packages are loaded in a group


u-fischer:
## Brief outline of the bug

Packages should not be loaded in a group as this can lead to misbeheaviour that can be hard to detect or understand. As an example hyperref fails with an internal socket error if a group has been opened by mistake: https://github.com/latex3/hyperref/issues/315.

\usepackage and \RequirePackage should check the group level and error directly.

clason:
Sorry to be "that guy", but this broke font packages that conditionally require other packages based on package options (e.g., `notomath`). Is this known, and are there mitigation strategies?

u-fischer:
@clason do you have an example where notomath breaks? I don't a package loaded in a group there.

clason:
```tex
\documentclass{article}
\usepackage[sfdefault]{notomath}
\begin{document}
Test
\end{document}
```

(The `sfdefault` option is what triggers the error.)

clason:
Seems to be this line in `notomath.sty` that is the culprit:
```tex
\ifnoto@sfdefault{\noto@MO{notosans}\else\noto@MO{noto}\fi
```

I don't quite see why, but commenting it out removes the error. (Voodoo debugging at its finest!)

clason:
Oh, of course, that's a typo (missing brace closing the first clause). 🤦 

Thanks for rubber-ducking, and sorry for wasting your time! Looks like this PR is doing its work already ;) I'll notify the maintainer.

u-fischer:
well that is a clear bug in the pakcage. The stray brace after `\ifnoto@sfdefault` shouldn't be there at all, and in earlier versions it would have lead to `### simple group (level 1) entered at line 54 ({)`. Notify the author.

PR #1187, opened 2023-11-26
add error test for #1185

davidcarlisle:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [X] Test file(s) added


josephwright (changes requested):
Do we need rollback here?

FrankMittelbach:
> Do we need rollback here?

I think yes, there may be some packages out that do by mistake load stuff in a group and until they get fixed a rollback will help users.

davidcarlisle:
so m3expl009 does

```
    \group_begin:
      #1
      \RequirePackage{expl3}
    \group_end:
```

and now triggers an error, do we pass that, or change the test or ???

FrankMittelbach:
fix I would suggest

base/doc/ltnews39.tex
```diff
@@ -151,6 +151,13 @@ \section{New or improved commands}
 
 \section{Code improvements}
 
+\subsection{Loading packages at the top level}
+Packages should only be loaded with \cs{usepackage} or \cs{RequirePackage}
+at the top level, not inside a group. Previously \LaTeX\ did not check this,
```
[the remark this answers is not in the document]
davidcarlisle: sigh, I added some permutation of the words I intended

base/doc/ltnews39.tex
```diff
@@ -151,6 +151,16 @@ \section{New or improved commands}
...
+which would often lead to low level errors later on if package declarations
+were reverted as a group ended.
+\LaTeX\ now checks the group level and an error is thrown
+if the package is loaded in a group.
+%
+\githubissue{549}
```
muzimuzhi, line 162:
```suggestion
\githubissue{1185}
```

base/doc/ltnews39.tex
```diff
@@ -151,6 +151,16 @@ \section{New or improved commands}
 
 \section{Code improvements}
 
+\subsection{Loading packages at the top level}
+Packages should only be loaded with \cs{usepackage} or \cs{RequirePackage}
```
muzimuzhi, line 155: Mind referring to `\RequirePackageWithOptions` as well?

base/ltclass.dtx
```diff
@@ -1752,10 +1752,30 @@
...
 %    \begin{macrocode}
+%</2ekernel>
+%<latexrelease>\IncludeInRelease{2024/06/01}%
+%<latexrelease>                 {\RequirePackage}{Check Group}%
+%<*2ekernel|latexrelease>
 \def\RequirePackage{%
+  \ifnum\currentgrouplevel>\z@
+     \@latex@error
+      {\noexpand\RequirePackage or \noexpand\usepackage in a group}%
+      {\noexpand\RequirePackage or \noexpand\usepackage should be at the top level}%
```
muzimuzhi, line 1766: `\noexpand\cmd` will leave an extra space after `\cmd`, `\string` won't.

If `\RequirePackageWithOptions` is needed too, perhaps
```suggestion
     \@latex@error
      {Package Loaded in a group}%
      {\string\usepackage, \string\RequirePackage, or
       \string\RequirePackageWithOptions \MessageBreak
       should be at the top level}%
```

base/testfiles/m3expl009.lvt
```diff
@@ -22,6 +22,7 @@
     \ExplSyntaxOff
     \cs_undefine:c { ver@expl3.sty }
     \group_begin:
+       \cs_set_eq:Nc\currentgrouplevel{z@}
```
muzimuzhi, line 25: expl3 loves spaces 😄
```suggestion
       \cs_set_eq:Nc \currentgrouplevel {z@}
```

muzimuzhi (changes requested):
Typos like wrong github issue number in ltnews, and whether to mention `\RequirePackageWithOptions`.

Does `\LoadClass` need a similar check?

base/doc/ltnews39.tex
FrankMittelbach: could be I guess

base/ltclass.dtx
FrankMittelbach: but you wouldn't want comma before "or" I guess

base/testfiles/m3expl009.lvt
FrankMittelbach: :-) yes

FrankMittelbach (approved):
I guess we should do the same to LoadClass (as suggested elswhere by @muzimuzhi) and yes we should probably mention all the relevant commands in the error message I guess.

Otherwise I think this is good to go

base/ltclass.dtx
davidcarlisle: I'd probably keep the error message shorter (but use \string)

base/doc/ltnews39.tex
davidcarlisle: added

davidcarlisle:
> I guess we should do the same to LoadClass (as suggested elswhere by @muzimuzhi) and yes we should probably mention all the relevant commands in the error message I guess.
> 
> Otherwise I think this is good to go

I suppose so although main use case  is end users calling packages in a group so `\usepackage` but I guess doing LoadClass as well would be more consistent

davidcarlisle:
If we want to catch `\documentclass`/`\LoadClass(WithOptions)` as well as `\usepackage` then rather than duplicate the code I suggest pushing it down to the shared code, and using a more generic message, something like

```
\def\@fileswithoptions#1{%
    \ifnum\currentgrouplevel>\z@
     \@latex@error
      {Loading a class or package in a group}%
      {Classes and packages should only be loaded be at the top level}%
  \fi
  \@ifnextchar[%]
    {\@fileswith@ptions#1}%
    {\@fileswith@ptions#1[]}}
```
instead of the current change to \RequirePackage`

and adjusting the rollback to match.  

@FrankMittelbach  @muzimuzhi  ?

FrankMittelbach:
> If we want to catch `\documentclass`/`\LoadClass(WithOptions)` as well as `\usepackage` then rather than duplicate the code I suggest pushing it down to the shared code, and using a more generic message, something like
> 
> ```
> \def\@fileswithoptions#1{%
>     \ifnum\currentgrouplevel>\z@
>      \@latex@error
>       {Loading a class or package in a group}%
>       {Classes and packages should only be loaded be at the top level}%
>   \fi
>   \@ifnextchar[%]
>     {\@fileswith@ptions#1}%
>     {\@fileswith@ptions#1[]}}
> ```
> 
> instead of the current change to \RequirePackage`
> 
> and adjusting the rollback to match.
> 
> @FrankMittelbach @muzimuzhi ?

sounds good to me. It is not more complicated but covers the possible mistakes better.

base/doc/ltnews39.tex
```diff
@@ -233,6 +233,17 @@ \subsection{pkg{doc}: better support for \pkg{upquote}}
 
 \section{Code improvements}
 
+\subsection{Loading packages at the top level}
+Classes and packages should only be loaded with \cs{documentclass}, \cs{usepackage},
+or class interface commands such as `\cs{LoadClass} or `\cs{RequirePackageWithOptions}
```
muzimuzhi, line 238:
```suggestion
or class interface commands such as \cs{LoadClass} or \cs{RequirePackageWithOptions}
```

base/doc/ltnews39.tex
```diff
@@ -233,6 +233,17 @@ \subsection{pkg{doc}: better support for \pkg{upquote}}
...
+at the top level, not inside a group. Previously \LaTeX\ did not check this,
+which would often lead to low level errors later on if package declarations
+were reverted as a group ended.
+\LaTeX\ now checks the group level and an error is thrown
+if the package is loaded in a group.
```
muzimuzhi, line 243:
```suggestion
if the class or package is loaded in a group.
```

muzimuzhi (changes requested):
Typo and minor improvement in ltnews.

base/doc/ltnews39.tex
davidcarlisle: oops stack exchange markdown habits, thanks

base/changes.txt
```diff
@@ -58,6 +60,9 @@ not part of the distribution.
 	* usrguide.tex (subsection{Performance}):
 	Document efficiency of ltcmd definitions
 
+2023-11-26  David Carlisle  <David.Carlisle@latex-project.org>
+	* ltclass.dtx: check that \RequirePackage is at top level (gh/1185)
```
FrankMittelbach, line 64: and similar commands (or list them)

base/changes.txt
```diff
@@ -58,6 +60,9 @@ not part of the distribution.
 	* usrguide.tex (subsection{Performance}):
 	Document efficiency of ltcmd definitions
 
+2023-11-26  David Carlisle  <David.Carlisle@latex-project.org>
```
FrankMittelbach, line 63: I would change the date to today (and also elsewhere) because it is kind of weird if an update shows up much earlier in the list compared to it actually showing up in the -dev format

base/ltclass.dtx
```diff
@@ -33,7 +33,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{ltclass.dtx}
-             [2023/04/14 v1.5h LaTeX Kernel (Class & Package Interface)]
+             [2023/11/26 v1.5i LaTeX Kernel (Class & Package Interface)]
```
FrankMittelbach, line 36: see above

base/ltclass.dtx
```diff
@@ -1950,12 +1950,34 @@
 % \end{macro}
 %
 % \begin{macro}{\@fileswithoptions}
+% \changes{v1.5i}{2023/11/26}{Test group level}
```
FrankMittelbach, line 1953: ditto

FrankMittelbach (approved):
looks good to me (with a few suggestions for change)

base/changes.txt
FrankMittelbach: remark: in other cases (like the pr I did yester which was also stale) there have been intermediate changes so date and/or version number were already updated and in that cases it becomes really confusing to keep the original text.

PR #1188, opened 2023-11-28
Promote `\NewDocumentCommand` in documentations

muzimuzhi:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [ ] Test file(s) added
- [x] <s>Version and</s> date string updated in changed source files

### DIFF
--- a/base/doc/usrguide.tex
+++ b/base/doc/usrguide.tex
@@ -43,7 +43,7 @@
     \texttt{usrguide.tex} for full details.}%
 }
 
-\date{2023-11-09}
+\date{2023-11-28}
 
 \NewDocumentCommand\cs{m}{\texttt{\textbackslash\detokenize{#1}}}
 \NewDocumentCommand\marg{m}{\arg{#1}}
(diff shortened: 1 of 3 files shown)


josephwright:
`changes.txt` entries?

base/doc/usrguide.tex
```diff
@@ -1161,7 +1161,7 @@ \section{Expandable floating point (and other) calculations}
 above. Nevertheless, in practice they are usually sufficient.  For
 example
 \begin{verbatim}
-\newcommand\calculateheight[1]{%
+\NewDocumentCommand\calculateheight{m}{%
```
FrankMittelbach, line 1164: My personal opinion is that as long as the "m" only case of \NewDocumentCommand isn't streamlined \newcommand does have its place (also when expansion is wanted) because the \NewDocumentMachinery is rather heavy when ending up with essentialls \protected\def in the end.  But this is kind of a different topic and fine with promoting \NewDocumentCommand here.

FrankMittelbach (approved):
I don't really think it warrants a changes entry (we usually don't do them for just doc changes).

base/doc/usrguide.tex
josephwright: @FrankMittelbach We could bring back shortcutting for all-`m` arguments

FrankMittelbach: did we ever had it, thought we only talked about it? But yes, I think it makes a lot of sense because it cuts down on the run time costs a lot moving them to the declaration.

josephwright: > did we ever had it, thought we only talked about it? But yes, I think it makes a lot of sense because it cuts down on the run time costs a lot moving them to the declaration.

We had something: I'll log an issue and put in the pool for 2024-06-01.

josephwright: Logged as https://github.com/latex3/latex2e/issues/1189

josephwright (approved):

PR #1013, opened 2023-03-11
Distinguish protected ltcmds with only m-type args

muzimuzhi:
**READ ME FIRST**: Please understand that in most cases we will not be able to merge a pull request because there are a lot of internal activities needed when updating the LaTeX2e sources. If you have a code suggestion please discuss it with the team first.

*Pull requests in this repository are intended for LaTeX Team members only.*

# Internal housekeeping

This PR resolves issue #1009.

PS: What's the main difference between `test2e.tex` and `regression-test.tex`? It's observed that all `testfiles-ltcmd/github-xxxx.lvt` load the former and `testfiles-ltcmd/ltcmdxxx.lvt` loads the latter, but usages are not that consistent in other test suites.

## Status of pull request

- Feedback wanted 
- Ready to merge

## Checklist of required changes before merge will be approved
- [x] Test file(s) added
- [x] Version and date string updated in changed source files
- [x] Relevant `\changes` entries in source included
- [x] Relevant `changes.txt` updated
- [x] `ltnewsX.tex` (and/or `latexchanges.tex`) updated

PR #1190, opened 2023-11-28
Optimise creation of document commands with m-type args

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [x] Test file(s) added

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,6 +6,10 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2023-11-28  Joseph Wright  <Joseph.Wright@latex-project.org>
+	* ltcmd.dtx (subsection{Declaring commands and environments}):
+	Optimise creation of simple document commands (gh/1189)
+
 2023-11-16  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
 	* ltpara.dtx (subsection{Providing hooks for paragraphs}):
 	Correct error message: hook left horizontal not vertical mode (gh/1182)
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,7 +6,7 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
-2023-11-29  Joseph Wright  <Joseph.Wright@latex-project.org>
+2023-11-30  Joseph Wright  <Joseph.Wright@latex-project.org>
 	* ltcmd.dtx (subsection{Declaring commands and environments}):
 	Optimize creation of simple document commands (gh/1189)
 	* usrguide.tex (subsection{Performance}):
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,7 +6,7 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
-2023-11-30  Joseph Wright  <Joseph.Wright@latex-project.org>
+2023-12-01  Joseph Wright  <Joseph.Wright@latex-project.org>
 	* ltcmd.dtx (subsection{Declaring commands and environments}):
 	Optimize creation of simple document commands (gh/1189)
 	* usrguide.tex (subsection{Performance}):
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,7 +6,7 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
-2023-11-28  Joseph Wright  <Joseph.Wright@latex-project.org>
+2023-11-29  Joseph Wright  <Joseph.Wright@latex-project.org>
 	* ltcmd.dtx (subsection{Declaring commands and environments}):
 	Optimize creation of simple document commands (gh/1189)
 	* usrguide.tex (subsection{Fully-expandable document commands}):
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -9,7 +9,7 @@ not part of the distribution.
 2023-11-29  Joseph Wright  <Joseph.Wright@latex-project.org>
 	* ltcmd.dtx (subsection{Declaring commands and environments}):
 	Optimize creation of simple document commands (gh/1189)
-	* usrguide.tex (subsection{Fully-expandable document commands}):
+	* usrguide.tex (subsection{Performance}):
 	Document efficiency of ltcmd definitions
 
 2023-11-16  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -8,7 +8,9 @@ not part of the distribution.
 
 2023-11-28  Joseph Wright  <Joseph.Wright@latex-project.org>
 	* ltcmd.dtx (subsection{Declaring commands and environments}):
-	Optimise creation of simple document commands (gh/1189)
+	Optimize creation of simple document commands (gh/1189)
+	* usrguide.tex (subsection{Fully-expandable document commands}):
+	Document efficiency of ltcmd definitions
 
 2023-11-16  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
 	* ltpara.dtx (subsection{Providing hooks for paragraphs}):
--- a/base/doc/ltnews39.tex
+++ b/base/doc/ltnews39.tex
@@ -210,10 +210,10 @@ Creating document commands using \cs{NewDocumentCommand}, etc., provides a very
 flexible way of grabbing arguments. When the document command only takes simple
 mandatory arguments, this has to-date added an overhead that could be avoided.
 We have now refined the internal code path such that \enquote{simple} document
-commands avoid any overhead at point-of-use, making the results as efficient as
-using \cs{newcommand} for low-level \TeX{} constructs. Note that
-as \cs{NewDocumentCommand} makes engine-robust commands, the direct
-equivalent to \cs{newcommand} is \cs{NewExpandableDocumentCommand}.
+commands avoid almost any overhead at point-of-use, making the results
+essentially as efficient as using \cs{newcommand} for low-level \TeX{}
+constructs. Note that as \cs{NewDocumentCommand} makes engine-robust commands,
+the direct equivalent to \cs{newcommand} is \cs{NewExpandableDocumentCommand}.
 %
 \githubissue{1189}
 
--- a/base/doc/ltnews39.tex
+++ b/base/doc/ltnews39.tex
@@ -204,6 +204,18 @@ the standard release information.
 %
 \githubissue{945}
 
+\subsection{Optimise creation of simple document commands}
+
+Creating document commands using \cs{NewDocumentCommand}, etc., provides a very
+flexible way of grabbing arguments. When the document command only takes simple
+mandatory arguments, this has to-date added an overhead that could be avoided.
+We have now refined the internal code path such that \enquote{simple} document
+commands avoid any overhead at point-of-use, making the results as efficient as
+using \cs{newcommand} or low-level \TeX{} constructs.
+%
+\githubissue{1189}
+
+
 %\section{Bug fixes}
 
 %\section{Changes to packages in the \pkg{amsmath} category}
--- a/base/doc/ltnews39.tex
+++ b/base/doc/ltnews39.tex
@@ -211,7 +211,9 @@ flexible way of grabbing arguments. When the document command only takes simple
 mandatory arguments, this has to-date added an overhead that could be avoided.
 We have now refined the internal code path such that \enquote{simple} document
 commands avoid any overhead at point-of-use, making the results as efficient as
-using \cs{newcommand} or low-level \TeX{} constructs.
+using \cs{newcommand} for low-level \TeX{} constructs. Note that
+as \cs{NewDocumentCommand} makes engine-robust commands, the direct
+equivalent to \cs{newcommand} is \cs{NewExpandableDocumentCommand}.
 %
 \githubissue{1189}
 
--- a/base/doc/ltnews39.tex
+++ b/base/doc/ltnews39.tex
@@ -204,7 +204,7 @@ the standard release information.
 %
 \githubissue{945}
 
-\subsection{Optimise creation of simple document commands}
+\subsection{Optimize creation of simple document commands}
 
 Creating document commands using \cs{NewDocumentCommand}, etc., provides a very
 flexible way of grabbing arguments. When the document command only takes simple
--- a/base/doc/usrguide.tex
+++ b/base/doc/usrguide.tex
@@ -776,7 +776,7 @@ available:
 \end{itemize}
 
 When expandable commands have only |(+)m|-type arguments, the internal
-structure created is as-efficient as that provided by |\newcommand(*)|.
+structure created is as efficient as that provided by |\newcommand(*)|.
 As such, \cs{\NewExpandableDocumentCommand} may be used to replace
 \cs{newcommand} without concerns about performance.
 
--- a/base/doc/usrguide.tex
+++ b/base/doc/usrguide.tex
@@ -777,7 +777,7 @@ available:
 
 When expandable commands have only |(+)m|-type arguments, the internal
 structure created is as efficient as that provided by |\newcommand(*)|.
-As such, \cs{\NewExpandableDocumentCommand} may be used to replace
+As such, \cs{NewExpandableDocumentCommand} may be used to replace
 \cs{newcommand} without concerns about performance.
 
 \subsection{Commands at the start of tabular cells}
--- a/base/doc/usrguide.tex
+++ b/base/doc/usrguide.tex
@@ -775,11 +775,6 @@ available:
     in the standard version.
 \end{itemize}
 
-When expandable commands have only |(+)m|-type arguments, the internal
-structure created is as efficient as that provided by |\newcommand(*)|.
-As such, \cs{NewExpandableDocumentCommand} may be used to replace
-\cs{newcommand} without concerns about performance.
-
 \subsection{Commands at the start of tabular cells}
 
 Creating commands that are used at the start of tabular cells imposes
@@ -798,6 +793,18 @@ a & b & c \\
 \MyMultiCol{stuff} \\
 \end{tabular}
 \end{verbatim}
+
+\subsection{Performance}
+
+For document commands where the argument specification is entirely
+comprised of |m| or |+m| entries (or is entirely empty), the internal structure
+created by \cs{NewDocumentCommand} is essentially as efficient 
+as provided by |\newcommand(*)|. As such, document commands may replace
+constructs arising from \cs{newcommand}, etc., without a need to be concerned
+about performance. Note that \cs{newcommand(*)} produces expandable
+results, so the direct replacement is \cs{NewExpandableDocumentCommand};
+in most cases, however, it is better to use \cs{NewDocumentCommand} to
+give more robust structures.
 
 \subsection{Details about argument delimiters}
 
--- a/base/doc/usrguide.tex
+++ b/base/doc/usrguide.tex
@@ -775,6 +775,11 @@ available:
     in the standard version.
 \end{itemize}
 
+When expandable commands have only |(+)m|-type arguments, the internal
+structure created is as-efficient as that provided by |\newcommand(*)|.
+As such, \cs{\NewExpandableDocumentCommand} may be used to replace
+\cs{newcommand} without concerns about performance.
+
 \subsection{Commands at the start of tabular cells}
 
 Creating commands that are used at the start of tabular cells imposes
--- a/base/ltcmd.dtx
+++ b/base/ltcmd.dtx
@@ -435,6 +435,26 @@
 %    \end{macrocode}
 % \end{macro}
 %
+% \begin{macro}[EXP]{\@@_all_m_check:n}
+% \begin{macro}[EXP]{\@@_all_m_check:w}
+%   A quick loop to check for all |(+)m|-type arguments.
+%    \begin{macrocode}
+\cs_new:Npn \@@_all_m_check:n #1
+  { \@@_all_m_check:w #1 \q_recursion_tail \q_recursion_stop }
+\cs_new:Npn \@@_all_m_check:w #1
+  {
+    \quark_if_recursion_tail_stop:n {#1}
+    \str_if_eq:nnF {#1} { m }
+      {
+        \str_if_eq:nnF {#1} { + }
+          { X }
+      }
+    \@@_all_m_check:w
+  }
+%    \end{macrocode}
+% \end{macro}
+% \end{macro}
+%
 % \begin{macro}{\@@_declare_cmd_code:Nnn}
 % \begin{macro}
 %   {
@@ -450,17 +470,17 @@
 %   optimised document command. This only applies to document commands,
 %   not creation of environments (which are more complex).
 %    \begin{macrocode}
-\cs_new_protected:Npn \@@_declare_cmd_code:Nnn
+\cs_new_protected:Npn \@@_declare_cmd_code:Nnn #1#2
   {
-    \bool_lazy_and:nnTF
-      { ! \l_@@_environment_bool }
+    \bool_lazy_all:nTF
       {
-        \str_if_eq_p:ee
-          { \exp_not:V \l_@@_signature_tl }
-          {
-            \prg_replicate:nn \l_@@_current_arg_int
-              { \exp_not:N \@@_expandable_grab_m:w }
-          }
+        { ! \l_@@_environment_bool }
+        {
+          ! \bool_lazy_and_p:nn
+            { \l_@@_some_short_bool }
+            { \l_@@_some_long_bool }
+        }
+        { \tl_if_blank_p:e { \@@_all_m_check:n {#2} } }
       }
       { \@@_declare_cmd_optimised:Nnn }
       {
@@ -468,18 +488,23 @@
           { \@@_declare_cmd_code_expandable:Nnn }
           { \@@_declare_cmd_code_aux:Nnn }
       }
+        #1 {#2}
    }
 %    \end{macrocode}
 %   The optimised version of commands just has to worry about whether to make
-%   them protected. The commands start with an expandable marker so that other
-%   parts of the kernel know these are set up by \pkg{ltcmd}
+%   them protected or long. The commands start with an expandable marker so
+%   that other parts of the kernel know these are set up by \pkg{ltcmd}.
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_declare_cmd_optimised:Nnn #1#2#3
   {
-    \bool_if:NTF \l_@@_expandable_bool
-      { \cs_generate_from_arg_count:NNnn #1 \cs_set_nopar:Npn }
-      { \cs_generate_from_arg_count:NNnn #1 \cs_set_protected_nopar:Npn }
-        \l_@@_current_arg_int { \@@_start_optimised: #3 }
+    \exp_args:NNc \cs_generate_from_arg_count:NNnn #1
+      {
+        cs_set
+        \bool_if:NF \l_@@_expandable_bool { _protected }
+        \bool_if:NF \l_@@_some_long_bool { _nopar }
+        :Npn
+      }
+      \l_@@_current_arg_int { \@@_start_optimised: #3 }
   }
 \cs_new:Npn \@@_start_optimised: { }
 %    \end{macrocode}
--- a/base/ltcmd.dtx
+++ b/base/ltcmd.dtx
@@ -35,7 +35,7 @@
 %
 %    \begin{macrocode}
 \def\ltcmdversion{v1.2b}
-\def\ltcmddate{2023-11-29}
+\def\ltcmddate{2023-11-30}
 %    \end{macrocode}
 %
 %<*driver>
@@ -459,7 +459,7 @@
 %     \@@_declare_cmd_code_expandable:Nnn
 %   }
 % \begin{macro}{\@@_start_optimized:}
-% \changes{v1.2b}{2023/11/29}
+% \changes{v1.2b}{2023/11/30}
 %   {Optimize cmd creation for all-\texttt{m} arguments}
 %   At this stage we can check for a short-cut possibility: if the signature
 %   is just made up of just |(+)m| tokens, and if all arguments are either
@@ -4516,7 +4516,7 @@
 %      \cs{NewCommandCopy} and \cs{ShowCommand} support}
 % \changes{v1.0l}{2022/03/18}{Fix \cs{@@_cmd_type_cases:NnnnnF}
 %"     prematurely expanding macros (gh/795)}
-% \changes{v1.2b}{2023/11/29}{Extend for optimized commands}
+% \changes{v1.2b}{2023/11/30}{Extend for optimized commands}
 % \begin{macro}{\@@_cmd_if_xparse_aux:N}
 %
 %    To determine whether the command is an \pkg{xparse} command check
--- a/base/ltcmd.dtx
+++ b/base/ltcmd.dtx
@@ -454,17 +454,17 @@
 % \begin{macro}{\@@_declare_cmd_code:Nnn}
 % \begin{macro}
 %   {
-%     \@@_declare_cmd_optimised:Nnn,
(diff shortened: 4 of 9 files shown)


josephwright:
No roll-back - I don't think we need it here

PhelypeOleinik:
> Not sure if we want to keep the one-step indirection? That would still mark these as '`ltcmd`-generated'?

I'd rather keep that, because then for other systems like `ltcmdhooks` we still know what generated the command and we can have some certainty when dealing with them.  Removing that indirection turns all into guesswork.  It isn't _that_ much slower, is it?

base/doc/ltnews39.tex
```diff
@@ -187,6 +187,16 @@ \subsection{Warn if counter names are problematic}
...
 \githubissue{823}
 
+\subsection{Optimise creation of simple document commands}
+
+Creating document commands using \cs{NewDocumentCommand}, etc., provides a very
+flexible way of grabbing arguments. When the document command only takes simple
+mandatory arguments, this has to-date added an overhead that could be avoided.
+We have now refined the internal code path such that \enquote{simple} document
+commands avoid any overhead at point-of-use, making the results as efficient as
+using \cs{newcommand} or low-level \TeX{} constructs.
```
FrankMittelbach, line 197: We could add something that explains that there is still a difference between \newcommand and \NewDocumentCommand as the former expands so corresponds to \NewExpandableDocumentCommand

FrankMittelbach (approved):
looks basically fine to me (assuming that it doesn't break any of the \ShowCommand.. \CopyCommand... thingies, but I can't see how nd the suite should tell us.

So other than perhaps a bit more words in ltnews and also in usrguide I guess (stating that one can use \NewExpanableDocumentCommand instead of \newcommand for consistency and get the same "speed" or something like that) it looks fine

FrankMittelbach:
> > Not sure if we want to keep the one-step indirection? That would still mark these as '`ltcmd`-generated'?
> 
> I'd rather keep that, because then for other systems like `ltcmdhooks` we still know what generated the command and we can have some certainty when dealing with them. Removing that indirection turns all into guesswork. It isn't _that_ much slower, is it?

I was wondering about that, but what do you gain from ltcmdhooks know that it was done on top-level with `\NewExpandableDocumentCommand` when all that is there in the end is `\def\foo#1#2#3{...}`. For adding hooks or showing the definition it seems fairly unimportant if it was produced by hand, by `\newcommand` or by `\NewExpandableDocumentCommand`, or not?  (unless in the "show" you want to explicitly say how it was produced)

PhelypeOleinik:
> For adding hooks or showing the definition it seems fairly unimportant if it was produced by hand, by `\newcommand` or by `\NewExpandableDocumentCommand`, or not?

Not :-)

When the code knows that the command `\foo` was defined with `\NewExpandableDocumentCommand` it can patch the command by expanding it with something along the lines of `\expandafter\definecmd\expandafter{\foo{##1}{##2}{##3}}`, because it knows there are no "weird" tokens in the parameter text.  If we don't know that, the code has to resort to patching with `\scantokens` which is... sketchy.

josephwright:
> > Not sure if we want to keep the one-step indirection? That would still mark these as '`ltcmd`-generated'?
> 
> I'd rather keep that, because then for other systems like `ltcmdhooks` we still know what generated the command and we can have some certainty when dealing with them. Removing that indirection turns all into guesswork. It isn't _that_ much slower, is it?

All doable. I am then wondering about `\__kernel_cmd_if_xparse:NTF`. First, shouldn't it be `\__kernel_cmd_if_ltcmd:NTF`? More importantly, why does it do all of these tests:
```latex
\cs_new_protected:Npn \@@_cmd_type_cases:NnnnnF #1 #2 #3 #4 #5 #6
  {
    \exp_args:Ne \str_case_e:nnF
      {
        \exp_args:Nf \tl_if_empty:nT { \cs_argument_spec:N #1 }
          { \exp_not:N \exp_not:n { \exp_not:e { \tl_head:N #1 } } }
      }
      {
        { \exp_not:N \@@_start:nNNnnn } {#2}
        { \exp_not:N \@@_start_expandable:nNNNNn } {#3}
        { \exp_not:N \@@_start_env:nnnnn } {#4}
        {
          \exp_after:wN \exp_not:N
            \cs:w environment~
              \exp_last_unbraced:Ne \use_none:nnn
                { \cs_to_str:N #1 } ~end~aux \cs_end:
        } {#5}
      }
      {#6}
  }
```
rather than simply check if `\foo code` exists? I'll need to adjust to allow for optimisation, so before I do can we agree on what this test should be?

josephwright:
Also similar in `\@@_copy:NN`, which again seems to test for the internal setup (even though the test happens before it's called too).

josephwright:
I've added one level of indirection. Note that this means that an error shows `\foo code` rather than `\foo` or `\foo[space]`: is that OK? We could fix with two levels of indirection, but I wonder if that is a bit much.

josephwright:
>  I am then wondering about `\__kernel_cmd_if_xparse:NTF`. First, shouldn't it be `\__kernel_cmd_if_ltcmd:NTF`? More importantly, why does it do all of these tests:

We might need to test for end-of-env, but all of the others have `\foo code`, no?

josephwright:
> > I am then wondering about `\__kernel_cmd_if_xparse:NTF`. First, shouldn't it be `\__kernel_cmd_if_ltcmd:NTF`? More importantly, why does it do all of these tests:
> 
> We might need to test for end-of-env, but all of the others have `\foo code`, no?

OK, I see that for copying there are different paths (so we will need to extend), but does this need to share the code with the yes/no test?

josephwright:
I need to adjust to skip optimisation for envs: currently doing that gives some very odd situation with building `source2e`. I need to track this down - may take a bit.

PhelypeOleinik:
> rather than simply check if `\foo code` exists?

The problem in this case is if someone does `\def\bar{something} \let\foo\bar`, then `\foo code` exists, but `\foo` is not an `ltcmd` command.  This case is particularly problematic for `\@@_show:N` and `\@@_copy:NN`, which rely heavily on the internal structure of the command being correct.

FrankMittelbach:
why not simply `\protected\def\foo{\flag\foocode}` with `\let\flag\@empty`(and a suitable internal name)?

josephwright:
> why not simply `\protected\def\foo{\flag\foocode}` with `\let\flag\@empty`(and a suitable internal name)?

That had also occurred to me: I was initially aiming for a simple structure, but a flag might serve as well as anything else.

car222222:
Not to be confused with an l3 flag:-).

base/doc/ltnews39.tex
```diff
@@ -187,6 +187,18 @@ \subsection{Warn if counter names are problematic}
...
 \githubissue{823}
 
+\subsection{Optimise creation of simple document commands}
+
+Creating document commands using \cs{NewDocumentCommand}, etc., provides a very
+flexible way of grabbing arguments. When the document command only takes simple
+mandatory arguments, this has to-date added an overhead that could be avoided.
+We have now refined the internal code path such that \enquote{simple} document
+commands avoid any overhead at point-of-use, making the results as efficient as
+using \cs{newcommand} for low-level \TeX{} constructs. Note that
```
muzimuzhi, line 197: With redirection reintroduced, description here need to adjust.

base/ltcmd.dtx
```diff
@@ -437,16 +437,52 @@
...
-      { \@@_declare_cmd_code_aux:Nnn }
+    \bool_lazy_and:nnTF
+      { ! \l_@@_environment_bool }
+      {
+        \str_if_eq_p:ee
+          { \exp_not:V \l_@@_signature_tl }
+          {
+            \prg_replicate:nn \l_@@_current_arg_int
+              { \exp_not:N \@@_expandable_grab_m:w }
+          }
```
muzimuzhi, line 463: Looks like this can be extended to cover all-`\@@_expandable_grab_m_long:w` signature too, which is closer to what `\newcommand` defines.

muzimuzhi (changes requested):
Ltnews adjustments for reintroduced redirection and possibility to optimise `+m`-arg only cases.

muzimuzhi:
> > rather than simply check if `\foo code` exists?
> 
> The problem in this case is if someone does `\def\bar{something} \let\foo\bar`, then `\foo code` exists, but `\foo` is not an `ltcmd` command.

Promoting `\NewCommandCopy` against `\let` to users is a long-term task...

PhelypeOleinik:
> Promoting `\NewCommandCopy` against `\let` to users is a long-term task...

Agreed, but in this case `\NewCommandCopy` wouldn't help, as `\bar` was defined with `\def` (or `\newcommand`, for that matter), so using `\NewCommandCopy` here would have the same effect as using `\let` plus some overhead processing.

base/doc/ltnews39.tex
josephwright: See latest ...

base/ltcmd.dtx
josephwright: Yes, did occur to me: I'll look at that once the first pass is done

Skillmon:
> > Not :-)
> 
> ok point taken but it should be a really light indirections that is basically a single extra expansion

Just `\cs_new_eq:NN \__ltcmd_defined: \prg_do_nothing:` and putting `\__ltcmd_defined:` at the start of the `\meaning` should be fine and no big overhead.

josephwright:
> > > Not :-)
> > 
> > 
> > ok point taken but it should be a really light indirections that is basically a single extra expansion
> 
> Just `\cs_new_eq:NN \__ltcmd_defined: \prg_do_nothing:` and putting `\__ltcmd_defined:` at the start of the `\meaning` should be fine and no big overhead. (so without the `\foocode` step).

More or less what I've done.

base/ltcmd.dtx
josephwright: Addressed by 62d512955111a94a6b8dae3d183de66ddeb0a48c - should make us as efficient as `\def`, more or less, for the common case.

base/doc/usrguide.tex
```diff
@@ -775,6 +775,11 @@ \subsection{Fully-expandable document commands\label{sec:ltcmd:expandable}}
     in the standard version.
 \end{itemize}
 
+When expandable commands have only |(+)m|-type arguments, the internal
```
muzimuzhi, line 778: Worth mentioning "without any other argument modifiers nor processors"?

base/doc/usrguide.tex
```diff
@@ -775,6 +775,11 @@ \subsection{Fully-expandable document commands\label{sec:ltcmd:expandable}}
...
+structure created is as efficient as that provided by |\newcommand(*)|.
+As such, \cs{\NewExpandableDocumentCommand} may be used to replace
```
muzimuzhi, line 780:
```suggestion
As such, \cs{NewExpandableDocumentCommand} may be used to replace
```

base/ltcmd.dtx
```diff
@@ -435,18 +435,75 @@
...
+%   {
+%     \@@_declare_cmd_optimised:Nnn,
+%     \@@_declare_cmd_code_aux:Nnn,
+%     \@@_declare_cmd_code_expandable:Nnn
+%   }
+% \begin{macro}{\@@_start_optimised:}
+% \changes{v1.2b}{2023/11/28}
+%   {Optimise cmd creation for all-\texttt{m} arguments}
+%   At this stage we can check for a short-cut possibility: if the signature
+%   is just made up of \cs{@@_expandable_grab_m:w}, then we can produce an
```
muzimuzhi, line 465: or `\cs{@@_expandable_grab_long_m:w}`

(Then ambiguity arises because the combination of short and long `m` are not supported. Fine, code comment won't be executed.)

base/ltcmd.dtx
```diff
@@ -435,18 +435,75 @@
...
+            { \l_@@_some_long_bool }
+        }
+        { \tl_if_blank_p:e { \@@_all_m_check:n {#2} } }
+      }
+      { \@@_declare_cmd_optimised:Nnn }
+      {
+        \bool_if:NTF \l_@@_grab_expandably_bool
+          { \@@_declare_cmd_code_expandable:Nnn }
+          { \@@_declare_cmd_code_aux:Nnn }
+      }
```
muzimuzhi, line 486: Only one branch of bool-expr is not flipped, would `\bool_lazy_or:nTF { {expr 1} {expr 2} {! expr_3} } {<normal>} {<optimized>}` be a little quicker (and more straight)?
```suggestion
    \bool_lazy_or:nTF
      {
        { \l_@@_environment_bool }
        {
          \bool_lazy_and_p:nn
            { \l_@@_some_short_bool }
            { \l_@@_some_long_bool }
        }
        { ! \tl_if_blank_p:e { \@@_all_m_check:n {#2} } }
      }
      {
        \bool_if:NTF \l_@@_grab_expandably_bool
          { \@@_declare_cmd_code_expandable:Nnn }
          { \@@_declare_cmd_code_aux:Nnn }
      }
      { \@@_declare_cmd_optimised:Nnn }
```

base/ltcmd.dtx
```diff
@@ -4376,27 +4476,29 @@
...
 %    \cs{@@_start_expandable:nNNNNn} (expandable command) or
+%    \cs{@@_start_optimised:} (optimised command) or
 %    \cs{@@_start_env:nnnnn} (environment) or
 %    \cs[no-index]{environment~\#1~end~aux} (environment end).
 %
 %    This conditional is needed in several kernel modules and is
 %    therefore has a kernel-internal name.
 %    \begin{macrocode}
-\cs_new_protected:Npn \@@_cmd_type_cases:NnnnnF #1 #2 #3 #4 #5 #6
+\cs_new_protected:Npn \@@_cmd_type_cases:NnnnnnF #1 #2 #3 #4 #5 #6#7
```
muzimuzhi, line 4501:
```suggestion
\cs_new_protected:Npn \@@_cmd_type_cases:NnnnnnF #1 #2 #3 #4 #5 #6 #7
```

base/testfiles-ltcmd/github-1009.tlg
```diff
@@ -4,32 +4,30 @@ Don't change this file in any respect.
 TEST 1: Non-expandable Document Commands with m-type arguments
 ============================================================
 > \foo=document command:
+ #1:m
 ->.
```
muzimuzhi, line 8: `\ShowCommand` works wrong.

- This `\foo` takes no arguments, hence shouldn't show `#1:m`.
- For `\foo` that does take `m` arguments, `\__cmd_start_optimised: .` is shown.

base/testfiles-ltcmd/ltcmd001.tlg
```diff
@@ -17,97 +17,73 @@ For immediate help type H <return>.
 l. ...}
 You have used \RenewDocumentCommand with a command that was never defined.
 LaTeX will ignore this entire definition.
-> \foo=\protected macro:->\__cmd_start_expandable:nNNNNn {}\foo  \foo  \foo
-code ?{}.
+> \foo=\protected macro:->\__cmd_start_optimised: First definition.
 <recently read> }
 l. ...}
-> \foo code=\protected\long macro:->First definition.
+> \foo code=undefined.
```
muzimuzhi, line 23: Polishing: perhaps the setting in `ltcmd001.lvt` can check if `\cs{foo code}` exists before showing it.
https://github.com/latex3/latex2e/blob/0f40cbb8ea0c43592db9d1c96dbb7575166b4a7b/base/testfiles-ltcmd/ltcmd001.lvt#L16-L19
```tex
% module also renamed from "xparse" to "ltcmd"
\cs_new:Npn \ltcmd_show:N #1 {
  \cs_show:N #1
  \cs_if_exist:cT { \cs_to_str:N #1 ~code }
    { \cs_show:c { \cs_to_str:N #1 ~code } }
}
```

muzimuzhi (changes requested):
`\ShowCommand` works wrong.

base/changes.txt
```diff
@@ -6,6 +6,12 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2023-11-28  Joseph Wright  <Joseph.Wright@latex-project.org>
+	* ltcmd.dtx (subsection{Declaring commands and environments}):
+	Optimize creation of simple document commands (gh/1189)
```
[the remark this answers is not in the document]
josephwright: I think for newer stuff we are sicking with Us spelling for realism reasons; also, 'ise' vs 'ize' isn't as clear-cut as say 'colour vs 'color' anyway.

base/testfiles-ltcmd/ltcmd001.tlg
josephwright: I thought about that, but isn't the entire point here that most `ltcmd` stuff does have this, so we should be explicit when it's not present?

base/ltcmd.dtx
josephwright: I've updated the comment

base/testfiles-ltcmd/github-1009.tlg
josephwright: I realise looking at this we do need a two-stage process - which I'd missed. Update coming up.

josephwright: Tricker than expected: I need to re-work the test for `ltcmd` support entirely

josephwright: OK, I have this right in `ltcmd` I think but there is still more to do :(

josephwright:
Turns out there is more to do here yet. The various code paths all assume that `ltcmd`-based stuff has indirection and no args. at the top level. I need to work out which way to address this: updates to come.

base/testfiles-ltcmd/ltcmd001.tlg
muzimuzhi: I'm OK to both ways.

base/ltcmd.dtx
```diff
@@ -435,17 +435,75 @@
...
+%   {
+%     \@@_declare_cmd_optimized:Nnn,
+%     \@@_declare_cmd_code_aux:Nnn,
+%     \@@_declare_cmd_code_expandable:Nnn
+%   }
+% \begin{macro}{\@@_start_optimized:}
+% \changes{v1.2b}{2023/11/29}
+%   {Optimise cmd creation for all-\texttt{m} arguments}
+%   At this stage we can check for a short-cut possibility: if the signature
+%   is just made up of just |(+)m| tokens, and if all arguments are either
```
muzimuzhi, line 465: Two "just". Not sure which one to drop.

```suggestion
%   is made up of just |(+)m| tokens, and if all arguments are either
```

base/doc/usrguide.tex
muzimuzhi: Like code comment for `\@@_declare_cmd_code:Nnn`, saying "argument specification" seems more rigorous. An argument specification consists of all things, argument types, modifiers, and processors.

> also we are dealing with expandable commands.

Thus `\NewDocumentCommand` and friends need a similar paragraph, or this paragraph, after adaption, can be move to an earlier place?

josephwright:
OK, now should be working properly: a one-level indirection is retained, but without grabbing anything more than once, so in practical terms the performance will be very very close to that of `\def\foo` - certainly much better than without the optimisation.

base/doc/usrguide.tex
josephwright: I've extended and revised the text a bit

base/testfiles-ltcmd/github-1009.tlg
josephwright: Should now be correct

base/testfiles-ltcmd/ltcmd001.tlg
josephwright: Latest revision is back using this internal ...

base/ltcmd.dtx
josephwright: @car222222 See the revised text

base/ltcmd.dtx
```diff
@@ -2530,6 +2622,47 @@
...
+  {
+    \exp_args:Nc \@@_show_optimized:NN
+      { \cs_to_str:N #1 \c_space_tl code }
+      #1
+  }
+\cs_new_protected:Npn \@@_show_optimized:NN #1#2
+  {
+    \cs_set:Npe \@@_show_optimized_aux:N ##1
+      {
+        \c_space_tl \c_hash_str ##1 :
```
muzimuzhi, line 2638: To be consistent with other cases. It's at line beginning here hence leading `~`s are always skipped. 
https://github.com/latex3/latex2e/blob/0f40cbb8ea0c43592db9d1c96dbb7575166b4a7b/base/ltcmd.dtx#L2736-L2740
```suggestion
        \c_space_tl \c_space_tl \c_hash_str ##1 :
```

muzimuzhi (changes requested):
One or two code fix(es) with typos and documentation discussions as usual :)

This fixes content of `github-1009.tlg`. I haven't (double-)checked `ltcmd001.tlg` and `ltcmd002.tlg` yet.

muzimuzhi:
Question: What's the status of `\DeclareRobustCommand`, currently documented in `clsguide` (not the historic version)? Is `\NewDocumentCommand` a replacement to it?

If so then in new paragraphs in ltnews and `usrguide` `\DeclareRobustCommand` needs to be mentioned, just like `\newcommand`.

base/changes.txt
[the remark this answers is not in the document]
muzimuzhi: Querying "optimise" on Cambridge Dictionary Online gives "optimize", which starts with

> optimize
> verb [[ T ]](https://dictionary.cambridge.org/help/codes.html) (UK usually optimise)

https://dictionary.cambridge.org/dictionary/english/optimize?q=optimise

Merriam-Webster Dictionary Online is more directly,

> optimisation, optimise
> op·​ti·​mi·​sa·​tion, op·​ti·​mise
> British spellings of [OPTIMIZATION](https://www.merriam-webster.com/dictionary/optimization), [OPTIMIZE](https://www.merriam-webster.com/dictionary/optimize)

https://www.merriam-webster.com/dictionary/optimise

FrankMittelbach:
> Question: What's the status of `\DeclareRobustCommand`, currently documented in `clsguide` (not the historic version)? Is `\NewDocumentCommand` a replacement to it?
> 
> If so then in new paragraphs in ltnews and `usrguide` `\DeclareRobustCommand` needs to be mentioned, just like `\newcommand`.

Don't think I would not go there this time around. `\DeclareRobustCommand` use a different approach but it is equally fast and for single optional arguments probably faster in execution. But if others feel differently, I'm happy to let myself getting convinced.

base/ltcmd.dtx
```diff
@@ -2530,6 +2622,47 @@
...
+          \bool_lazy_or:nnF
+            { \token_if_protected_macro_p:N #1 }
+            { \token_if_protected_long_macro_p:N #1 }
+              { expandable ~ } document~command:
+        \iow_newline:
+        \int_step_function:nN
+          {
+            \int_div_truncate:nn
+              { \tl_count:n { \cs_parameter_spec:N #1 } }
+              { 2 }
```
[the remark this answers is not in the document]
josephwright: I'll go with adding a variant: we can sweep that out again later.

base/ltcmd.dtx
josephwright: Changed (done as a commit by me as also needs `.tlg` changes).

josephwright:
We could optimise all multi- `m` grabbing, particularly if it's after any other arg. types. It's all a question of balance of complexity. I suggest we go now with this change, see the real-world impact and can revisit `om...` in a future round (we might need to re-work more of `ltcmd` to do that.)

josephwright:
> Also it seems `\ShowCommand` needs more tests. `github-1009.lvt` only tests it against zero or one `m`-types.

I've extended to cover the case of two `m`-type: I think this shows the pattern works well.

base/doc/ltnews39.tex
josephwright: That was true even with a one-level approach unless you make it equivalent to the `\def` version - I guess we _could_ do that if we then track commands created by `ltcmd` manually (slightly risky: vulnerable to a `\def` after creation).

josephwright: I've revised the text a bit.

base/testfiles-ltcmd/github-1009.tlg
```diff
@@ -31,6 +43,30 @@ l. ...}
...
+  #1:+m
+  #2:+m
+->.
+<recently read> }
+l. ...}
+> \foo=document command:
+  #1:m
+  #2:+m
+->.
+<recently read> }
```
muzimuzhi, line 62: `m +m` shouldn't be optimized.

josephwright: It's not: I added the test purely as things would go sideways if it was optimized accidentally.

muzimuzhi: Ignore me. `\ShowCommand` works as expected here.

base/testfiles-ltcmd/ltcmd001.tlg
```diff
@@ -700,14 +668,14 @@ TEST 17: (ab)using xparse commands in csnames
 l. ...  }
 The control sequence marked <to be read again> should
 not appear between \csname and \endcsname.
-\test- \xparse function is not expandable \bool_set_false:N \l__cmd_environment_bool \__cmd_start_aux:NNnnnn \foo code {D...}{\__cmd_grab_D:w ..}{{.}}{}....\cs_end: 
+\test- \xparse function is not expandable \bool_set_false:N \l__cmd_environment_bool \__cmd_start_aux:NNnnnn \foo  \foo code {D...}{\__cmd_grab_D:w ..}{{.}}{}....\cs_end:
```
muzimuzhi, line 671: This change (`\foo␣code {D...}` -> `\foo␣ \foo␣code {D...}`) took place because previously for expandable command `\cmd`, `\cmd␣` is also defined (and if that command has both short and long arguments, `\cmd␣␣` is defined in addition). All happen in `\__cmd_declare_cmd_code_expandable:Nnn`.

Prior to this PR, in `ltcmd001.lvt` lines 363 and 368--369,
- first `\DeclareDocumentCommand { \foo } { m m } { #1-#2 }`  defines `\foo␣` to `macro:#1#2->#1{#2}`,
- then `\DeclareDocumentCommand { \foo } { D... } { test-#1 }` leaves `\foo␣` intact,
- finally in `\TYPE { \exp_not:c { test- \foo .... } }`,  `\foo␣ \foo␣code {D...}` expands to `\foo␣code {D...}`, which is then written to log.
 
Now `\foo␣` is made `\relax` by `\DeclareDocumentCommand { \foo } { m m } { #1-#2 }` (see `\__cmd_declare_cmd_internal:Nnnn`), hence left in log.

https://github.com/latex3/latex2e/blob/0f40cbb8ea0c43592db9d1c96dbb7575166b4a7b/base/testfiles-ltcmd/ltcmd001.lvt#L361-L375

Several questions are involved:
- Can optimized expandable commands leave `\foo␣` `\relax`ed?
- Should `\(Renew|Declare)DocumentCommand` and friends clear command internals before redefinition? See also #1191.
- In what extent should (ab)using document commands in csnames (name of the test these log lines correspond to) be guarded and tested?
  Do optimized commands need to add a such guard? This may lead to a new `\__cmd_start_optimized_expandable:` and the old `\__cmd_start_optimized:` containing (full expansion of) something like `\exp_not:c { xparse~function~is~not~expandable }`.

These are all not firmly related to current PR hence I think they can be discussed and probably resolved in separate issue(s) and PR(s).

base/doc/ltnews39.tex
```diff
@@ -187,6 +187,18 @@ \subsection{Warn if counter names are problematic}
 %
 \githubissue{823}
 
+\subsection{Optimise creation of simple document commands}
+
```
FrankMittelbach, line 191: We should go with the US spelling (which is Optimize)

base/testfiles-ltcmd/ltcmd001.tlg
josephwright: I've moved part of the setup as it avoids creating the csname before it's required. That's of course separate from the other queries you raise.

base/doc/ltnews39.tex
josephwright: Corrected: thought I'd got all of these

FrankMittelbach (approved):

base/testfiles-ltcmd/ltcmd001.tlg
muzimuzhi: > * In what extent should (ab)using document commands in csnames (name of the test these log lines correspond to) be guarded and tested?
>   Do optimized commands need to add a such guard? This may lead to a new `\__cmd_start_optimized_expandable:` and the old `\__cmd_start_optimized:` containing (full expansion of) something like `\exp_not:c { xparse~function~is~not~expandable }`.

This one is not needed, as document commands having all-`m` arguments always start with `\__cmd_start_expandable:nNNNNn`, never `\__cmd_start:nNNnnn`, see for example #1013.

base/ltcmd.dtx
```diff
@@ -4376,27 +4510,29 @@
...
 % \begin{macro}{\__kernel_cmd_if_xparse:NTF}
-% \begin{macro}{\@@_cmd_type_cases:Nnnnn}
+% \begin{macro}{\@@_cmd_type_cases:Nnnnnn}
 % \changes{v1.0d}{2021/04/19}{Renamed \cs{__cmd_cmd_if_xparse:NTF} to
 %      \cs{__kernel_cmd_if_xparse:NTF} for cross-module usage}
 % \changes{v1.0d}{2021/07/30}{Added \cs{@@_cmd_type_cases:NnnnnF} for
 %      \cs{NewCommandCopy} and \cs{ShowCommand} support}
 % \changes{v1.0l}{2022/03/18}{Fix \cs{@@_cmd_type_cases:NnnnnF}
 %"     prematurely expanding macros (gh/795)}
+% \changes{v1.2b}{2023/12/01}{Extend for optimized commands}
```
muzimuzhi, line 4520: Not related to this PR and not worth a separate PR too.
```suggestion
%      prematurely expanding macros (gh/795)}
% \changes{v1.2b}{2023/12/01}{Extend for optimized commands}
```

FrankMittelbach:
> Now `\group_align_safe_begin:` and `\group_align_safe_end:` are dropped in optimized commands. `\conditionally@traceoff` and `\conditionally@traceon` are dropped in optimized non-expandable commands in addition.

I missed that, where are they dropped?

muzimuzhi:
@FrankMittelbach I worried too much. Optimized commands need neither `\group_align_safe_(begin|end):` and `\conditionally@trace(off|on)`.

Skillmon:
@muzimuzhi it has the potential to change behaviour (check with the current behaviour):

```
\documentclass{article}

\NewDocumentCommand\mygobble { m } {}
\newcommand\gobble[1]{}

\begin{document}
\begin{tabular}{ll}
  a\mygobble&&b \\ % fine
  c\gobble&&d % error
\end{tabular}
\end{document}
```

josephwright:
> @muzimuzhi it has the potential to change behaviour (check with the current behaviour):
> 
> ```
> \documentclass{article}
> 
> \NewDocumentCommand\mygobble { m } {}
> \newcommand\gobble[1]{}
> 
> \begin{document}
> \begin{tabular}{ll}
>   a\mygobble&&b \\ % fine
>   c\gobble&&d % error
> \end{tabular}
> \end{document}
> ```

That’s protected status , which is unchanged

Skillmon:
> That’s protected status , which is unchanged

The same behaviour change would occur with `\NewExpandableCommand`, currently it uses `\group_align_safe_...` so would work in above MWE, with the shortcut behaviour dropping `\group_align_safe_...` this would change (though unlikely to have any real effect in the field, it's a behaviour change that I wanted to make sure was noted).

FrankMittelbach:
> this would change (though unlikely to have any real effect in the field, it's a behaviour change that I wanted to make sure was noted).

I think a comment on that difference next to  the code definition might be sensible, but I wouldn't go further and document it on the level of the user documentation because being able able to grab and destroy a `&` is not really an intended function for such command

josephwright:
> > this would change (though unlikely to have any real effect in the field, it's a behaviour change that I wanted to make sure was noted).
> 
> I think a comment on that difference next to the code definition might be sensible, but I wouldn't go further and document it on the level of the user documentation because being able able to grab and destroy a `&` is not really an intended function for such command

Done

PR #1192, opened 2023-11-29
Remove \tex_lowercase:D in ltcmd

josephwright:
This will be slightly slower but means we stick
to our own rules.

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added


josephwright:
Cool - I'll wait to see if there are any other comments before merging

FrankMittelbach:
I'm not 100% sure that I'm keen on sticking to our own rules as you phrase it

Basically, I see kernel code as necessarily optimized for speed which means, for example, I rather prefer to see `:o` then `:V` when I deal with opening a tl variable. I know I'm an oldtimer in that respect (perhaps in others, too) who started when space and speed where really important, but even now I think we have to be careful and keep the kernel really really optimized at least its core functions.

josephwright:
> Basically, I see kernel code as necessarily optimized for speed which means, for example, I rather prefer to see `:o` then `:V` when I deal with opening a tl variable.

I know, and I leave those alone, although they rely on implementation detail ... as you know, I prefer to use `o` only when I explicitly need one expansion (like `\use_none:n` for some argument gobbling).

> I know I'm an oldtimer in that respect (perhaps in others, too) who started when space and speed where really important, but even now I think we have to be careful and keep the kernel really really optimized at least its core functions.

Well 'kernel' is a movable feast: here we are talking `ltcmd` (relatively high-level), which already uses `\char_generate:nn` and in this case is using `\tex_lowercase:D` for argument splitting - not a core function we need for every doc command. (I hope the other PR about speeding up doc commands shows I'm not ignoring that aspect.)

josephwright:
I was mainly bothered about the use of `^^@`: it's what attracted my attention.

josephwright:
If we want to gain performance in this area in `ltcmd`, I'd look at https://github.com/latex3/latex2e/blob/0f40cbb8ea0c43592db9d1c96dbb7575166b4a7b/base/ltcmd.dtx#L3033-L3051.

josephwright:
> I know, and I leave those alone, although they rely on implementation detail ... as you know, I prefer to use `o` only when I explicitly need one expansion (like `\use_none:n` for some argument gobbling).

I mean of course in my own code, where in `siunitx` for example it's all `V` except where strictly required.

josephwright:
> I'm not 100% sure that I'm keen on sticking to our own rules as you phrase it

There's a reason I went with a PR and wanted to wait for opinions :)

josephwright:
> If we want to gain performance in this area in `ltcmd`, I'd look at
> 
> https://github.com/latex3/latex2e/blob/0f40cbb8ea0c43592db9d1c96dbb7575166b4a7b/base/ltcmd.dtx#L3033-L3051
> .

Checking, we avoid this in the common case ...

josephwright:
> If we want to gain performance in this area in `ltcmd`, I'd look at
> 
> https://github.com/latex3/latex2e/blob/0f40cbb8ea0c43592db9d1c96dbb7575166b4a7b/base/ltcmd.dtx#L3033-L3051
> .

(This gets called if we find the opening token for the optional argument: it's not used if there is no optional arg.)

davidcarlisle:
I was going to make a comment simmilar to @FrankMittelbach , the code looks OK, but I think for functions that are in the format I'd rather go the other way and do any optiomisations that don't obfuscate the code too much.

If you replace a screenfull of low level hackery by a single high level function, a small time penalty may be worth the code clarity, but here the higher level code isn't really any shorter or clearer, so I'm not sure I see the gain.

josephwright:
> I was going to make a comment simmilar to @FrankMittelbach , the code looks OK, but I think for functions that are in the format I'd rather go the other way and do any optiomisations that don't obfuscate the code too much.
> 
> If you replace a screenfull of low level hackery by a single high level function, a small time penalty may be worth the code clarity, but here the higher level code isn't really any shorter or clearer, so I'm not sure I see the gain.

Well I can but ask: we will have a record of the decision one way or the other :)

josephwright:
@FrankMittelbach  That's not really needed :) If the overall feeling is a 'no', I'll close the PR and add an explicit comment to the sources documenting this.

FrankMittelbach:
@josephwright sure, but the meaning of 2 oldtimers may not be enough :-) and also I would like to spend a few minutes on discussing what we generally think we want to do or not do in kernel code.

FrankMittelbach:
> I'm (relatively) young,

trust me ... that will change over time :-)

car222222:
But it's not too bad . . . (??)

Issue #1198, opened 2023-12-06
Misleading ltcmd message "Invalid argument prefix '!'"


muzimuzhi:
## Brief outline of the bug

When an argument prefix `!` is applied to a non-trailing optional argument, the resulting error message is misleading.

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{article}

\NewDocumentCommand{\mycmd}{!m}{def}   % clear error message
\NewDocumentCommand{\mycmd}{!o m}{def} % misleading error message

\begin{document}
\end{document}
```

Compiling example above throws two copies of a similar error message. For the second copy, `You tried to apply it to 'm'.` is not a clear description of the problem in argument specification `!o m`.
```
! LaTeX cmd Error: Invalid argument prefix '!' in command '\mycmd'.

...                                        

The prefix '!' is only allowed for trailing optional arguments. You tried to
apply it to 'm'.

LaTeX will ignore this entire definition.
```

## Log file (required) and possibly PDF file  

[latex2e-mislading-message-invalid-bang.log](https://github.com/latex3/latex2e/files/13578409/latex2e-mislading-message-invalid-bang.log)

FrankMittelbach:
maybe it could say instead
```
The prefix '!' is only allowed for trailing optional arguments. 
There is a non-optional 'm' argument afterwards.
```

PR #1199, opened 2023-12-06
Clarify error message for one cause of invalid argument prfix `!`

muzimuzhi:
Implements #1198

Rollback needed?

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development

- [ ] Test file(s) added


base/changes.txt
```diff
@@ -6,6 +6,11 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2023-12-06  Yukai Chou <muzimuzhi@gmail.com>
+	* ltcmd.dtx (subsection{Normalizing the argument specifications})
+	Use a clarified error message when an argument prefix `!' is applied to
```
josephwright, line 11: I've approved in review, but 'Clarify error message when ...' would be a little more natural

muzimuzhi: Oh I forgot to update all these places. Thanks.

FrankMittelbach (commented):
I think that this adds in some sense unnecessary processing and just changing the explanation slightly would do as well (as I suggesed elsewhere). But if the common feeling is that this way it is more helpful, I'm not against that version.

josephwright:
@FrankMittelbach Sure but we are in an error branch here so performance shouldn't be an issue? I agree it would be reasonable to simply make the description more accurate as an alternative fix.

FrankMittelbach (approved):
true only when erroring anyway, I misread that.

base/ltcmd.dtx
```diff
@@ -4747,7 +4754,14 @@
...
-\msg_new:nnnn { cmd } { invalid-bang }
+\msg_new:nnnn { cmd } { invalid-bang-before-mandatory }
+  { Invalid~argument~prefix~'!'~in~#1. }
+  {
+    The~prefix~'!'~is~only~allowed~for~trailing~optional~arguments.~
+    You~tried~to~apply~it~to~an~optional~argument~before~mandatory~'#2'.
+    \c_@@_ignore_def_tl
+  }
+\msg_new:nnnn { cmd } { invalid-bang-on-mandatory }
   { Invalid~argument~prefix~'!'~in~#1. }
```
muzimuzhi, line 4765: I kinda changed my mind.

These two messages share the same `<text>` (`#3` of `\msg_new:nnnn`) and only differ in `<more text>` (`#4` of `\msg_new:nnnn`). Thus they can be combined.

muzimuzhi: Done in https://github.com/latex3/latex2e/pull/1199/commits/6ce39c0b3a29d7712c4be08bc7f880d3edd4f6e8, along with rebasing on `develop`.

PR #1209, opened 2023-12-16
Tagging support for array, tabularx, and longtable

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready for merging (further development and cleanup later)

- [x] Test file(s) added


base/lttagging.dtx
```diff
@@ -0,0 +1,789 @@
...
+%
+%    \begin{macrocode}
+\ExplSyntaxOn
+%    \end{macrocode}
+%
+%
+% \section{General support for tagged output}
+%
+%  \DescribeMacro\SuspendTagging
+%  \DescribeMacro\ResumeTagging
```
u-fischer, line 84: the commands take an argument. That is missing here in the description. (In the next tagpdf the argument of \tag_stop:n is only a debugging help, and is no longer used to handle the nesting).

base/lttagging.dtx
```diff
@@ -0,0 +1,789 @@
...
+%  activities, e.g., when we are doing trial typesetting that it is done
+%  several times. In such a case one must tag only the final version
+%  that is actually used, otherwise tagging structures are allowed
+%  which then do not end up in the PDF and confuse the mechanism. For
+%  this we have two commands that can be used in packages:
+%  \cs{SuspendTagging} and \cs{ResumeTagging}. They are available as
+%  part of the \LaTeX{} kernel, so that they can be safely used in
+%  packages whether or not tagging is requested. By default they do
+%  nothing.
+%
```
u-fischer, line 95: Currently tagging should not be stopped while the output routine is called (as header/footer/footnoterule etc should always be tagged as artifact). We probably will have to guard the output routine, but for now this restriction should at least be documented.

base/lttagging.dtx
```diff
@@ -0,0 +1,789 @@
...
+%  It is possible to use the tagging support sockets with
+%  \cs{UseSocket} directly, but in this case the socket remains active
+%  if \cs{SuspendTagging} is in force. There my be reasons for doing
+%  that but in general we expect to always use \cs{UseTaggingSocket}.
+%
+%  The L3 programming layer versions \cs{tag_socket_use:n} and
+%  \cs{tag_socket_use:nn} are slightly more efficient than
+%  \cs{UseTaggingSocket} because they do not have to determine how
+%  many arguments the socket takes when disabling it, so in code that
+%  is using the L3 progamming layer we recomment to use them instead
```
u-fischer, line 133: typos: programming   recommend

base/lttagging.dtx
```diff
@@ -0,0 +1,789 @@
...
+%    \end{macrocode}%
+%    Next line is is incomplete (only disables table sockets) and
+%    should instead be done in the above command instead of disabling them
+%    individually because that will otherwise forever change if new
+%    sockets get used.
+%    \begin{macrocode}
+      \__tag_tbl_disable:
+    }
+%    \end{macrocode}
+%
```
u-fischer, line 165: is the tbl_disable: needed? doesn't the table code use \UseTaggingSocket?

base/lttagging.dtx
```diff
@@ -0,0 +1,789 @@
...
+%    \begin{macrocode}
+    \cs_set:Npn \ResumeTagging #1 {
+      \tag_start:n {#1}
+      \__tag_tbl_enable:
+    }
```
u-fischer, line 170: and here similar

base/lttagging.dtx
```diff
@@ -0,0 +1,789 @@
...
+}
+%    \end{macrocode}
+%  \end{macro}
+%
+%  \begin{macro}{\tag_socket_use:n,
+%                \tag_socket_use:nn,
+%                \UseTaggingSocket,
+%               }
+%    Again this is not the final definition for the kernel; it is just
+%    a version to get going while some partsof the kernel support are
```
u-fischer, line 181: typo: parts of

base/lttagging.dtx
```diff
@@ -0,0 +1,789 @@
...
+}
+%    \end{macrocode}
+%  \end{macro}
+%
+%
+%
+%  \begin{macro}{\tbl_count_missing_cells:n}
+%    We might have the situation that some table package has not
+%    implemented the \cs{} in which case \cs{g_@@_table_cols_tl} would
+%    always be zero and we would get an error below when we try to
```
u-fischer, line 485: \cs{} ? command name missing?

base/lttagging.dtx
```diff
@@ -0,0 +1,789 @@
...
+%    column counters, if they stay internal we need something like
+%    this here (perhaps using \texttt{gincr} etc, or perhaps some
+%    other names in the first place).
+%    \begin{macrocode}
+\cs_new_protected:Npn \tbl_init_row_count: {
+  \int_gzero:N \g_@@_row_int
+}
+\cs_new_protected:Npn \tbl_incr_row_count: {
+  \int_gincr:N \g_@@_row_int
+}
```
u-fischer, line 637: should the command name use gincr?

base/lttagging.dtx
```diff
@@ -0,0 +1,789 @@
...
+%
+%    TODO: move to right .dtx file
+%    \begin{macrocode}
+\let\@kernel@refstepcounter\refstepcounter
+%    \end{macrocode}
+%    And no patch by hyperref (should be in longtable in the end)
+%    \begin{macrocode}
+\def\hyper@nopatch@longtable{}
+%    \end{macrocode}
+%
```
u-fischer, line 757: should the moves be done now?

base/lttagging.dtx
```diff
@@ -0,0 +1,789 @@
...
+%    Account for removed file (for now)
+%    \begin{macrocode}
+\ifcsname bitset_new:Nn\endcsname
+  \declare@file@substitution{l3bitset.sty}{null.tex}
+\fi
+%    \end{macrocode}
+%
```
u-fischer, line 764: this could go if a current pdfmanagement and tagpdf is moved into the texmf.

required/tools/array.dtx
```diff
@@ -2144,10 +2327,61 @@
 %    start the wanted =\halign=.
 %    \changes{1994/12/08}{v2.3b}{add \cs{tabularnewline}}
 %    \begin{macrocode}
-  \let\\\@arraycr \let\tabularnewline\\\let\par\@empty \@preamble}
+  \let\\\@arraycr \let\tabularnewline\\\let\par\@empty
```
u-fischer, line 2330: around 10 lines above (where I can't comment ...) is \m@th used. In the current version of the math-code it disables math grabbing which we don't want here (as it disables math inside all cells), so that should be changed into `\mathsurround\z@`.

required/tools/array.dtx
```diff
@@ -2486,11 +2766,41 @@
 %    =\@array=. To save \textsf{token} space we then redefine
 %    =\@preamble=
 %    because its replacement text isn't longer needed.
+%
```
u-fischer, line 2769: The $ around 20 lines above should be surrounded by something like `\FakeMathStart$\FakeMathStop `   to avoid that it triggers math tagging. The commands can be defined as dummies (in lttagging.dtx?) and the math code can then redefine them.

base/lttagging.dtx
FrankMittelbach: From a conceptional point of view tagging inside the OR is independent of tagging in the document, e.g., the OR has to establish its on tagging --- the interface  needs to be decided but it has to be available. The restriction as you call it isn't really a restriction that the user can control unless everything is inside a box, so rather than to document it we have to resolve it soon

base/lttagging.dtx
FrankMittelbach: at this point in time `\tag_start:n` doesn't handle the sockets so I had to do something for them. All of these are only stubs that tagpdf should provide real definitions for. Once they are there non of this other than the dummy kernel definitions would remain.

u-fischer: What do you mean? The sockets in the table module are set with `\UseTaggingSockets` so enabling/disabling should simply work with \tagstart/\tagstop without changing the plugs.

base/lttagging.dtx
u-fischer: yes I agree that the OR has to guard itself. On the other side, suspending  (completly stopping) tagging should imho be done only in small, controlled places, mostly if boxes are involved e.g. for measuring.

base/lttagging.dtx
FrankMittelbach: I mean that my version doesn't do that yet ... maybe yours does already, and once it does generally this code should be removed.

base/lttagging.dtx
FrankMittelbach: I already made a remark to that effect, but I think the interface names are wrong anyway and depending on whether or not we decide that the counters are directly available to programmers none of thesee commands are needed at all.

base/lttagging.dtx
FrankMittelbach: ??? if that is no longer necessary then for sure it can go.

required/tools/array.dtx
FrankMittelbach: where you can't comment?  why not? you should be able to see more of the surrounding by clicking on one of the arrows and then you shouldd be able to comment (or amy I mssing something?). But otherwise point taken :-)

required/tools/array.dtx
FrankMittelbach: We have to talk about that suggested  fakemath interface, I think it is wrong, but that isn't for this pull.

base/lttagging.dtx
u-fischer: no I meant, shouldn't we move that directly into the right places (ltxref.dtx and longtable.dtx probably)?

required/tools/array.dtx
u-fischer: well I can't. Comments can be added around 3 or four lines around a changed line but not in arbitrary places.

required/tools/array.dtx
u-fischer: It would be really good if there were directly an interface (can be internal)  to suppress the math grabbing here. Otherwise it is hard to tag math inside tabulars (one has to disable it outside the tabular and reenable it in every cell) and I need that for the larger math example ... (also it would resolve an open issue with tabularx).

base/lttagging.dtx
u-fischer: hm I'm missing something here. Your code has `\UseTaggingSocket` everywhere and if the definition is correct it should simply work ;-). I will have to test that.

u-fischer: simply \tag_start/\tag_stop works fine for me in your branch. And one could also remove the `\tl_put_right:Nn \@kernel@before@head {\__tag_tbl_disable:}` in the table code: as the header and footer contain a tag_stop tabular tagging is disabled automatically too.

base/lttagging.dtx
FrankMittelbach: agreed, but this is why in a new context (OR and possibly also floats) tagging should be re-established if it was active before  \SuspendTagging was executed. So it needs 2 control variables, one for the overall status (tagging=yes/no) and one for temporary suspension.

u-fischer: as I used now an int to keep track of start/stop one can use that to check for temporary suspension. Which means I should release the next tagpdf version ;-)

required/tools/array.dtx
FrankMittelbach: actually, if the latex-lab-math code is correctly implemented then \m@th should only trigger a signal that we have fake math here and that shouldn't stop grab inner math only the math at this level. Right now that's not the case, but it should be in the end (as \m@th is only a heuristic).

required/tools/array.dtx
FrankMittelbach: Again, I can't put anything in that is only in your private version of tagpdf, because if that then goes out as latex-dev it will blow up. Anway, in my opinion the interface should  just say `\FakeMathStart` hiding the `$` inside (and whatever is necessary to ensure that inner math is still grabbed and tagged --- which may be nothing if the command is protected)

base/lttagging.dtx
FrankMittelbach: yes but only if it includes proper definitions for `\SuspendTagging` and friends so we should be sure first that we want those with these names etc. Then we can drop the redefinitions in lttagging and only provide the dummies.

base/lttagging.dtx
FrankMittelbach: you are right, I thought that more is necessary, but it isn't

base/lttagging.dtx
FrankMittelbach: not sure I get it. this stops hyperref from doing patching, so it needs to be around when hyperref is loaded which may be well before longtable --- I haven't looked how the patching is done but usually by pushing something into begin document so it survives even if longtable is loaded later, no?

required/tools/array.dtx
u-fischer: Well we do *not* have fake math here and this occurence of `\m@th` should not trigger a signal. So we should use here `\mathsurround` or remove the redefinition of `\m@th` from the math code (or do both). (I don't think that removing the redefinition will lead to problems, url is still correctly tagged. ).  Which one do you prefer.

required/tools/array.dtx
u-fischer: Sorry? There is nothing about this in my private tagpdf. We simply need here something that allows us to say "this $ should not do math grabbing" and does not affect following math.

base/lttagging.dtx
u-fischer: ah yes, you are right. putting it in longtable could be too late. And in the latex-lab-table code wouldn't work if that is not loaded. So it should probably stay here until hyperref is adapted to detect the format.

FrankMittelbach: ah good, at least right on one account :-)

FrankMittelbach: so it stays in the kernel with a remark intil hyperref is adjusted and it can go out.

required/tools/array.dtx
FrankMittelbach: sorry not in your tagpdf but in your branch for math coding which is not released

u-fischer: but why should it blow up? I only asked that there is something  around the dollar which does nothing by default but can be redefined by the math code to suppress the math grabbing in this place.

required/tools/array.dtx
muzimuzhi: > well I can't. Comments can be added around 3 or four lines around a changed line but not in arbitrary places.

Support for this feature is on GitHub's public roadmap, but in category "Future".
https://github.com/github/roadmap/issues/281

Also I find in GitHub Rest API to [Create a review comment for a pull request](https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#create-a-review-comment-for-a-pull-request), the value of `line` parameter is limited to

> The line of the blob in the pull request diff that the comment applies to. For a multi-line comment, the last line of the range that your comment applies to.

base/lttagging.dtx
FrankMittelbach: now changed to g...

base/lttagging.dtx
FrankMittelbach: and when exactly is that?

u-fischer: it could be done an any time. I prepared a branch to move them into develop a few days ago, but got hold up as the tests could not run. I can run the tests now and merge into develop and you could pull into your branch. OK?

base/doc/ltnews39.tex
```diff
@@ -147,6 +147,33 @@ \section{Introduction}
...
+support for tagged tabulars. Some of the necessary support code has
+now been moved from \texttt{latex-lab} to the corresponding packages
+(using sockets and plugs) and to the \LaTeX{} kernel (for those parts
+that are also necessary for other aspects of tagging).
+
+The kernel code specific to tagging is implemented in the file
+\texttt{lttagging.dtx}.  For now it contains \cs{UseTaggingSocket}, a
+special invocation command for sockets that are specific to
+tagging. This enables us to also provide \cs{SuspendTagging} and
+\cs{ResumeTagging}, i.e., a very efficient way to temporary disable
```
davidcarlisle, line 162: temporarily

base/doc/ltnews39.tex
```diff
@@ -423,6 +450,20 @@ \subsection{Declaring appropriate sub-encodings for \texttt{TS1} symbol fonts}
 
 \section{Changes to packages in the \pkg{tools} category}
 
+
+\subsection{\pkg{array}, \pkg{longtable}, \pkg{tabularx}: support tagging}
+
+The three packages are now enabled to produce tagged tabulars upon
```
davidcarlisle, line 456: The three packages are now extended to enable producing tagged tabulars upon

base/doc/source2e.tex
```diff
@@ -342,6 +342,8 @@
 
  \DocInclude{ltoutput} % Output routine
 
+ \DocInclude{lttagging}% Tagging support
```
davidcarlisle, line 345: not an issue for this PR but since we went past a-z-A-Z for source2e files, the sorting of files in the code index is off as a-zA-Zaa-zz doesn't work with the current makeindex style.

required/latex-lab/latex-lab-table.dtx
```diff
@@ -740,65 +736,57 @@
...
+            \__tag_attr_new_entry:ee
+              {colspan-\int_eval:n{#1}}
+              {/O /Table /ColSpan~\int_eval:n{#1}}
+          }
+       \tl_set:Ne \l_@@_cellattribute_tl
+          {colspan-\int_eval:n{#1}}
+     }
+  }
+%    \end{macrocode}
+%    TODO: move to tagpdf
```
davidcarlisle, line 761: can you not move already?

required/tools/longtable.dtx
```diff
@@ -1066,7 +1081,42 @@
...
 %    \begin{macrocode}
 \def\LT@array[#1]#2{%
-  \refstepcounter{table}\stepcounter{LT@tables}%
+%    \end{macrocode}
+%
+%    With respect to tagging we have a complicated situation with
+%    longtable. When at the begin the \cs{endhead},
+%    \cs{endfirsthead}, \cs{endfoot} and \cs{endlastfoot} are used to
+%    setup head and foot they create each a structure subtree with one or
+%    more rows. From this structures we want to keep at most two (head
```
davidcarlisle, line 1090: From these

required/tools/longtable.dtx
```diff
@@ -1555,6 +1732,13 @@
 %    \begin{macrocode}
   \global\advance\LT@rows\@ne
   \ifnum\LT@rows=\LTchunksize
+%    \end{macrocode}
+%    At the end of the chunk \verb=\\= is doing something special and
+%    so we loose \verb=\tbl_count_missing_cells:n=. Below is about the
```
davidcarlisle, line 1737: lose

davidcarlisle (commented):
got to end of longtable.dtx (only made doc comments) out of time this morning but will pick up from here, but pushing these comments now

required/tools/longtable.dtx
```diff
@@ -1066,7 +1081,42 @@
...
+%    longtable. When at the begin the \cs{endhead},
+%    \cs{endfirsthead}, \cs{endfoot} and \cs{endlastfoot} are used to
+%    setup head and foot they create each a structure subtree with one or
+%    more rows. From this structures we want to keep at most two (head
+%    and foot) and move the foot to the end of the table. When the head
+%    and foot boxes are (re)inserted on following pages we want to mark
+%    them up as artifact with the exception of the head at the begin and
+%    the foot box at the end.
+%
+%    TODO: When a line is killed the structure subtree is there already
```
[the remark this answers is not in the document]
FrankMittelbach: yes, maybe , but it is probably not too hard for @u-fischer ... we'll see. I added a further note.

required/latex-lab/latex-lab-table.dtx
FrankMittelbach: probably may in fact has happened (this is the pr before Ulrike worked on it in her branch).

base/doc/source2e.tex
FrankMittelbach: ahh, good point. Could you open a separate pr that this should be looked at at some point?

required/latex-lab/latex-lab-table.dtx
u-fischer: I added it now to tagpdf and will remove it then here later, when tagpdf has been updated.

required/tools/longtable.dtx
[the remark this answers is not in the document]
u-fischer: hm, not easy. Currently \kill (with table-tagging) errors, and if I avoid the error it doesn't kill the row (don't know why). And if that problem can be resolved: getting tagging right is not trivial either, it would probably require some label and an additional counter to keep track on the rows that should be ignored. I'm inclined to make that a simply error with "sorry not supported with tagging".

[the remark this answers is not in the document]
FrankMittelbach: I guess part of the problem is that it acts "after the fact" so to speak. That means one would basically either wait with a lot of things to see what comes at the end of a row or maybe record the fact that there was a kill in row X and next time around do the killing  properly.
Instead if this we could perhaps have a \killstart command to be added at the beginning of the row which would tell the tagging what's up. That would still mean some manual work with such documents but at least the functionality could be preserved.

Perhaps @davidcarlisle  has some thoughts?

In a first version, just stating that this doesn't work with tagging is certainly ok.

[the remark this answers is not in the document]
davidcarlisle: I was going to suggest same, whether `\xkill ..... \\` would be easier so that knew up front that this was a killed row. Can you handle `\kill` in the kernel `tabbing` ?  (on which the longtable version was based)

davidcarlisle (approved):

u-fischer (approved):

PR #1229, opened 2023-12-29
Define `\ShowHookInfo` and `\LogHookInfo`.

jlaurens:

The actual `\ShowHook` is too verbose, cryptic and rather inefficient.

The `lthooks-info.tlg` contains many examples of `\ShowHook` output vs `\ShowHookInfo` output.
For example, we have 
```
-> The generic hook 'env/lthooks.regression test/before':
> The hook is not declared.
> The hook is empty.
```
for `\ShowHook` and
```
==> Hook: 'env/lthooks.regression test/before'
> buitin generic.
```
for `\ShowHookInfo`.

For non empty hooks, we have
```
-> The hook 'EMAN KOOH':
> Code chunks:
>     LABEL 1 -> \typeout {[code 1]}
>     LABEL 2 -> \typeout {[code 2]}
>     LABEL 3 -> \typeout {[code 3]}
> Document-level (top-level) code (executed first):
>     -> \typeout {[TOP LEVEL]}
> Extra code for next invocation:
>     ---
> Rules:
>     LABEL 3|LABEL 1 with relation <
>     LABEL 3|LABEL 2 with default relation <
> Execution order (after reversal and applying rules):
>     LABEL 2, LABEL 3, LABEL 1.
```
versus
```
==> Hook: 'EMAN KOOH'
> Code chunks:
>     'LABEL 1' -> \typeout {[code 1]}
>     'LABEL 2' -> \typeout {[code 2]}
>     'LABEL 3' -> \typeout {[code 3]}
>     top level -> \typeout {[TOP LEVEL]}
> Default rules, to be mirrored:
>     'LABEL 3' before 'LABEL 2'
> Rules:
>     'LABEL 3' before 'LABEL 1'
> Execution order:
>     1: 'top level'
>     2: 'LABEL 2'
>     3: 'LABEL 3'
>     4: 'LABEL 1'.
```

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [X] Test file(s) added


muzimuzhi:
Is backward compatibility such a heavy burden now, so that it's preferable to provide new commands like `\ShowHookInfo` rather than improving existing ones?

FrankMittelbach:
Backward compatibility is an important concept, but this doesn't have to be applied to debugging  commands that are temporarily inserted into to documents, e.g., there is no need to invent a new command name if we would change the display output of something like `\ShowHook`.

jlaurens:
Except that `\ShowHook` is more than a debugging facility as suggested by its rather weird documentation. It is also used for testing purposes (in nearly 40 files). Changing its output will certainly modify the purpose of the tests, thus possibly breaking the testing policy.
If testing with `\ShowHook` is not important, it is no big deal to change `\ShowHookInfo` into `\ShowHook`, and update the `.lgt` files once the new definition is accepted. In the meanwhile, having the 2 commands in parallel helps in making the decision because one can compare the outputs.
If testing with `\ShowHook` is important, it is certainly possible to replace in the `.lvt` files `\ShowHook` by `\TestShowHook`, which is only defined in an included test setup file. In that case, the `.lgt` files would not change.

As a starting point: is the `\ShowHookInfo` output OK?

FrankMittelbach:
> Except that `\ShowHook` is more than a debugging facility as suggested by its rather weird documentation.

not sure I like that statement given that it might have been me who wrote that documentation,  but let's keep that aside :-)

> It is also used for testing purposes (in nearly 40 files). Changing its output will certainly modify the purpose of the tests, thus possibly breaking the testing policy. 

It should certainly not change the "purpose"; it would change the result but in a controlled manner and that makes it easy to update the files (even though it means checking them)

> If testing with `\ShowHook` is not important, it is no big deal to change `\ShowHookInfo` into `\ShowHook`, and update the `.lgt` files once the new definition is accepted.

Or beforehand, which is simple enough given that there are to shell scripts that run the relevant tests and update the file so that one can look locally at the diffs. I don't really fancy pulling in something that fails the test suite unless it is on a separate branch.

>  In the meanwhile, having the 2 commands in parallel helps in making the decision because one can compare the outputs. 

true

> If testing with `\ShowHook` is important, it is certainly possible to replace in the `.lvt` files `\ShowHook` by `\TestShowHook`, which is only defined in an included test setup file. In that case, the `.lgt` files would not change.

not keen on that approach

> As a starting point: is the `\ShowHookInfo` output OK?

I noticed at least one typo  `buitin generic` and it should probably say 'top-level'. But I haven't looked  closely at the whole thing because I have a deadline coming up before the end of the year

car222222:
:-) @FrankMittelbach But then much of your text is "rather weird" -- be proud of it! :-)??

jlaurens:
One commit to get the best of two worlds ([\ShowHookInfo→ \ShowHook](https://github.com/latex3/latex2e/pull/1229/commits/c22dc04dda11798bf98c95bfc97a63db9722c71f))
- all checks passed with absolutely no ".lgt" change
- No new command created
- comparisons between old `\ShowHook` and new `\ShowHook` is available in `lthooks-show.tlg`.

base/testfiles-lthooks/helpers/lthooks-setup.tex
muzimuzhi, line 1: To me it's OK to update existing `.tlg` files due to changes from `\ShowHook` and `\LogHook`. The improvements in their outputs can be told from `git diff` on those existing `.tlg` files, hence not limited to the new `lthooks-show.tlg`.

Introducing duplicate `lthooks-setup.tex` files and adding `\input{./lthooks-setup.tex}` lines to _every_ `.lvt` files containing `\ShowHook`/`\LogHook` only make the test suite a bit more intricate.

u-fischer:
> `Hook: 'env/lthooks.regression test/before'`
> `Hook: 'EMAN KOOH'`

I'm not really keen on examples that uses spaces and dots in hook names (even if that is hidden in testfiles). And  `lthooks.regression test` looks definitively odd as an environment name.

muzimuzhi:
Sorry but I feel [recent commits](https://github.com/latex3/latex2e/pull/1229/files/c8489110e9b43bc1fbd268ac7e12e9504d9dac55..f3a3856b318a635bb94f318fc20fa0041a270d9b) which add and adjust `lthooks` documentation sections make this PR less on topic. It's not that the proposed changes in those commits are in poor quality, just that perhaps they can be proposed in a separate PR (under a different title), to  make each PR be better on-topic and easier to review.

jlaurens:
Intermediate commits dropped.

josephwright:
As it stands, the PR feels like it's becoming too broad. I think we all like the idea of extending the logging info, so a focussed PR for that would be best. The suggested doc changes need separate discussion, and likely a change in `l3doc` if they are to be integrated: _ad hoc_ stuff doesn't really work so well (and yes, I know there is some in places - but we'd like to get things consistent as far as possible).

Could you either trim down or (perhaps easier) open a new PR with just the logging stuff updated?

jlaurens:
This is on the way

josephwright:
The core idea here is good but we do need a focussed PR just adding the new functionality. Please open a new one that does just that job - I'm closing as otherwise this PR will sit for a long time unchanged.

PR #1235, opened 2024-01-07
Fix mistakes

ivan-pan:
I fixed some mistakes in the documentation such as `one the` (instead of `on the`).


required/tools/varioref.dtx
```diff
@@ -169,7 +169,7 @@
 % comes out as ``\ldots~see the example which shows~\ldots'', which
 % could be misleading.
 %
-% A space in front of |\vpageref| it will be ignored if
+% A space in front of |\vpageref| will be ignored if
 % the command doesn't produce any text at all.
```
muzimuzhi, line 173: Why? The original sentence looks right to me.

"A space (in from of ...) will be ignored (if ...)."

required/tools/varioref.dtx
```diff
@@ -324,7 +324,7 @@
 % \ChangedIn{2019}
 % introduced the star forms as a means to generating references
 % without hyperlinks (which is clearly the more important
-% application).
+% application)).
```
muzimuzhi, line 327: Or drop the nested parentheses,

```
% without hyperlinks, which is clearly the more important
% application).
```

required/tools/varioref.dtx
```diff
@@ -2344,7 +2344,7 @@
 %    the argument of |\is@pos@num@|. Note, that the added space in
```
muzimuzhi, line 2344: Since other two "Note, that ..." are already changed to "Note that ...", maybe also

```suggestion
%    the argument of |\is@pos@num@|. Note that the added space in
```

Though to me, a non-native speaker, these two forms both look OK.

FrankMittelbach:
Despite the name of the pr this isn't only about varioref but also about other areas of the code. So I guess @davidcarlisle should take a look at the graphics changes. But it might be better limit the amount of unrelated changes per pr in the future.

@ivan-pan I see that you are continue to make updates, could you say when it is worth doing a review?

required/tools/varioref.dtx
```diff
@@ -2131,7 +2131,7 @@
          \reftextfaraway{#3}}%
   \fi
 %    \end{macrocode}
-%    Finally we generate the internal label so that it can be check on
+%    Finally we generate the internal label so that it can be checked on
```
FrankMittelbach, line 2134: probably should be "Finally,"

required/tools/varioref.dtx
```diff
@@ -2252,7 +2252,7 @@
 %    The star form is easy (and just a convenient shortcut), just call
 %    |\ref| followed by |\vpageref| and pass the optional argument to
 %    the latter. We don't support the second optional argument of
-%    |\vpageref|, it that is needed one has to call both commands
+%    |\vpageref|, if that is needed one has to call both commands
 %    explicitly
```
FrankMittelbach, line 2256: explicitly.

FrankMittelbach (commented):
I'm ok with the changes to varioref (with the two or three comments added).
I haven't looked at the changes to the files in the graphic bundle, perhaps @davidcarlisle can do that?

required/graphics/epsfig.dtx
```diff
@@ -51,7 +51,7 @@
 \RequirePackage{graphicx}
 %    \end{macrocode}
 % \subsection{Emulation of `psfig' syntax}
-% Emulate "epsfig.sty", and most varieties of psfig
+% Emulate ``epsfig.sty", and most varieties of psfig
```
davidcarlisle, line 54: \texttt{epsfig.tex} not `\`\`.."` I think

required/graphics/epsfig.dtx
```diff
@@ -83,7 +83,7 @@
 %    \end{macrocode}
 % \end{macro}
 % \subsection{Emulation of `epsf' syntax}
-% Emulate Rokicki's "epsf.tex" supplied with the ever-popular dvips.
+% Emulate Rokicki's ``epsf.tex" supplied with the ever-popular dvips.
```
davidcarlisle, line 86: \texttt as above

davidcarlisle (changes requested):
all looks good except a couple of file names which would be better in tt rather than quotes (even though they were in bad quotes before, I think the `"..."` was intended to be shortverb)

required/tools/varioref.dtx
```diff
@@ -323,7 +323,7 @@
 % However, that too has problems (one is that \pkg{hyperref}
 % \ChangedIn{2019}
 % introduced the star forms as a means to generating references
-% without hyperlinks (which is clearly the more important
+% without hyperlinks, which is clearly the more important
 % application).
```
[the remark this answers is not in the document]
muzimuzhi: No, the matching beginning ( is on line 323, "(one is that hyperref introduced ...)".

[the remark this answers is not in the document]
car222222: Much better not to use parentheses here! In line 323 use : instead of the ( .  “ . . . problems: one is that . . . “

required/graphics/epsfig.dtx
car222222: And add , “ . . . , supplied . . . “

required/graphics/grfguide.tex
```diff
@@ -710,7 +710,7 @@ \subsection{Other commands in the \package{graphics} package}
  as the directory separator, even on Windows.)
 
  The default setting of this path is
- |\input@path| that is: graphics files will be found wherever \TeX\
+ |\input@path|, that is: graphics files will be found wherever \TeX\
```
car222222, line 713: Again better: " so that graphics . . . "

car222222 (changes requested):
In the graphics files, there are probably many other adjustments needed, to the original.  For example, many cases of "However" requiring a following ",".

required/tools/varioref.dtx
```diff
@@ -2264,7 +2264,7 @@
 %
 %  \begin{macro}{\vrefformat,\vrefdefaultformat}
 %    Japanese needs a different word order in \cs{vref@star} so this
-%    is separated out o that it can be changed on language level.
+%    is separated out so that it can be changed on language level.
```
car222222, line 2267: ==>  . . . changed on the language . . .  

And would "at" be better than "on" ?

(Many similar cases.)

required/tools/varioref.dtx
```diff
@@ -2389,7 +2389,7 @@
 %
 %  \begin{macro}{\fullrefformat,\fullrefdefaultformat}
 %    Japanese needs a different word order in \cs{fullref}  so this
-%    is separated out o that it can be changed on language level.
+%    is separated out so that it can be changed on language level.
```
car222222, line 2392: See above.

required/tools/varioref.dtx
```diff
@@ -2523,7 +2523,7 @@
 %
 %  \begin{macro}{\vrefrangeformat,\vrefrangedefaultformat}
 %    Japanese needs a different word order in \cs{vrefrange}  so this
-%    is separated out o that it can be changed on language level.
+%    is separated out so that it can be changed on language level.
```
car222222, line 2526: And again!

required/tools/varioref.dtx
```diff
@@ -2558,7 +2558,7 @@
 %
 %  \begin{macro}{\Vrefformat,\Vrefdefaultformat}
 %    Japanese needs a different word order in \cs{Vref}  so this
-%    is separated out o that it can be changed on language level.
+%    is separated out so that it can be changed on language level.
```
car222222, line 2561: Finally!

required/tools/varioref.dtx
```diff
@@ -1200,10 +1200,10 @@
   }}
 %    \end{macrocode}
 %    There are no good variants for German (I think and still think but
-%    this is a matter of taste :-).
+%    this is a matter of taste :-)).
 % \changes{v1.0b}{1994/01/31}{Replace in incorrect `def by `let}
 %    The following definitions were recently suggested to me but since
-%    the original are in for a long time i don't want to change them
+%    the original are in for a long time I don't want to change them
```
car222222, line 1206: are in for  ==>  has been there for

car222222, line 1206: "original" or "originals" ?

So maybe this shout be:

the originals have been there for ...

car222222 (changes requested):
This time I also looked more carefully at varioref.

required/tools/varioref.dtx
```diff
@@ -320,11 +320,11 @@
 % versions of the macros which do not add any space before the
 % generated text (they do nevertheless remove space at the left).
 %
-% However, that too has problems (one is that \pkg{hyperref}
+% However, that too has problems: one is that \pkg{hyperref}
 % \ChangedIn{2019}
 % introduced the star forms as a means to generating references
```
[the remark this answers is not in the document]
FrankMittelbach: you mean "of the label"? can't see any other "to"

required/tools/varioref.dtx
```diff
@@ -1036,7 +1036,7 @@
     \let\vrefrangeformat\vrefrangedefaultformat
   }}
 %    \end{macrocode}
-%    Default string for dutch have been contributed by Frank Poppe
+%    Default strings for dutch have been contributed by Frank Poppe
```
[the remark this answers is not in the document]
FrankMittelbach: I guess it would be correct to use always uppercase for language names. That is just lousy/lazy writing on my parts (or maybe copying in what others sent me)

required/tools/varioref.dtx
FrankMittelbach: I wouldn't know, but "at" sounds correct

FrankMittelbach (approved):
guess I agree with all the suggestions by Chris for varioref

required/tools/varioref.dtx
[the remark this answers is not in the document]
car222222: Clarification:

. . . as a means of generating . . .

required/tools/varioref.dtx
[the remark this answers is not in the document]
car222222: Easy to get this wrong!
The rule is, I think, clear when it is a noun, but using "french" as an adjective only sometimes seems to deserve Uppercase. 

And then there are cases when it is unclear whether it is being used as a noun or as an adjective!

required/tools/varioref.dtx
car222222: Again I was unclear!

The important thing is to use "the language level".

Since this is a technical term, it could be normal to use "on", but "at" sounds better to me.

josephwright:
Status here?

car222222:
I am unsure what, if anything, is left for me to review.

FrankMittelbach:
> I am unsure what, if anything, is left for me to review.

I don't think there is, my understanding is that this can be applied after fixing the grfguide conflict (which is trivial).

car222222:
Good to know.

davidcarlisle:
@ivan-pan  thanks for the edits

Issue #1238, opened 2024-01-12
ltkeys: Global options starting surrounded by spaces in the input are not removed from the unused option list


Skillmon:
## Brief outline of the bug

The current code used to remove keys from the unused options list doesn't first strip spaces from the key name, which results in false-positives in the unused options list.

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!

\begin{filecontents}{bugreport.sty}
\DeclareKeys{foo.store=\myfoo, bar.store=\mybar, baz.store=\mybaz}
\ProcessKeyOptions
\ProvidesPackage{bugreport}
\end{filecontents}

\documentclass[foo =bar, bar=baz, baz = foo]{article}

  % Any preamble code goes here
\usepackage{bugreport}

\typeout{%
  foo: \myfoo^^J%
  bar: \mybar^^J%
  baz: \mybaz^^J%
}

\begin{document}
\end{document}
```

## Log file (required) and possibly PDF file  

[bugreport.log](https://github.com/latex3/latex2e/files/13924262/bugreport.log)

Skillmon:
First found thanks to https://tex.stackexchange.com/questions/706935/why-do-i-get-unused-global-options-warning-for-options-forwarded-to-a-package

Skillmon:
It is fixed in `main` see https://github.com/latex3/latex2e/pull/1239. The `ltnews39` entry is still missing. Also, currently the fix wasn't merged into `develop`, so the `-dev` formats still throw the warning, while the normal formats don't.

tweh:
As far as I understood this issue is fixed however I came across this while implementing an expl3 class and the following MWE still show this issue (with TeXLive 2024, up-to-date on macOS, up-to-date)

```tex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!

\begin{filecontents}{democlass.cls}
\ProvidesExplClass {democlass} {2024/07/22} {1} {demo}
\LoadClassWithOptions { scrbook }
\tl_new:N \l_demo_word_tl
\keys_define:nn { democlass } {
  word .tl_set:N = \l_demo_word_tl,
}
\ProcessKeyOptions [ democlass ]
\newcommand{\myword}{\tl_use:N \l_demo_word_tl}
\end{filecontents}

\documentclass[word =xx]{democlass}

% Any preamble code goes here
\usepackage{bugreport}

\begin{document}
Test: \myword
[...]

PR #1239, opened 2024-01-12
Fix global options spaces (#1238)

Skillmon:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [x] Test file(s) added

### DIFF
--- a/base/testfiles-1run/github-1238.lvt
+++ b/base/testfiles-1run/github-1238.lvt
@@ -1,0 +1,18 @@
+\input{test2e}
+\START
+\AUTHOR{Jonathan P. Spratte}
+
+\OMIT
+\makeatletter
+\def\@unusedoptionlist{foo,bar,baz}
+\def\@raw@classoptionslist{foo =foo, bar=bar, baz =baz}
+\def\@currname{github-1238}
+\def\@currext{sty}
+\DeclareKeys{foo.store=\myfoo, bar.store=\mybar, baz.store=\mybaz}
+\TIMO
+
+\typeout{"\@unusedoptionlist"}%
+\ProcessKeyOptions
+\typeout{"\@unusedoptionlist"}%
+
+\stop
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -10,6 +10,10 @@ not part of the distribution.
 All changes above are only part of the development branch for the next release.
 ================================================================================
 
+2024-01-13 Jonathan P. Spratte <jspratte@yahoo.de>
+	* ltkeys.dtx: 
+	Trim spaces off key names taken from the raw-lists (gh1238)
+
 #########################
 # 2023-11-01 Release
 #########################
--- a/base/doc/ltnews38.tex
+++ b/base/doc/ltnews38.tex
@@ -552,15 +552,7 @@ details. They were essentially debugging tools which with hindsight should
 not have been made available directly in the kernel.
 
 
-\section{Bug fixes}
-
-\subsection{Handling of global keys with spaces}
-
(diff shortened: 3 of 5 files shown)


FrankMittelbach (changes requested):
I think it could do with a short ltnews entry given that this is user facing. Otherwise looks fine.

Skillmon:
I hope I picked the correct `ltnews`-file, and the short entry suffices.

FrankMittelbach (approved):

FrankMittelbach:
> I hope I picked the correct `ltnews`-file, and the short entry suffices.

actually, not ... it has to go into ltnews39 not 38 (which was for last release). Otherwise length is fine

FrankMittelbach (changes requested):
mvoe it to ltnews39

Skillmon:
There is no `ltnews39.tex` in `main`, I guess it's inside develop. Joseph told me that merging against main would be best for a patch-level, and against develop for the next release in summer. Since I thought this is no real change in behaviour, just fixing a minor bug, this could go with the next patch-release, and merged against main.

Any idea what to do in this case? I could change the base-branch to develop (and pull it in), and then add this section to `ltnews39`, or I add another PR, that just adds that text to 39 (or one of the team just moves the text).

FrankMittelbach:
I overlooked that you tried to patch into main. On LaTeX2e we never do this unless we do a hotfix and actually plan to have a patch-level release. I'm not sure we should do this for this issue -- maybe others see this differently.

FrankMittelbach:
If @josephwright  wants to make this a patch-level release for LaTeX, then I guess the simplest solution is that you take the change to ltnews38 out, and Joseph merges the rest (plus what is necessary for a patch release) then copies it over into develop and only then we put something into ltnews39.

Skillmon:
I can rework it to be merged into develop. It was more or less my decision to go for the patch, not Joseph's, he formulated more as if I should decide it, but maybe I misunderstood him. Happy to make any adaptation necessary here, so just decide.

davidcarlisle:
@FrankMittelbach It's only a warning but I think maybe we should do a hotfix/patch release. Avoiding "option clash" errors by using the new handler is a major selling pont but if you always get unused option warnings using the new code in practice (when kv options are often on separate lines) it may harm adoption.

FrankMittelbach:
> @FrankMittelbach It's only a warning but I think maybe we should do a hotfix/patch release. Avoiding "option clash" errors by using the new handler is a major selling pont but if you always get unused option warnings using the new code in practice (when kv options are often on separate lines) it may harm adoption.

if you are both in favor ... then do a patch release ... didn't have one in a while :-)

base/testfiles-1run/github-1238.lvt
```diff
@@ -0,0 +1,18 @@
+\input{test2e}
+\START
+\AUTHOR{Jonathan P. Spratte}
+
+\OMIT
+\makeatletter
+\def\@unusedoptionlist{foo,bar,baz}
+\def\@raw@classoptionslist{foo =foo, bar=bar, baz =baz}
+\def\@currname{github-1238}
+\def\@currext{sty}
```
muzimuzhi, line 10: I think it's OK to have a `filecontents` env in a test, to avoid all the internal macros used here.

There're already plenty uses of `filecontents` in existing test files.
```
$ rg --no-heading -F '\begin{filecontents}' -g 'testfiles*/**' | wc -l

     182
```

josephwright: True but I think not a blocker

PR #1246, opened 2024-01-19
lthooks doc: unified code display

jlaurens:
No `LaTeX` code change in this PR, only documentation is modified.

Actually, `verbatim` and `quote` environments are both used in `lthooks` documentation to display code chunks as well as terminal output. Both have pros and cons but the main fact is that they are not formatted in a consistent way.

The new environment `lthooksCode` aims to take the best of both worlds, and is consistent with the `syntax` environment.
The definition of `lthooksCode` makes heavy use of hooks, in many different aspects, which is interesting as a use case IMHO. However, the `lthooksCode/tool` hook may seem overkill at first glance, but this design anticipates forthcoming changes.

The main example of benefit is around line 1200 which corresponds to the explanation of the `\ShowHook` output at pages 13-14 of the [lthooks.pdf](https://filesender.renater.fr/?s=download&token=0fe38001-ecfb-402e-91dc-7797c173eab6) of this PR (available until 02/03). Notice that code line numbers are now links. In addition to the red color of line numbers, there is also an arrow mark  because some people are not sensitive to color changes.

This somehow reflects my personal tastes, suggestions are welcome.

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [ ] Test file(s) added
- [ X] Version and date string updated in changed source files

### DIFF
--- a/base/lthooks.dtx
+++ b/base/lthooks.dtx
@@ -32,7 +32,7 @@
 %
 %    \begin{macrocode}
 \def\lthooksversion{v1.1g}
-\def\lthooksdate{2024/01/17}
+\def\lthooksdate{2024/01/23}
 %    \end{macrocode}
 %
 %<*driver>
@@ -44,6 +44,7 @@
 
 \EnableCrossrefs
 \CodelineIndex
+
 \begin{document}
   \DocInput{lthooks.dtx}
 \end{document}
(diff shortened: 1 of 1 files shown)


u-fischer:
> The definition of lthooksCode makes heavy use of hooks, in many different aspects, which is interesting as a use case IMHO.

Sorry but I do think that this gives the wrong message. As the introduction says, hooks and hook rules help to coordinate code additions of  "*different* packages that do not know about each other". It doesn't look right to use them heavily to implement options for an environment that only we will use.  

> but the main fact is that they are not formatted in a consistent way.

But if you use a new, different *ad-hoc* environment in lthooks it means that the various LaTeX documentations are no longer formatted in a consistent way. 

Apart from this: the main project of the team is currently the Tagged PDF project and at some time in the not so far future we want to create tagged documentations. This will require some redefinitions and clean up, e.g. to correct lazy code formatting, but it also means that new environments should be defined so that they are prepared for tagging and not only give a visual pleasing output but also create a semantically sensible structure.

jlaurens:
As the name suggests, `lthooksCode` is an environment private to the source code `lthooks.dtx` such that no people except latex developers that look into this source code is aware of its existence, there is no worldwide message associate with it. This is just an API and if in some future `l3doc` ever provides us with a dedicated environment, aware of tagged pdfs, it will be extremely easy to use that environment either directly or as `lthooksCode` background. Meanwhile, this will allow for a more efficient documentation.

Regarding code display, we can see that the whole latex documentation is already a little inconsistent : `syntax` environment typeface is smaller than code examples, whereas it is more important. Which are the _exact_ explicit formatting rules expected, before `l3doc` gives a solution?

Here, the hook usage inside `lthooks.dtx` is not limited to only one environment as might appear at first glance. In fact, the same options can be set at the document level, at the environment level or at the line level and hook rules help in defining the relationships between these levels, which are different depending on the options. More generally, using hooks for one private environment is not always wrong, in particular when this environment has good chances to ever become public. Similarly, other places in the hook documentation explain that hooks are not limited to coordinate packages that "do not know about each other", despite this notion is a little difficult to understand.

jlaurens:
After comments form U. Fisher, the new output is very very close to the actual one, as far as quote and verbatim can be considered close of course. Some improvements in the background, with comments, mainly for future changes.
New [pdf available until 02/06](https://filesender.renater.fr/?s=download&token=254d152e-c2d7-486b-961f-1c51969335ad).

PR #1253, opened 2024-01-24
Adding tabs support for \verbatiminput*

### LINKED ISSUE #1245: \verbatiminput* from the verbatim package vs TABs
## Brief outline of the bug

Given that the kernel fixed `\verb*|A<tab>B|` to show the TAB as a space, it would be nice to also make `\verbatiminput*` mark TABs as spaces.

In the PDF we can see the space is marked, but the TAB is not.

First make a file with a tab between two letters, forexample

    A<tab>B<space>C

(`filecontens` converts TABs so cannot be used, and tabs are not shown one the web)

We'll assume the file is called `verb.txt` (I've also attached it below)

```latex
\documentclass{article}
\usepackage{verbatim}
\begin{document}
\verbatiminput*{verb.txt}
\end{document}
```

[verb.txt](https://github.com/latex3/latex2e/files/13960813/verb.txt)

[bug.log](https://github.com/latex3/latex2e/files/13960815/bug.log)

[bug.pdf](https://github.com/latex3/latex2e/files/13960818/bug.pdf)

daleif:
This fixes #1245 by adding tab marking support to \verbatiminput*

# Internal housekeeping

## Status of pull request

- Ready to merge

- [X] Test file(s) added


josephwright (approved):
Looks good to me: I guess the only question is whether something is needed in LaTeX News

FrankMittelbach:
> Looks good to me: I guess the only question is whether something is needed in LaTeX News

I think there should be a short one, after all it is a change of behavior.

daleif:
> > Looks good to me: I guess the only question is whether something is needed in LaTeX News
> 
> I think there should be a short one, after all it is a change of behavior.

Is that something I need to add?

FrankMittelbach:
> > > Looks good to me: I guess the only question is whether something is needed in LaTeX News
> > 
> > 
> > I think there should be a short one, after all it is a change of behavior.
> 
> Is that something I need to add?

need? probably not, but it would be appreciated :-), it needs doing by somebody and it is normally best to this in one go.

car222222:
If you can add this yourself, then yes please!  Or you could post your suggested text here (or in an email) and we can add it to the next latex news file.   Thanks!

daleif:
I would just write something like

Following the kernel update (XX) that allowed `\verb*` to mark tabs as spaces, the `\Lpack{verbatim}` package have been updated such that `\verbatiminput*` marks tabs as spaces as well.

FrankMittelbach:
thanks, will do the text in ltnews39 separately

PR #1268, opened 2024-02-09
Unprotect usesocket

u-fischer:
This moves the unprotected version of \socket_use:nw from latex-lab to ltsockets.

**Edit** Needs some more thoughts: the plugs must be unprotected too, and the debug message must be removed.

# Internal housekeeping

## Status of pull request

X Ready to merge

- [ ] Test file(s) added (not needed imho)
- [x ] Relevant `changes.txt` updated
- [ ] Rollback provided (if necessary)? (imho unneeded)
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated (imho unneeded)

### DIFF
--- a/base/ltsockets.dtx
+++ b/base/ltsockets.dtx
@@ -31,8 +31,8 @@
 %%% From File: ltsockets.dtx
 %
 %    \begin{macrocode}
-\def\ltsocketsversion{0.9a}
-\def\ltsocketsdate{2024-02-04}
+\def\ltsocketsversion{0.9b}
+\def\ltsocketsdate{2024-02-09}
 %    \end{macrocode}
 %
 %<*driver>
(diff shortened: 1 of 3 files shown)


base/ltsockets.dtx
```diff
@@ -919,10 +920,10 @@
...
 %
 %    \begin{macrocode}
-\cs_new_protected:Npn \socket_use:nw #1 {
-  \@@_debug_term:n
-    { Socket~ '#1'~ containing~ plug~
-      '\str_use:c { l_@@_#1_plug_str }'~ used. }
+\cs_new:Npn \socket_use:nw #1 {
+%  \@@_debug_term:n
+%    { Socket~ '#1'~ containing~ plug~
+%      '\str_use:c { l_@@_#1_plug_str }'~ used. }
```
FrankMittelbach, line 926: we should probably say why debugging can't be used. In fact we might want to consider supporting a special
\UseAlignSocket as I did had one time --- something to ponder (and perhaps make a remark here).

FrankMittelbach (changes requested):
two doc comments

josephwright:
Can you rebase onto current dev?

PR #1270, opened 2024-02-11
Gh1267

davidcarlisle:
## Status of pull request

- Ready to merge

- [ ] Test file(s) added
- [X] Version and date string updated in changed source files  (only date not version, in all cases)

### DIFF
--- a/base/ltcmdhooks.dtx
+++ b/base/ltcmdhooks.dtx
@@ -1,6 +1,6 @@
 % \iffalse meta-comment
 %
-%% File: ltcmdhooks.dtx
+%%% From File: ltcmdhooks.dtx
 %% Copyright (C) 2020-2023
 %% Frank Mittelbach, Phelype Oleinik, The LaTeX Project
 %
@@ -12,16 +12,20 @@
 %    https://www.latex-project.org/lppl.txt
 %
 %
-%%% From File: ltcmdhooks.dtx
+% \fi
 %
-\def\ltcmdhooksversion{v1.0i}
-\def\ltcmdhooksdate{2023/11/28}
-%
-%
+% \iffalse
+%%% From File: lthooks.dtx
 %
 %<*driver>
+% \fi
+\ProvidesFile{ltcmdhooks.dtx}
+             [2024/02/11 v1.0i LaTeX Kernel (Command hooks)]
+% \iffalse
+%
 \documentclass{l3doc}
-%\usepackage{ltcmdhooks}
+\GetFileInfo{ltcmdhooks.dtx}
+% \usepackage{ltcmdhooks}
 \EnableCrossrefs
 \CodelineIndex
 \begin{document}
@@ -38,7 +42,7 @@
 % \providecommand\phoinline[1]{\begin{quote}\itshape\footnotesize PhO: #1\end{quote}}
 %
 % \title{The \texttt{ltcmdhooks} module\thanks{This file has version
-%    \ltcmdhooksversion\ dated \ltcmdhooksdate, \copyright\ \LaTeX\
+%    \fileversion\ dated \filedate, \copyright\ \LaTeX\
 %    Project.}}
 % \author{Frank Mittelbach \and Phelype Oleinik}
 %
--- a/base/ltdirchk.dtx
+++ b/base/ltdirchk.dtx
@@ -36,8 +36,10 @@
 %</unstripped>
 %<*driver>
 % \fi
+% \fi
 \ProvidesFile{ltdirchk.dtx}
-             [2021/12/11 v1.3a LaTeX Kernel (System Dependent Parts)]
+             [2024/02/11 v1.3a LaTeX Kernel (System Dependent Parts)]
+% \iffalse
 % \iffalse
 \documentclass{ltxdoc}
 \GetFileInfo{ltdirchk.dtx}
--- a/base/ltfilehook.dtx
+++ b/base/ltfilehook.dtx
@@ -30,6 +30,8 @@
 %
 %%% From File: ltfilehook.dtx
 %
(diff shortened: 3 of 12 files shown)


base/ltcmdhooks.dtx
```diff
@@ -12,16 +12,20 @@
...
 %<*driver>
+% \fi
+\ProvidesFile{ltcmdhooks.dtx}
+             [2024/02/11 v1.0i LaTeX Kernel (Command hooks)]
+% \iffalse
+%
 \documentclass{l3doc}
-%\usepackage{ltcmdhooks}
+\GetFileInfo{ltcmdhooks.dtx}
+% \usepackage{ltcmdhooks}
```
FrankMittelbach, line 28: why not throw it out?

base/ltdirchk.dtx
```diff
@@ -36,8 +36,10 @@
 %</unstripped>
 %<*driver>
 % \fi
+% \fi
 \ProvidesFile{ltdirchk.dtx}
-             [2021/12/11 v1.3a LaTeX Kernel (System Dependent Parts)]
+             [2024/02/11 v1.3a LaTeX Kernel (System Dependent Parts)]
+% \iffalse
```
FrankMittelbach, line 42: how comes we need two \fi and two \iffalse? (might be correct ...can't check now but looks odd)

base/ltfilehook.dtx
```diff
@@ -30,15 +30,16 @@
...
-\providecommand\ltfilehookdate{2024/02/04}
-%    \end{macrocode}
-%
+%<structuredlog>\providecommand\ltfilehookdate{2024/02/11}
+%<structuredlog>\providecommand\ltfilehookversion{v1.0o}
 %<*driver>
-
+% \fi
+\ProvidesFile{ltfilehook.dtx}
+             [2024/02/11 v1.0o LaTeX Kernel (hooks)]
```
FrankMittelbach, line 38: I don't think we want to have to specify the date in more than one place -- or am I reading this diff wrong?

base/ltluatex.dtx
```diff
@@ -23,16 +23,14 @@
 %<tex>  \ProvidesFile\undefined\begingroup\def\ProvidesFile
 %<tex>  #1#2[#3]{\endgroup\immediate\write-1{File: #1 #3}}
 %<tex>\fi
-%<plain>\ProvidesFile{ltluatex.tex}%
+%<plain>\ProvidesFile{ltluatex.tex}[2024/02/11 v1.2c LuaTeX support for plain TeX (core)]
 %<*driver>
-\ProvidesFile{ltluatex.dtx}
+% \fi
+\ProvidesFile{ltluatex.dtx}[2024/02/11 v1.2c LuaTeX Kernel (LuaTeX support)]
```
FrankMittelbach, line 29: again the date in 2 places (that would always be the same). This is one of the reaons why I made macros for the date.

base/ltshipout.dtx
```diff
@@ -31,16 +31,16 @@
...
+% \iffalse
+%
 \documentclass{l3doc}
+\GetFileInfo{ltshipout.dtx}
 
-% bug fix fo l3doc.cls
+% bug fix for l3doc.cls
 \ExplSyntaxOn
 \cs_set_protected:Npn \__codedoc_macro_typeset_one:nN #1#2
   {
```
FrankMittelbach, line 46: is that bug fix still relevant? or fixed?

FrankMittelbach (changes requested):
not really a thorough review ... have by boarding ticket already, but not yet packed so running out of time ...

base/ltcmdhooks.dtx
FrankMittelbach: the commented out line I mean

davidcarlisle: yes I left it in to keep the diff focussed on the issue rather than rolling in unrelated cleanup, but it can certainly go.

base/ltdirchk.dtx
davidcarlisle: as above I left it that way  as that's what we did for all the original ones `ltclass.dtx`, `ltdefns.dtx`  etc all have  separate but adjacent `\iffalse` around the preamble text then the driver block. I think historically it allowed some experiments with moving the way the docstrip guards worked.

base/ltfilehook.dtx
davidcarlisle: I wondered if you'd say that:-)  I ended up with duplicated dates in two files (ltfilehook and ltluatex)  it may be possible to avoid it by adjusting the docstrip guards, or as you say we could keep the macros, I'll look again.

davidcarlisle: re-arranged to use \ltfilehookdate so date only given once

base/ltluatex.dtx
davidcarlisle: rearranged docstrip guards so date only given once

base/ltshipout.dtx
josephwright: Good questions: I'll check that (independent of PR)

josephwright:
Looks good to me - there are some additional tidy-ups it reminds us of, but I think we can merge.

josephwright (approved):

base/ltshipout.dtx
muzimuzhi: The missing-`\MacroFont` bug has been fixed in `l3doc.dtx` in latex3/latex3@34338536 committed on 2020-08-21, and then simplified in https://github.com/latex3/latex3/commit/a67bf5341092ba962c88681c1618d768f8f360c4.

Thus the redefinition of `\__codedoc_macro_typeset_one:nN` in `ltshipout.dtx` can be safely removed.

PR #1284, opened 2024-02-29
Add firstaid for acro

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/testfiles-lthooks/ltcmdhooks-001.tlg
+++ b/base/testfiles-lthooks/ltcmdhooks-001.tlg
@@ -184,6 +184,12 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid = 0 -> 
 Handled code for firstaid
+Update code for hook 'package/acro/after' on input line ...:
+Code labels for sorting:
+ firstaid
+Data structure for label rules:
+ firstaid = 0 -> 
+Handled code for firstaid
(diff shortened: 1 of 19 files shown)


muzimuzhi:
With l3kernel 2024-02-18 (or maybe 2024-02-20), loading `acro` won't raise errors any more, and since `acro.sty` always [clears `\l__acro_tmpa_prop` before setting to it](https://github.com/cgnieder/acro/blob/9a56e414b0eb82ff89241decc66da9620f2292df/code/acro.sty#L1114-L1129), it seems to me there's no urgent need to firstaid `acro`.

josephwright:
@muzimuzhi  Bruno's patch to make that work is meant to be temporary. I'm trying to cover the case where someone installs the TL'24 ISO then updates just `expl3`: ideally that should 'just work' even if we remove Bruno's patch (for performance reasons).

FrankMittelbach (approved):
Can't see anything wrong with this and giving the timeline I would apply the firstaid.

josephwright:
@FrankMittelbach  Good to go for all three?

FrankMittelbach:
guess so, but they probably don't play if merged one after another (without further updating) or do they?

josephwright:
No but I'll sort that :)

Issue #58, opened 2018-07-15
Calc : \global assignment of lenght does not work anymore. 


maieul:
## Brief outline of the bug
Normaly, a `\global` before a `\setlength` make a global assignment.
With the calc package, it does not work anymore,

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}
\documentclass{article}
\usepackage{calc}
\begin{document}
\newlength{\toto}
\begingroup
\global\setlength{\toto}{2pt}
\endgroup
\the\toto
\end{document}



```

## Log file (required) and possibly PDF file  
[toto.log](https://github.com/latex3/latex2e/files/2195922/toto.log)
[toto.pdf](https://github.com/latex3/latex2e/files/2195923/toto.pdf)

davidcarlisle:
sorry this is not a bug `\global` is never supported with `\setlength` although without `calc` it sort of half works by accident. With `calc` it does not work at all,

Issue #12, opened 2018-01-25
multicol with footnotes all in one column get outside of text area


joseluisquiroga:
!! Please fill out all sections after reading CONRIBUTING guidelines !!

## Brief outline of the bug

multicol with footnotes all in one column get outside of text area

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
%\documentclass{article}

  % Any preamble code goes here
\documentclass{book}
%\usepackage[showframe]{geometry}
%\usepackage[para,perpage]{footmisc}
\usepackage{multicol}


\begin{document}

  % Demonstration of issue here

\begin{multicols}{2}

Future autem ex multis
et per alia media concludit idem. Hic enim primo ostendit, quod anima \footnote{my note 7 at the margin}
rationalis \footnote{my note 8 at the margin} non est virtus in corpore, ita quod secundum esse vel
operationem vel utrumque ad \footnote{my note 9 at the margin} corporis harmoniam dependeat. \\

Future autem ex multis
et per alia media concludit idem. Hic enim primo ostendit, quod anima \footnote{my note 7 at the margin}
rationalis \footnote{my note 8 at the margin} non est virtus in corpore, ita quod secundum esse vel
operationem vel utrumque ad \footnote{my note 9 at the margin} corporis harmoniam dependeat. \\

Future autem ex multis
et per alia media concludit idem. Hic enim primo ostendit, quod anima \footnote{my note 7 at the margin}
rationalis \footnote{my note 8 at the margin} non est virtus in corpore, ita quod secundum esse vel
operationem vel utrumque ad \footnote{my note 9 at the margin} corporis harmoniam dependeat. \\

Future autem ex multis
et per alia media concludit idem. Hic enim primo ostendit, quod anima \footnote{my note 7 at the margin}
rationalis \footnote{my note 8 at the margin} non est virtus in corpore, ita quod secundum esse vel
operationem vel utrumque ad \footnote{my note 9 at the margin} corporis harmoniam dependeat. \\

Future autem ex multis
et per alia media concludit idem. Hic enim primo ostendit, quod anima \footnote{my note 7 at the margin}
rationalis \footnote{my note 8 at the margin} non est virtus in corpore, ita quod secundum esse vel
operationem vel utrumque ad \footnote{my note 9 at the margin} corporis harmoniam dependeat. \\

Future autem ex multis
[...]

PR #1291, opened 2024-03-11
Latex lab/tagging #12, #58 and by now #62

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


u-fischer:
The code no longer errors, but the tagging is not correct if the hang option of footmisc is used. Both the label and the text are in the same MC and so also in the same structure. I will try to look tomorrow.

FrankMittelbach:
> The code no longer errors, but the tagging is not correct if the hang option of footmisc is used. Both the label and the text are in the same MC and so also in the same structure. I will try to look tomorrow.

@u-fischer true. The hang code was not adding sockets for the label. But it looks as if we don't have a test for this (i.e., changing that didn't alter any  existing tests). Please take a look if it is now ok or if the socket is better applied differently.

u-fischer:
This works more or less with pdftex but fails with luatex: as the box is saved it has the attributes from the earlier context and so the mc is placed into the wrong structure. In the simple example from the issue one can reset the mc-attributes 

~~~~
\UseSocket{tagsupport/fntext/mark}{\@nameuse{tag_mc_reset_box:N}\@tempboxa\llap{\box\@tempboxa}}%
~~~~

But that wouldn't catch cases where the label contains additional structures (e.g. an \em or something like that). So perhaps one should additionally store the box with tagging disabled, at least for now.

u-fischer (approved):

PR #1302, opened 2024-03-19
Latex lab/blocks fix \itemindent bug and firstaid for verse

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


FrankMittelbach:
@wspr added you as a reviewer, becaus the firstaid could (and should) probably go into verse.sty (i.e., repeating the env definition at begin document)

u-fischer:
> @wspr added you as a reviewer, becaus the firstaid could (and should) probably go into verse.sty (i.e., repeating the env definition at begin document)



> @wspr added you as a reviewer, becaus the firstaid could (and should) probably go into verse.sty (i.e., repeating the env definition at begin document)

couldn't we instead move the redefinition of verse out of the beginenvironment hook and load it only in the standard classes?

FrankMittelbach:
> couldn't we instead move the redefinition of verse out of the beginenvironment hook and load it only in the standard classes?

would that help anything? Point is verse is defined in the classes not but it is defined basically in each and every class and as far as I know more or less always identically through copy and paste. So this way we enable it for all classes out there.

u-fischer:
> > couldn't we instead move the redefinition of verse out of the beginenvironment hook and load it only in the standard classes?
> 
> would that help anything? Point is verse is defined in the classes not but it is defined basically in each and every class and as far as I know more or less always identically through copy and paste. So this way we enable it for all classes out there.

but if tagging works with the original definition of verse.sty as you say in this PR (as \list is tagging aware) it should also work for most other other original verse definitions. The main improvement from the block code is currently only the additional optional argument, isn't it?

FrankMittelbach:
> > > couldn't we instead move the redefinition of verse out of the beginenvironment hook and load it only in the standard classes?
> > 
> > 
> > would that help anything? Point is verse is defined in the classes not but it is defined basically in each and every class and as far as I know more or less always identically through copy and paste. So this way we enable it for all classes out there.
> 
> but if tagging works with the original definition of verse.sty as you say in this PR (as \list is tagging aware) it should also work for most other other original verse definitions. The main improvement from the block code is currently only the additional optional argument, isn't it?

come to think of it that is probably true, so yes, what the block code does is that it adds the ability to adjust the layout via the key/val argument. So you are right, we could only do it in the standard clasess. However, that would then still break version so if somebody uses article.cls + verse then the block code would overwrite (unless we provide replacements for the standard classes ... after which something like "at begin doc" would no longer happen but we are far from providing new classes right now).

u-fischer:
> However, that would then still break version so if somebody uses article.cls + verse then the block code would overwrite

I'm not sure if I understand the argument. If we would move the verse redefinition into the class/article/after hook, then this combination should work fine, it would do what the firstaid does now as the verse package would come later and its redefinition would win.

u-fischer (approved):

Issue #1051, opened 2023-04-19
Missing documentation for `\DeclareFontShapeChangeRule`, `\fontseriesforce`, and `\fontshapeforce`


muzimuzhi:
## Brief outline of the enhancement

**LaTeX2e generally cannot add new features without an extreme amount of care to accommodate backwards compatibility. Please do not be offended if your request is closed for being infeasible.**

2020-02-02 release introduced several new commands related to font declaration. Thanks to @dr-scsi, some (or, most) of them are already documented in `fntguide`, see
- https://github.com/latex3/latex2e/pull/347
- https://github.com/latex3/latex2e/pull/349
- 213516d (updated documentation on \updefault and \ulcdefault, 2020-11-17)
- https://github.com/latex3/latex2e/issues/717 resolved by https://github.com/latex3/latex2e/pull/720

or the git history of [`base/doc/fntguide.tex`](https://github.com/latex3/latex2e/commits/develop/base/doc/fntguide.tex).

I find `\DeclareFontShapeChangeRule`, `\fontseriesforce`, and `\fontshapeforce` are undocumented yet.

FrankMittelbach:
@muzimuzhi that's a good argument ... well if it stays long enough and you have trouble with that then eventually ... but if you make an attempt that would help

FrankMittelbach:
@muzimuzhi still on point to produce a first doc draft?

josephwright:
@muzimuzhi  Status here?

dr-scsi:
In order to get this rolling, I suggest to start with:

```
diff --git a/base/doc/fntguide.tex b/base/doc/fntguide.tex
index 7cf5d239..0c8ef7cc 100644
--- a/base/doc/fntguide.tex
+++ b/base/doc/fntguide.tex
@@ -1997,6 +1997,55 @@ For example,
 would use |sb| (semi-bold) when |\rmfamily\bfseries| is requested in
 document.

+\subsection{Handling of current and requested font series and shape}
+
+In the original NFSS implementation, the series was a single attribute
+stored in |\f@series| and so one always had to specify both weight and
+width together.  Hence, it was impossible to typeset a paragraph in a
+condensed font and inside have a few words in bold weight (but still
+condensed) without doing this manually by requesting
+|\fontseries{bc}\selectfont|.
+
+\NEWfeature{2020/02/02}
+The new implementation now works differently by looking both at the
+current value of |\f@series| and the requested new series and out of
[...]

PR #1303, opened 2024-03-20
Describe \DeclareFont(Series|Shape)ChangeRule macros

dr-scsi:
* base/doc/fntguide.tex (subsection{Handling of current and requested font series and shape}): New section describing the macros \DeclareFontSeriesChangeRule and
\DeclareFontShapeChangeRule. (#1051)

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development

- [n/a] Test file(s) added

### DIFF
--- a/base/doc/fntguide.tex
+++ b/base/doc/fntguide.tex
@@ -2045,7 +2045,7 @@ A special value is |m| which is used to reset both weight and width.  In
 order to reset only one of them, the special values |?m| (reset width)
 and |m?| (reset weight) are provided, e.g.:
 \begin{verbatim}
-   \DeclareFontSeriesChangeRule{c}{m?}{c}{}
+   \DeclareFontSeriesChangeRule{bc}{m?}{c}{}
 \end{verbatim}
 
 The corresponding macro |\DeclareFontShapeChangeRule| is also provided
@@ -2063,6 +2063,16 @@ An example would be:
 If italics is the current shape and small caps is requested, switch to
 |scit| (small caps italics) and if that doesn't exist, try |scsl| (small
 caps slanted).
+
+Finally, it is also possible to overrule the entries in the lookup
+tables and forcibly select a series or shape with:
+\begin{decl}
+  |\fontseriesforce| \arg{series} \qquad |\fontshapeforce| \arg{shape}
+\end{decl}
+
+With the example above for the |c| series, issueing
+|\fontseriesforce{b}| means that series switches to |b| and not to |bc|.
+Same applies to |\fontshapeforce|.
 
 \subsection{Handling of nested emphasis}
 
--- a/base/doc/fntguide.tex
+++ b/base/doc/fntguide.tex
@@ -2028,8 +2028,10 @@ exist.  The example above now looks like this:
    \DeclareFontSeriesChangeRule{c}{b}{bc}{}
 \end{verbatim}
 which means: switch to the |bc| series if |c| is current and |b| is
-(additionally) requested, and do nothing if the current font doesn't
-have the combination.
+(additionally) requested, and if the current font doesn't have the
+combination, start the normal font substitution, i.e., switch back shape
+to |n| and if this combination doesn't succeed, switch back series to
+|m| as well, ending up with |m/n|.
 
 Another example is:
 \begin{verbatim}
@@ -2037,11 +2039,11 @@ Another example is:
 \end{verbatim}
 which means: if the current series is bold condensed (|bc|) and
 semi-condensed (|sc|) is requested (additionally), try bold
-semi-condensed (|bsc|) if available and stick to bold condensed if not.
+semi-condensed (|bsc|) if available but stay with bold condensed if not.
 
 A special value is |m| which is used to reset both weight and width.  In
-order to reset just the width, |?m| is provided and |m?| to reset just
-the weight, e.g.:
+order to reset only one of them, the special values |?m| (reset width)
+and |m?| (reset weight) are provided, e.g.:
 \begin{verbatim}
    \DeclareFontSeriesChangeRule{c}{m?}{c}{}
 \end{verbatim}
@@ -2051,8 +2053,16 @@ for setting database entries for font shapes:
 \begin{decl}
   |\DeclareFontShapeChangeRule| %
   \arg{current shape} \arg{requested shape} \\
-  \leavevmode\hfill    \arg{result} \arg{alternative result}
+  \leavevmode\hfill   \arg{result} \arg{alternative result}
 \end{decl}
+
+An example would be:
+\begin{verbatim}
+   \DeclareFontShapeChangeRule{it}{sc}{scit}{scsl}
+\end{verbatim}
+If italics is the current shape and small caps is requested, switch to
+|scit| (small caps italics) and if that doesn't exist, try |scsl| (small
+caps slanted).
 
 \subsection{Handling of nested emphasis}
 
--- a/base/doc/fntguide.tex
+++ b/base/doc/fntguide.tex
@@ -42,7 +42,7 @@
 
 \title{\LaTeXe{} font selection}
 
-\author{\copyright~Copyright 1995--2023, \LaTeX\ Project
+\author{\copyright~Copyright 1995--2024, \LaTeX\ Project
   Team.\thanks{Thanks to Arash Esbati for documenting the
     newer NFSS features of 2020}\\
   All rights reserved.%
@@ -52,7 +52,7 @@
    \texttt{fntguide.tex} for full details.}%
 }
 
-\date{October 2023}
+\date{March 2024}
 
 \begin{document}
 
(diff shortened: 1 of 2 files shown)


base/doc/fntguide.tex
```diff
@@ -1997,6 +1997,63 @@ \subsection{Font series defaults per document family}
...
+\arg{requested series} is the new series requested, \arg{result} is the
+combined value if it exists for the given font family, and
+\arg{alternative result} is a fallback in case \arg{result} doesn't
+exist.  The example above now looks like this:
+\begin{verbatim}
+   \DeclareFontSeriesChangeRule{c}{b}{bc}{}
+\end{verbatim}
+which means: switch to the |bc| series if |c| is current and |b| is
+(additionally) requested, and do nothing if the current font doesn't
+have the combination.
```
FrankMittelbach, line 2032: The second part of the statement is not correct or at least not understandable. What happens is that  "c" + "b" unconditionally switches to "bc" but if that doesn't exist in the font then the normal font substitution kicks in, i.e., shape is switched back to "n" (usually) if that doesn't help series switches back to "m" (usually).

source2e says:

Only a few entries have “alternative” values and perhaps most of them should get dropped. Or maybe not . . . needs some thought perhaps.

>The idea is that you don’t want the normal substitution to kick in because that would reset the shape first and it may be better to stay with b when a change to c is requested and bc doesn’t exist, than to go to first change the shape to n and then find that bc/n doesn’t exist either and thus ending up with m/n.

base/doc/fntguide.tex
```diff
@@ -1997,6 +1997,63 @@ \subsection{Font series defaults per document family}
...
+
+Another example is:
+\begin{verbatim}
+   \DeclareFontSeriesChangeRule{bc}{sc}{bsc}{bc}
+\end{verbatim}
+which means: if the current series is bold condensed (|bc|) and
+semi-condensed (|sc|) is requested (additionally), try bold
+semi-condensed (|bsc|) if available and stick to bold condensed if not.
```
FrankMittelbach, line 2040: but stay with bold condensed

base/doc/fntguide.tex
```diff
@@ -1997,6 +1997,63 @@ \subsection{Font series defaults per document family}
...
+
+A special value is |m| which is used to reset both weight and width.  In
+order to reset just the width, |?m| is provided and |m?| to reset just
```
FrankMittelbach, line 2043: to reset only one of them the special values |?m| (reset width) and |m?| (reset weight) are provided,e.g.,

base/doc/fntguide.tex
```diff
@@ -1997,6 +1997,63 @@ \subsection{Font series defaults per document family}
...
+\end{verbatim}
+
+The corresponding macro |\DeclareFontShapeChangeRule| is also provided
+for setting database entries for font shapes:
+\begin{decl}
+  |\DeclareFontShapeChangeRule| %
+  \arg{current shape} \arg{requested shape} \\
+  \leavevmode\hfill    \arg{result} \arg{alternative result}
+\end{decl}
+
```
FrankMittelbach, line 2056: \end{decl}

An example would be 
```
\DeclareFontShapeChangeRule {it}{sc}  {scit}  {scsl}
```
If italics is the current shape and small caps is requested switch to |scit| (small caps italics) and if that doesn't exist, try |scsl| (small caps slanted).

base/doc/fntguide.tex
```diff
@@ -42,7 +42,7 @@
 
 \title{\LaTeXe{} font selection}
 
-\author{\copyright~Copyright 1995--2023, \LaTeX\ Project
+\author{\copyright~Copyright 1995--2024, \LaTeX\ Project
   Team.\thanks{Thanks to Arash Esbati for documenting the
     newer NFSS features of 2020}\\
```
FrankMittelbach, line 47: Esbati and .... for documenting

FrankMittelbach (changes requested):

dr-scsi:
@FrankMittelbach - Thnx, I've incorporated your suggestions.  What is the preferred way to proceed, amend the changes to my last commit and force push, or just make a new commit with the adjustments?  I don't know how GitHub works in this regard.

FrankMittelbach:
> @FrankMittelbach - Thnx, I've incorporated your suggestions. What is the preferred way to proceed, amend the changes to my last commit and force push, or just make a new commit with the adjustments? I don't know how GitHub works in this regard.

you just make the changes in your version and then commit that would update the pull request and then some of the comments would then show as outdated.

FrankMittelbach:
when you are done just ask for a nother review via the github interface

FrankMittelbach:
if you like there could be a small section in ltnews39 stating that the commands are now documented in fntguide (but I could do that if you prefer).

dr-scsi:
> when you are done just ask for a nother review via the github interface

Thanks, I think I've got it 👍

dr-scsi:
> if you like there could be a small section in ltnews39 stating that the commands are now documented in fntguide (but I could do that if you prefer).

Yes, I can make a suggestion as well.  But would you add something about the macros `\fontseriesforce` and `\fontshapeforce` to `fntguide.tex`?  I had a look at their definitions in source2e.pdf, but I'm afraid I only have a vague idea about what they do, so before you re-write the whole thing ...

FrankMittelbach:
> Yes, I can make a suggestion as well. But would you add something about the macros `\fontseriesforce` and `\fontshapeforce` to `fntguide.tex`? I had a look at their definitions in source2e.pdf, but I'm afraid I only have a vague idea about what they do, so before you re-write the whole thing ...

you are right they should be described as well. They simply switch to the requested series or shape and do not look at the rules, so  current "c"  and `\fontseries{b}`  go to "bc" while ` \fontseriesforce{b}`  would go to "b"

base/doc/fntguide.tex
```diff
@@ -1997,6 +1997,73 @@ \subsection{Font series defaults per document family}
...
+which means: if the current series is bold condensed (|bc|) and
+semi-condensed (|sc|) is requested (additionally), try bold
+semi-condensed (|bsc|) if available but stay with bold condensed if not.
+
+A special value is |m| which is used to reset both weight and width.  In
+order to reset only one of them, the special values |?m| (reset width)
+and |m?| (reset weight) are provided, e.g.:
+\begin{verbatim}
+   \DeclareFontSeriesChangeRule{c}{m?}{c}{}
+\end{verbatim}
```
FrankMittelbach, line 2049: what's that example supposed to show? It requests changing c into c. better use something like `{bc}{m?}{c}{}`

Actually I do have such nonsense at least once in the table :-) (for "x")

FrankMittelbach (changes requested):
if you would add the \...force commands and an entry in ltnews that would be great

dr-scsi:
@FrankMittelbach - Thank you for your comments and clarification.  Please have a look again, I hope we're all set now.

base/doc/fntguide.tex
```diff
@@ -1997,6 +1997,83 @@ \subsection{Font series defaults per document family}
...
+|scit| (small caps italics) and if that doesn't exist, try |scsl| (small
+caps slanted).
+
+Finally, it is also possible to overrule the entries in the lookup
+tables and forcibly select a series or shape with:
+\begin{decl}
+  |\fontseriesforce| \arg{series} \qquad |\fontshapeforce| \arg{shape}
+\end{decl}
+
+With the example above for the |c| series, issueing
```
FrankMittelbach, line 2073: issuing

base/doc/fntguide.tex
```diff
@@ -1997,6 +1997,83 @@ \subsection{Font series defaults per document family}
...
+|\fontseriesforce{b}| means that series switches to |b| and not to |bc|.
```
FrankMittelbach, line 2074: that the series

base/doc/ltnews39.tex
```diff
@@ -513,7 +513,15 @@ \subsection{Declaring appropriate sub-encodings for \texttt{TS1} symbol fonts}
 %
 \githubissue{1257}
 
+\section{Documentation improvements}
 
+\subsection{Further updates to the guides}
+
+We reported about the updated versions of \texttt{usrguide} and
+\texttt{clsguide} in \LaTeX{} News~37~\cite{39:ltnews37}.  We have
```
FrankMittelbach, line 521: We have now

FrankMittelbach (approved):
3 minor points I noticed still open

dr-scsi:
@FrankMittelbach - Thanks, should be fixed now.

FrankMittelbach (approved):
thanks for the efforts ... as one can see (and not many people realize) this is always more work than they assume ... which is why things do not go as fas as people expect them to and wonder why we are always so slow :-)

dr-scsi:
> thanks for the efforts ...

Thanks again for guiding me through this.

> as one can see (and not many people realize) this is always more work than they assume ... which is why things do not go as fas as people expect them to and wonder why we are always so slow :-)

Well, things could have been faster if I've managed to get it right in the first place `;-)`.  But given how much you have on your plate, I wouldn't say things are slow.

PR #1305, opened 2024-03-21
firstaid for cleveref (five arguments in newlabel if xr-hyper ist used)

u-fischer:
This implements a first aid for cleveref for compability with xr-hyper and labels with five arguments.

See https://tex.stackexchange.com/a/708445/2388

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x ] Test file(s) added
- [x ] Relevant `changes.txt` updated


required/firstaid/latex2e-first-aid-for-external-files.dtx
```diff
@@ -655,6 +655,69 @@
...
+             \expandafter\@firstoffive#2}}% <-------- five
+       \def\cpageref@getref#1#2{%
+         \expandafter\let\expandafter#2\csname r@#1@cref\endcsname%
+         \expandafter\expandafter\expandafter\def%
+           \expandafter\expandafter\expandafter#2%
+           \expandafter\expandafter\expandafter{%
+             \expandafter\@secondoffive#2}}% <----------- five  
+%    \end{macrocode}
+%  We also need to redefine the internal label commands of cleveref.
+%  This must be done after cleveref, so we use an unique label.
```
FrankMittelbach, line 686: not sure I understand this: without a rule I don't think this is guaranteed to follow what cleveref puts into begindocument (it will initially, but ...).

I would guess you either have to do \RemoveFromHook{begindocument}[cleveref] or you use the cleveref label to add it in which case it should be appended.

FrankMittelbach (changes requested):

davidcarlisle (commented):
Looks good once Frank's coment re hook label addressed

required/firstaid/latex2e-first-aid-for-external-files.dtx
u-fischer: I would have expected that the order stays, but you are quite right, it makes much sense to use the existing cleveref label and to append to this chunk. (You can't remove first, there is other code in the hook which must stay)

FrankMittelbach: you are right, it uses \AtBeginDocument many times ... I didn't check for that. So yes, appending to it is the right thing to do (there the order is well-defined)

u-fischer:
should now be ok and can be merged.

required/firstaid/latex2e-first-aid-for-external-files.dtx
```diff
@@ -685,7 +685,7 @@
 %  We also need to redefine the internal label commands of cleveref.
 %  This must be done after cleveref, so we use an unique label.
 %    \begin{macrocode}
```
FrankMittelbach: This must be done after \pkg{cleveref} has made its changes in \texttt{begindocument} so we add it to the same hook using the hook label used by \pkg{cleveref}. This way it is guaranteed to overwrite the definitions.

FrankMittelbach (approved):
docu should change too, after that it could be merged I guess.

PR #1315, opened 2024-03-26
add key-value support for A-4F and A-4E

u-fischer:
This add the key-value definitions for `pdfstandard=A-4E` and `pdfstandard=A-4F`
(currently they differ from A-4 only in the xmp-metadata setting, if more is needed it will be add to l3pdfmeta.)

## Status of pull request

- Ready to merge

- [X] Test file(s) added
- [] `ltnewsX.tex` (and/or `latexchanges.tex`) updated not needed imho


FrankMittelbach:
I wonder what should happen with a request for A-4 + math or some other usages of AF files:
- just provide actual text support instead of AFs?
- change it (with a warning?) to A-4f?
- just carry on and let it fail with validators? as it would now?

u-fischer:
> I wonder what should happen with a request for A-4 + math or some other usages of AF files:
> 
> * just provide actual text support instead of AFs?
> * change it (with a warning?) to A-4f?
> * just carry on and let it fail with validators? as it would now?

we should at first clarify if the failures are actually ok and what is expected here. Even after rereading more than once the spec I don't get it.

u-fischer:
@FrankMittelbach  do you have any idea why the tests failed here? They work fine locally.

PR #1317, opened 2024-03-27
Sanitize artifact name

muzimuzhi:
In current CI settings, `/` would be part of an artifact name for some jobs, but `/` is one of [invalid characters in artifact names](https://github.com/zauguin/l3build-failure-artifacts/pull/1), as it cannot be part of a file name. Thus the step named "Archive failed test output" would fail for such jobs.

This PR adds a step which sanitizes artifact name, by replacing `/` with `_` .

See also https://github.com/zauguin/l3build-failure-artifacts/pull/1.

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [ ] Test file(s) added


u-fischer:
We should really test that, I so often have missing artifacts because of the `/` and that makes checking test failures tiresomes.

PR #1318, opened 2024-03-27
Exclude dups and non-docs from Documentation artifacts

muzimuzhi:
Before
  ```
  Documentation-base-1
  ├── base
  │   └── doc
  │       └── source2e.pdf
  ├── build
  │   └── doc
  │       └── source2e.pdf
  └── texmf
      └── tex
          └── latex
              └── mwe
                  ├── example-image-a.pdf
                  ├── example-image-b.pdf
                  ├── example-image-duck.pdf
                  └── example-image.pdf
  ```
After
  ```
  Documentation-base-1
  └── base
       └── doc
           └── source2e.pdf
  ```

Similar to https://github.com/latex3/latex3/pull/1410.

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -195,7 +195,11 @@ jobs:
         with:
           # example names: "Documentation-base-1", "Documentation-requires"
           name: Documentation-${{ matrix.module }}${{ matrix.environment && format('-{0}', env.LTX_DOC_COMPONENT) || ''}}
-          path: "**/*.pdf"
+          path: |
+            **/*.pdf
+            !build
+            !texmf
+            !**/testfiles*
           # Decide how long to keep the test output artifact:
           retention-days: 21
   # GitHub automatically informs the initiator of any action about the result, but
--- a/.github/workflows/pretest.yaml
+++ b/.github/workflows/pretest.yaml
@@ -173,6 +173,10 @@ jobs:
         with:
           # example names: "Documentation-base-1", "Documentation-requires"
           name: Documentation-${{ matrix.module }}${{ matrix.environment && format('-{0}', env.LTX_DOC_COMPONENT) || ''}}
-          path: "**/*.pdf"
+          path: |
+            **/*.pdf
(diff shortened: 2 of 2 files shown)


FrankMittelbach:
just looking at the suggested outcome above, I would think that this isn't right, simply because it is build/source2e.pdf that was failing not doc/...  Yes, they should be identical but something may have gone wrong elsewhere and the doc/ version isn't representing the failure.

For the same reason I'm not sure it is wise to drop the local texmf that was used, again it might contain rubbish and that could be important. 

Is this actually gaining much to do exclusions? After all the artifacts do not stay around a long time.

muzimuzhi:
> just looking at the suggested outcome above, I would think that this isn't right, simply because it is build/source2e.pdf that was failing not doc/... Yes, they should be identical but something may have gone wrong elsewhere and the doc/ version isn't representing the failure.

Like some `l3build` issue causing differences between same named PDFs in `build/` and `doc/`?

Also I doubt uploading PDFs in local `./texmf` would help. They are uploaded only because of their `.pdf` extension. If local `./texmf` is important in debugging generated docs, then maybe more parts of or even the whole `./texmf` directory should be uploaded.

I find myself only took snapshot-purpose into consideration when preparing this PR, in part because documentation artifacts are always uploaded, no matter previous steps succeeded or failed.

- When a user just wants to have a look at the in-developing latest LaTeX2e docs, a simpler directory structure in artifacts may be easier to use.
- When a developer wants to modify and/or debug the latest docs, their `.tex`/`.dtx` source files are the first ones they need. This quickly suggests either remote-editing or cloning repository and editing locally.

In both cases, I think getting one copy of generated docs is enough. The question of which copy (`./build` vs `./**/doc`) is more reliable is answered by `l3build`.

josephwright:
I think this looks OK: we would have to have a very odd situation for the artefacts to differ.

PR #1325, opened 2024-04-10
Correct mark-up for non-documented cmds

josephwright:


base/ltfilehook.dtx
```diff
@@ -750,9 +750,9 @@
 %  \LaTeXe{} core commands, because of some circular dependencies in the
 %  kernel we  do this only now and not in \texttt{ltfiles}.
 %
-% \begin{macro}{\InputIfFileExists}
-% \begin{macro}{\@input@file@exists@with@hooks}
-% \begin{macro}{\unqu@tefilef@und}
+% \begin{macro}[no-user-doc]{\InputIfFileExists}
```
FrankMittelbach, line 753: sure we don't want to document that one given that it is CamelCase.

TLC3 documents both \IfFileExists and this one and says

The command \InputIfFileExists tests not only whether file exists, but also inputs it immediately after executing then-code. The name file is then added to the list of files to be displayed by \listfiles .

But looking a bit further I guess it is just that the documentation for this command is in another file, so I guess making it no-user-doc here is correct.

josephwright: It's described in `ltclass.dtx`: we re-define it in `ltfilehooks.dtx`, but that means there's no user documentation _here_ - I guess we need a 'described elsewhere' marker but for the present `no-user-doc` seemed the best fit.

FrankMittelbach: yes that's what I was thinking also when I realised the the corresponding section was called "patching ....".
Probably a good idea to also add `described-elsewhere`and use that for such cases, or perhaps even `description=ltclass.dtx`

josephwright: OK, so perhaps leave these ones 'non-fixed' at the moment until we have a version of `l3doc` that covers this?

FrankMittelbach: Agreed, might be better to change l3doc first, and also fix the missing \MaintainedBy stuff there and then update the 2e .dtx file (for the latter there is an open issue)

josephwright:
We got a bit bogged down here - I've rebased and suspect we should merge as-is to at least address the easier cases.

FrankMittelbach:
> We got a bit bogged down here - I've rebased and suspect we should merge as-is to at least address the easier cases.

yes do I would say

PR #1327, opened 2024-04-10
Gh1262 and 1222

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


base/ltclass.dtx
```diff
@@ -1102,6 +1103,99 @@
...
+%    \end{macrocode}
+%    For rollback/rollforward pretend everything was available since
+%    the beginning of dawn.
+%    \begin{macrocode}
+%</2ekernel|latexrelease>
+%<latexrelease>\EndIncludeInRelease
+%<latexrelease>\IncludeInRelease{0000/00/00}%
+%<latexrelease>                 {\IfPackageLoadedT}{More conditionals}%
+%<latexrelease>
+%<latexrelease>\def\IfPackageLoadedT #1#2{\IfPackageLoadedTF{#1}{#2}{}}
```
davidcarlisle, line 1164: shouldn't this be undefining all these?

davidcarlisle (approved):
one comment inline re rollback

base/ltclass.dtx
FrankMittelbach: It is what we normally do in cases like this. We keep these kind of definitions around because in case of a rollback  we want them still available if used in packages that in themselves do not rollback, e.g. you have package abc that was available in 2010 but now you get the 2024 version that has \IfPackageLoadedT and in case of a rollback to 2015 you still run the 2024 package version, so it better keeps that command around in the hope that it does still work.

what additional text got lost or where?

davidcarlisle: > it is what we normally do in cases like this.
really? I'm not aware of anywhere else where we "rollback" to identical code to the new version. But I didn't search for any:-) Mostly in 0000/00/00 blocks I see things like `\let\obeyedline\@undefined`

the text of my comment originally got posted with just one line, despite the fact I'd written something like the text you see now. As I say I don't mind, but that argument could apply to any rollback it may break packages using newer stuff but there's no clear right and wrong here and I guess undefining won't actually help any use cases so I'm happy to go with it as it is.

FrankMittelbach:
> one comment inline re rollback

sorry must be dense ... I only see additions no drops in the commits

base/ltclass.dtx
muzimuzhi: > really? I'm not aware of anywhere else where we "rollback" to identical code to the new version.

For example, the rollback code for

- [`\If(Class|Package)Loaded(WithOptions)TF`](https://github.com/latex3/latex2e/blob/96fcbe83fcec7b3bcd720a0c70b1db7dd9607c62/base/ltclass.dtx#L1072-L1100)
- [`\IfExplAtLeastTF`](https://github.com/latex3/latex2e/blob/96fcbe83fcec7b3bcd720a0c70b1db7dd9607c62/base/ltexpl.dtx#L522-L540)

davidcarlisle: @muzimuzhi  ah thanks:-)

davidcarlisle:
> > one comment inline re rollback
> 
> sorry must be dense ... I only see additions no drops in the commits

that's just a reference to the (now resolved) comment on rollback use, so all good from me now.

PR #1330, opened 2024-04-17
Switch from \cs_argument_spec:N to \cs_parameter_spec:N

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added N/A
- [ ] Rollback provided (if necessary)? N/A
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated N/A

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -7,6 +7,10 @@ not part of the distribution.
 ================================================================================
 
 2024-04-17 Joseph Wright  <Joseph.Wright@latex-project.org>
+	* ltexpl.dtx, ltdefns.dtx
+	Rename \@expl@cs@argument@spec@@N to \@expl@cs@parameter@spec@@N (gh/1014)
+
+2024-04-17 Joseph Wright  <Joseph.Wright@latex-project.org>
 	* lttemplates.dtx
 	Use \IfInstanceExistsTF with an 's'
 
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -9,6 +9,8 @@ not part of the distribution.
 2024-04-17 Joseph Wright  <Joseph.Wright@latex-project.org>
 	* ltexpl.dtx, ltdefns.dtx
 	Rename \@expl@cs@argument@spec@@N to \@expl@cs@parameter@spec@@N (gh/1014)
+	* ltexpl.dtx, ltcmd.dtx, ltcmdhooks.dtx
+	Use \cs_parameter_spec:N in place of \cs_argument_spec:N (gh/1014)
 
 2024-04-17 Joseph Wright  <Joseph.Wright@latex-project.org>
 	* lttemplates.dtx
--- a/base/ltexpl.dtx
+++ b/base/ltexpl.dtx
@@ -33,7 +33,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{ltexpl.dtx}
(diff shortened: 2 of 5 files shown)


base/ltexpl.dtx
```diff
@@ -356,9 +356,19 @@
...
+%         {Update name of \pkg{expl3} function}
+% \changes{v1.3h}{2024/04/17}
+%         {Add a kernel-level copy of \cs{cs_parameter_spec:N}}
 %    \begin{macrocode}
 \cs_gset_eq:NN \@expl@cs@prefix@spec@@N \cs_prefix_spec:N
-\cs_gset_eq:NN \@expl@cs@argument@spec@@N \cs_argument_spec:N
+\cs_if_exist:NTF \cs_parameter_spec:N
+  { \cs_gset_eq:NN \@expl@cs@parameter@spec@@N \cs_parameter_spec:N }
+  { \cs_gset_eq:NN \@expl@cs@parameter@spec@@N \cs_argument_spec:N }
+\cs_gset_eq:NN \__kernel_cs_parameter_spec:N \@expl@cs@parameter@spec@@N
```
muzimuzhi, line 371: Does the new `\__kernel_cs_parameter_spec:N` need to be undefined in rollback code?

base/changes.txt
```diff
@@ -6,6 +6,12 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2024-04-17 Joseph Wright  <Joseph.Wright@latex-project.org>
+	* ltexpl.dtx, ltdefns.dtx
+	Rename \@expl@cs@argument@spec@@N to \@expl@cs@parameter@spec@@N (gh/1014)
+	* ltexpl.dtx, ltcmd.dtx, ltcmdhooks.dtx
+	Use \cs_parameter_spec:N in place of \cs_argument_spec:N (gh/1014)
```
muzimuzhi, line 13: Mention `\__kernel_cs_parameter_spec:N` too?

base/ltexpl.dtx
josephwright: I don't think so - this is strictly kernel-internl

base/changes.txt
josephwright: I don't think the changes needs to go to that level of detail myself

FrankMittelbach: We don't normally roll back internals unless there is a very good reason. Being internal they should not confliced with anything if they stay around.

FrankMittelbach: ups comment in the wrong place :-)

base/ltexpl.dtx
muzimuzhi: I don't insist on adding it, but is `\__kernel_cs_parameter_spec:N` more kernel-internal than `\@expl@cs@parameter@spec@@N`? The latter is undefined in rollback code.

FrankMittelbach: We usually undefine in rollback if there is a real chance that the new command was used in the past in documents or code. That really isn't the case for `\@expl@cs@parameter@spec@@N` so I gues it doesn't have to be undefined.  -- Guess it doesn't hurt either way

FrankMittelbach (approved):

PR #1335, opened 2024-04-22
Drop default option "info" of textcomp

### LINKED ISSUE #1333: Loading `textcomp` turns `LaTeX Info` messages into `Package textcomp Info` messages
## Brief outline of the enhancement

If you use a `TS1` symbol that is not provided by the current font, you get a `LaTeX Info` message by default:

```latex
\RequirePackage{latexbug}
\documentclass{article}

\begin{document}

\fontfamily{pplx}\selectfont \textrightarrow

\end{document}
```

```
LaTeX Info: Symbol \textrightarrow not provided by
            font family pplx in TS1 encoding.
            Default family used instead on input line 6.
```

However, if you load the package `textcomp`, you get a `Package textcomp Info` message instead:

```latex
\RequirePackage{latexbug}
\documentclass{article}
\usepackage{textcomp}

\begin{document}

\fontfamily{pplx}\selectfont \textrightarrow

\end{document}
```

```
Package textcomp Info: Symbol \textrightarrow not provided by
(textcomp)             font family pplx in TS1 encoding.
(textcomp)             Default family used instead on input line 8.
```

I find this inconsistent and it does not correspond to the description in [LaTeX News 31](https://www.latex-project.org/news/latex2e-news/ltnews31.pdf), which explicitly mentions LaTeX warnings and LaTeX errors:

> With the text companion symbols as part of the
kernel, it is normally no longer necessary to load the
textcomp package, but for backwards compatibility this
package will remain available. There is, however, one
use case where it remains useful: if you load the package
with the option error or warn then substitutions will
change their behavior and result in a LaTeX error or a
LaTeX warning (on the terminal), respectively. Without
the package the substitution information only appears
in the .log file. If you use the option quiet, then even
the information in the transcript is suppressed (which is
not really recommended).

As described above, the functionality of the `textcomp` package has been moved to the LaTeX kernel, so that the processes that generate the messages are now executed by the LaTeX kernel. The package itself is only there to change the status of the messages to "warning" or "error". However, even if the `textcomp` package is loaded without options (as in the example above), the `LaTeX Info` messages turn into `package textcomp Info` messages, as the `textcomp` code contains `\ExecuteOptions{info}`.

At least if the package `textcomp` is loaded without options, the `LaTeX Info` messages should not turn into `Package textcomp Info` messages, as the package really does nothing in this case. This often happens in real documents, as many packages still contain `\RequirePackage{textcomp}` just for compatibility.

muzimuzhi:
Closes #1333.

Change is tested by existing test file `github-0262.tlg`.

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [ ] Test file(s) added


base/lttextcomp.dtx
```diff
@@ -1527,14 +1527,11 @@
 %
 %    \begin{macrocode}
 %<*TS1sty>
-\providecommand\DeclareRelease[3]{}
-\providecommand\DeclareCurrentRelease[2]{}
-
-\DeclareRelease{}{2018-08-11}{textcomp-2018-08-11.sty}
```
FrankMittelbach, line 1533: I would keep that line (even though technically not necessary) and also add a line of docu like "For any rollback request before 2018-08-11 we make an attempt by loading the 2018  version."

muzimuzhi: Added back that line.

FrankMittelbach:
thanks once more!

PR #1338, opened 2024-04-27
Fix for gh1336

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Not yet  ready to merge 

- [x] Test file(s) added

There is an issue with rolling back graphics. It fails in `graphics.cfg` for luatex with `\outputmode` being undefined. Is this an issue with the graphics code?

A second potential issue are the lines
```
Package amsmath Warning: Unable to redefine math accent \hat.
Package amsmath Warning: Unable to redefine math accent \check.
Package amsmath Warning: Unable to redefine math accent \tilde.
...
```
Are those okay?


davidcarlisle:
> There is an issue with rolling back graphics. It fails in graphics.cfg for luatex with \outputmode being undefined. Is this an issue with the graphics code?

I don't think so I think it's just a limitation of rollback that the rolled back files can't always work in the context of a current installation.

graphics-cfg/graphics.cfg has

```
   \ifnum\luatexversion>85
      \ifnum\outputmode>0
      \chardef\x=4\relax
      \fi
    \else
      \ifnum\pdfoutput>0
      \chardef\x=1\relax
      \fi
```

so  if the engine is older than luatex 0.85 the engine has pdftex primitive names and pdftex.def s used by default.
If the engine is later than that then the luatex.def is used by default.  If you roll back graphics to a version that no longer tests for luatex but use it with a luatex that does not define pdftex names like \pdfoutput  it fails.

It would be possible to try to "roll back" the engine changes and re-enable the pdftex names cf the luatex85 package, but this seems more than expected for a package rollback.

davidcarlisle:
On Sat, 27 Apr 2024 at 10:19, Frank Mittelbach ***@***.***>
wrote:

> Not quite getting that. We run in an engine that is current even if we
> roll back and with one for which the .cfg does work without rollback. So
> what does rollback do that makes \outputmode undefined?
>

I'll trace exactly where it fails, but in general if you roll back graphics
(or latex) to a version that did not specifically test for luatex and
assumes that pdftex primitives are available, then things will go wrong if
you use that code with a current luatex where they are not available.

That's why I think I prefer 0000/00/00 date we use elsewhere as a general
indication of start date, it doesn't imply we tested with tex engines
current at that time. Using a specific date of 2001/03/03 or whatever gives
a false expectation that rolling back that far is a reliable emulation
even if you are just rolling back one package but all the binaries and
installed packages are current,

I started to test this but seems we more generally have an issue

```
\RequirePackage[2015-01-01]{latexrelease}
\documentclass{article}
```

fails in lualatex for that or any earlier date

```
LaTeX2e <2023-11-01> patch level 1
L3 programming layer <2024-03-14>
 (/usr/local/texlive/2024/texmf-dist/tex/latex/base/latexrelease.sty
! Undefined control sequence.
***@***.*** ...e \expandafter \ifx \lastnamedcs
                                                  \relax \else
***@***.***@d...

l.1358 ***@***.***{ExplSyntaxOff}
                                 ***@***.***}
?
````

Assuming rollback stops working at date XXX for graphics then perhaps we

**Comment by davidcarlisle** [comment]
> Not quite getting that. We run in an engine that is current even if we
> roll back and with one for which the .cfg does work without rollback. So
> what does rollback do that makes \outputmode undefined?
>

I'll trace exactly where it fails, but in general if you roll back graphics
(or latex) to a version that did not specifically test for luatex and
assumes that pdftex primitives are available, then things will go wrong if
you use that code with a current luatex where they are not available.

That's why I think I prefer 0000/00/00 date we use elsewhere as a general
indication of start date, it doesn't imply we tested with tex engines
current at that time. Using a specific date of 2001/03/03 or whatever gives
a false expectation that rolling back that far is a reliable emulation
even if you are just rolling back one package but all the binaries and
installed packages are current,

I started to test this but seems we more generally have an issue


```
\RequirePackage[2015-01-01]{latexrelease}
\documentclass{article}
```

fails in lualatex for that or any earlier date

```
LaTeX2e <2023-11-01> patch level 1
L3 programming layer <2024-03-14>
 (/usr/local/texlive/2024/texmf-dist/tex/latex/base/latexrelease.sty
! Undefined control sequence.
\@ifundefined ...e \expandafter \ifx \lastnamedcs 
                                                  \relax \else \@ifundefin@d...

l.1358 \@ifundefined{ExplSyntaxOff}
                                 {}{\latexrelease@postltexpl}
? 
```

**Comment by davidcarlisle** [comment]
If you roll back earlier than 2015 in luatex you do

```
\ifx\directlua\undefined\else
\IncludeInRelease{2015/10/01}{\luatexluafunction}
                             {LuaTeX (prefixed names)}%
    \directlua{tex.enableprimitives("",%
                 tex.extraprimitives("omega", "aleph", "luatex"))}
\EndIncludeInRelease
\IncludeInRelease{0000/00/00}{\luatexluafunction}
                             {LuaTeX (prefixed names)}%
\directlua{
  tex.enableprimitives(
    "luatex",
```

or rather it's not that adding the luatex prefixed ones makes things fail we go further and do

```
 end
  for _,i in pairs(t) do
    tex.print(
      "\noexpand\\let\noexpand\\" .. i
        .. "\noexpand\\undefined"
    )
```

and undefine all the unprefixed names.

So either we need to test for that everywhere or (much simpler) just be less agressive with the emulation there and define the prefixed names in addition to the unprefixed ones
which means all the luatex primitives are unavailable without the `\luatex` prefix and then most things, including `latexrelease.sty` itself, fail.

**Comment by FrankMittelbach** [comment]
> I'll trace exactly where it fails, but in general if you roll back graphics
> (or latex) to a version that did not specifically test for luatex and
> assumes that pdftex primitives are available, then things will go wrong if
> you use that code with a current luatex where they are not available.

but wouldn't that be resolvable if latexrelease loads simply loads luatex85 if in luatex and a rollback prior to a certain date is requested?

**Comment by davidcarlisle** [comment]
> but wouldn't that be resolvable if latexrelease loads simply loads luatex85 if in luatex and a rollback prior to a certain date is requested?

yes probably, but first we need to fix that you can't roll back before 2015 at all with luatex currently. even luatex85.sty  assumes the luatex primitives are available unprefixed

**Comment by FrankMittelbach** [comment]
> > 
> 
> yes probably, but first we need to fix that you can't roll back before 2015 at all with luatex currently. even luatex85.sty assumes the luatex primitives are available unprefixed

or we accept that as a cutoffpoint for LuaTeX and put a warning into latexelease that rolling back before 2015 with the luatex engine is not supported (and errors are likely to happen).

**Comment by davidcarlisle** [comment]
OK to cutoff but shouldn't just make a warning we should make sure that at least it avoids the undefined command error loading latexrelease, but simplest way to do that I think is just remove the lua loop undefining all the primitives. we could undefine them all except \lastcsnamedcs  so at least latexrelease.sty gets to the end, but that's starting to be weird.

**Comment by FrankMittelbach** [comment]
> OK to cutoff but shouldn't just make a warning we should make sure that at least it avoids the undefined command error loading latexrelease, but simplest way to do that I think is just remove the lua loop undefining all the primitives. we could undefine them all except \lastcsnamedcs so at least latexrelease.sty gets to the end, but that's starting to be weird.

Couldn't we just drop the undefining completely? I mean we do that elsewhere too in cases where newer commands are likely to be used in packages that do not roll back. Yes, I know e undefine in case somebody had defined the commands in their code or document, but maybe keeping the primitives here is better overall causing less isses on the whole in the rollback case.

**Comment by davidcarlisle** [comment]
> Couldn't we just drop the undefining completely?

yes I think removing the entire undefine loop works best (but as I just commented in #1336 it's still not enough to make graphics work in -dev versions)

**Comment by davidcarlisle** [comment]
I redid tests with a clean build just removing the undefine loop and I think all works as expected, if you roll back far enough you get

```
LaTeX Warning: Suspicious rollback/min-date date given
               A minimal date of 2009/07/12 has been specified for package
               'epstopdf-base'.
               But this is in conflict with a rollback request to 2000-01-01.

```

but it still works, and that would not be fixable here.


@FrankMittelbach I could push that change to this branch?

**Comment by FrankMittelbach** [comment]
Yes please do.

**Review by davidcarlisle — APPROVED** [review]
some changes already pushed...

File: required/tools/multicol.dtx [hunk 1]
```diff
@@ -92,13 +92,13 @@
 %<package>\DeclareRelease{v1.8}{2019-10-01}{multicol-2019-10-01.sty}
 %<package>\DeclareCurrentRelease{}{2021-11-15}
 %<package>
+%<package>\NeedsTeXFormat{LaTeX2e}[2018-04-01]
+%<package>
```
user227621, line 96: Is `\NeedsTeXFormat{LaTeX2e}[2018-04-01]` really needed here? It was changed from `\NeedsTeXFormat{LaTeX2e}[1997/12/01]` in https://github.com/latex3/latex2e/commit/921f5fb086325f3e1ff2a36efd7c10e4c7ccdee5 but only because the package rollback mechanism was introduced in that commit. Maybe this should be changed back into `\NeedsTeXFormat{LaTeX2e}[1997-12-01]`?

required/tools/multicol-2019-10-01.sty
```diff
@@ -103,12 +103,6 @@
 %%
 \NeedsTeXFormat{LaTeX2e}[2018-04-01]
```
user227621, line 104: Same here

required/tools/multicol.dtx
FrankMittelbach: good point. It probably should now

FrankMittelbach: On second thought I don't think it should (or needs to be). The file  does use the rollback commands which aren't there if you really try use it with a release prior to 2018. Of course it now comes after  getting undefined csname then, but as a record the date is still not wrong.

In the rollback situation it doesn't apply because the other file is loaded instead.

required/tools/multicol-2019-10-01.sty
FrankMittelbach: here it should go otherwise one gets a warning between 2017 and 2018 rollback

required/tools/multicol.dtx
user227621: Wouldn't `\NeedsTeXFormat{LaTeX2e}[2018-04-01]` then also have to be added to all other files that contain rollback declarations? These do not yet contain such a line, although they do not really differ from `multicol` in terms of rollback.

On the other hand, I think that `\NeedsTeXFormat{LaTeX2e}[2018-04-01]` only has a very limited benefit. It only comes into effect if you use the new `multicol` is together with a pre-2018 kernel _and_ continue after two `Undefined control sequence` errors (`\DeclareRelease` and `\DeclareCurrentRelease`) ...

required/amsmath/amsmath.dtx
```diff
@@ -76,17 +76,15 @@ Bug reports can be opened (category \texttt{#1}) at\\%
...
 %
-%    Providing a rollback to earlier version(s)
+%    Providing a rollback to earlier version(s). Anything prior to
+%    2018 we attempt to fulfill by using the version from 2018.
 %    \begin{macrocode}
-\providecommand\DeclareRelease[3]{}
-\providecommand\DeclareCurrentRelease[2]{}
-%
-\DeclareRelease{}{2018-12-01}{amsmath-2018-12-01.sty}
+\DeclareRelease{}{1994-06-01}{amsmath-2018-12-01.sty}
```
user227621, line 82: For `textcomp` you have argued (https://github.com/latex3/latex2e/pull/1335#discussion_r1577494948) that it is better to keep the line `\DeclareRelease{}{2018-08-11}{textcomp-2018-08-11.sty}`, even though it is not technically necessary, so that you have

```latex
\DeclareRelease{}{1997-12-01}{textcomp-2018-08-11.sty}
\DeclareRelease{}{2018-08-11}{textcomp-2018-08-11.sty}
```

Is there a reason why this is not applied uniformly? E. g. for `amsmath`

```latex
```

required/graphics/graphics.dtx
```diff
@@ -17,17 +17,16 @@
           \ProvidesFile{graphics.dtx}
 %</dtx>
 %<+package>\NeedsTeXFormat{LaTeX2e}[1995/12/01]
-%<+package>\providecommand\DeclareRelease[3]{}
-%<+package>\providecommand\DeclareCurrentRelease[2]{}
 %<+package>
-%<+package>\DeclareRelease{}{2017-06-25}{graphics-2017-06-25.sty}
+% Try to fulfill rollback requests to early years by using the 2017 version:
+%<+package>\DeclareRelease{}{1994-06-01}{graphics-2017-06-25.sty}
```
user227621, line 22: Same here

required/tools/array.dtx
```diff
@@ -25,10 +25,10 @@
 %% Package `array' to use with LaTeX 2e
 %% Copyright (C) 1989-1998 Frank Mittelbach, all rights reserved.
 %<+package>\NeedsTeXFormat{LaTeX2e}[1995/06/01]
-%<+package>\providecommand\DeclareRelease[3]{}
-%<+package>\providecommand\DeclareCurrentRelease[2]{}
 %<+package>
-%<+package>\DeclareRelease{}{2016-10-06}{array-2016-10-06.sty}
+% For anything before 2016-10-06 we load the 2016 version and hope for the best:
+%<+package>\DeclareRelease{}{1994-06-01}{array-2016-10-06.sty}
```
user227621, line 30: Same here

required/tools/multicol.dtx
```diff
@@ -86,20 +86,19 @@
...
           \ProvidesFile{multicol.dtx}
 %</dtx>
-%<package>\NeedsTeXFormat{LaTeX2e}[2018-04-01]
-%<package>
-%<package>\providecommand\DeclareRelease[3]{}
-%<package>\providecommand\DeclareCurrentRelease[2]{}
 %<package>
-%<package>\DeclareRelease{}{2017-04-11}{multicol-2017-04-11.sty}
+% For any rollback before 2017 load the 2017 version and hope for the best:          
+%<package>\DeclareRelease{}{1994-06-01}{multicol-2017-04-11.sty}
```
user227621, line 91: Same here

required/tools/showkeys.dtx
```diff
@@ -29,17 +29,15 @@
 %</dtx>
 %<package>\NeedsTeXFormat{LaTeX2e}
 %<package>
-%<package>\providecommand\DeclareRelease[3]{}
-%<package>\providecommand\DeclareCurrentRelease[2]{}
-%<package>
-%<package>\DeclareRelease{}{2014-10-28}{showkeys-2014-10-28.sty}
+% For every rollback request before 2022 use the 2014 version:
+%<package>\DeclareRelease{}{1994-06-01}{showkeys-2014-10-28.sty}
```
user227621, line 33: Same here

required/tools/varioref.dtx
```diff
@@ -33,14 +33,14 @@
 %%
 %
 %<package>\NeedsTeXFormat{LaTeX2e}[1995/05/16]
-%<package>\providecommand\DeclareRelease[3]{}
-%<package>\providecommand\DeclareCurrentRelease[2]{}
 %<package>
-%<package>\DeclareRelease{}{2016-02-16}{varioref-2016-02-16.sty}
+%
+% For any rollback before 2019 load the 2016 version and hope for the best:
+%<package>\DeclareRelease{}{1994-06-01}{varioref-2016-02-16.sty}
```
user227621, line 39: Same here

PR #1350, opened 2024-05-19
ltcmd: implement SplitKeyval argument processor

Skillmon:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [x] Test file(s) added


josephwright:
We'll look at this after the June 2024 release ...

josephwright:
Could you rebase onto develop?

josephwright:
Was just looking back at this -could you rebase and update dates/`changes.txt` and I'll look at it properly.

josephwright (approved):
Looks good to me, though we will want a `changes.txt` entry and probably something for `ltnews` (they can be added later).

Skillmon:
Done.

Another question: `expkv-cs` has the `...` notation to collect any undefined keys encountered in an argument instead of throwing an error. Is that something I should add here as well (would be similar to `\keys_set_known:nnN` but forward the undefined keys as another element in the split arguments)?

base/doc/ltnews41.tex
```diff
@@ -597,6 +597,32 @@ \subsection{Collecting environment bodies verbatim}
 \end{verbatim}
 \end{center}
 
+\subsection{Simple key=value argument processor}
+
+For very simple key=value setups with only a handful simple keys a new argument
```
davidcarlisle, line 602:
```suggestion
For very simple key=value setups with only a handful simple keys, a new argument
```

base/doc/ltnews41.tex
```diff
@@ -597,6 +597,32 @@ \subsection{Collecting environment bodies verbatim}
...
+processor \cs{SplitKeyval} is added that works similar to \cs{SplitArgument} but
```
davidcarlisle, line 603:
```suggestion
processor \cs{SplitKeyval} is added that works in a similar way to \cs{SplitArgument} but
```

base/doc/ltnews41.tex
```diff
@@ -597,6 +597,32 @@ \subsection{Collecting environment bodies verbatim}
...
+values will be the special \verb|-NoValue-| marker or the value given to it when
+it is declared. The values will be forwarded in the same order the keys were
+declared. Thus
+\begin{verbatim}
+  \NewDocumentCommand\MyKeyvalCmd
+    {>{\SplitKeyval{keyA=A,keyB,key=C}}m}
+    {\FunctionOfThree#1}
+\end{verbatim}
+defines a macro that knows three keys. If during usage any key doesn't get a
+value an empty value will be forwarded. We can then use the macro and get the
```
davidcarlisle, line 615:
```suggestion
value an empty value, will be forwarded. We can then use the macro and get the
```

davidcarlisle (changes requested):
The code seems Ok I just made some comments on the text in the news.

But more generally it would be good to document the motivation for this. The documentation comments in a few places it should only be used for small lists of keys, but doesn't really say what advantage it has for small lists over the presumed alternative that would work for larger key sets of passing `#1` to a standard keyval processor so saving the three keys as \tmpa \tmpb and \tmpc then calling \ExpArgs{ooo}\FunctionOfThree{\tnpa}[\tmbb}{\tmpc}`

Is this quicker?

Skillmon:
> Is this quicker?

No, it isn't. In fact it's slower by roughly a factor 2 because it does the key-setup and then parsing on each usage instead of a one-off setup.

But it's much simpler, so is easier for document authors (I'd say this isn't really meant for package authors), compare the following:

```latex
\NewDocumentCommand\processor{>{\SplitKeyval{keyA=A,keyB,keyC=C}}m}
  { \use_none:nnn #1 }
```

vs

```latex
\keys_define:nn { test }
  {
     keyA .tl_set:N = \l_test_keyA_tl
    ,keyA .initial:n = A
    ,keyB .tl_set:N = \l_test_keyB_tl
    ,keyB .initial:V = \c_novalue_tl
    ,keyC .tl_set:N = \l_test_keyC_tl
    ,keyC .initial:n = C
  }
\NewDocumentCommand\keysusing{m}
  {
    \group_begin:
      \keys_set:nn {test} {#1}
      \exp_args:NNVVV
    \group_end:
    \use_none:nnn \l_test_keyA_tl \l_test_keyB_tl \l_test_keyC_tl
  }
\exp_args_generate:n {NVVV}
```

(even if the group handling shouldn't apply, the code becomes only marginally simpler, and for users it could involve expansion handling, depending on their use case.)

So the *only* selling point is indeed simplicity.

base/doc/ltnews41.tex
Skillmon: I think the comma should be after the first "value" here, not after the second?

davidcarlisle (approved):
I guess simplicity argument is OK since it isn't adding a lot of code

base/doc/ltnews41.tex
davidcarlisle: yes, sorry

FrankMittelbach:
I am a bit concerned that code like this ends up in packages and then people complain about L3/ LaTeX being slow. So I guess it's a big disclaimer that this is only for user level and not for package code, including advise what to do on the package level instead.

davidcarlisle:
@FrankMittelbach  @Skillmon  that's my concern as well. I don't think "typical document use" is as complicated as the above comment suggests, in particular you don't need to drop into L3 syntax

```
\documentclass{article}

\NewDocumentCommand\threeargs{mmm}{-#1-#2-#3-}
\DeclareKeys[me]{
  key1 .store = \tmpa,
  key2 .store = \tmpb,
  key3 .store = \tmpc
}
\NewDocumentCommand\keyargs{O{}}{%
  \renewcommand\tmpa{defa}%
  \renewcommand\tmpb{defb}%
  \renewcommand\tmpc{defc}%
  \SetKeys[me]{#1}%
  \ExpandArgs{ooo}\threeargs\tmpa\tmpb\tmpc
  }
\begin{document}

\keyargs

\keyargs[key2=TWO]

\end{document}
```

This isn't really significantly more complicated than the version here.
It doesn't group the temp commands but the calling command can group if that us needed.

Skillmon:
Feel free to drop it :) I could write a fast version by doing some caching, setting it up during command-definition etc., but that would turn out to get rather complicated for something that's generally just a small and less powerful variant of something already there.

So I don't have any bad feelings here. TBH, I already forgot about it; it just came up last week in the TeX.SX chat because some user asked me whether this was available nowadays.

josephwright:
OK, based on the review comments I'll close.

PR #1360, opened 2024-05-30
Gh1359

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added

This is a straight bug in code that hasn't been used that much yet, so I think a rollback isn't really needed.

I don't understand one test file change: github-1041, why that goes from no room for insert to no room for count in luaTeX, maybe @davidcarlisle can verify that as it is a test for code written by him.

Otherwise I think this would be ready to go.


josephwright:
I took a look at the failure in `github-1041` and I think I see what's happening. With LuaTeX, we can create inserts in the extended pool, which means that the code path deviates from the classical one. Doing that, we call `\extrafloats`, and that does some allocation. What's happening I think is the extra count being consumed is meaning that rather than hit the buffers in `\newinsert`, we take a path in `\extrafloats` that tries to allocate just a count - and that now fails.

The code for allocators is all a bit baroque, not least because there are a lot of conditionals for pre-e-TeX, etc. I think for the present we live with the change of `.tlg` here, but perhaps should look to re-work the allocator code a bit more widely to make the logic clearer. (Not exactly a high priority, but would help with this sort of edge case thing.)

josephwright (approved):

davidcarlisle (approved):
as @josephwright  says test 1041 will be affected by any change in the number of counts registers allocated  by the end of the document class loading, so an update here is expected.

PR #1361, opened 2024-05-31
Gh1359 b

FrankMittelbach:
Internal housekeeping

## Status of pull request

- Ready to merge (for this or the next release)

- [x] Test file(s) added


josephwright (approved):
Looks good. Will need to go into develop and main: I will do that if necessary.

davidcarlisle (approved):
looks good to me too modulo ongoing discussion of whether to remove the ids. In most cases it should be enough to simply expand the mark so let the ids go, so I think this is OK.

FrankMittelbach:
> looks good to me too modulo ongoing discussion of whether to remove the ids. In most cases it should be enough to simply expand the mark so let the ids go, so I think this is OK.

but this *is* removing the marks, eg if you do `\FirstMark...` what is returned is the original  data without the mark now. one could use the data without that removal but this way it is simpler for use cases like the one Pieter outlined.

davidcarlisle:
> > looks good to me too modulo ongoing discussion of whether to remove the ids. In most cases it should be enough to simply expand the mark so let the ids go, so I think this is OK.
> 
> but this _is_ removing the marks, eg if you do `\FirstMark...` what is returned is the original data without the mark now. one could use the data without that removal but this way it is simpler for use cases like the one Pieter outlined.

sorry use:v  yes all good then,

Issue #1380, opened 2024-06-17
missing character warnings from \xetex \showhyphens


davidcarlisle:
## Brief outline of the bug

xelatex warns of missing characters from `\showhyphens` despite  `lltfssbas` documentation saying they are suppressed, adding  `\tracinglostchars\z@`  would fix

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}
\documentclass[12pt]{article}

\usepackage{fontspec}
\setmainfont{FreeSans}
\begin{document}

\makeatletter
\DeclareRobustCommand\Xshowhyphens[1]{% remove the X to see fix
  \setbox0\vbox{%
    \usefont{TU}{lmr}{m}{n}%
    \hsize 1sp %
    \hbadness\@M
    \hfuzz\maxdimen
    \tracingonline\z@
    \tracinglostchars\z@%%%%%%%%%%<<<<<<<
    \everypar={}%
    \leftskip\z@skip
    \rightskip\z@skip
    \parfillskip\z@skip
    \hyphenpenalty=-\@M
    \pretolerance\m@ne
    \interlinepenalty\z@
    \clubpenalty\z@
    \widowpenalty\z@
    \brokenpenalty1127 %
    \setbox\z@\hbox{}%
    \noindent
    \hskip\z@skip
    #1%
    \par
     \loop
     \@tempswafalse
     \ifnum\lastnodetype=11\unskip\@tempswatrue\fi
     \ifnum\lastnodetype=12\unkern\@tempswatrue\fi
     \ifnum\lastnodetype=13 %
      \count@\lastpenalty
      \unpenalty\@tempswatrue
    \fi
    \ifnum\lastnodetype=\@ne
     \setbox\tw@\lastbox\@tempswatrue
     \setbox0\hbox{\unhbox\tw@\unskip\unskip\unpenalty
                   \ifnum\count@=1127 \else\ \fi
                   \unhbox0}%
     \count@\z@
    \fi
    \if@tempswa
    \repeat
   \hbadness\z@
   \hsize\maxdimen
   \showboxdepth\z@
   \tolerance\m@ne
   \hyphenpenalty\z@
   \noindent\unhbox\z@
}}

\showhyphens{ինչու ինչպես որտեղ հիվանդանոց հիվանդանոցում Երևան Երևանում Նամակ Երևանիս ԵՐԵՎԱՆ
[...]

PR #1381, opened 2024-06-17
set tracinglostchars for #1380

davidcarlisle:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [X] Test file(s) added
- [n/a] Rollback provided (if necessary)? not needed I think, there is already rollback to pre 2017 version

### DIFF
--- a/base/testfiles-TU/github-1380.luatex.tlg
+++ b/base/testfiles-TU/github-1380.luatex.tlg
@@ -1,0 +1,6 @@
+This is a generated file for the LaTeX2e validation system.
+Don't change this file in any respect.
+luaotfload | db : Font names database loaded from ../luaotfload-names.luc.gz
+Underfull \hbox (badness 10000) in paragraph at lines 5--5
+[] \TU/lmr/m/n/10 ի󰀀ն󰀀չ󰀀ո󰀀ւ󰀀 ի󰀀ն󰀀չ󰀀պ󰀀ե󰀀ս󰀀
+\hbox(0.0+0.0)x16383.99998, glue set 4902.88477, direction TLT []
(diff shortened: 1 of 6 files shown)


base/testfiles-TU/github-1380.luatex.tlg
```diff
@@ -0,0 +1,6 @@
+This is a generated file for the LaTeX2e validation system.
+Don't change this file in any respect.
+luaotfload | db : Font names database loaded from ../luaotfload-names.luc.gz
```
FrankMittelbach, line 3: should we try to supress line 3 or would it always come out like this? E.g., move the test after \begin{document}?

FrankMittelbach (approved):
looks fine to me (with one query)

base/testfiles-TU/github-1380.luatex.tlg
davidcarlisle: I think it's just luaotfload being chatty, every luatlatex has that line somewhere in the log, it just got picked up here as it's the first typesettung done (and inconvenient to pre-use the font before `\START` as this was in the preamble, and the format wasn't quite the same as xetex anyway so it would still have needed a `.luatex.tlg` even without this line

FrankMittelbach: if it is always the same line then it doesn't make a difference I agree, I was just wondering if it might change, in which case one could have moved the hyphenation test after begin document (or so I thought) and use \START then

davidcarlisle: i adjusted to avoid the luaotfload message

Issue #1383, opened 2024-06-19
TS1 fd file input every run with Unicode engines


davidcarlisle:
## Minimal example showing the current behaviour

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{article}

\begin{document}
x
\end{document}
```

This inputs

```
/usr/local/texlive/2024/texmf-dist/tex/latex/base/ts1cmr.fd
```

at begin document as latex checks the default setup for TS1 encoding. 8 bit formats have already input this file but for Unicode engines ts1lmr.fd is preloaded but the cm one is then loaded at run time.

We could preload the fd file during format creation at the time other fd files are preloaded.

PR #1384, opened 2024-06-19
Gh1383

davidcarlisle:
## Status of pull request

- Ready to merge

- [X] Test file(s) added
- [ ] Rollback provided (if necessary)? no rollback here I think
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) essentially internal change

Preload ts1cmr.fd in unicode engines for #1383


base/fontdef.dtx
```diff
@@ -411,6 +411,18 @@
 \endgroup
 %    \end{macrocode}
 %
+% \changes{v3.0j}{2024/06/24}{load ts1 cmr fd file in Unicode engines}
+% Even though Unicode engines default to |lm| load |ts1cmr| 
+% as this may be used for fallback for TS1 encoding.
```
FrankMittelbach, line 416: THinking about that, I don't think there is anything wrong to set the fallback to lm in case of Unicode engines if that font is loaded into the format. But the way it is now is probably ok as well.

FrankMittelbach (approved):
with a comment

base/fontdef.dtx
davidcarlisle: yes I did wonder but probably we say in several places that cm is the default so for now at least I just changed the preload (which shouldn't change any output as the same fd file got loaded at run time anyway) I'll merge

Issue #1386, opened 2024-06-20
rollback error


u-fischer:
## Minimal example showing the bug

```latex
\RequirePackage{latexbug}
\RequirePackage[2024-05-01]{latexrelease}

\documentclass{article}
\begin{document}

This is NOT an item.

\begin{itemize}
\item This is an item.
\end{itemize}

\end{document}
```

errors with 

~~~~
! LaTeX Error: Something's wrong--perhaps a missing \item.
~~~~

This rollback is not applied:
~~~
Already applied: [2021/06/01] minipage fix on input line 9924.
~~~

and so the toks register at the end of `\g__para_standard_everypar_tl` is missing.

## Log file (required) and possibly PDF file  
[test-utf8.log](https://github.com/user-attachments/files/15915278/test-utf8.log)

muzimuzhi:
The rollback code for `\g__para_standard_everypar_tl` labeled `minipage fix` has two versions. Rolling back to 2024/05/01 applies the 2023/06/01 version, thus it's OK that the 2021/06/01 version logs "Already applied".
```
Applying: [2023/06/01] minipage fix on input line 9899.
Already applied: [2021/06/01] minipage fix on input line 9924.
```

It seems the 2023/06/01 version lacks something like
https://github.com/latex3/latex2e/blob/022e71422ed86784d086d252e0c3e47b4e51238e/base/ltpara.dtx#L851-L857

FrankMittelbach:
what I'm not so sure about is the question of whether or not this can wait until the fall release (not being able to roll back to the 2023-11 release) or wther that should go into a  hotfix

josephwright:
We should fix asap I guess: PL is really no major issue

FrankMittelbach:
In that case the code should go into a hotfix and the ltnews entry only in develop with a mentioning that this was already resolved as a hotfix in PL2

FrankMittelbach:
in PL2

PR #1387, opened 2024-06-21
Append \everypar toks to \g__para_standard_everypar_tl

muzimuzhi:
Fix rollback 2023/06/01 of \g__para_standard_everypar_tl, labeled "minipage fix". See #1386.

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [x] Test file(s) added

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,6 +6,12 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2024-06-23  Yukai Chou  <muzimuzhi@gmail.com>
+
+	* ltpara.dtx
+	Append \everypar toks to \g__parar_standard_everypar_tl, rollback
+	2023/06/01 (gh/1386)
+
 ================================================================================
 All changes above are only part of the development branch for the next release.
 ================================================================================
--- a/base/doc/ltnews39.tex
+++ b/base/doc/ltnews39.tex
@@ -727,8 +727,8 @@ logic for the strut changes when vertical mode is detected.
 
 \subsection{Fix a ``missing \cs{item}'' rollback error}
 
-If \LaTeX{} is rolled back to a date not newer than 2023/06/01 and
-older than 2024/06/01, any list-based environment would raise an error
+If \LaTeX{} is rolled back to a date between 2023/06/01 (inclusive) and
+2024/06/01 (exclusive), any list-based environment would raise an error
 \begin{verbatim}
 ! LaTeX Error: Something's wrong--perhaps a missing \item.
 \end{verbatim}
--- a/base/doc/ltnews39.tex
+++ b/base/doc/ltnews39.tex
@@ -725,6 +725,19 @@ logic for the strut changes when vertical mode is detected.
 \pkg{bigfoot})}
 
 
+\subsection{Fix a ``missing \cs{item}'' rollback error}
+
+If \LaTeX{} is rolled back to a date not newer than 2023/06/01 and
+older than 2024/06/01, any list-based environment would raise an error
+\begin{verbatim}
+! LaTeX Error: Something's wrong--perhaps a missing \item.
+\end{verbatim}
+This has now been corrected as a hotfix in patch level 2, by enhancing
+a 2023/06/01 version rollback code of new paragraph mechanism.
+%
+\githubissue{1386}
+
+
 
 \section{Changes to packages in the \pkg{amsmath} category}
 
--- a/base/ltpara.dtx
+++ b/base/ltpara.dtx
@@ -17,7 +17,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{ltpara.dtx}
-             [2024/05/16 v1.0m LaTeX Kernel (paragraph hooks)]
+             [2024/06/23 v1.0n LaTeX Kernel (paragraph hooks)]
 % \iffalse
 %
 \documentclass{l3doc}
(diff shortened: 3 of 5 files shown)


base/ltpara.dtx
```diff
@@ -819,6 +819,18 @@
   \@@_handle_indent:
 % \the \everypar           % <--- done differently below
 }
+%    \end{macrocode}
+% \changes{v1.0n}{2024/06/21}
+%         {Append \cs{everypar} toks to \cs{g_@@_standard_everypar_tl},
+%          rollback 2023/06/01 (gh/1386)}
+%    \begin{macrocode}
+%<latexrelease>\cs_set:Npn \@@_tmp:w #1#2#3#4#5 { }
+%<latexrelease>\tl_gput_right:Nx \g_@@_standard_everypar_tl {
```
josephwright, line 828: `\tl_gpur_right:Ne`

muzimuzhi: Do you also suggest to change other two `\tl_gput_right:Nx`s in `ltpara.dtx`?

davidcarlisle: @josephwright  I think it's Ok to use `e` (as @muzimuzhi  has already changed it) but this is rollback so rolling back to what it was would have been more natural  I think?

muzimuzhi: Ah this is rollback.

I just checked that `\tl_gput_right:Ne` was not provided until l3kernel 2023-10-10 (see latex3/latex3@fe0eaaa7fbfc59ce9226c7613036a76ef9694c94). So maybe I should keep using `x` in that rollback code for 2023/06/01.

FrankMittelbach: For rollback I suggest we stick to roll back to what we had back then and not do some thing like x to e (or whatever) even if it appears to be safe. Any unnecessary change would at least alter a test with `\CheckCommand` (even if that is unlikely). Of course they may be a few cases where we want to divert from that general rule, but as a best practice I think that is better.

davidcarlisle: @muzimuzhi  sorry can you revert back to `x`

muzimuzhi: Done. I dropped the 'x to e' commit https://github.com/latex3/latex2e/commit/1c23fde6c347c8d9f0e59657642e2256962819cd.

muzimuzhi:
Need a ltnews40 entry?

base/changes.txt
```diff
@@ -6,12 +6,19 @@ to completeness or accuracy and it contains some references to files that are
...
+	Append \everypar toks to \g__parar_standard_everypar_tl, rollback
+	2023/06/01 (gh/1386)
+
 2024-06-19  David Carlisle  <David.Carlisle@latex-project.org>
 
 	* fontdef.dtx:
 	preload ts1lmr.fd in Unicode engines
 
-2024-06-19 Joseph Wright  <Joseph.Wright@latex-project.org>
+2024-06-19  Joseph Wright  <Joseph.Wright@latex-project.org>
```
josephwright, line 20: Why the unrelated change here?

josephwright (approved):

base/changes.txt
muzimuzhi: To unify whitespaces. I can drop this unrelated change.

base/changes.txt
```diff
@@ -17,8 +17,7 @@ not part of the distribution.
 	* fontdef.dtx:
 	preload ts1lmr.fd in Unicode engines
 
-2024-06-19  Joseph Wright  <Joseph.Wright@latex-project.org>
-
```
FrankMittelbach: it has no been reverted, which isn't a problem either way, but I agree with @muzimuzhi 's approch @josephwright to clean up such things on the fly rather than making unnecessary pull requests for such a minor cleanup. I usually do that too when I notice one of those.

FrankMittelbach: can't type as usual. I meant: "If it has been reverted, so be it, but I agree with ...

muzimuzhi:
The base branch of this PR is changed from `develop` to `main`, and I've moved ltnews entry to `ltnews39.tex`, adapted changes entry, and updated date.

base/doc/ltnews39.tex
```diff
@@ -725,6 +725,19 @@ \subsection{Struts at the end of footnotes or \texttt{p} columns}
 \pkg{bigfoot})}
 
 
+\subsection{Fix a ``missing \cs{item}'' rollback error}
+
+If \LaTeX{} is rolled back to a date not newer than 2023/06/01 and
+older than 2024/06/01, any list-based environment would raise an error
```
FrankMittelbach, line 731: not newer 2023 but older than 2024/06 ? I think you meant something like

rolled back to a date between ...

maybe?

davidcarlisle: between also was not is as it's used with would so

If \LaTeX{} was rolled back to a date between  2023/06/01 and
2024/06/01, any list-based environment would raise an error

muzimuzhi: The original "not newer than 2023/06/01" should be "not older than ...".

To express the constraint "2023/06/01 <= date < 2024/06/01", how about

```suggestion
If \LaTeX{} is rolled back to a date between 2023/06/01 (inclusive) and
2024/06/01 (exclusive), any list-based environment would raise an error
```

FrankMittelbach: looks fine

FrankMittelbach (approved):

base/doc/ltnews39.tex
```diff
@@ -725,6 +725,19 @@ \subsection{Struts at the end of footnotes or \texttt{p} columns}
 \pkg{bigfoot})}
 
 
+\subsection{Fix a ``missing \cs{item}'' rollback error}
+
+If \LaTeX{} is rolled back to a date between 2023/06/01 (inclusive) and
+2024/06/01 (exclusive), any list-based environment would raise an error
```
muzimuzhi, line 731: Maybe nowadays YYYY-MM-DD form is more preferable?

FrankMittelbach: hmm, I personally prefer the old one, but I don't mind either way

muzimuzhi: Then let it be.

PR #1396, opened 2024-07-01
Add numerous `\DeclareFontSeriesChangeRule` entries

user227621:
LaTeX is currently missing some `\DeclareFontSeriesChangeRule` entries. Consider the following example:

```latex
\documentclass{article}

\begin{document}

\fontfamily{NotoSans-TLF}\selectfont

\fontseries{lsc}\selectfont This is light semicondensed.

\fontseries{b}\selectfont This should be bold semicondensed. %but you get bold regular

\end{document}
```

Furthermore, it is documented that not all widths are yet supported:

```
%    Also: while I did set up all nine standard weight values from
%    \texttt{ul} to \texttt{ub} I only bothered to provide entries for
%    \texttt{ec},  \texttt{sc}, \texttt{c} and \texttt{x}, because other levels of
%    compression/expansion are not in any real fonts that I know.
%
%    Could and perhaps should be eventually extended to cover the
%    whole set.
```

This PR adds rules so that the full range of weights (from `ul` to `ub`) and widths (from `uc` to `ux`) is covered. The entries are sorted first by weight (from `ul` to `ub`) and then by width (from `uc` to `ux`). (Perhaps this gives a little structure to the long list of entries.)

# Internal housekeeping

## Status of pull request

- Ready to merge
- rollback needed?
- test files needed?

- [ ] Test file(s) added


FrankMittelbach:
I'm going to look at this once you say it is ready, but not before returning from Prague, ie not before August. Fixing the conflict is probably not sensible right now as it will conflict again during the next weeks.

FrankMittelbach:
And it should get a (draft) entry in ltnews40, at least a placeholder so that it is not forgotten, thanks.

base/ltfssaxes.dtx
```diff
@@ -132,8 +132,9 @@
 % The rules set up use explicit series values not \cs{..default}
 % indirections; my current feeling is that this is in fact better.
 %
-% With 9 weights and 9 width classes this table is getting a bit large
-% in the end (324 entries) but on the other hand it doesn't change and
+% With 9 weight values (from \texttt{ul} to \texttt{ub}) and 9 width values
+% (from \texttt{uc} to \texttt{ux}), this table is getting a bit large in
+% the end (1345 entries), but on the other hand it doesn't change and
 % accessing speed and it is fast this way.
```
user227621, line 138: An incomplete sentence: "but on the other hand it doesn't change and accessing speed and it is fast this way." What was that originally supposed to mean?

base/ltfssaxes.dtx
```diff
@@ -264,7 +264,6 @@
 \DeclareFontSeriesChangeRule {ebc}{x}{ebx}{}
 %    \end{macrocode}
 %
-%
 %    \begin{macrocode}
 \DeclareFontSeriesChangeRule {ec}{bx} {bx}  {b} 	%<-----
```
user227621: There are a number of entries of the form `\DeclareFontSeriesChangeRule{...}{...}{bx}{b}`. I suspect that they exist because for the Computer Modern fonts 'bold' is `bx` and not `b`, so that you have a reasonable substitution when `bx` is applied to another font. Wouldn't it be better if this was handled in a general way, e.g., in `\merge@font@series@`, instead of handling this case (incompletely) via random `\DeclareFontSeriesChangeRule` entries?

base/ltfssaxes.dtx
```diff
@@ -905,12 +905,8 @@
 %    \end{macrocode}
 %
 %    \begin{macrocode}
-\DeclareFontSeriesChangeRule {m}{l}{l}  {m}                 %<-----
 \DeclareFontSeriesChangeRule {m}{sb}{sb}  {b}               %<-----
 \DeclareFontSeriesChangeRule {m}{b}{b}  {bx}                %<-----
-\DeclareFontSeriesChangeRule {m}{ec}{ec}  {m}               %<-----
-\DeclareFontSeriesChangeRule {m}{c}{c}  {m}                 %<-----
-\DeclareFontSeriesChangeRule {m}{sc}{sc}  {m}               %<-----
 \DeclareFontSeriesChangeRule {m}{bx}{bx}  {b}               %<-----
```
user227621: There are a number of entries of the form `\DeclareFontSeriesChangeRule{...}{...}{bx}{b}`. I suspect that they exist because for the Computer Modern fonts 'bold' is `bx` and not `b`, so that you have a reasonable substitution when `bx` is applied to another font. Wouldn't it be better if this was handled in a general way, e.g., in `\merge@font@series@`, instead of handling this case (incompletely) via random `\DeclareFontSeriesChangeRule` entries?

base/ltfssaxes.dtx
```diff
@@ -280,23 +570,69 @@
...
 \DeclareFontSeriesChangeRule {l}{ec}{lec}  {l}   % ?        %<-----
 \DeclareFontSeriesChangeRule {l}{c}{lc}  {l}   % ?          %<-----
 \DeclareFontSeriesChangeRule {l}{sc}{lsc}  {l}   % ?        %<-----
+\DeclareFontSeriesChangeRule {l}{sx}{lsx}  {l}   % ?        %<-----
 \DeclareFontSeriesChangeRule {l}{x}{lx}  {l}   % ?          %<-----
+\DeclareFontSeriesChangeRule {l}{ex}{lex}  {l}   % ?        %<-----
+\DeclareFontSeriesChangeRule {l}{ux}{lux}  {l}   % ?        %<-----
+\DeclareFontSeriesChangeRule {l}{sb}{sb}  {b}   % ?         %<-----
+\DeclareFontSeriesChangeRule {l}{b}{b}  {bx}                %<-----
 \DeclareFontSeriesChangeRule {l}{bx}{bx}  {b}               %<-----
```
user227621: There are a number of entries of the form `\DeclareFontSeriesChangeRule{...}{...}{bx}{b}`. I suspect that they exist because for the Computer Modern fonts 'bold' is `bx` and not `b`, so that you have a reasonable substitution when `bx` is applied to another font. Wouldn't it be better if this was handled in a general way, e.g., in `\merge@font@series@`, instead of handling this case (incompletely) via random `\DeclareFontSeriesChangeRule` entries?

user227621:
@FrankMittelbach I've added an entry in ltnews40. Minor open questions in the comments above, otherwise this PR is ready for review.

base/ltfssaxes.dtx
davidcarlisle: @user227621  

I assume the intended meaning is along the lines of this, which might be clearer?

this table is now rather large  (1345 entries), but on the other hand, the table doesn't change and accessing rules is fast using a table implemented in this way.

user227621: I have rephrased the sentence accordingly.

cfr42:
Ignore this comment. It's only here because I can't figure out how to watch a PR. (If this gets merged, I can probably remove some code from `nfssext-cfr-nnfss.sty`.)

user227621:
@cfr42 There is this button in the right-hand bar `;-)`:

temp

(If you're opening GitHub on a mobile device, it's at the bottom of the page.)

cfr42:
> @cfr42 There is this button in the right-hand sidebar `;-)`:
> 
> temp

Thanks! For some reason, I did not make that terminological connection, despite seeing the button. (I tried various other things, but not that one.)

base/ltfssaxes.dtx
FrankMittelbach: It would not, I think, because the rules are adaptable, so if oyu have a special job that needs changes in that respect you can overwrite the rule. If it would be done internally your hand is forced.

But in the opposite direction this means that there is no point in adding such alternatives in situations where the extended shapes play no role, e.g., if ultra light is an existing series then we have a font like noto and then falling back to bx or b doesn't make much sense

FrankMittelbach:
> @FrankMittelbach I've added an entry in ltnews40. Minor open questions in the comments above, otherwise this PR is ready for review.

sorry that got somewhat burried and it seems simpler by now if I'm marging manually myself rather than you trying to get the PR in shape, e.g., it would need to go to ltnews41 now and some other changes makes this not work cleanly (so this is what I'm currently doing).

However I have a couple of questions on your choices, some of which I don't understand. For example,
```
\DeclareFontSeriesChangeRule {l}{uc} {luc} {l}          %<-----
\DeclareFontSeriesChangeRule {l}{ec} {lec} {l}          %<-----
\DeclareFontSeriesChangeRule {l}{c}  {lc}  {l}            %<-----
\DeclareFontSeriesChangeRule {l}{sc} {lsc} {l}          %<-----
\DeclareFontSeriesChangeRule {l}{sx} {lsx} {l}          %<-----
\DeclareFontSeriesChangeRule {l}{x} {lx}   {l}           %<-----
\DeclareFontSeriesChangeRule {l}{ex} {lex} {l}         %<-----
\DeclareFontSeriesChangeRule {l}{ux} {lux} {l}         %<-----
\DeclareFontSeriesChangeRule {l}{sb} {sb}  {b}           %<-----
\DeclareFontSeriesChangeRule {l}{b}  {b}   {bx}           %<-----
\DeclareFontSeriesChangeRule {l}{bx} {bx}  {b}           %<-----
```
Why the fallback to `l` or `b` here?  That is not done for any other weight (or if then only for a few). What is the rationale for not just leaving it `{}`?

I can see some sense in saying that some weights and widths are more likely to be implemented so if you come across something that isn't available then fall back to a more likely neighbor, but that would then need to be more generally done, not just for a single weight.

user227621:
These seven rules were already there before I changed anything (some of them marked with a question mark):

```latex
\DeclareFontSeriesChangeRule {l}{sb}{sb}  {b}   % ?         %<-----
\DeclareFontSeriesChangeRule {l}{b}{b}  {bx}                %<-----
\DeclareFontSeriesChangeRule {l}{ec}{lec}  {l}   % ?        %<-----
\DeclareFontSeriesChangeRule {l}{c}{lc}  {l}   % ?          %<-----
\DeclareFontSeriesChangeRule {l}{sc}{lsc}  {l}   % ?        %<-----
\DeclareFontSeriesChangeRule {l}{x}{lx}  {l}   % ?          %<-----
\DeclareFontSeriesChangeRule {l}{bx}{bx}  {b}               %<-----
```

So I thought that someone thought that this would be useful and I just added analogous rules for the other weights. I don't find these fallbacks useful either, so I will delete them.

FrankMittelbach:
> So I thought that someone thought that this would be useful and I just added analogous rules for the other weights. I don't find these fallbacks useful either, so I will delete them.

I ust realized that they are partly in the old sources, so must have been my doing ...

But regardless of that, please do not spend further time on this PR (but leave it open).  As I said I'm going to do the merging myself and have already moved most of it in a new branch at my end. Once that is done I'll ask you to review.

base/ltfssaxes.dtx
user227621: So you think it would be better to delete the entries of the form `\DeclareFontSeriesChangeRule{...}{...}{bx}{b}` (`\DeclareFontSeriesChangeRule{...}{...}{b}{bx}`) altogether?

user227621:
Ok (just pushed two commits before I read your comment)

base/ltfssaxes.dtx
FrankMittelbach: I have mixed feelings about the b / bx. I think it is probably helpful to have the rules that ask for `bx` fall back to `b` because the `bx`might just come from a wrong `\bfseriesdefault`.  Similar, if CM or LM fonts + some others are used, then the setting might have changed to `b`and so fails with CM or LM. Thus, I think keeping those makes some sense. The situation is different if we are already in some series that is clearly outside the main ones, then they are much less useful, though they don't hurt.

FrankMittelbach (commented):
There are a few (minor) things that I would do differently, but due to the fact that we left the PR too long alone, it would be hard to merge now, as the corrections to make it work are numerous due to some changes in ltfssaxes.

I therefore decided to start a new branch essentially takingg the work here rather than trying to rebase.

This is https://github.com/latex3/latex2e/pull/1583 and it is ready for checking.

PR #1403, opened 2024-07-07
replace `\em` with `\emph` in alltt doc

mbertucci47:
~This PR replaces the use of old-style font command `\em` in alltt.dtx with `\emph` since the latter is recommended nowadays.~

As I was informed by @u-fischer, `\em` is not one of the old-style font commands. However, for the reasons expressed in my comment below, I still think this is a useful change.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/alltt.dtx
+++ b/base/alltt.dtx
@@ -40,7 +40,7 @@
 %<package>\ProvidesPackage{alltt}
 %<driver>\ProvidesFile{alltt.drv}
 %\ProvidesFile{alltt.dtx}
-              [2024/02/08 v2.0g defines alltt environment]
+              [2024/07/07 v2.0g defines alltt environment]
 %
 %<*driver>
 \documentclass{ltxdoc}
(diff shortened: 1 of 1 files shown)


u-fischer:
`\em` is not an old-style font like `\bf` or `\it`. It is the standard font switch variant of `\emph`.

mbertucci47:
My apologies, I thought it was similar to `\it` and `\bf`. Before closing this, I'll make another argument for the change. Since the `alltt` environment treats spaces differently, the user can't simply copy and paste existing text with `{\em...}` into it. Compare:
```tex
\documentclass{article}
\usepackage{alltt}

\begin{document}
abc {\em emphasized text\/} abc

abc{\em emphasized text\/} abc

\begin{alltt}
abc {\em emphasized text\/} abc

abc{\em emphasized text\/} abc
\end{alltt}
\end{document}
```
The "correct" form for normal text produces an extra space in `alltt`, which is expected but possibly unnecessarily confusing for a documented suggestion. The `\emph` form does not have this problem.
```tex
\documentclass{article}
\usepackage{alltt}

\begin{document}
abc \emph{emphasized text} abc

\begin{alltt}
abc \emph{emphasized text} abc
\end{alltt}
\end{document}
```

car222222:
But in what practical ways is `\em` different from `\it` etc?  

OK, so it is a switch rather than being absolute, but is this of any import or relevance in this context (i.e., the contrast with `\emph`)?

davidcarlisle:
@car222222 

> But in what practical ways is \em different from \it etc?

The main practical difference is that `\em` is defined in the format, and `\it` is not.

car222222:
Aha!  "Not many people know that!":-)

Issue #1413, opened 2024-07-12
Size and hash of historic version of a package


muzimuzhi:
## Brief outline of the bug

If LaTeX rollback causes loading of historic version of a package, in `\listfiles[sizes, hashes]` output, the logged package info (date, version, etc.) is of historic version, while the size and hash are of the  latest version.

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
% \RequirePackage[2023-11-01]{latexrelease}
\documentclass{article}
\usepackage{array}

\listfiles[sizes, hashes]

\begin{document}
\end{document}
```
gives
```
 *File List*
...
   array.sty    2024/06/14 v2.6d Tabular extension package (FMi)
    (size 14067, hash 1383BD6EE23AB18AE0E796155404DEB7)
...
```
and uncomment the `\RequirePackage[2023-11-01]{latexrelease}` gives
```
 *File List*
...
latexrelease.sty    2022/11/14 v1.0p LaTeX release emulation and tests (includi
ng releases up to 2024-06-01)
    (size 710925, hash 156BFC1ADA060CE40F75D50ACCE69C2D)
...
   array.sty    2023/10/16 v2.5g Tabular extension package (FMi)
    (size 14067, hash 1383BD6EE23AB18AE0E796155404DEB7)
...
```

Note that the date and version of `array.sty` changed, while the size and hash didn't. Thus the package info and size/hash don't match.

For rolled back packages maybe each of their two versions (here `array.sty` and `array-2023-11-01.sty`) need an entry in `\listfiles` output.

## Log file (required) and possibly PDF file  

[sizes-and-hashes-of-rolled-back-packages.log](https://github.com/user-attachments/files/16193771/sizes-and-hashes-of-rolled-back-packages.log)

muzimuzhi:
> For rolled back packages maybe each of their two versions (here `array.sty` and `array-2023-11-01.sty`) need an entry in `\listfiles` output.

Refer to how `\@input@` is defined, adding `\@addtofilelist{#1}` before a historic version is loaded by `\@@input <filename>\relax` seems to fix the bug here.
https://github.com/latex3/latex2e/blob/c2f7848d5e58b40a209f5af1d920158a42906bbf/base/ltfiles.dtx#L1531-L1533

Side question: how to test against the change of `\listfiles[sizes, hashes]` output in a portable and smooth (no need to update testfile after new package release) way?

```tex
[...]

PR #1425, opened 2024-08-02
Add selected pkg/cls release to the file list, gh1413

muzimuzhi:
See #1413.

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [ ] Test file(s) added


required/graphics/testfiles/utf8-filenames-002.tlg
```diff
@@ -91,6 +91,7 @@ l3backend-dvips.def    ....-..-.. L3 backend support: dvips
 graphicx.sty    ....-..-.. v... Enhanced LaTeX Graphics (DPC,SPQR)
   keyval.sty    ....-..-.. v... key=value parser (DPC)
 graphics.sty    ....-..-.. v... Standard LaTeX Graphics (DPC,SPQR)
+graphics-....-..-...sty
```
muzimuzhi, line 94: As I commented in https://github.com/latex3/latex2e/issues/1413#issuecomment-2229488694, the version test of base version (here `\ver@graphics.sty`) is still that of the historic version.

Since in the base file (here `graphics.sty`) `\DeclareRelease` (can be one or more) and `\DeclareCurrentRelease` all came before `\ProvidesPackage`, if a historic release is selected the loading of base file is ended before that `\ProvidesPackage` is invoked. Therefore currently there is no way to get the base version.

I once tried with appending some explanatory short text like ` (rolled back)` to the version text `\ver@graphics.sty`, but found that not very helpful.

What do you think?

josephwright: I think we are getting what we need to know - the effective date of the code being used. Yes, it's a little tricky but we quite deliberately are not loading the full `foo.sty` but are switching to `foo-YYYY-MM-DD.sty`, and that is clear in the `.log`.

josephwright (approved):

PR #1431, opened 2024-08-09
Support token-by-token math grabbing for tabulars

josephwright:
This provides a token-by-token approach to grabbing inline math mode to allow this to work in tabulars. The mechanism is covered in one commit: a subsequent one will be needed for activation (although for testing one can simply apply globally - it works in all cases but is simply slower).

### DIFF
--- a/required/latex-lab/latex-lab-math.dtx
+++ b/required/latex-lab/latex-lab-math.dtx
@@ -19,8 +19,8 @@
 % for those people who are interested or want to report an issue.
 %
 %
-\def\ltlabmathdate{2024-07-05}
-\def\ltlabmathversion{0.5j}
+\def\ltlabmathdate{2024-08-12}
+\def\ltlabmathversion{0.6a}
 %
 %<*driver>
 \documentclass{l3doc}
@@ -1349,25 +1349,24 @@
 % \subsection{Content grabbing}
 %
 % \begin{macro}{\@@_grab_dollar:w}
-%   Grab up to a single |$|, for inline math mode, suppressing
-%   any processing if the token is \tn{m@th} found in the content.
-%
-%    \fmi{what's that test doing?}
-%
-%    \car{It is some kind of fix, to avoid the remote
-%      possibility that the math is empty, making the code
-%      produce an unwanted \texttt{\$\$}.}
-%
-%\car{cf.~the code for this in \cs{@ensuredmath}}
-%
-%\car{It is harmless but unnecessary in the dollardollar grabbing below.}
-%
+% \begin{macro}{\@@_grab_dollar:n}
+%   Top-level function to handle grabbing of inline math mode delimited by
+%   |$| tokens. We provide two different ways to do that: a token-by-token
+%   one that can be used everywhere, and a fast delimited one that does not
+%   work anywhere that the end |$| token may be hidden, most obviously in
+%   tabulars. The function here is therefore set up as a variable starting
+%   point.
 %    \begin{macrocode}
-\cs_new_protected:Npn \@@_grab_dollar:w % $
-  #1 $
+\cs_new_protected:Npn \@@_grab_dollar:w { \@@_grab_dollar_delim:w }
+%    \end{macrocode}
+%   After grabbing inline math material, there is again common processing
+%   independent of mechanism of collection.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_grab_dollar:n #1
   {
 %    \end{macrocode}
-%    \fmi{what's that test doing?}
+%   To avoid further work with entirely empty math segments, we test before
+%   doing anything more.
 %    \begin{macrocode}
     \tl_if_blank:nF {#1}
       {
@@ -1389,6 +1388,16 @@
         }         
       }
   }
+%    \end{macrocode}
+% \end{macro}
+% \end{macro}
+%
+% \begin{macro}{\@@_grab_dollar_delim:w}
+%   Grab up to a single |$|, for inline math mode, suppressing
+%   any processing if the token is \tn{m@th} found in the content.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_grab_dollar_delim:w #1 $ % $
+  { \@@_grab_dollar:n {#1} }
 %    \end{macrocode}
 % \end{macro}
 % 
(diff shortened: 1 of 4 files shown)


required/latex-lab/latex-lab-math.dtx
```diff
@@ -1349,25 +1349,25 @@
...
-  #1 $
+\cs_new_protected:Npn \@@_grab_dollar:w { \@@_grab_dollar_delim:w }
+%    \end{macrocode}
+%   After grabbing inline math material, there is again common processing
+%   independent of mechanism of collection.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_grab_dollar:n #1
   {
 %    \end{macrocode}
 %    \fmi{what's that test doing?}
```
FrankMittelbach, line 1368: so my question should be dropped I guess. But is that actually a common case that warrants testing for? (not that it hurt)

required/latex-lab/latex-lab-math.dtx
```diff
@@ -1463,6 +1473,164 @@
...
+%   Needed to count up the number of nested environments encountered.
+%    \begin{macrocode}
+\int_new:N \l_@@_grab_env_int
+%    \end{macrocode}
+% \end{variable}
+%
+% \begin{macro}{\@@_grab_dollar_loop:}
+% \begin{macro}{\@@_grab_loop:}
+%   The lead-off here establishes a group: we need that as we will have to
+%   be careful in the way \tn{cr} is handled and ensure this is only
```
FrankMittelbach, line 1502: \cr and & ?

required/latex-lab/latex-lab-math.dtx
```diff
@@ -1463,6 +1473,164 @@
...
+% \end{macro}
+% \end{macro}
+%
+% \begin{macro}{\@@_grab_loop_token:N}
+% \begin{macro}
+%   {
+%      \@@_grab_loop_$:              ,
+%      \@@_grab_loop_\\:             ,
+%      \@@_grab_loop_\begin:         ,
+%      \@@_grab_loop_\begin :        ,
```
FrankMittelbach, line 1545: thatlooks pretty odd ... warrants an explanation? or is it a mistake?

required/latex-lab/latex-lab-math.dtx
```diff
@@ -1463,6 +1473,164 @@
...
+      { \@@_grab_loop: }
+  }
+%    \end{macrocode}
+% \end{macro}
+% \end{macro}
+%
+% \begin{macro}{\@@_grab_loop_newline:}
+%   To allow collection of tokens in the part of the \tn{halign} template after
+%   |#|, we need \TeX{} to see the primitive with the loop token in the right
+%   place. That is done by re-defining \tn{cr} at present: to discuss further.
```
FrankMittelbach, line 1607: temporarily re-defining \tn{cr} ...      % that redefinition is fairly local

FrankMittelbach: maybe mention the plan to use a socket in array (and longtable?) etc

FrankMittelbach (commented):
looks ok to me but

 - what about the \cr's in longtable
 - the PR should contain test file(s)

josephwright:
>     * the PR should contain test file(s)

Currently the code is not activated, hence no tests beyond the current ones (i.e. that nothing is broken) - I was going to test once we have an agreed mech for actually using this.

required/latex-lab/latex-lab-math.dtx
josephwright: `&` doesn't need any handling as we never pick it up: TeX inserts the end-of-template tokens, we read the ending `$` and the loop stops

required/latex-lab/latex-lab-math.dtx
josephwright: If `\begin` is at the start of a cell, it expands to `\protect\begin[space]` - the reason I asked if we want to 'tidy up'. Unless we do, there has to be coverage of the space-added macro

FrankMittelbach: not sure how you will tidy that up, which is why I suggested explaining the case

required/latex-lab/latex-lab-math.dtx
FrankMittelbach: treu

required/latex-lab/latex-lab-math.dtx
josephwright: Tidy up is as I said earlier: find a `\protect` token, grab the next token, see if it is `\foo[space]` and `\foo` exists, process just `\foo`.

davidcarlisle: or make `\begin` etex protected rather than `\protect` ....

required/latex-lab/latex-lab-math.dtx
davidcarlisle: It's true that a primitive `&` isn't seen but there are other things beside `\\ ` that hide an end of cell, notably packages using active `&` (tikz matrix of nodes being a common example). In many ways tabular and `\\ ` is just a common case of the underlying problem of end of math being behind a macro, which can occur anywwhere not just in tables. Probably should not try to handle the general case here but instead at some point (not in this PR) provide packages with a way of declaring inline math constructs that do safely grab the content.

FrankMittelbach: well, maybe the case of active & actually warrants detection and handling though. Couldn't that been done by simply expanding it if enountered?

required/latex-lab/latex-lab-math.dtx
josephwright: @davidcarlisle Oh yes _please_

FrankMittelbach: maybe that indeed.

josephwright:
> > ```
> > * the PR should contain test file(s)
> > ```
> 
> Currently the code is not activated, hence no tests beyond the current ones (i.e. that nothing is broken) - I was going to test once we have an agreed mech for actually using this.

I'll look to add some tests based on my local hacked-together stuff today

u-fischer:
I get a failure with the `\dots` command if it is followed by an `&`. It doesn't matter which grabbing method is activated, it fails with both.

~~~~
\DocumentMetadata{testphase={math}}
\documentclass{article}
\ExplSyntaxOn
%\cs_set_protected:Npn \__math_grab_dollar:w { \__math_grab_dollar_loop:}
\cs_set_protected:Npn \__math_grab_dollar:w { \__math_grab_dollar_delim:w}
\ExplSyntaxOff
\begin{document}
\begin{equation}
\begin{pmatrix} 
\dots&\\
\end{pmatrix}
\end{equation}
\end{document}
~~~~

~~~~
! Extra \else.
\stripprotect@ ...ifx #1\protect \meaning #2\else 
                                                  \meaning \@let@token \fi 
l.218 \end
          {equation}
~~~~

josephwright:
@u-fischer You got the latest version? I'd made a couple of errors that I've fixed - it's working for me ('release' latex-dev + latex-lab from GitHub)

u-fischer:
@josephwright it fails with amsmath from github, with amsmath from release it is fine. Looks as if the *Extend support for \cs{dots}* is the problem?

josephwright:
@u-fischer Indeed: we are not even needing the new grabbing code here  :)

FrankMittelbach:
I can't reproduce this. If I use amsmath from dev without any grabbing or latex-dev stuff it works as far as I can see

josephwright:
Shorter example:
```latex
\DocumentMetadata{pdfversion = 2.0, testphase = {phase-III,title,math}, lang = en}
\documentclass{article}
\usepackage{amsmath}
\begin{document}
\makeatletter\@firstofone{$\begin {pmatrix} \dots &\\ \end {pmatrix}$}
\end{document}
```

josephwright:
I see the `\dots` issue: when we grab the math mode content, collection means that the `$` at the end of the material is not there, and a look-ahead sees `\textonly@unskip`. The test inside `\dots` is then not designed to handle expanding a primitive conditional.

FrankMittelbach:
> I see the `\dots` issue: when we grab the math mode content, collection means that the `$` at the end of the material is not there, and a look-ahead sees `\textonly@unskip`. The test inside `\dots` is then not designed to handle expanding a primitive conditional.

I guess `\textonly@unskip` could be protected, however, question is does the dots definition need to be more cautious and if so how?

davidcarlisle:
> > I see the `\dots` issue: when we grab the math mode content, collection means that the `$` at the end of the material is not there, and a look-ahead sees `\textonly@unskip`. The test inside `\dots` is then not designed to handle expanding a primitive conditional.
> 
> I guess `\textonly@unskip` could be protected, however, question is does the dots definition need to be more cautious and if so how?

```
\documentclass{article}

\usepackage{amsmath}

\begin{document}

$ a \dots \iffalse b \else c \fi $
\end{document}
```

hmm

davidcarlisle:
I think we could do

```
\xdef\meaning@{\@xp\stripprotect@\@let@token\z@=\z@\fi....\stripprotect@. .........}%
```

using `z@=\z@\fi` rather than `....` as throw away tokens, so if `\@let@token` is an if  of any sort it'll get closed safely.
But this would need setting up in a helper macro as that definition is already inside some conditional \if...\fi which would be thrown off by the unmatched fi there.

But maybe more sensible to say that commands using primitve conditionals should be protected so that dots never gets to see the conditionals  rather than trying to make them safe.

josephwright:
> But maybe more sensible to say that commands using primitve conditionals should be protected so that dots never gets to see the conditionals rather than trying to make them safe.

You are deliberately using `\expandafter` to remove `\protected` status ...

davidcarlisle:
> > But maybe more sensible to say that commands using primitve conditionals should be protected so that dots never gets to see the conditionals rather than trying to make them safe.
> 
> You are deliberately using `\expandafter` to remove `\protected` status ...

oh yes, we really ought to f-expand before testing so then these conditionals would just go but @FrankMittelbach  and I
 looked at that the other day and passed on it, as it's quite a big change.

required/latex-lab/latex-lab-math.dtx
josephwright: I've dropped the comment - I think the cost of the test is not high and should mean we don't need to worry about empty structures from 'fake' math or similar, so would favour keeping it.

required/latex-lab/latex-lab-math.dtx
josephwright: I guess active-`&` can be picked up by `\futurelet` comparison with the appropriate definition - but I think we want to see if this is really an issue before adjusting.

required/latex-lab/latex-lab-math.dtx
josephwright: I've explained in a comment why the case is covered at present: if we can switch to `\protected\def` for `\begin`, this special case could be removed.

josephwright:
I think in terms of mechanism/comments on mechanism we are sorted for a first pass - we have just the question of how to activate. There are outstanding issues about `\cr`, etc., but my feeling is we should be thinking of merging/activating then being able to see how they work out 'in the wild'.

required/latex-lab/latex-lab-math.dtx
FrankMittelbach: don't we already know that it will be needed for, say, tikz stuff as david mentioned? But I'm ok with merging without it at this point in time and adjust later

FrankMittelbach (approved):
I'm fine with merging. Activation can be done for now through low-level patching and we can later improve this by using a suitable socket.

josephwright:
@FrankMittelbach OK, go with patch `\@array` for the present? Or something else?

josephwright:
(I've squashed and force-pushed in prep for a merge)

FrankMittelbach:
> @FrankMittelbach OK, go with patch `\@array` for the present? Or something else?

with a note next to it, that this is going to be replaced, yes.

u-fischer:
I'm fine with a merge, but what about the \dots business (as that was from a test "in the wild": I tried with our amsldoc example.)

josephwright:
@u-fischer I think we need a separate issue for that: it's not about how the math mode material is grabbed

josephwright:
I think I will always log one for the `\begin` business

FrankMittelbach:
> @u-fischer I think we need a separate issue for that: it's not about how the math mode material is grabbed

agreed. (and I hope @davidcarlisle can look at that ine and update the amsmath code)

josephwright:
I've merged: on the activation, I could either

- Add to `\@array` (for `longtable`, we'd want to add to `\LT@array` I guess)
- Add to `tbl/init`
- Provide an interface that can be added to `tbl/init`

Perhaps the last plan is the best - avoids needing internals of one mechanism in the wrong place?

u-fischer:
Using tagsupport/tbl/init feels a bit wrong to me. Generally, as all other math grabbing (e.g. from amsmath) is not activated/deactivated with tagging. And practically as the socket can be disabled by users if one doesn't want table tagging and then math parsing would break. But for a first step one could use (if there is a todo to revisit this later).

FrankMittelbach:
if the user doesn't want table tagging then what happens? the table cells aren't tagged? then why should then math grabbing happen? shouldn't in that case math grabbing be suppressed too?

anyway, I thought for now we add it to `\@array`and worry about a socket to use later

PR #1435, opened 2024-08-11
set para-instance in verbatim explicitly. 

u-fischer:
This sets the para-instance in verbatim to justify explicitly and so (partly) resolves issue [latex3/tagging-project#119 ](https://github.com/latex3/tagging-project/issues/119)
 
# Internal housekeeping

## Status of pull request

x Ready to merge

- [x] Test file(s) added

### DIFF
--- a/required/latex-lab/changes.txt
+++ b/required/latex-lab/changes.txt
@@ -1,3 +1,6 @@
+2024-08-11 Ulrike Fischer <Ulrike.Fischer@latex-project.org>
+	* latex-lab-block.dtx: preset verbatim para-instance, issue #119
+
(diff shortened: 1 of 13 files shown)


FrankMittelbach (approved):
why do you say the issue 119 is only partly resolved?

u-fischer:
> why do you say the issue 119 is only partly resolved?

because I wanted to keep the info from the comments that the other lists need to be handled too. But as in the meantime I opened a new issue about that we can then close 119.

FrankMittelbach:
agreed, I think it is better if theis has its own issue.

Issue #10, opened 2018-01-14
babel: \babeltags broken


lemzwerg:
[babel 3.16]

Have a look at the following example.

```latex
\documentclass{article}

\usepackage[italian, ngerman]{babel}
\babeltags{italian=italian}

\begin{document}
"~foo

\foreignlanguage{italian}{bar}

"~foo

\italian{bar}

"~foo
\end{document}
```

which gives the following output.

![mwe](https://user-images.githubusercontent.com/1630245/34919899-5d75e13a-f96a-11e7-9083-ea35bb028c72.png)

According to the documentation, the `\italian` command defined by `\babeltags` should be equivalent to `\foreignlanguage`.  As can be seen, this is not the case: shorthands are not properly respected.  It's even worse since the deactivation of the shorthands is not local to the scope of `\italian`.

blefloch:
As far as I can tell the right use of `\italian` is `{\italian bar}` or `\begin{italian}bar\end{italian}`.  That seems to work.

lemzwerg:
Ah, ok.  Right.  I misread the documentation.  It's a bit terse (examples would be nice), but concise.

Closing.  Sorry for the noise.

PR #1442, opened 2024-08-14
add luamml support to math module

u-fischer:
## Status of pull request

- Ready to merge

- [x] Test file(s) added


required/latex-lab/latex-lab-math.dtx
```diff
@@ -685,8 +712,148 @@
...
+% to avoid that nothing is written in the first run, we must activate the sockets:
+%    \begin{macrocode}
+   \bool_gset_true:N\g__tag_math_mathml_AF_bool
+   \AssignSocketPlug{tagsupport/math/struct/begin}{test-mathml}
+   \AssignSocketPlug{tagsupport/math/struct/end}{test-mathml}
+   \AssignSocketPlug{tagsupport/math/substruct/begin}{single}
+   \AssignSocketPlug{tagsupport/math/substruct/end}{single} 
+   \int_set:Nn \l__luamml_pretty_int { 5 }
+   \RegisterFamilyMapping\symsymbols{oms}
+   \RegisterFamilyMapping\symletters{oml}
```
davidcarlisle, line 814: This is OK for initial merge but I think the fact that unicode-math leaves `\varCapitalGreek` defined and using legacy 8 bit fonts is a bug in that package we shoudl fix there. then these mappings would not be needed.

required/latex-lab/latex-lab-math.dtx
```diff
@@ -685,8 +712,148 @@
...
+ {
+%    \end{macrocode}
+% to avoid that nothing is written in the first run, we must activate the sockets:
+%    \begin{macrocode}
+   \bool_gset_true:N\g__tag_math_mathml_AF_bool
+   \AssignSocketPlug{tagsupport/math/struct/begin}{test-mathml}
+   \AssignSocketPlug{tagsupport/math/struct/end}{test-mathml}
+   \AssignSocketPlug{tagsupport/math/substruct/begin}{single}
+   \AssignSocketPlug{tagsupport/math/substruct/end}{single} 
+   \int_set:Nn \l__luamml_pretty_int { 5 }
```
davidcarlisle, line 812: The pretty printing is a bit spacy and not consistent with mathml-core which doesn't trim white space so

```
<mi>
x
</mi>
```

would be better as

```
<mi>x</mi>
```

and only add pretty-print newlines between elements not in text content.  But this is a comment for luamml and shuould not affect this merge

required/latex-lab/testfiles-math-luatex/mathml-luamml-1.tlg
```diff
@@ -0,0 +1,112 @@
+This is a generated file for the l3build validation system.
+Don't change this file in any respect.
+-------- mathml-luamml-1-luamml-mathml.html (start) ---------
+(mathml-luamml-1-luamml-mathml.html) <!DOCTYPE html>^^M
```
davidcarlisle, line 4: all these `^^M` suggests (I think)  that on windows you are getting mixed format files with some unix and some dos line ends, which indicates an issue with the file writing somewhere which could probably be more cross platform

davidcarlisle (approved):
I left some comments, but mostly on luamml conversions, approving the merge

required/latex-lab/testfiles-math-luatex/mathml-luamml-1.tlg
u-fischer: the test passes also on git so it can't be windows dependant, can it?

required/latex-lab/latex-lab-math.dtx
```diff
@@ -316,6 +323,20 @@
...
+%  \subsection{Automatic mathml creation with luamml}\label{sec:luamml}
+%  
+%  If lualatex is used, if the package \texttt{unicode-math} is used
+%  and if the package \texttt{luamml} is found\footnote{Until it is 
+%  on CTAN and it must be installed from \url{https://github.com/zauguin/luamml}.}
+%  then it will automatically generate the file |\jobname-luamml-mathml.html| with mathml representations
+%  of all math formulas.
+%  The generation of the file can be suppressed with |math/mathml/luamml=false|.
+%  The package is still quite experimental and the output should be checked, e.g. in
+%  a browser. The file is then used in subsequent compilations and works also with
```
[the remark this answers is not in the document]
u-fischer: hm, any suggestion how to formulate that?

required/latex-lab/testfiles-math-luatex/mathml-luamml-1.tlg
davidcarlisle: yes I know, it also passes for me using linux line ends, but that is because the lines in the file have unix line endings, just `#10` but the html part has been serialised with a `^^M` at the end of every line whereas the standard l3build  and tagpdf messages do not.

```
This is a generated file for the l3build validation system.
Don't change this file in any respect.
-------- mathml-luamml-1-luamml-mathml.html (start) ---------
(mathml-luamml-1-luamml-mathml.html) <!DOCTYPE html>^^M
<html>^^M
^^M
<div>^^M
<h2>\mml 1</h2>^^M
<p>\begin {equation}\sqrt {x^2}=\lvert x\rvert \end {equation}</p>^^M
<p>97E0ED65D7DB327BEE6496118E96C5B1</p>^^M
^^M
<math display="block" xmlns="http://www.w3.org/1998/Math/MathML">^^M
 <mo lspace="0" rspace="0"/>^^M
 <msqrt>^^M
 <msup>^^M
 <mi>^^M
....
</div>^^M
^^M
</html>^^M
-------- mathml-luamml-1-luamml-mathml.html (end) -----------
Inserting mathml with Hash 97E0ED65D7DB327BEE6496118E96C5B1
====>subpart splitting deactivated
====>grabbed math=macro:->\sqrt {x^2}=\lvert x\rvert 
Inserting mathml with Hash 626AD6FB6CF94CF1BE7FF5A7B8335B53
====>subpart splitting deactivated
====>grabbed math=macro:->\sqrt {x^2}=\lvert x\rvert 
Inserting mathml with Hash 59E43DDDF43026AE640CBD33D3EF4739
====>subpart splitting deactivated
====>grabbed math=macro:->\varGamma 
[1
] (mathml-luamml-1.aux)
MathML statistic
================
==> 3 MathML fragments read
==> 3 different MathML fragments
==> 3 math fragments found
==> 3 fitting MathML AF found
==> 3 MathML AF attached
Package tagpdf Info: Finalizing the tagging structure:
(tagpdf)             Writing out ~12 structure objects
(tagpdf)             with ~9 'MC' leaf nodes.
(tagpdf)             Be patient if there are lots of objects!
Package tagpdf Info: writing ParentTree
Package tagpdf Info: writing IDTree
Package tagpdf Info: writing RoleMap
Package tagpdf Info: writing ClassMap
Package tagpdf Info: writing NameSpaces
Package tagpdf Info: writing StructElems
Package tagpdf Info: writing Root

```

required/latex-lab/latex-lab-math.dtx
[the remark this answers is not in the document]
davidcarlisle: Not really any good idea which is why I didn't suggest a text originally...


perhaps change

> The package is still quite experimental and the output should be checked, e.g. in a browser.

to

> The package is still quite experimental and the output should be checked. The `-luamml-mathml.html` file may be previewed in a browser although you may need to add additional css or javascript declarations to enable browser support for all mathml constructs.



we could additionally adjust the default html file to (say) reference the stix2 opentype math font (improves mathml rendering in chrome a lot, firefox can cope better with legacy fonts)  or potentially reference mathjax to polyfill full mahml3 support, but I'd rather not do either I think. Hopefully the hint that something is needed is enough here, and we'll have some examples available online somewhere that may show what's needed.

in edge for example the image above changes to

image


if you add

```
<script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?config=TeX-AMS-MML_SVG"></script>
```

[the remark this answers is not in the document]
car222222: I doubt that manual checking will be sufficient here: but what would I know?  
Maybe that is what @davidcarlisle is saying, inter alia? 

I have some other suggestions on a methodology for improving, and hopefully extending, these abilities.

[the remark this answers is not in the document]
car222222: But, without a standard to check against, what can be checked beyond "looks good", which is not the same as "the math information is identical"?

required/latex-lab/latex-lab-math.dtx
car222222: One of many "strangenesses" -- but we already know that unicode-math needs a lot of work before we recommend its use for anything non-trivial.

required/latex-lab/latex-lab-math.dtx
car222222: Do we want to adhere to mathml-core???

required/latex-lab/latex-lab-math.dtx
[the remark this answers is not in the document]
davidcarlisle: @car222222  manual checking obviously dousn't scale if say converting the back catalogue of arxiv, but is a necessary first step and anyway needed in authoring stages of new documents.

"manual" however doesn't mean "by eye". The html format used here was developed while working on the [wtpdf examples](https://github.com/latex3/tagging-project/discussions/72) and allows for example the math to be checked with a suitable screen reader (I used NVDA/MathCAT) at an early stage of the process, and without requiring a pdf reader that can extract the associated files and pass to AT.

Less important with luamml conversions but if using external conversions via tex4ht or latexml etc the other thing that frequently needs correcting and is much easier to do by eye than by an automated test is to check that the correct mathml has been paired up with the correct tex fragment as it's easy to get these out of sync.

required/latex-lab/latex-lab-math.dtx
davidcarlisle: In practice unicode-math is pretty stable and we do need to recommend its use for all math tagging at least for the medium term, probably in the end we'll want some version loaded by default in lualatex formats but we are not there yet.

required/latex-lab/latex-lab-math.dtx
davidcarlisle: > Do we want to adhere to mathml-core???

I think in the end we will need an option to configure luamml to target core.

As far as I can tell, the PDF specifications are largely silent about what mathml may be used either in associated files or if using the direct mathml structure tagging. So in practice it probably depends on what the screen readers implement.

If you target full mathml3 (including the "elementary math" elements for displayed sums and products, long division etc), then current generation browsers can display it, but using relatively large amounts of javascript to ployfill in the needed support. It is, to say the least, unclear how much of that would be accessible to a screen reader using mathml embedded in a pdf.  mathml-core is (designed to be) directly displayed in a browser requiring no javascript support and just requirng an opentype math font. So in practice is likely to be the base level of support targeted by screen readers and so probably the base level that should be targeted by systems generating mathml associated files to be used with an unknown reading application.

required/latex-lab/latex-lab-math.dtx
```diff
@@ -247,8 +247,15 @@
 %  It can be suppressed by an option with
 %  \texttt{math/tex/AF=false}, see below. 
 %  
-%  The MathML is attached if a file |\jobname-mathml.html| is found and
-%  if it contains a suitable MathML snippet for the current formula. 
+%  The MathML is attached if the files |\jobname-mathml.html| and/or 
+%  |\jobname-luamml-mathml.html| are found 
+%  and if they contains a suitable MathML snippet for the current formula. 
+%  If the files contain more than one suitable snippet (as identified by the hash)
+%  the first one is used.
```
josephwright, line 254: Fine for the present, but normally I'd expect to have overrides based on _later_ files, i.e. that the last version would win (so you read the auto-generated code first then move on to manual adjustments).

required/latex-lab/latex-lab-math.dtx
josephwright: @car222222 I'm not sure what you mean - as @davidcarlisle says, it's pretty stable and whilst some refinement may be desirable, some things may be constrained by decisions in the early 1990s (cf. ConTeXt, where they've gone entirely Unicode but with a very different approach to some aspects of stability). In my experience as a user, `unicode-math` 'just works'.

required/latex-lab/latex-lab-math.dtx
u-fischer: the current code already uses the first version as it makes processing easier: We embed the first fragment for a hash and then mark the hash as used and can ignore all the following. If we would have to wait until the end, we would have to keep all fragments in memory.

josephwright (approved):

required/latex-lab/latex-lab-math.dtx
```diff
@@ -247,8 +247,15 @@
 %  It can be suppressed by an option with
 %  \texttt{math/tex/AF=false}, see below.
```
davidcarlisle, line 248: @u-fischer if I want to correct or improve the generated mathml , eg add `intent` attributes to improve generated text is the recommeded way to use `false` here and copy the `-luamml-mathml.html` file to `mathml.html` so it gets picked up by the existing mechanism? or is there a key specifically to disable the file being re-written but still allow its use.

Longer term one may hope to add such attributes via optional arguments in the tex source, but while testing, remediating the generated mathml is likely needed,

u-fischer: @davidcarlisle yes the current idea is that you copy the file (or parts of it) into the mathml.html. You can also use some other name and then set the `math/mathml/sources={file1,file2}` (without extensions). You can disable the writing of the luamml with `math/mathml/luamml=false`, the existing file is still read. But imho adapting the luamml-file is rather fragile if you extend your document you can not reenable the automation.

required/latex-lab/latex-lab-math.dtx
u-fischer: @car222222  unicode-math is used wildly for lots of non-trivial mathematic and normally works fine. Only during the tests I found that it doesn't support all amsmath commands see https://github.com/latex3/unicode-math/issues/442

car222222:
Is it possible to explain the term "luamml-file" please?

u-fischer:
@car222222  the luamml-file is the file created by luamml (we german like to create words with hyphens).

required/latex-lab/latex-lab-math.dtx
car222222: Please can we see some of this wild, wide use of unicode-math? 

How does this relate to the fact that unicode-math itself documents many of its problems, etc? 
Are these concerns in fact not problems, or what? 

All very confusing!  But it works for everyone and everything, then no matter.

required/latex-lab/latex-lab-math.dtx
car222222: "As far as I can tell, the PDF specifications are largely silent about what mathml may be used either in associated files or if using the direct mathml structure tagging."

Well, on the last part: there is a url reference to some particular MathML namespace, but that can easily be amended.

There is no reason to restrict the content of an associated file; but it is probably useful, and maybe compulsory, to sate what type of file it is, at some level of formality.

car222222:
I think that I am beginning to understand now: that this is mostly concerned with getting MathML that "looks about right".  Which is fine, but leaves many questions marks what should be required (or even mandated?) concerning the relationship between the look of the MathML (core, of course:-) and that of the (primary) PDF formatting.

car222222:
@u-fischer Sure, or better! even without these pesky hyphens.  

So is this is an HTML file?  Or a PDF file with added HTML-tags?

Maybe a better question is:
is there any readable, and prefrerably comprehensive, documentation of the current status and possible future of luamml and related ideas?

required/latex-lab/latex-lab-math.dtx
davidcarlisle: @car222222 

> Well, on the last part: there is a url reference to some particular MathML namespace, but that can easily be amended.

All versions of MathML from 1.0 to 4.0 and mathml-core all use the same mathml namespace, so that namespace URI is just a slightly more formal way of saying "it's mathml" 

> There is no reason to restrict the content of an associated file;
If the reason for associating the  file is to improve accessibility then you need to use some format that AT systems can use. That is currently largely unspecified I think. In practice the only current pdf readers that can use the MathML AF in a pdf and do anything with it ultimately rely on mathcat to generate the speech which means that a largeish subset of mathml3 presentation is supported along with the intent attribute (which is a draft mathml4 feature being added specifically to improve accessibility and generated speech text in particular)

required/latex-lab/latex-lab-math.dtx
davidcarlisle: @car222222  you could for example look at the WTPDF examples where all the math uses unicode-math including the arxiv papers, the Bohr collection, the amsmath guide etc. or for more "in the wild" use [stackexchange unicode-math tag](https://tex.stackexchange.com/questions/tagged/unicode-math)

required/latex-lab/latex-lab-math.dtx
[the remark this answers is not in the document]
car222222: But do we not really need to check the mml code itself, in detail?  

Rather than just eyeballing the output: does it really matter so much what it looks like? The details of what it actually is are surely much more important, when actually using it, than its prettiness?

required/latex-lab/latex-lab-math.dtx
car222222: What does "look at" mean here?
I am sure they are exquisitely beautiful to look at, but so what?

car222222: How can unicode-math be described as in any sense "stable" when it is full of warnings and notes on the many unfinished areas and unfixed problems.

Or am I looking at a different version thereof?

required/latex-lab/testfiles-math-luatex/mathml-luamml-1.tlg
car222222: What test, and how is passing decider for it?

PR #1446, opened 2024-08-21
Empty math

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/required/latex-lab/changes.txt
+++ b/required/latex-lab/changes.txt
@@ -1,3 +1,6 @@
+2024-08-21 Joseph Wright <Joseph.Wright@latex-project.org>
+	* latex-lab-math.dtx: correct handling of empty math processing
+
 2024-08-14 Ulrike Fischer <Ulrike.Fischer@latex-project.org>
 	* latex-lab-math.dtx: add luamml support for automatic mathml generation
 
--- a/required/latex-lab/changes.txt
+++ b/required/latex-lab/changes.txt
@@ -1,4 +1,4 @@
-2024-08-21 Joseph Wright <Joseph.Wright@latex-project.org>
+2024-08-22 Joseph Wright <Joseph.Wright@latex-project.org>
 	* latex-lab-math.dtx: correct handling of empty math processing
 
 2024-08-14 Ulrike Fischer <Ulrike.Fischer@latex-project.org>
--- a/required/latex-lab/latex-lab-math.dtx
+++ b/required/latex-lab/latex-lab-math.dtx
@@ -1508,6 +1508,7 @@
 %
 % \begin{macro}{\@@_grab_dollar:w}
 % \begin{macro}{\@@_grab_dollar:n}
+% \changes{v0.6c}{2024-08-22}{Correct handling of empty math segments}
 %   Top-level function to handle grabbing of inline math mode delimited by
 %   |$| tokens. We provide two different ways to do that: a token-by-token
 %   one that can be used everywhere, and a fast delimited one that does not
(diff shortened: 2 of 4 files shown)


FrankMittelbach (approved):
aha :-)
but why is that caser not caught by the quick test that we talked about recently when grabbing makes a fast check for empty content before really starting to roll?

one general point. I think we shold start making \changes entries in such files not just adding a changelog entry. I always found them rather useful at later stages.

josephwright:
>but why is that caser not caught by the quick test that we talked about recently when grabbing makes a fast check for empty content before really starting to roll?

Seems nothing was testing empty math mode, in particular testing for errors _after_ grabbing the text (the grabber itself is fine).
 
> one general point. I think we shold start making \changes entries in such files not just adding a changelog entry. I always found them rather useful at later stages.

Sure, if you'd like (as you can likely tell, I'm happy with `changes.txt` ~ `CHANGELOG.md` and the Git log.)

car222222:
This could be related to whatever Frank has added above?

﻿﻿Maybe my logic is wrong, but it seems possible that the current code here 
is still not correct (for the empty #1 case): 

If #1 is empty, then  \tl_if_blank_p:n {#1}  is true:
therefore the “if_any” test is true 
so this line gets executed:

        { #1 $ }

which I guess is likely to result in an unwanted $$.  (It is not clear from this part, but I am assuming here that this is executed immediately after an explicit $ token is scanned.) 

In other words, if #1 is blank then it is safer not to process the  inline math at all, just throw it away!   This is what the code was doing before this change.  

Or not? 

Extra thought:
Is there a test with lots of cases of “empty math”, to ensure that 
no such “wicked $$” ever get created?

car222222:
Adding `\scan_stop` may be OK.

But I thought that the general policy was simply to drop any "empty inline math"; not that it really matters since this case is right out on the edge.

At least the possibility of the "unexpected $$" is now avoided . . . here!  

Do we need to check also the other "types of inline math" in the new tagging setup?

josephwright:
@car222222  We do avoid tagging empty math mode, but I'd missed that as we use `\everymath` to start grabbing, we are always inside math mode so we have to clean up - even if there is nothing to see. That's showing in tables as the other 'stuff' that LaTeX adds (`\ignorespaces`, `\unskip`) is simply dropped during grabbing.

davidcarlisle:
> @car222222 Problem is there is an opening `$` that's already happened, so we have to close math mode. But you are right about `$$`: I guess add a `\scan_stop:`.

@josephwright if the first `$` has already triggered `\everymath` , even if the math is empty, if the `\everymath` code just inserts a `$` it has to end the math doesn't it? how could it be interpreted as `$$` and start a display?

josephwright:
@davidcarlisle  Ah, yes - I'll adjust

car222222:
@davidcarlisle That is sound reasoning . . . probably.

Maybe there were/are cases where the tokens being constructed get reprocessed (including the opening $)?

This emphasizes the need for tests of all such possibly empty cases, does it not?

davidcarlisle:
> This emphasizes the need for tests of all such possibly empty cases, does it not?

well tests are always good, but I don't see any reason to particularly look for artificial empty cases to test. They'll come up as they come up. More important is to test real world non-empty cases, eg just looking at the first of the project WTPDF examples when passed through math grabbing+luamml has shown issues with equation labels.

car222222:
A misunderstanding, it seems! 

I did not mean "tests to find empty inline math", rather the idea is to add specific tests to ensure that, where it matters, cases where such empty math is explicitly present (and where such may be produced by the new grabbing methods) do not lead to disaster.

PR #1452, opened 2024-08-30
Document \IfPackageAtLeastTF, etc.

josephwright:


josephwright:
I've not covered \IfPackageWithOptionsTF yet: do they fit here?

base/doc/clsguide.tex
```diff
@@ -977,6 +977,30 @@ \subsection{Passing options around}
...
 explicitly declared.
 
+\subsection{Testing dates of other code}
+
+\begin{decl}
+  |\IfPackageAtLeast(TF)| \arg{package} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfClassAtLeast(TF)| \arg{class} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfFileAtLeast(TF)| \arg{file} \arg{date} \arg{true code} \arg{false code}
+\end{decl}
+There are occasions when you need to test if one package if another package has
```
car222222, line 987: Should be:
. . . test in one package if . . . 

Not sure about "has" here? Does a package "have a version"?

base/doc/clsguide.tex
```diff
@@ -977,6 +977,30 @@ \subsection{Passing options around}
...
+a particular version (assuming it has been loaded). This can be done using the
+commands |\IfPackageAtLeastTF|, etc. Separate commands available to test
```
car222222, line 989: What about:
. . . commands are available ...

base/doc/clsguide.tex
```diff
@@ -977,6 +977,30 @@ \subsection{Passing options around}
...
+packages, classes and general files: in all cases, they rely on the existence
+of a |\Provides...| line in the appropriate file being queried. Three versions
```
car222222, line 991: Remove "appropriate" ?

base/doc/clsguide.tex
```diff
@@ -977,6 +977,30 @@ \subsection{Passing options around}
...
+of each command are available, ending |T|, |F| and |TF|, which indicate the
+outcome branches required. Thus for example |\IfPakcageAtLeastT| does not need
```
car222222, line 993: Maybe:
. . . which specify the branches provided."

car222222 (approved):

base/doc/clsguide.tex
josephwright: Rephrased

base/doc/clsguide.tex
josephwright: Changed

base/doc/clsguide.tex
josephwright: I was trying to get at that you need particular arguments - not sure if 'provided' covers that.

car222222: OK, so say something about the arguments.
Maybe a reference to the very many other similar "T, F or TF variant" cases would be best.

base/doc/clsguide.tex
```diff
@@ -977,6 +977,30 @@ \subsection{Passing options around}
...
 explicitly declared.
 
+\subsection{Testing dates of other code}
+
+\begin{decl}
+  |\IfPackageAtLeast(TF)| \arg{package} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfClassAtLeast(TF)| \arg{class} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfFileAtLeast(TF)| \arg{file} \arg{date} \arg{true code} \arg{false code}
+\end{decl}
+There are occasions when you need to test if one package if another package is
```
mrpiggi, line 987: Typo mentioned by @car222222 is still present:
> when you need to test if one package if another package is

And suggested was
> when you need to test *in* one package if another package is

But I think, this does not hold only for packages so should be removed completely

> when you need to test if another package is

josephwright: Changed

base/doc/clsguide.tex
```diff
@@ -977,6 +977,30 @@ \subsection{Passing options around}
 handler, but this handler is not used for |landscape| because that option is
 explicitly declared.
 
+\subsection{Testing dates of other code}
+
+\begin{decl}
+  |\IfPackageAtLeast(TF)| \arg{package} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfClassAtLeast(TF)| \arg{class} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfFileAtLeast(TF)| \arg{file} \arg{date} \arg{true code} \arg{false code}
```
muzimuzhi, line 985: I think these parentheses can be dropped.
- Existence of no-parentheses forms (at least one occurrence for each command) in PDF help when searching.
- `\arg{true code} \arg{false code}` corresponds to `TF` only.

Suppose there will be more `clsguide` subsections describing `\If...TF` conditionals, a general mention of the single-branch forms could be added at the first of those subsections.

```suggestion
  |\IfPackageAtLeastTF| \arg{package} \arg{date} \arg{true code} \arg{false code}\\
  |\IfClassAtLeastTF| \arg{class} \arg{date} \arg{true code} \arg{false code}\\
  |\IfFileAtLeastTF| \arg{file} \arg{date} \arg{true code} \arg{false code}
```

base/doc/clsguide.tex
```diff
@@ -977,6 +977,30 @@ \subsection{Passing options around}
...
+commands |\IfPackageAtLeastTF|, etc. Separate commands are available to test
+packages, classes and general files: in all cases, they rely on the existence
+of a |\Provides...| line in the file being queried. Three versions
+of each command are available, ending |T|, |F| and |TF|, which indicate the
+outcome branches required. Thus for example |\IfPakcageAtLeastT| does not need
+a \meta{false code} argument. The \meta{date} argument can be given either in
+ISO (|YYYY-MM-DD|) or traditional \LaTeX{} (|YYYY/MM/DD|) formats.
+
+\begin{decl}
+  |\IfFormatAtLeast(TF)| \arg{date} \arg{true code} \arg{false code}
```
muzimuzhi, line 998:
```suggestion
  |\IfFormatAtLeastTF| \arg{date} \arg{true code} \arg{false code}
```

Also `\IfFormatAtLeastTF` doesn't match the title of current subsection
```tex
\subsection{Testing dates of other code}
```

Perhaps
```tex
\subsection{Testing dates of the \LaTeX{} format and \pkg{expl3}}
\IfFormatAtLeastTF
\IfExplAtLeastTF

\subsection{Testing dates of other code}
\IfPackageAtLeastTF
\IfClassAtLeastTF
\IfFileAtLeastTF
```

josephwright: Well the kernel is 'other code' :)

josephwright: I didn't include `\IfExplAtLeaseTF` since @car222222 is not happy with the name

base/doc/clsguide.tex
josephwright: We use the `(TF)` convention in `expl3` and I felt it was clearer - but let's see what others feel.

base/doc/clsguide.tex
```diff
@@ -977,6 +977,30 @@ \subsection{Passing options around}
...
+\begin{decl}
+  |\IfPackageAtLeast(TF)| \arg{package} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfClassAtLeast(TF)| \arg{class} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfFileAtLeast(TF)| \arg{file} \arg{date} \arg{true code} \arg{false code}
+\end{decl}
+There are occasions when you need to test if a package is
+at least a particular date (assuming it has been loaded). This can be done using the
+commands |\IfPackageAtLeastTF|, etc. Separate commands are available to test
+packages, classes and general files: in all cases, they rely on the existence
+of a |\Provides...| line in the file being queried. Three versions
```
muzimuzhi, line 991: Strictly speaking, the `\Provides...` line also need to specify the date in the optional argument.

base/doc/clsguide.tex
muzimuzhi, line 985: I think these parentheses can be dropped.
- Existence of no-parentheses forms (at least one occurrence for each command) in PDF help when searching.
- `\arg{true code} \arg{false code}` corresponds to `TF` only.

Suppose there will be more `clsguide` subsections describing `\If...TF` conditionals, a in-detail syntax and description of the single-branch forms could be added at the first of those subsections, for the first command, which then can be referred to in following sections.

```suggestion
  |\IfPackageAtLeastTF| \arg{package} \arg{date} \arg{true code} \arg{false code}\\
  |\IfClassAtLeastTF| \arg{class} \arg{date} \arg{true code} \arg{false code}\\
  |\IfFileAtLeastTF| \arg{file} \arg{date} \arg{true code} \arg{false code}
```

base/doc/clsguide.tex
muzimuzhi, line 998:
```suggestion
  |\IfFormatAtLeastTF| \arg{date} \arg{true code} \arg{false code}
```

Also `\IfFormatAtLeastTF` doesn't match the title of current subsection
```tex
\subsection{Testing dates of other code}
```

Perhaps
```tex
\subsection{Testing dates of the \LaTeX{} format and \pkg{expl3}}
\IfFormatAtLeastTF
\IfExplAtLeastTF

\subsection{Testing dates of other code}
\IfPackageAtLeastTF
\IfClassAtLeastTF
\IfFileAtLeastTF
```

base/doc/clsguide.tex
```diff
@@ -977,6 +977,30 @@ \subsection{Passing options around}
...
+of each command are available, ending |T|, |F| and |TF|, which indicate the
+outcome branches required. Thus for example |\IfPakcageAtLeastT| does not need
+a \meta{false code} argument. The \meta{date} argument can be given either in
+ISO (|YYYY-MM-DD|) or traditional \LaTeX{} (|YYYY/MM/DD|) formats.
```
muzimuzhi, line 995: In all other three places of [`clsguide.tex`](https://github.com/latex3/latex2e/blob/1f18053be5081bbe8b4a45193d26e101eaeac4f6/base/doc/clsguide.tex#L392-L394) where the date format is mentioned, only the `YYYY-MM-DD` format is mentioned and it's typeset in `\textsc{yyyy-mm-dd}`. See the first place,

https://github.com/latex3/latex2e/blob/1f18053be5081bbe8b4a45193d26e101eaeac4f6/base/doc/clsguide.tex#L392-L394

base/doc/clsguide.tex
```diff
@@ -977,6 +977,30 @@ \subsection{Passing options around}
...
 
+\subsection{Testing dates of other code}
+
+\begin{decl}
+  |\IfPackageAtLeast(TF)| \arg{package} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfClassAtLeast(TF)| \arg{class} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfFileAtLeast(TF)| \arg{file} \arg{date} \arg{true code} \arg{false code}
+\end{decl}
+There are occasions when you need to test if a package is
+at least a particular date (assuming it has been loaded). This can be done using the
```
muzimuzhi, line 988: Worth mentioning that `\IfPackageAtLeastTF` always uses the false branch if the specified package is not loaded (yet)?

muzimuzhi:
> I've not covered \IfPackageWithOptionsTF yet: do they fit here?

The correct names are `\If(Class|Package)LoadedWithOptionsTF`.

Also the simpler ones `\If(Class|Package|File)LoadedTF` are not documented either.

josephwright:
> > I've not covered \IfPackageWithOptionsTF yet: do they fit here?
> 
> The correct names are `\If(Class|Package)LoadedWithOptionsTF`.

I've never used so would have to check :)

> Also the simpler ones `\If(Class|Package|File)LoadedTF` are not documented either.

Ah, I'd not realised we'd not covered those.

base/doc/clsguide.tex
josephwright: Possibly - but it would be odd to test the version without knowing it's loaded

base/doc/clsguide.tex
josephwright: I can change that - but I think we do need to say you can use the traditional form too

base/doc/clsguide.tex
muzimuzhi: > We use the `(TF)` convention in `expl3` and I felt it was clearer - but let's see what others feel.
> <sub>_Originally posted by @josephwright in https://github.com/latex3/latex2e/pull/1452#discussion_r1738892279_</sub>

But in syntax blocks in `interface3.pdf`, there're no `\func...:...(TF) ... {<true code>} {<false code>}`, just `\func...:...TF ... {<true code>} {<false code>}`.

(So sorry I posted reviews as separate comments, then deleted the comments and reposted them as part of a review, _again_.)

base/doc/clsguide.tex
muzimuzhi: > Well the kernel is 'other code' :)
> <sub>_Originally posted by @josephwright in https://github.com/latex3/latex2e/pull/1452#discussion_r1738885024_</sub>

base/doc/clsguide.tex
muzimuzhi: True. Though in `interface3.pdf`, even the marginal versions don't print _explicit_ parentheses enclosing `TF`. We need a better way to document `TF` forms in `clsguide`.

image

image

base/doc/clsguide.tex
muzimuzhi: Not a bad idea!

base/doc/clsguide.tex
muzimuzhi: I can't recall details of how I encountered such odd uses before. Just guessing a less odd case: when the package name was misspelled.

Also I find this is the special case of saying "`\IfPackageAtLeastTF` never errors." Invalid date format leaves true branch, package not-loaded (yet) leaves false branch. Maybe these should be classified as undefined behavior.

base/doc/clsguide.tex
muzimuzhi, line 985: Use the same-named placeholders to be consistent with docs of `\Provides...` and others
```suggestion
  |\IfPackageAtLeast(TF)| \arg{package-name} \arg{date} \arg{true code} \arg{false code}\\
  |\IfClassAtLeast(TF)| \arg{class-name} \arg{date} \arg{true code} \arg{false code}\\
  |\IfFileAtLeast(TF)| \arg{file-name} \arg{date} \arg{true code} \arg{false code}
```

Also, similar to the doc of `\ProvidesFile`, maybe repeat here that "the full filename, including the extension, must be given" for `<file-name>`.
https://github.com/latex3/latex2e/blob/1f18053be5081bbe8b4a45193d26e101eaeac4f6/base/doc/clsguide.tex#L718-L723

muzimuzhi (changes requested):

josephwright:
I'm revising the content a bit ...

josephwright:
OK, I've generalised the TF business, added `\If...LoadedTF`, covered the business about optional date info, what happens if things are not loaded - anything else?

base/doc/clsguide.tex
```diff
@@ -977,29 +978,46 @@ \subsection{Passing options around}
...
-packages, classes and general files: in all cases, they rely on the existence
-of a |\Provides...| line in the file being queried. Three versions
-of each command are available, ending |T|, |F| and |TF|, which indicate the
-outcome branches required. Thus for example |\IfPakcageAtLeastT| does not need
-a \meta{false code} argument. The \meta{date} argument can be given either in
-ISO (\textsc{yyyy-mm-dd}) or traditional \LaTeX{} (\textsc{yyyy/mm/dd}) formats.
+These tests are used to check if a package or class has been loaded with at
+least a given \meta{date}. The \meta{date} is compared with that in the
+optional argument to \cs{Provides...} (as described in Section~\ref{sec:ID}).
+As for \cs{Provides...}, the \meta{date} should be given in the form
```
davidcarlisle, line 1011: I think somewhere probably at the referenced Provides.... section we need to say yyyy/mm/dd form can be used, as far as I can tell after this diff, all mention of that form is gone, but if anyone looks at any existing examples that's mostly what they will see,

davidcarlisle (approved):
approving but with a comment about /

base/doc/clsguide.tex
josephwright: I'll adjust again - I'll add something in there (looking at the blame, I wrote the bit without the traditional date format).

(As an aside, I do wonder if we should normalise the dates across the repos ...)

muzimuzhi: Perhaps: Both forms are accepted and the yyyy-mm-dd form is recommended in new codes.

base/doc/clsguide.tex
```diff
@@ -977,6 +980,47 @@ \subsection{Passing options around}
...
+`\textsc{yyyy-mm-dd}' or `\textsc{yyyy/nmm/dd}'. If the package or class is not
+loaded, the \meta{false} branch will be taken.
+
+\begin{decl}
+  |\IfFormatAtLeastTF| \arg{date} \arg{true code} \arg{false code}
+\end{decl}
+Tests the release \meta{date} of the \LaTeX{} format and selects the
+appropriate branch. This will reflect any roll forward or back that has been
+applied: the apparent date of the format will be that after roll back.
```
muzimuzhi, line 1022: Reminds me that this also applies to rolled back packages and classes. Although the `\DeclareRelease` and `\DeclareCurrentRelease` (introduced in LaTeX 2018-04-01, general rollback concept) are not documented either.

muzimuzhi (approved):

base/doc/clsguide.tex
josephwright: Fancy making a PR?

base/doc/clsguide.tex
josephwright: I'm wondering where that would fit: with `\ProvidesPackage`?

FrankMittelbach: well, in many cases I prefer the traditional /../../ notation

base/doc/clsguide.tex
```diff
@@ -390,7 +390,8 @@ \subsection{Identification%
    \NeedsTeXFormat{LaTeX2e}
    \ProvidesClass{article}[2022-06-01 Standard LaTeX class]
 \end{verbatim}
-The \m{date} should be given in the form `\textsc{yyyy-mm-dd}' and must be
+The \m{date} should be given in the form `\textsc{yyyy-mm-dd}' or
+`\textsc{yyyy/nmm/dd}', and must be
```
mrpiggi, line 394: Should read /mm/ not /nmm/ and sprawled in several places

base/doc/clsguide.tex
FrankMittelbach: invalid date means the date becomes an small number typically zero so ends up being earlier than any reasonable date, hence the results depending on what you ask for. But I agree these cases should be listed as undefined behavior.

base/doc/clsguide.tex
FrankMittelbach: but if I remember correctly everybody else was happy enough with it given that Expl3 is not possible and ExplIII or Expliii is horrible and the longer names do not do well either. My vote is on keeping the name.

base/doc/clsguide.tex
```diff
@@ -977,6 +980,47 @@ \subsection{Passing options around}
...
+\meta{file-name} must be provided; in contrast, no extension should be given if
+testing for a package or class.
+
+\begin{decl}
+  |\IfPackageAtLeastTF| \arg{package-name} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfClassAtLeastTF| \arg{class-name} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfFileAtLeastTF| \arg{file-name} \arg{date} \arg{true code} \arg{false code}
+\end{decl}
+These tests are used to check if a package or class has been loaded with at
+least a given \meta{date}. The \meta{date} is compared with that in the
```
car222222, line 1011: I f not think that a file, etc., is "loaded with a date", is it? Rather the file, etc., has a date that is declared in its |\Provides ...| line.

So this needs rewording.

base/doc/clsguide.tex
FrankMittelbach: Maybe just use \If...TF and have a convention section up front that has a subsection on conditionals mentioning that "most"  (I'm reculant to say all really) TF commands also have variants with T or F for cases where the other branch is unnecessary.

base/doc/clsguide.tex
```diff
@@ -977,6 +980,47 @@ \subsection{Passing options around}
...
+optional argument to \cs{Provides...} (as described in Section~\ref{sec:ID}).
```
car222222, line 1012: What happens if a file, etc., has no such optional argument?
Maybe this is covered elsewhere?

base/doc/clsguide.tex
```diff
@@ -977,6 +980,47 @@ \subsection{Passing options around}
...
+As for \cs{Provides...}, the \meta{date} should be given in the form
+`\textsc{yyyy-mm-dd}' or `\textsc{yyyy/nmm/dd}'. If the package or class is not
+loaded, the \meta{false} branch will be taken.
+
+\begin{decl}
+  |\IfFormatAtLeastTF| \arg{date} \arg{true code} \arg{false code}
+\end{decl}
+Tests the release \meta{date} of the \LaTeX{} format and selects the
+appropriate branch. This will reflect any roll forward or back that has been
```
car222222, line 1021: Ok, but how will it "reflect" this? 
To what exactly is any rollback applied? and what effect does this have on the date used in the comparison?

car222222 (changes requested):
This time I found a lot of questionable content, much of it well beyond the currently changed parts being reviewed.

base/doc/clsguide.tex
car222222: > I didn't include `\IfExplAtLeaseTF` since @car222222 is not happy with the name

Is he not?  I think that was in a somewhat different context. 
As @FrankMittelbach indicates, below, in this case the alternative is something like `\IfExplThreeAtLeaseTF`. 

It would be useful to explain that we use just "Expl" in places like this where a "3" is not permitted.

base/doc/clsguide.tex
```diff
@@ -977,6 +980,47 @@ \subsection{Passing options around}
...
+Here, we document only the |TF| versions: the other forms are however
+available.
+
+\begin{decl}
+  |\IfPackageLoadedTF| \arg{package-name} \arg{true code} \arg{false code}\\
+  |\IfClassLoadedTF| \arg{class-name} \arg{true code} \arg{false code}\\
+  |\IfFileLoadedTF| \arg{file-name} \arg{true code} \arg{false code}
+\end{decl}
+These commands test if the named package, class or file is loaded: this
+requires that a matching |\Provides...| line has been given, as described in
```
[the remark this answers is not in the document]
FrankMittelbach: given were such \Provides...  decls are used, eg for Class in class files for File in files ...

I find this understandable

base/doc/clsguide.tex
FrankMittelbach: yeah a bit perhaps:

  ... package, class or file ...

 ... with a date in \Provides...  which is \meta{date} or newer

base/doc/clsguide.tex
FrankMittelbach: Perhaps:

If no \Provides... declaration is given then this is equivalent to the date 0000/00/00 with respect to the the comparison.

base/doc/clsguide.tex
FrankMittelbach: isn't that what the rest of the sentence says? e.g if you roll back to yyyy/mm/dd then the conditional uses that date as the format date for comparison.

base/doc/clsguide.tex
FrankMittelbach: I think that should be brief and for further details refer to the TUB article on that and to texdoc latexrelease

FrankMittelbach (approved):
a few comments and suggestions

FrankMittelbach:
> a few comments and suggestions

some of the comments are reealy replies to comments made by others, but unfortunately in the above summary that is lost.

base/doc/clsguide.tex
```diff
@@ -389,7 +390,8 @@ \subsection{Identification}
    \NeedsTeXFormat{LaTeX2e}
    \ProvidesClass{article}[2022-06-01 Standard LaTeX class]
 \end{verbatim}
-The \m{date} should be given in the form `\textsc{yyyy-mm-dd}' and must be
+The \m{date} should be given in the form `\textsc{yyyy-mm-dd}' or
+`\textsc{yyyy/nmm/dd}', and must be
 present if the optional argument is used (this is also true for the
 |\NeedsTeXFormat| command). Any derivation from this syntax will result in
```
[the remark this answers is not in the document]
car222222: I think that this very detailed requirement, that there must be exactly two digits in these two locations (and four for the year) needs to explained very clearly in detail, either here or elsewhere.

base/doc/clsguide.tex
```diff
@@ -977,6 +980,47 @@ \subsection{Passing options around}
 handler, but this handler is not used for |landscape| because that option is
 explicitly declared.
 
+\subsection{Useful status tests}
+
+A number of utility status tests are available which can be used by package and
+class authors to query the presence and date of other code, the current kernel
+release and other related ideas. Three forms of each test are provided, one
```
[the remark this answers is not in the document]
car222222: OK.
But the "or other" replaces the "and other" in the first sentence.

base/doc/clsguide.tex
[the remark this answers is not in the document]
car222222: Understandable, yes! 
But maybe improvable too?

base/doc/clsguide.tex
car222222: Yes, the explanation is indeed there.

But the first part needs to be more precise, e.g. thus:

The date used for the format is adjusted to incorporate any rollback that has been applied to it, . . .

base/doc/clsguide.tex
car222222: Maybe this could be combined with a careful explanation of the important point that there must be exactly 8 digits used (4+2+2) for the date (using either / or -).

josephwright: I've added a note

base/doc/clsguide.tex
josephwright: I think that should be separate - I'll come back to it in a separate PR later on today

base/doc/clsguide.tex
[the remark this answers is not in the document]
josephwright: Agreed: as the optional argument to `\NeedsTeXFormat` is only there for a date, I think we can drop.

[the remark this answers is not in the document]
josephwright: I've fixed the typo and as stated put in a note about dates

base/doc/clsguide.tex
```diff
@@ -389,7 +390,8 @@ \subsection{Identification}
    \NeedsTeXFormat{LaTeX2e}
    \ProvidesClass{article}[2022-06-01 Standard LaTeX class]
 \end{verbatim}
-The \m{date} should be given in the form `\textsc{yyyy-mm-dd}' and must be
+The \m{date} should be given in the form `\textsc{yyyy-mm-dd}' or
+`\textsc{yyyy/nmm/dd}', and must be
 present if the optional argument is used (this is also true for the
 |\NeedsTeXFormat| command). Any derivation from this syntax will result in
 low-level \TeX{} errors---the commands expect a valid syntax to speed up the
```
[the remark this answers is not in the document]
josephwright: I've re-worked this paragraph further - hopefully makes more sense now.

josephwright:
OK, hopefully I've got everything - I've squashed into one commit and force-pushed as we don't need me fixing loads of typos in a PR :)

josephwright:
Further comments? I've explicitly re-asked @car222222 but others may want to chime back in too.

base/doc/clsguide.tex
```diff
@@ -366,7 +366,8 @@ \section{The structure of a class or package}
    declaring new variables, commands and fonts; and loading other files.
 \end{description}
 
-\subsection{Identification}
+\subsection{Identification%
+  \label{sec:ID}}
```
muzimuzhi, line 370: Use `Sec` prefix which is used in all but one labels in clsguide.
https://github.com/latex3/latex2e/blob/1f18053be5081bbe8b4a45193d26e101eaeac4f6/base/doc/clsguide.tex#L1085-L1086

```suggestion
\subsection{Identification%
  \label{Sec:ID}}
```
or maybe even move `\label` out of `\subsection`
```suggestion
\subsection{Identification}
\label{Sec:ID}
```

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
+
+\begin{decl}
+  |\IfPackageAtLeastTF| \arg{package-name} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfClassAtLeastTF| \arg{class-name} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfFileAtLeastTF| \arg{file-name} \arg{date} \arg{true code} \arg{false code}
+\end{decl}
+These tests are used to check wheteher the date information given in the
+|\Provides...| line of a package or class is least a given \meta{date}.
+The \meta{date} is compared with that in the
+optional argument to \cs{Provides...} (as described in Section~\ref{sec:ID});
```
muzimuzhi, line 1023:
```suggestion
optional argument to \cs{Provides...} (as described in Section~\ref{Sec:ID});
```

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
+available.
+
+\begin{decl}
+  |\IfPackageLoadedTF| \arg{package-name} \arg{true code} \arg{false code}\\
+  |\IfClassLoadedTF| \arg{class-name} \arg{true code} \arg{false code}\\
+  |\IfFileLoadedTF| \arg{file-name} \arg{true code} \arg{false code}
+\end{decl}
+These commands test whether the named package, class or file is loaded: this
+requires that this information is available in an appropriate |\Provides...|
+line, as described in Section~\ref{sec:ID}. In the case of \cs{IfFileLoadedTF},
```
muzimuzhi, line 1002:
```suggestion
line, as described in Section~\ref{Sec:ID}. In the case of \cs{IfFileLoadedTF},
```

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
+the full \meta{file-name} must be provided; in contrast, no extension should be
+given if testing for a package or class.
+
+\begin{decl}
+  |\IfPackageLoadedWithOptionsTF| \arg{package-name} \arg{options} \arg{true code} \arg{false code}\\
+  |\IfClassLoadedOptionsTF| \arg{class-name} \arg{options} \arg{true code} \arg{false code}
+\end{decl}
+These commands test if the named package or class was loaded with
```
muzimuzhi, line 1010: Maybe s/was/is/
```suggestion
These commands test if the named package or class is loaded with
```

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
+\emph{exactly} the \meta{options} specified. Thus the package or class must be
+loaded (giving a \texttt{true} result for \cs{If...LoadedT}) and the option
```
muzimuzhi, line 1012: Need an attributive here?

Thus (to get a \texttt{true} result) the package or class must be ... and ....

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
+list used when loading it must be identifical to the \meta{options}.
+
+\begin{decl}
+  |\IfPackageAtLeastTF| \arg{package-name} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfClassAtLeastTF| \arg{class-name} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfFileAtLeastTF| \arg{file-name} \arg{date} \arg{true code} \arg{false code}
+\end{decl}
+These tests are used to check wheteher the date information given in the
+|\Provides...| line of a package or class is least a given \meta{date}.
```
muzimuzhi, line 1021:
```suggestion
|\Provides...| line of a package, class or file is least a given \meta{date}.
```

Not sure if an "at" right before "least" is necessary.

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
+The \meta{date} is compared with that in the
+optional argument to \cs{Provides...} (as described in Section~\ref{sec:ID});
+if \cs{Provides...} was missing or had no optional argument, the reference
+date is treated as |0000/00/00|.
+As for \cs{Provides...}, the \meta{date} should be given in the form
+`\textsc{yyyy-mm-dd}' or `\textsc{yyyy/mm/dd}'. If the package or class is not
```
muzimuzhi, line 1027:
```suggestion
`\textsc{yyyy-mm-dd}' or `\textsc{yyyy/mm/dd}'. If the package, class or file is not
```

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
+loaded, the \meta{false} branch will be taken; in contrast, if the \meta{date}
+is not of the required form, the behavior is formally undefined.
+
+\begin{decl}
+  |\IfFormatAtLeastTF| \arg{date} \arg{true code} \arg{false code}
+\end{decl}
+Tests the release \meta{date} of the \LaTeX{} format and selects the
+appropriate branch. The date used by the format is adjusted to incorporate any
+roll forward or back that has been applied to it, so that the the apparent date
```
muzimuzhi, line 1036:
```suggestion
roll forward or back that has been applied to it, so that the apparent date
```

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
+of the format will be that after roll back.
```
muzimuzhi, line 1037:
```suggestion
of the format will be that after roll forward or back.
```

muzimuzhi (changes requested):

base/doc/clsguide.tex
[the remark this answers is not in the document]
FrankMittelbach: given that some of these commands exists for 30 years (with slighlty different names) and no developer ever had a problem, my assumption is that stating it must be exactly `yyyy/mm/dd` or `yyyy-mm-dd` is precise enough.

base/doc/clsguide.tex
FrankMittelbach: actually I would go the other way around  and switch to `sec:` everywhere because that is the typical prefix and clsguide using `Sec:` is anomaly

base/doc/clsguide.tex
car222222: Or "has been" ?

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
 
+\subsection{Useful status tests}
+
+A number of status tests are available which can be used by package and
+class authors to query the presence and date of other code, the current kernel
+release and other related ideas. Three forms of each test are provided, one
+requiring arguments for both \meta{true} and \meta{false} branches, one
+requiring only a \meta{true} branch and one requiring only a \meta{false}
+branch. These are indicated in their names as |TF|, |T| and |F|, respectively.
+Here, we document only the |TF| versions: the other forms are however
```
car222222, line 992: . . . are also available.

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
+available.
+
+\begin{decl}
+  |\IfPackageLoadedTF| \arg{package-name} \arg{true code} \arg{false code}\\
+  |\IfClassLoadedTF| \arg{class-name} \arg{true code} \arg{false code}\\
+  |\IfFileLoadedTF| \arg{file-name} \arg{true code} \arg{false code}
+\end{decl}
+These commands test whether the named package, class or file is loaded: this
+requires that this information is available in an appropriate |\Provides...|
```
car222222, line 1001: Maybe:
. . . is loaded: this is done by using information that must be contained in an . . .

base/doc/clsguide.tex
car222222, line 1021: Better:
. . . is no earlier than the given . . .

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
+  |\IfPackageAtLeastTF| \arg{package-name} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfClassAtLeastTF| \arg{class-name} \arg{date} \arg{true code} \arg{false code}\\
+  |\IfFileAtLeastTF| \arg{file-name} \arg{date} \arg{true code} \arg{false code}
+\end{decl}
+These tests are used to check wheteher the date information given in the
+|\Provides...| line of a package or class is least a given \meta{date}.
+The \meta{date} is compared with that in the
+optional argument to \cs{Provides...} (as described in Section~\ref{sec:ID});
+if \cs{Provides...} was missing or had no optional argument, the reference
+date is treated as |0000/00/00|.
```
car222222, line 1025: This is very unclear: 
which date is the "reference date"? 
Does this mean the date that is missing from any "Provides line"?

base/doc/clsguide.tex
```diff
@@ -977,6 +981,61 @@ \subsection{Passing options around}
...
+As for \cs{Provides...}, the \meta{date} should be given in the form
+`\textsc{yyyy-mm-dd}' or `\textsc{yyyy/mm/dd}'. If the package or class is not
+loaded, the \meta{false} branch will be taken; in contrast, if the \meta{date}
+is not of the required form, the behavior is formally undefined.
```
car222222, line 1029: Maybe:
. . . specified in the required . . .

car222222 (changes requested):
Quite a lot of extra queries, and some +1s.

base/doc/clsguide.tex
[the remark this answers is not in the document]
car222222: OK @FrankMittelbach ! But all we are doing here is making it extra clear what "exactly" means here.
This is needed because this is a "document of record" (i.e., it precisely specifies 
such rules) and not just a reminder of what has been true (and well-known) for 30 years.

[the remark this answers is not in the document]
FrankMittelbach: all I'm saying is that I thought stating yyyy/mm/dd is prettly clear in my opinion, but for sure with Joseph's additional footnote it is even clearer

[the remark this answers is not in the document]
car222222: Yes sure: it is definitely "pretty" clear (at least to you and maybe others), but is it "sufficiently complete and accurate" for everyone?

base/doc/clsguide.tex
josephwright: Sorted. The `\label` is in `\section` as we know this is needed in some cases - roll on keyval optional `\section[label = ...]{...}` :)

base/doc/clsguide.tex
```diff
@@ -389,12 +390,14 @@ \subsection{Identification}
...
-low-level \TeX{} errors---the commands expect a valid syntax to speed up the
-daily usage of the package or class and make no provision for the case that the
-developer made a mistake!
+The \m{date} should be given in the form `\textsc{yyyy-mm-dd}' or
+`\textsc{yyyy/mm/dd}' and must be present if the optional argument is used.
+Exactly four digits are required for the year and two each for the month and
+day. Where necessary, zeros should be added to pad the month and day
+appropriately. If digits or separators are missing, the date will likely be
+misinterpreted: the commands expect a valid syntax to speed up the daily usage
+of the package or class and make no provision for the case that the developer
```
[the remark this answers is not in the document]
josephwright: I've re-worded, but I think 'developer' is clear - we are in `clsguide` after all.

josephwright:
Further rephrasing done.

base/doc/clsguide.tex
```diff
@@ -977,6 +982,62 @@ \subsection{Passing options around}
...
+branch. These are indicated in their names as |TF|, |T| and |F|, respectively.
+Here, we document only the |TF| versions: the other forms are however also
+available.
+
+\begin{decl}
+  |\IfPackageLoadedTF| \arg{package-name} \arg{true code} \arg{false code}\\
+  |\IfClassLoadedTF| \arg{class-name} \arg{true code} \arg{false code}\\
+  |\IfFileLoadedTF| \arg{file-name} \arg{true code} \arg{false code}
+\end{decl}
+These commands test whether the named package, class or file is loaded: this is
```
muzimuzhi, line 1001: After the last round of review, in doc of `\IfPackageLoadedWithOptionsTF` we use "has been loaded", then here
```suggestion
These commands test whether the named package, class or file has been loaded: this is
```

base/doc/clsguide.tex
```diff
@@ -977,6 +982,62 @@ \subsection{Passing options around}
...
+
+\begin{decl}
+  |\IfPackageLoadedWithOptionsTF| \arg{package-name} \arg{options} \arg{true code} \arg{false code}\\
+  |\IfClassLoadedOptionsTF| \arg{class-name} \arg{options} \arg{true code} \arg{false code}
+\end{decl}
+These commands test whether the named package or class has been loaded with
+\emph{exactly} the \meta{options} specified. In order to take the \meta{true}
+branch, the package or class must be loaded (giving a \meta{true} result for
+\cs{If...LoadedT}) and the option list used when loading it must be identical
+to the \meta{options}.
```
muzimuzhi, line 1015: Only the non-global options count, do they?

base/changes.txt
```diff
@@ -6,6 +6,11 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2024-09-02  Joseph Wright  <Joseph.Wright@latex-project.org>
+
+	* clsguide.tex:
+	Document \IfPackageAtLeastTF, etc.
```
muzimuzhi, line 12: Maybe also mention `\IfPackageLoadedTF`.
```suggestion
	Document \IfPackageLoadedTF, \IfPackageAtLeastTF, etc.
```

(I noticed it in the last round but didn't post it.)

muzimuzhi (commented):
Hope this is the last round. 😄

base/doc/clsguide.tex
```diff
@@ -389,12 +390,14 @@ \subsection{Identification}
...
-daily usage of the package or class and make no provision for the case that the
-developer made a mistake!
+The \m{date} should be given in the form `\textsc{yyyy-mm-dd}' or
+`\textsc{yyyy/mm/dd}' and must be present if the optional argument is used.
+Exactly four digits are required for the year and two each for the month and
+day. Where necessary, zeros should be added to pad the month and day
+appropriately. If digits or separators are missing, the date will likely be
+misinterpreted: the commands expect a valid syntax to speed up the routine usage
+of the package or class and make no provision for the case that the developer
+made a mistake!
```
car222222, line 400: Try:
. . . in this identification. 
Then it becomes clear which "developer" is being referenced here.

car222222 (changes requested):
Only one new idea this time.

But there are some other suggestions, from various people, that are not yet resolved.

base/doc/clsguide.tex
car222222: So should this be marked as "resolved"?

car222222:
One query for @muzimuzhi .

Otherwise looks GTG (Good to Go)!

base/doc/clsguide.tex
```diff
@@ -977,6 +982,62 @@ \subsection{Passing options around}
...
+\end{decl}
+These commands test whether the named package, class or file has been loaded: this is
+done by using information that must be contained in an appropriate
+|\Provides...| line, as described in Section~\ref{sec:ID}. In the case of
+\cs{IfFileLoadedTF}, the full \meta{file-name} must be provided; in contrast,
+no extension should be given if testing for a package or class.
+
+\begin{decl}
+  |\IfPackageLoadedWithOptionsTF| \arg{package-name} \arg{options} \arg{true code} \arg{false code}\\
+  |\IfClassLoadedOptionsTF| \arg{class-name} \arg{options} \arg{true code} \arg{false code}
```
muzimuzhi, line 1009:
```suggestion
  |\IfClassLoadedWithOptionsTF| \arg{class-name} \arg{options} \arg{true code} \arg{false code}
```

muzimuzhi (changes requested):
Really not intentional but a remaining typo is (finally) caught.

All others look good to me.

base/doc/clsguide.tex
muzimuzhi: Yes I think it can be marked as "resolved" for now.

car222222:
There is always "just one more", it seems!

car222222:
Phew!  Thanks.

PR #1455, opened 2024-09-02
Fix for tagging36+tagging641

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


u-fischer:
imho the `\@noitemerr` in the `\end` part is in the wrong place. It is in `\__block_list_end:` but that is only used with active tagging and so this doesn't warn about an empty bibliography:

~~~~
\DocumentMetadata
  {
    lang=en-US,
    pdfversion=2.0,
    pdfstandard=ua-2,
    testphase={block}
  }
\documentclass{article}

\begin{document}
text

\begin{thebibliography}{9} % or this
\end{thebibliography}

\end{document}
~~~~

FrankMittelbach:
Am 03.09.24 um 17:21 schrieb Ulrike Fischer:
> imho the ***@***.***| in the |\end| part is in the wrong place. It is 
> in |\__block_list_end:| but that is only used with active tagging and so 
> this doesn't warn about an empty bibliography:

tricky, but you are right, that was wrong from the start (I only added 
the switch change).

About time, we change that file to sockets, then things get clearer and 
I wouldn't have added anything into a tagging socket or at least put it 
in all plugs for that socket

frank

u-fischer:
> About time, we change that file to sockets, 

yes ;-) Perhaps a first step would be to always use `tag` in the name of internal commands that are only used in tagging.

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2210,12 +2271,26 @@
...
+%    If we encounter horizontal material before the first \cs{item} we
+%    do want a \cs{@noitemerr} straight away, because afterwards we
+%    end up with tagging structure faults  and the cause is the
+%    missing \cs{item}. So we setup up \cs{@@_item_everypar:} to test
+%    for this. When the first \cs{item} is encountered this is then
+%    reset. This is only relevant for vertical lists, if we have
+%    inline lists one would need to test for somethingg else to
+%    identify that there is horizontal material between list start and
+%    the first \cs{item}, maybe some \cs{spacefactor} trick could be
+%    used then.\fmi{Think about a better implementation at some point.}
```
josephwright, line 2289: As we know the first thing after the start of the env. must be `\item`, I suspect a look-ahead might be easier ...

josephwright (approved):
Looks good to me

required/latex-lab/latex-lab-block.dtx
u-fischer: that is not true, e.g. this is quite allowed:
~~~~
\begin{enumerate} 
\setcounter{enumi}{100}
\item blub
\end{enumerate}
~~~~

FrankMittelbach:
> yes ;-) Perhaps a first step would be to always use `tag` in the name of internal commands that are only used in tagging.

well first step is to do this in a separate PR. but also revisit the whole tagging/notagging situation. Right now the recipes aren't set up if tagging is not active, which is fine if the whole files is without tagging but probably wrong if tagging is temporarily disabled so could be that the whole setup is faulty in that case.

Actually I was wrong, I moved the test for `@newlist`to avoid an extra test, so it is my fault.

required/latex-lab/latex-lab-block.dtx
josephwright: Well that should clearly be `\begin{enumerate}[start = 100]` or similar :)

More seriously, one could check that the result of boxing material between the start of the env and the first `\item` has no width.

FrankMittelbach: I doubt it. the item might be hidden as in `\bibitem` or worse. On the other hand, it might be the only thing reliable. The spacefact trick should work most of the time except for characters with spacefactor 0. Might be acceptable, might be not ... we'll see

josephwright: I'm reasonably serious about looking ahead - it would be easy enough to cover `\bibitem` as well as `\item`, and one could have redirection for older documents but insist that lists have to start `\(bib)item` with `\DocumentMetadata`

josephwright: Anyway, that's an aside - at present the PR doesn't change the behaviour here

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2528,6 +2604,16 @@
 }
 %    \end{macrocode}
 %
+%    This is the setting for \cs{@@_item_everypar:} before the first
+%    \cs{item} is encountered.
+%  \changes{v0.8q}{2024/09/02}{Call \cs{@noitemerr} if hmode is
+%    started before the first item}
+%    \begin{macrocode}
+\cs_new:Npn \@@_item_everypar_first: {
```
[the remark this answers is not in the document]
FrankMittelbach: somehow I'm not a big fan of protected when it is not needed, I probably should as I can't see that no protection makes it faster, I guess there are other places then as well.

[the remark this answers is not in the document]
josephwright: There was a smiley - I know not everyone agrees, I can live with it

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: > ;-). But then `\makebox[0pt]{blub}` would be allowed ... or a full overlay tikzpicture.

not true. any box changes spacefactor to 1000 ... see TeX book. So if you start with spacesfactor that is out of range you see any hmode stuff unless it is ")" and one or two others

required/latex-lab/latex-lab-block.dtx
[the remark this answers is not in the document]
FrankMittelbach: my point is, I'm doing it probably for the wrong reason (my brain says it is faster (which it most likely isn't) so the only remaining reason is it is less to type :-)

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: > Well that should clearly be `\begin{enumerate}[start = 100]` or similar :)
> 
> More seriously, one could check that the result of boxing material between the start of the env and the first `\item` has no width.

Testing is neat idea, except for a bad optimisation of TeX: anything that goes in an \hbox gets no discretionaries added and unbox it later doesn't help unfortunately if I remember correctly.

required/latex-lab/latex-lab-block.dtx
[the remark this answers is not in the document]
Skillmon: @FrankMittelbach which makes it faster, just not during runtime. More seriously the team having a clear line when to use protected would be a good thing.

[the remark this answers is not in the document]
FrankMittelbach: @Skillmon what makes what faster? and why?

required/latex-lab/latex-lab-block.dtx
Skillmon: Nope, `\discretionary` works fine:

```latex
\documentclass{article}

\begin{document}
\hsize=2cm
\sbox0{This is important crabbable.}\unhbox0

\sbox0{This is im\discretionary{i}{}{}portant crab\discretionary{i}{}{}bable.}\unhbox0
\end{document}
```

required/latex-lab/latex-lab-block.dtx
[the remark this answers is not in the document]
Skillmon: The less characters to type makes writing the code faster. The first part of my message was a bad joke, sorry. The second part I was serious though, please be consistent with using protected (of course some might slip through, but as far as humanly possible).

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: nope  partly .... :-) I didn't mean explicit hyphens, but I remembered incorrectly. it is not the hyphen that has a problem (as they are added only when the paragraph is finally built) but the language nodes that are omitted in an hbox.

Try this:
```
\documentclass{article}

\begin{document}

\def\sample{%
A few words in one language \language1
A few words in one language \language2
A few words in one language \language0}

\sample

\setbox0\hbox{\sample}

\unhbox0

\showoutput

\end{document}
```

required/latex-lab/latex-lab-block.dtx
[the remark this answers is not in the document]
FrankMittelbach: my consistency in the past was: don't use it unless it is needed because the majority of internal commands do not need protection.

u-fischer (approved):

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2210,12 +2298,26 @@
       \tl_set_eq:NN \@itemlabel \l_@@_item_label_tl
     }
 %    \end{macrocode}
-%    finally, we signal that we are at the start of a new list (which
+%    Finally, we signal that we are at the start of a new list (which
 %    effects how the first \cs{item} is handled and how \cs{par}
```
car222222, line 2302: "affects" is (I think!) needed here: effect is a noun only(??), the verb is "affect".

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2210,12 +2298,26 @@
...
 %    commands are interpreted.
 %    \begin{macrocode}
     \legacy_if_gset_true:n { @newlist }
 %    \end{macrocode}
+%    If we encounter horizontal material before the first \cs{item} we
+%    do want a \cs{@noitemerr} straight away, because afterwards we
+%    end up with tagging structure faults  and the cause is the
```
car222222, line 2309: Better? : 
. . . faults whose cause is . . .

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2210,12 +2298,26 @@
...
+%    missing \cs{item}. So we setup up \cs{@@_item_everypar:} to test
+%    for this. When the first \cs{item} is encountered this is then
```
car222222, line 2311: Maybe 
". When" ==> "; when" 
And:
"is then" ==> "will get"

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2210,12 +2298,26 @@
...
+%    reset. This is only relevant for vertical lists, if we have
+%    inline lists one would need to test for somethingg else to
```
car222222, line 2313: gg ==> g

And:

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2210,12 +2298,26 @@
...
 %    commands are interpreted.
 %    \begin{macrocode}
     \legacy_if_gset_true:n { @newlist }
 %    \end{macrocode}
+%    If we encounter horizontal material before the first \cs{item} we
+%    do want a \cs{@noitemerr} straight away, because afterwards we
+%    end up with tagging structure faults  and the cause is the
+%    missing \cs{item}. So we setup up \cs{@@_item_everypar:} to test
+%    for this. When the first \cs{item} is encountered this is then
+%    reset. This is only relevant for vertical lists, if we have
```
car222222, line 2312: "if we have" ==> "when dealing with" 
OR ==> "for"

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2210,12 +2298,26 @@
...
+%    inline lists one would need to test for somethingg else to
+%    identify that there is horizontal material between list start and
```
car222222, line 2314: "list start" ==> "the start of the list"

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2210,12 +2298,26 @@
...
+%    the first \cs{item}, maybe some \cs{spacefactor} trick could be
```
car222222, line 2315: Better:
. . . item (maybe some . . . should be used in this case).

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2457,26 +2560,26 @@
 %
 %
 %
-% \begin{macro}{\@@_item_everypar:, \@@_item_everypar_std:}
+% \begin{macro}{\@@_item_everypar:, \@@_item_everypar_std:, \@@_item_everypar_first:}
 %    The \cs{@@_item_everypar:} command is executed as part of \hook{para/begin}
 %    but most of the time does nothing, i.e., it has the following
-%    default definition.
+%    default definition outside of lists and most of the time within lists.
```
car222222, line 2566: Better:
. . . outside lists (and, most of the time, also within lists).

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2457,26 +2560,26 @@
...
+\AddToHook{para/begin}[items]{\@@_item_everypar:}
 %    \end{macrocode}
 %
 %    Note that we have to make sure that the above code is executed
 %    after the hook chunk from \texttt{tagpdf} because the latter uses
 %    \texttt{@inlabel} to make a decision.
 %
 %    By the end of the day both should probably move into the kernel
-%    hook instead!
+%    hook instead or into sockets really.
```
car222222, line 2580: What does "really" mean here?
Maybe? : 
instead, or even into sockets. 
Or:
or, better, into sockets.

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2528,6 +2631,16 @@
 }
 %    \end{macrocode}
 %
+%    This is the setting for \cs{@@_item_everypar:} before the first
```
car222222, line 2634: setting for ==> meaning of

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3971,15 +4092,38 @@
 %    Finally, at the list end we have to close the open
 %    \texttt{LBody}, \texttt{LI}, \texttt{L}, and possibly a
 %    \struct{text} if the last item ends with a list.
+%    However, if the user forgot to add an \cs{item} there is no
```
car222222, line 4095: ==> . . . then there will be no . . .

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3971,15 +4092,38 @@
...
+%    containing any \cs{bibitem}s. For that reason \cs{@noitemerr} is
+%    redefined inside that environment to only produce a warning;
+%    hence we have to produce correct tag structures in that case.
+%  \changes{v0.8q}{2024/09/02}{Do not close LI and LBody if they
+%    never were opened.}
 %    \begin{macrocode}
 \cs_set:Npn \@@_list_end: {
-  \legacy_if:nT { @endpe }
+%    \end{macrocode}
+%    If \texttt{@newlist} is true (ie when we have an error or warning
```
car222222, line 4113: i.e.,

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3971,15 +4092,38 @@
...
+%    situation) there is not much to close. If \verb=<L></L>= is not
+%    valid syntax we could add an empty \verb=<LI>= at this point instead.\fmi{check}
```
car222222, line 4115: Info: Currently it is clearly stated that any PDF structure can be empty, but this is a question that is under active, often heated, discussion!

car222222 (commented):
This is very hairy stuff: I wonder whether it needs to be so convoluted?
Meanwhile: 
Lots and lots here!  Including some answers. 

I do not seem to be allowed to approve, or otherwise, this one! Strange.

required/latex-lab/latex-lab-block.dtx
car222222: "the language nodes that are omitted in an hbox."
How weird! "No bad language please, as we are boxed in!"

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1792,6 +1795,13 @@
 %    \begin{macrocode}
   \__kernel_displayblock_end:
 %    \end{macrocode}
+%    Resetting the \texttt{@newlist} switch is also only done if the
```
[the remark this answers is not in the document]
FrankMittelbach: because 3 sentences earlier something is only done for lists and not unconditionally. That is a reference back to that.

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: One of the few (I would say) misguided optimizations, I guess the idea was that in hboxes there will be no hyphenation thus avoid extra processing.

PR #1457, opened 2024-09-04
Suspend tagging in \@settodim

u-fischer:
## Status of pull request

x Ready to merge

- [x] Test file(s) added
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated imho not needed


base/ltlength.dtx
```diff
@@ -133,13 +133,31 @@
 % \begin{macro}{\@settodim}
 % \changes{LaTeX2e}{1993/11/22}{Macro added}
 % \changes{v1.0a}{1994/03/07}{(DPC) Extra group for colour}
+% \changes{v1.1e}{2024/09/04}{(UF) Suspend tagging} 
 %    The obvious analogs of |\settowidth|.
 %    \begin{macrocode}
-\def\@settodim#1#2#3{\setbox\@tempboxa\hbox{{#3}}#2#1\@tempboxa
+%</2ekernel>
+%<*2ekernel|latexrelease>
+%<latexrelease>\IncludeInRelease{2024/09/04}%
```
FrankMittelbach, line 141: wrong: it is included in the release 2024/11/01

base/ltlength.dtx
```diff
@@ -133,13 +133,31 @@
...
+%<latexrelease>                 {\@settodim}{suspend tagging}%
+\def\@settodim#1#2#3{\setbox\@tempboxa\hbox
+  {{\SuspendTagging{@settodim}#3\ResumeTagging{@settodim}}}#2#1\@tempboxa
```
FrankMittelbach, line 144: Maybe not for this PR but we should consider always writing something like \string\@settodim as that is faster: 2 tokens instead of 9 each time

FrankMittelbach (approved):

FrankMittelbach (changes requested):
sorry was to supposed to request a change

base/ltlength.dtx
u-fischer: hm, not sure that I understand. `\string@settodim` are 10 tokens aren't they?

josephwright: @u-fischer  No, one token assuming it's something like `\edef\string@settodim{\detokenize{@settodim}}` - TeX only needs to read the one macro token when skipping ...

FrankMittelbach: On the other hand maybe this is never needed but  if we always detokenize and write `\SuspendTagging{\foo}` then this has the danger that `\foo` might not be protected while `\SuspendTaggging` is. Not sure what is better. Somehow `\SuspendTagging{\@settodim}` is nicer and shorter.

u-fischer: well `\SuspendTagging` passes the argument to `\tag_stop:n` and that expands the argument in the debug message. So `\SuspendTagging{\foo}` would need a change in tagpdf first. But looking at it: Why is `\SuspendTagging` not protected?

FrankMittelbach: and in `\tag_stop:n` it makes probably sense to allow expansion; so we should either use `\string`on toplevel or detokenize the arg in `\SuspendTagging on the way to `\tagstop`. Or perhaps, expansion is not needed in `tag_stop` in practice in which case it could be done there. Anyway, looks like a separate PR.

FrankMittelbach: Bottom line, I guess you should figure out if you do need expansion in `\tag_stop:n`, if yes I would suggest to use `\string\...`as the debug text. Or we do something like
```
    \cs_set:Npn \SuspendTagging #1 {
      \tag_stop:n {\exp_not:n{#1}}
    }
    \cs_set:Npn \ResumeTagging #1 { \tag_start:n {\exp_not{#1}} }
```
in lttagging. If there is no use case for expanding the arg  of`\tag_stop:n`then I would correct it there.

u-fischer: I don't think that there was a concrete reason to expand the argument, more the general idea that one perhaps want to create a label programatically with e.g. a number. I can change that easily and then we can replace the various cases later in one go. OK?

FrankMittelbach: well, that's why I said maybe one wants it there. So I think it might be better to make the change in lttagging as we certainly don't need expansion in that case.

Why don't you do that in this pr (it wouldn't need rollback imho)

u-fischer: but expansion costs time even if there is nothing to expand. So I think it is more effective not to expand in \tl_stop:n, and if we really get a case where expansion is wanted to generate \tl_stop:e.

FrankMittelbach: but it is expansion that happens only when debugging or not? So it doesn't matter (much).

FrankMittelbach: Of course, your `:e`approach would work too, whatever you prefer.

josephwright:
Status here?

FrankMittelbach:
> Status here?

its waiting for `\SuspendTagging{@settodim}` to become `\SuspendTagging{\@settodim}`

u-fischer:
I have uploaded a new tagpdf to ctan. I have already updated the relevant latex-lab files in a branch and will merge that tomorrow and then update this PR so that it can be merged too.

PR #1477, opened 2024-09-25
add support for amsthm theorems via firstaid

u-fischer:
This resolves e.g. https://github.com/latex3/tagging-project/issues/83

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [x ] Version and date string updated in changed source files

### DIFF
--- a/required/latex-lab/latex-lab-firstaid.dtx
+++ b/required/latex-lab/latex-lab-firstaid.dtx
@@ -16,8 +16,8 @@
 %
 % for those people who are interested or want to report an issue.
 %
-\def\ltlabfirstaiddate{2024-08-11}
-\def\ltlabfirstaidversion{0.85e}
+\def\ltlabfirstaiddate{2024-09-26}
+\def\ltlabfirstaidversion{0.85f}
 %<*driver>
 \documentclass{l3doc}
 \EnableCrossrefs
@@ -87,7 +87,6 @@
       } }
   \exp_args:Ncx\str_if_eq:onT{ver@#1.#2}{#3}
 }
-\ExplSyntaxOff
 %    \end{macrocode}
 %  \end{macro}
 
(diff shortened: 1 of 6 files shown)


required/latex-lab/latex-lab-firstaid.dtx
```diff
@@ -105,7 +104,168 @@
...
 %    \end{macrocode}
 %
+% \subsection{ams classes and amsthm}
+% The amsart, amsbook and amsproc classes redefine the theorem code
+% and this breaks the tagging added by the block code. The following
+% reenables tagging. It does \emph{not} give a completly identical output
+% (similar to the new theorem code, 
+%  see \url{https://github.com/latex3/tagging-project/issues/715}). 
+%  The code also does not try to use sockets yet, as the theorem definitions
+%  in the block code don't do that yet either.
```
FrankMittelbach, line 114: maybe we should record that task in a separate issue (or in 715?) to make sure that we do fix the socket usage at some point in all places

FrankMittelbach (approved):

required/latex-lab/latex-lab-firstaid.dtx
u-fischer: I would suggest a dedicated "review firstaid" issue that we edit if we add or changes something + cross references to specific issues if they exist. Does that sound ok? (Can one add a reminder on an issue so that one looks at it every month or so?)

FrankMittelbach: ask the experts :-) I'm not sure you can. My problem with summary issues is that after a while they typically get messy and that way fairly useless, because you can't easily drop things done if in comments. So that would only work if you always just edit the main issue text. My feeling is individual issues for individual identified  tasks is better, but that's just me perhaps.

u-fischer: well yes, I prefer issues with a single topic too, but that also means that one shouldn't add a remark to an existing issue, but open new ones for each firstaid-review task.

FrankMittelbach: yes and no. if it is sensible to resolve a task/remark as part of the existing issue then i think it is useful to stay just with a single issue, but I made that remark above without looking at what the other issue was explicitly saying, so yes, int that case it doesn't really should be added to the existing one.

PR #1488, opened 2024-10-09
latex-lab  handling \item and \itemsep

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added

Normalizing key names (a breaking change) will happen at a later stage


required/latex-lab/latex-lab-block.dtx
```diff
@@ -390,7 +392,7 @@
 %              {Instance of type \texttt{item} to be used to format
 %               the label string}{basic}
 %  \TemplateKey{item-skip}{skip}{The space in front of an item in the
-%               list.\fmi{May need to be on a different template level}}{\cs{itemsep}}
+%               list. If not specified the value specified in the block template instance is used}{}
```
muzimuzhi, line 395: Unpaired braces due to removal of `\fmi{`.

FrankMittelbach: sorry don't see what you mean

u-fischer (commented):
I think the definitions `\def\propagate@doendpe` and `\def\@endpetrue` can go. They are already in the kernel. And probably also the following redefinition of `end `. 

On line 1340 a variable `\l_@@_env_params_tl`  is declared that is  imho not used (the used variable name is `\l_@@_legacy_env_params_tl` which is declared elsewhere.

Apart from this it looks good.

FrankMittelbach:
> I think the definitions `\def\propagate@doendpe` and `\def\@endpetrue` can go. They are already in the kernel. And probably also the following redefinition of `end `.

yes, but `\propagate@doendpe` has extra debugging code not in the kernel version. Perhaps now is the right point to drop that. Thoughts?

> 
> On line 1340 a variable `\l_@@_env_params_tl` is declared that is imho not used (the used variable name is `\l_@@_legacy_env_params_tl` which is declared elsewhere.

true

u-fischer:
> Perhaps now is the right point to drop that.

Well I would try to cleanup when possible. This kind of stuff gets forgotten and then you wonder why after a change you don't get the expected result.

FrankMittelbach:
> > Perhaps now is the right point to drop that.
> 
> Well I would try to cleanup when possible. This kind of stuff gets forgotten and then you wonder why after a change you don't get the expected result.

okay, done now

PR #1500, opened 2024-10-20
Make two argument tagging sockets preserve the second arg when suspended and add expandable variants

zauguin:
First part synchronized with 7a2b09d51edb7ee5619c45e33e126e48f1d3b406 in tagpdf.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added
- n/a Rollback provided (if necessary)?

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,10 +6,12 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
-2024-10-21  Marcel Krüger  <Marcel.Krueger@latex-project.org>
+2024-10-27  Marcel Krüger  <Marcel.Krueger@latex-project.org>
 	* lttagging.dtx
 	Adapt two argument tagging sockets to preserve second argument when the socket is disabled.
 	Add expandable versions of tagging socket use commands.
+	* ltsocket.dtx
+	Add expandable versions of \socket_use_expandable:n
 
 2024-10-26  Yukai Chou <muzimuzhi@gmail.com>
 	* ltcounts.dtx (subsection{Environment Counter Macros}):
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,6 +6,10 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2024-10-21  Marcel Krüger  <Marcel.Krueger@latex-project.org>
+	* lttagging.dtx
+	Adapt two argument tagging sockets to preserve second argument when the socket is disabled.
+
 2024-10-26  Yukai Chou <muzimuzhi@gmail.com>
 	* ltcounts.dtx (subsection{Environment Counter Macros}):
 	Fully expand counter name in \theH<counter> commands (gh/1508)
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -8,7 +8,8 @@ not part of the distribution.
 
 2024-10-27  Marcel Krüger  <Marcel.Krueger@latex-project.org>
 	* lttagging.dtx
-	Adapt two argument tagging sockets to preserve second argument when the socket is disabled.
+	Change tagging sockets with two arguments to drop the first argument
+	but return the second argument when the socket is disabled with \SuspendTagging.
 	Add expandable versions of tagging socket use commands.
 	* ltsocket.dtx
 	Add expandable versions of \socket_use_expandable:n.
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -11,7 +11,8 @@ not part of the distribution.
 	Adapt two argument tagging sockets to preserve second argument when the socket is disabled.
 	Add expandable versions of tagging socket use commands.
 	* ltsocket.dtx
-	Add expandable versions of \socket_use_expandable:n
+	Add expandable versions of \socket_use_expandable:n.
+	Make internal plug definitions non-protected to allow expandable use.
 
 2024-10-26  Yukai Chou <muzimuzhi@gmail.com>
 	* ltcounts.dtx (subsection{Environment Counter Macros}):
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -9,6 +9,7 @@ not part of the distribution.
 2024-10-21  Marcel Krüger  <Marcel.Krueger@latex-project.org>
 	* lttagging.dtx
 	Adapt two argument tagging sockets to preserve second argument when the socket is disabled.
+	Add expandable versions of tagging socket use commands.
 
 2024-10-26  Yukai Chou <muzimuzhi@gmail.com>
 	* ltcounts.dtx (subsection{Environment Counter Macros}):
--- a/base/lttagging.dtx
+++ b/base/lttagging.dtx
@@ -104,9 +104,7 @@
 %  debugging strings\ldots
 %
 %
-% \DescribeMacro\UseExpandableTaggingSocket
 % \DescribeMacro\UseTaggingSocket
-% \DescribeMacro\tag_socket_use_expandable:n
 % \DescribeMacro\tag_socket_use:n
 % \DescribeMacro\tag_socket_use:nn
 % \DescribeMacro\tag_socket_use:nnn
@@ -141,6 +139,8 @@
 % if \cs{SuspendTagging} is in force. There may be reasons for doing
 % that but in general we expect to always use \cs{UseTaggingSocket}.
 %
+% \DescribeMacro\UseExpandableTaggingSocket
+% \DescribeMacro\tag_socket_use_expandable:n
 %  For special cases like in some \cs{halign} contexts we need a fully expandable
 %  version of the commend. For these cases, \cs{UseExpandableTaggingSocket} can be
 %  used. To allow being expandable, it does not output any debugging information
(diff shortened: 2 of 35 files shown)


base/lttagging.dtx
```diff
@@ -104,49 +104,55 @@
...
-%  arguments are ignored (including the case when there is a single
-%  one), i.e., all sockets used with \cs{UseTaggingSocket} behave as
-%  if the \text{noop} plug is assigned.
+%  Beside being slightly shorter, the big advantage is that this way
+%  we can change \cs{UseTaggingSocket} to do nothing by switching a boolean
+%  instead of changing the plugs of the tagging support sockets back and forth.
+%
+% Usually, these sockets have exactly two plugs defined.
+% When using \cs{UseTaggingSocket} the first plug is always
+% used if tagging is suspended.
```
FrankMittelbach, line 130: That description doesn't really make sense to me. The point of \SuspendTagging is that no plugs are changed but the the behavior of \UseTaggingSocket is changed. It is not a defined "plug" that is used but rather that no plug is used but that the socket stops acting like a socket more or less. While one can describe the behavior as if there is a plug but in practice the plug may not exist and on it only behaves as if the plug is assigned (which is what I wrote in my version of the description).

base/lttagging.dtx
```diff
@@ -104,49 +104,55 @@
...
+% For the sockets with zero and one argument
+% this is the \texttt{noop} plug, so they gobble the
+% argument in such a case. The assigned plug use when tagging is enable.
```
FrankMittelbach, line 133: sentence at the end of the line is no sentence :-) but again I think the whole para need rewriting

base/lttagging.dtx
```diff
@@ -104,49 +104,55 @@
...
+% There may be more plugs, e.g., tagging with special debugging, etc.,
+% but right now it is usually just on or off.
 %
 %  It is possible to use the tagging support sockets with
-%  \cs{UseSocket} instead of \cs{UseTaggingSocket}, but in this case
-%  the socket remains active if \cs{SuspendTagging} is in force. There
-%  may be reasons for doing that, but in general we expect to always
-%  use \cs{UseTaggingSocket}.
+%  \cs{UseSocket} directly, but in this case the socket remains active
+%  if e.g. \cs{SuspendTagging} is in force. There may be reasons for doing
```
FrankMittelbach, line 143: e.g. means "for example"  I don't think it should be here

base/lttagging.dtx
```diff
@@ -104,49 +104,55 @@
...
+%  that but in general we expect to always use \cs{UseTaggingSocket}.
 %
-%  The L3 programming layer versions \cs{tag_socket_use:n} and
-%  \cs{tag_socket_use:nn} are slightly more efficient than
+%  For special cases like in some \cs{halign} contexts we need a fully expandable
+%  version of the commend. For these cases, \cs{UseExpandableTaggingSocket} can be
+%  used.
```
FrankMittelbach, line 148: Add something like:

It does not uotput any debugging info (to remain expandable) if \DebugSocketsOn is used and should therefore only used if really necessary.

base/lttagging.dtx
u-fischer: I agree that it is misleading. I changed the wording in the tagpdf documentation ([tagpdf-user, currently in the taggingsocket branch](https://github.com/latex3/tagpdf/commit/6522fabb82092e98b3b2073458f11e62f1a07347)). If that sounds ok, it can be copied here.

FrankMittelbach: I made one suggestion for change in your text @u-fischer, otherwise I think that's fine

base/lttagging.dtx
FrankMittelbach: or the other way around:  

If \DebugSocketsOn is used it does not output any debugging info (to remain expandable) und should therefore ...

josephwright:
Bump

base/lttagging.dtx
```diff
@@ -209,6 +219,19 @@
...
+  \int_case:nnF
+      { \int_use:c { c__socket_tagsupport/#1_args_int } }
+         {
+           0 \prg_do_nothing:
+           1 \use_none:n
+           2 \use_ii:nn
+         }
+         \ERRORusetaggingsocket  % that should get a proper error message
+}
+%    \end{macrocode}
```
u-fischer, line 234: Do we need expandable tagging sockets with arguments? (There is only one l3layer version without argument). And should there be an error condition? If not the definition could be simplified to `\cs_new:Npn \UseExpandableTaggingSocket #1 {\prg_do_nothing:}`

u-fischer (commented):
Apart from one question and documentation issue it is imho fine.

u-fischer:
Actually trying to implement the equivalent in tagpdf I realize that an `\socket_use_expandable:n` is missing. (and imho we should restrict that for now to the zero argument case if no use case appear.

So in ltsockets we need also 

~~~~
\cs_new:Npn \socket_use_expandable:n #1 {
  \use:c { @@_#1_plug_ \str_use:c { l_@@_#1_plug_str } :w }
}
~~~~

base/lttagging.dtx
zauguin: Adjusted following Ulrike's `tagpdf` change.

base/lttagging.dtx
zauguin: Currently I do not think that we have a need for them. I don't think that the command should be simplified like that though since then attempts to use them would silently do the wrong thing. Instead they should either work or have a clear error. Given that it's not really more work to just support the case here and it's more consistent, I would support arguments in `\UseExpandableTaggingSocket`, but I also wouldn't mind changing this to error instead.

u-fischer: I don't mind either way, only asking. But in any case we need underlying expandable sockets in ltsockets, as with tagging active I must be able to call the socket really.

u-fischer:
a changes entry for ltsockets + changes.txt?

zauguin:
> a changes entry for ltsockets + changes.txt?

I added them now.

u-fischer:
I'm not sure if the expandable version is correct, it doesn't fully expand if I do this:

~~~~
\documentclass{article}

\NewSocket{blub}{0}
\NewSocketPlug{blub}{abc}{abc}
\AssignSocketPlug{blub}{abc}
\begin{document}
\ExplSyntaxOn
\show\socket_use_expandable:n
\edef\blub{\socket_use_expandable:n{blub}}
\show\blub
\ExplSyntaxOff
\end{document}
~~~~

but I have to go now.

zauguin:
The problem was that the plugs are defined as protected commands so they do not expand. I now changed them to be non-protected which should not make any difference otherwise since they are nomally only referenced from protected commands, but it would be great if someone double checks that I didn't miss anything.

u-fischer (approved):

base/changes.txt
```diff
@@ -6,6 +6,14 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2024-10-27  Marcel Krüger  <Marcel.Krueger@latex-project.org>
+	* lttagging.dtx
+	Adapt two argument tagging sockets to preserve second argument when the socket is disabled.
```
FrankMittelbach, line 11: Change tagging sockets with two arguments todrop the first argument but return the second argument when the socket is disabled with \SuspendTagging.

base/lttagging.dtx
```diff
@@ -104,49 +104,54 @@
...
+% The sockets with zero arguments do nothing. The sockets with one argument
+% gobble their argument. The sockets with two arguments
+% will drop their first argument and pass the second unchanged.
+%
+% It is possible to use the tagging support sockets with
+% \cs{UseSocket} directly, but in this case the socket remains active
+% if \cs{SuspendTagging} is in force. There may be reasons for doing
+% that but in general we expect to always use \cs{UseTaggingSocket}.
+%
+%  For special cases like in some \cs{halign} contexts we need a fully expandable
```
FrankMittelbach, line 144: I would move

\DescribeMacro\UseExpandableTaggingSocket
\DescribeMacro\tag_socket_use_expandable:n

to highlight that these are special and should not be used unless really needed (as the text says)

base/lttagging.dtx
FrankMittelbach: I think keeping them aligned as suggested by @zauguin is fine (even if we don't use them right now. We clear do not intend to provide a different  behavior other than one is epeandabel and the other is not.

FrankMittelbach (approved):
approved with a few suggestions for altering the documentation

FrankMittelbach:
> a changes entry for ltsockets + changes.txt?

Maybe we should have a news entry as well? explaining the behavior of \UseTaggingSocket (with different kinds of arguments). After all it is a breaking change even though it hasn't been used yet other than by us.

FrankMittelbach:
> The problem was that the plugs are defined as protected commands so they do not expand. I now changed them to be non-protected which should not make any difference otherwise since they are nomally only referenced from protected commands, but it would be great if someone double checks that I didn't miss anything.

I think we are good here.

zauguin:
> > a changes entry for ltsockets + changes.txt?
> 
> Maybe we should have a news entry as well? explaining the behavior of \UseTaggingSocket (with different kinds of arguments). After all it is a breaking change even though it hasn't been used yet other than by us.

I added a news entry for the breaking change of two argument tagging sockets but did not mention the expandable sockets there since they are neither a breaking change nor recommended for external users.

FrankMittelbach (approved):
good to go now I would say

PR #1509, opened 2024-10-26
Fully expand counter name in `\theH<counter>` commands

### LINKED ISSUE #1508: Missing full expansion on counter name in `\theH<counter>` representation
## Brief outline of the bug

When defining `\theH<counter>` representation, no full expansion is made on the counter name, which is inconsistent with the previous behavior in `hyperref` and causes problems, see https://github.com/latex3/hyperref/issues/357 for instance.

```latex
\documentclass{article}

\def\fooname{foo}
\newcounter{\fooname}

\show\theHfoo

\begin{document}
\end{document}
```

```tex
\show\theHfoo

expected: ->\the \value {foo}.
actual:   ->\the \value {\fooname }.
```

[latex2e-theHcounter-representation.log](https://github.com/user-attachments/files/17528942/latex2e-theHcounter-representation.log)

muzimuzhi:
As the fix is for unreleased latex-dev release, no rollback is needed I think.

Resolves #1508.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


base/ltcounts.dtx
```diff
@@ -371,7 +374,9 @@
 %<latexrelease>                             {provide theHfoo commands}%
 %<*2ekernel|latexrelease>
 \def\@addtoreset#1#2{\expandafter\@cons\csname cl@#2\endcsname {{#1}}%
-  \expandafter\gdef\csname theH#1\endcsname{\csname theH#2\endcsname.\the\value{#1}}%
+  \expandafter\xdef\csname theH#1\endcsname{%
+    \expandafter\noexpand\csname theH#2\endcsname.%
+    \noexpand\the\noexpand\value{#1}}%
```
FrankMittelbach, line 379: better perhaps?

\expandafter\gdef\csname theH#1\expandafter\endcsname\expandafter{\csname theH#2\endcsname.\the\value{#1}}%

FrankMittelbach (changes requested):
not 100% sure this is faster, so really just a suggestion

FrankMittelbach:
> As the fix is for unreleased latex-dev release, no rollback is needed I think.

agreed, nowhere to rollback to :-)

josephwright (approved):

base/ltcounts.dtx
FrankMittelbach: or maybe not in case `#1` also needs expanding

u-fischer: ah, I see the problem, I used \value to avoid that weird redefinitions of `\@arabic`  break the pdfstrings, but then one no longer has the expansion from the \csname. The xdef is imho fine.

u-fischer (approved):

base/ltcounts.dtx
muzimuzhi: > or maybe not in case `#1` also needs expanding

@FrankMittelbach The main point here is that `#1` needs expanding. The one-step expansion on `\csname theH#2\endcsname` is just an incidental result.

An alternative solution is to fully expand both counter names (in `\newcounter{<ctr name1>}[<ctr name2>]`) as early as possible, but that needs redefine more commands. Or, are the raw, unexpanded counter names used in any error messages?

FrankMittelbach (approved):

Issue #1495, opened 2024-10-17
bug in longtable (2024-04-26 v4.20) after a pagebreak


Reinhard-Jhr:
## After a pagebreak: \pagegoal=\maxdimen. 
\endlongtable -> .... \dimen@\pagegoal\advance\dimen@\ht\LT@foot\pagegoal\dimen@ ....

After a pagebreak you cannot add a positive value to \dimen@. This will lead to the error message: dimension too large.
I don't understand why you need this code, because the next command is an \endgroup and \pagegoal will be resetted. 

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{article}

  % Any preamble code goes here

\begin{document}

  % Demonstration of issue here
  
\end{document}
```

davidcarlisle:
Could you please supply a test file, the code shown in your description is empty

Setting the value just before `\endgroup` does look suspicous but not necessarily wrong it depends when the output routine is fired I'd need to check, and a test file would help...

This specific arrangement is relatively new (v4.14) from 819fda96d so your test file presumably doesn't show the issue if you use

```
\usepackage{longtable}[=v4.13]
```
?

davidcarlisle:
The group is not necessarily wrong (`\pagegoal` settings are global so survive the `\endgroup`)  However it is certainly true that  you can not increase pagegoal when it is `\maxdimen` so something is not right here. A test file would help, as noted above.

Reinhard-Jhr:
Excuse me, but the problem occurs in a very complex table. Here, longtable is only used indirectly in an environment that also allows page breaks within table cells and more. But in the end it is only longtable code. Unfortunately I was not able to create a simple example that shows the error. Therefore, I am sending an excerpt from the log file created with \tracingall, from which the problem can be seen.

[longtablebug.log](https://github.com/user-attachments/files/17466731/longtablebug.log)

version 4.13 does not show the error. \endlongtable is defined completely differently there.

davidcarlisle:
I can understand that making a minimal example is hard for this case. If you could send me privately the test file or put it in a private gh repository or whatever you find convenient, this would be really helpful This is a tricky area and making any change without a test file showing the problem is a bit stressful:-).

That said, it looks like line 255

```
    \dimen@\pagegoal\advance\dimen@\ht\LT@foot\pagegoal\dimen@
```

should look like  line 220

```
[...]

PR #1510, opened 2024-10-27
guard for #1495

davidcarlisle:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added

The issue  #1495 did not provide a test file, and I failed to generate one,  however the existing code could clearly, if called at a bad time, hit TeX's ` `\maxdimen` marker for `\pagegoal` at the start of a page and lead to a dimension too large error,

It's possible this is a symptom of a logical flaw in the code, but this code has been widely publicised as a patch since around 2002 before being incorporated into the main release, so I do not want to do major changes without any test files.

This change restricts to adding  a single guard to prevent `\pagegoal` being increased when it has the top of page value, so will have no effect except in cases that currently error. As such no rollback or test cases are added.

### DIFF
--- a/required/tools/changes.txt
+++ b/required/tools/changes.txt
@@ -5,6 +5,11 @@ completeness or accuracy and it contains some references to files that
 are not part of the distribution.
 =======================================================================
 
+2024-10-27  David Carlisle  <David.Carlisle@latex-project.org>
+
+	* longtable.dtx Guard against increasing \pagegoal past \maxdimen (gh/1495)
+
+
 2024-10-12  Joseph Wright  <Joseph.Wright@latex-project.org>
 
 	* array.dtx (section{The insertion of declarations ...}):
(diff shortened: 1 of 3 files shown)


required/tools/changes.txt
```diff
@@ -5,6 +5,11 @@ completeness or accuracy and it contains some references to files that
 are not part of the distribution.
 =======================================================================
 
+2024-10-27  David Carlisle  <David.Carlisle@latex-project.org>
+
+	* longtable.dtx Guard against increasing \pagegoal past \maxdimen (gh/1495)
+
+
```
josephwright, line 12: Stray line ....

josephwright (approved):
One v. minor formatting issue - we can tidy after a merge

FrankMittelbach (approved):
ltnews entry perhaps?

davidcarlisle:
> ltnews entry perhaps?

I could do if we want but I thought not, especially as I couldn't really say when this could occur, at the level of ltnews couldn't really say much more than "made a change to address gh1495" which isn't really very informative.

davidcarlisle:
> ltnews entry perhaps?

I added a line

PR #1512, opened 2024-10-28
Ltnews40 intro, corrections and some reordering

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

## Checklist of required changes before merge will be approved
- [n/a] Test file(s) added
- [n/a] Version and date string updated in changed source files
- [n/a] Relevant `\changes` entries in source included
- [n/a] Relevant `changes.txt` updated
- [n/a] Rollback provided (if necessary)?
- [x] `ltnewsX.tex` (and/or `latexchanges.tex`) updated

Looks like the ltugboat change hasn't yet made it to all mirrors, so at the moment the firstaid test file errors (after David's change to it). Not much that one can do about it other that wait, I guess.

Looks like the pull request ran through, so it is really an issue that will go away by tomorrow (hopefully).

PR #1515, opened 2024-10-28
Skip mapping over undeclared prop in `\ShowHook` etc.

muzimuzhi:
I'm not so familiar with hooks implementation, so maybe there're more uncaught similar cases.

Rollback needed? Only `\ShowHook`, `\LogHook` and friends are affected.

`ltnews40.tex` entry not added. Maybe after #1512 gets merged?

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- Ready to merge

- [x] Test file(s) added

### DIFF
--- a/base/lthooks.dtx
+++ b/base/lthooks.dtx
@@ -32,7 +32,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{lthooks.dtx}
-             [2024/10/21 v1.1j LaTeX Kernel (hooks)]
+             [2024/10/29 v1.1k LaTeX Kernel (hooks)]
 % \iffalse
 %
 \documentclass{l3doc}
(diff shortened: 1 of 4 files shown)


josephwright (changes requested):
Code and changes look fine, and rollback is not needed here, but I think a short news entry would normally be included - just something like `Showing hooks gave a low-level error if the hook was undeclared; this has now been corrected.`

FrankMittelbach (approved):
+ ltnews as suggested by Joseph

base/lthooks.dtx
```diff
@@ -6183,7 +6185,9 @@
       { #2 { The~hook~is~empty } }
       {
         \@@_log_line:x { Code~chunks: }
-        \prop_if_empty:cTF { g_@@_#1_code_prop }
+        \bool_lazy_or:nnTF
+          { ! \prop_if_exist_p:c { g_@@_#1_code_prop } }
+          { \prop_if_empty_p:c { g_@@_#1_code_prop } }
```
FrankMittelbach, line 6190: actually, it might be worth putting in a short comment why this is necessary, ie that a non-existent prop would return "empty" = true but that fail when mapping over it

FrankMittelbach: rubbish, would be empty = false ....

PR #1525, opened 2024-11-01
Introduce an f-specification argument in ltcmd

josephwright:
See latex3/latex3#591.

# Internal housekeeping

## Status of pull request

- [x] Feedback wanted 
- Under development
- Ready to merge

- [x] Test file(s) added


josephwright:
Currently this is a draft: there are some clear open questions to answer before it is ready to merge

- Do we want to cover chars 0-32 as `filecontents` does?
- Do we want a more complex approach to a preceding optional argument. For example, `fancyvrb` sets the catcode of `^^M` prior to the `\@ifnextchar` to ensure it is tokenizes in a recognisable way
- Do we want to auto-apply `!` to any preceding optional arg
- Is the letter `f` the best choice? I can see an argument for `c` ('collected'), `x` (charcode of `b` + `v`), `w` (v-squared`v`), ...

base/doc/usrguide.tex
```diff
@@ -858,6 +858,42 @@ \subsection{Typesetting verbatim-like material}
...
+  % Some code is show here
+  $y = mx + c$
+\end{MyVerbatim}
+
+As grabbing the entire contents verbatim, newlines are always permitted: if you
+use \texttt{f} rather than \texttt{+f}, a warning will be issued and \LaTeX{}
+will assume you meant to include the \texttt{+}. As for the \texttt{v}
+specification, new lines are stored as \cs{ObeyedLine}. In a similar fashion to
+the \texttt{b}~specification, by default \emph{newlines} are trimmed at both
+ends of the body. Putting the prefix |!| before \texttt{b} suppresses
```
Skillmon, line 889: Is `\texttt{b}` correct here or should that be `f`?

Skillmon:
I'd maybe drop the warning on `+f` and make `f` just behave correctly. There is no good reason not to, imho, it's also not `+b` for multiline bodies of environments.

Skillmon:
Is it possible in the architecture of `ltcmd` to change the behaviour of an argument grabber based on the next argument type? If so the optional argument should be searched with the full category code regime of `f` in place, and if it's found the regime should be reverted, the optional argument collected, the verbatim catcode regime reactivated and the body collected. And an optional argument should have otherwise the same observable rules as it would have with a `b` body (so should be found even with a new line between `\begin{foo}` and the optional argument).

Cf. the handling of the optional argument in `enverb` (which as the only strange point also finds the opening bracket of the optional argument after multiple blank lines -- maybe I should revise this :)).

Skillmon:
As for the argument name, what about an upper case `V`?

base/doc/usrguide.tex
josephwright: Sorted

josephwright:
> I'd maybe drop the warning on `+f` and make `f` just behave correctly. There is no good reason not to, imho, it's also not `+b` for multiline bodies of environments.

Thinking here is that reading the arg. spec. should be clear: collecting an environment will allow newlines, so should be `+f` not just `f`. That's why I've currently made it a warning - at a technical level, it's not an issue.

josephwright:
> As for the argument name, what about an upper case `V`?

I thought about that - but the pattern we have at the moment is all uppercase specs relate to the matching lowercase in that they take an extra `{<default>}` argument. That's not the situation here, which is why at the moment I went with something else.

josephwright:
> Is it possible in the architecture of `ltcmd` to change the behaviour of an argument grabber based on the next argument type? If so the optional argument should be searched with the full category code regime of `f` in place, and if it's found the regime should be reverted, the optional argument collected, the verbatim catcode regime reactivated and the body collected. And an optional argument should have otherwise the same observable rules as it would have with a `b` body (so should be found even with a new line between `\begin{foo}` and the optional argument).

Whilst we don't currently do this, it would be possible to do a 'pre-scan' to establish if special settings are needed. The issue I think is that unlike say `fancyvrb` or `enverb`, we don't know the nature of the optional argument(s) - they don't have to be delimited by `[..]`. So for example if someone has `+d^$ +f`, we need to be able to cope with that in a (somewhat) predictable way. We also have the case of multiple optional arguments, so it's all a bit more 'interesting' - in the end this could all be done, it's a question of what we want to provide and how well that fits more widely into the LaTeX landscape.

FrankMittelbach:
> > As for the argument name, what about an upper case `V`?
> 
> I thought about that - but the pattern we have at the moment is all uppercase specs relate to the matching lowercase in that they take an extra `{<default>}` argument. That's not the situation here, which is why at the moment I went with something else.

that pattern is strong and easy to understand so I don't think we should break it, thus that would rule `V` out in my opinion

FrankMittelbach:
Maybe a dumb Saturday evening idea, but what about using a special sort of processor? Right now `>{...}`is a processor that is applied after grabbing, but we could have `<{...}` for before, or less general `c{named-catcode-regiment}`.  The you would have  `c{verbatim}b+` for proces the body with "verbatim" catcodes. And fancyvrb could define its own named set. The whole thing would then work also with other arg specifiers. Handwaving ...

FrankMittelbach:
> Whilst we don't currently do this, it would be possible to do a 'pre-scan' to establish if special settings are needed. The issue I think is that unlike say `fancyvrb` or `enverb`, we don't know the nature of the optional argument(s) - they don't have to be delimited by `[..]`. So for example if someone has `+d^$ +f`, we need to be able to cope with that in a (somewhat) predictable way. We also have the case of multiple optional arguments, so it's all a bit more 'interesting' - in the end this could all be done, it's a question of what we want to provide and how well that fits more widely into the LaTeX landscape.

I would think this is bound to produce problems forever. Besides, what you do with respect to chars being `%`?  Could be done, as you say but would be slow and probably riddled with errors in weird cases that we hadn't thought about.

josephwright:
> Maybe a dumb Saturday evening idea, but what about using a special sort of processor? Right now `>{...}`is a processor that is applied after grabbing, but we could have `<{...}` for before, or less general `c{named-catcode-regiment}`. The you would have `c{verbatim}b+` for proces the body with "verbatim" catcodes. And fancyvrb could define its own named set. The whole thing would then work also with other arg specifiers. Handwaving ...

I see the idea, but we have a few things that mean I suspect it's not quite right

- We need to fiddle with `\endlinechar`
- We may want to 'set and reset' the catcodes (see other replies)
- It leaves finding the end point a bit awkward
- There's the entire question of nesting

However, it does suggest something we could do - see parallel reply.

josephwright:
> > Whilst we don't currently do this, it would be possible to do a 'pre-scan' to establish if special settings are needed. The issue I think is that unlike say `fancyvrb` or `enverb`, we don't know the nature of the optional argument(s) - they don't have to be delimited by `[..]`. So for example if someone has `+d^$ +f`, we need to be able to cope with that in a (somewhat) predictable way. We also have the case of multiple optional arguments, so it's all a bit more 'interesting' - in the end this could all be done, it's a question of what we want to provide and how well that fits more widely into the LaTeX landscape.
> 
> I would think this is bound to produce problems forever. Besides, what you do with respect to chars being `%`? Could be done, as you say but would be slow and probably riddled with errors in weird cases that we hadn't thought about.

I think @Skillmon's approach could work, with a logic for seeking an optional arg:

1. Set all specials and `^^M` verbatim
2. Reset the catcode of the peek token to normal
3. Peek ahead - loop to remove spaces/newlines if needed
4. Reset catcodes and grab the optional arg if found

This could be done by adding an additional specifier (lets say `*`) which means 'use a verbatim-safe approach to grabbing an optional arg'. That would allow us to avoid needed to unpick everything to decide if to use this slower approach.

Skillmon:
Performance is not that bad, it's just `\begingroup\expanded{\noexpand\dospecials\catcode`\peekchar=\the\catcode`\peekchar}\@ifnextchar\peekchar` (sort of) and then, if the argument is found `\endgroup` and collect it, else grab the verbatim argument. And: To match current behaviour of `ltcmd` you should only strip the first newline (because two consecutives are normally a `\par` hence don't match the optional argument's start-token). And depending on the stripping of the first newline in the `f`-argument you need to conditionally add the ignored newline to the verbatim argument if no optional argument is found.

josephwright:
> Performance is not that bad, it's just `\begingroup\expanded{\noexpand\dospecials\catcode`\peekchar=\the\catcode`\peekchar}\@ifnextchar\peekchar` (sort of) and then, if the argument is found `\endgroup` and collect it, else grab the verbatim argument. And: To match current behaviour of `ltcmd` you should only strip the first newline (because two consecutives are normally a `\par` hence don't match the optional argument's start-token). And depending on the stripping of the first newline in the `f`-argument you need to conditionally add the ignored newline to the verbatim argument if no optional argument is found.

I don't think we can use a group, and we have to deal with looping for spaces, but yes, I guess you are broadly right about performance.

josephwright:
> And: To match current behaviour of `ltcmd` you should only strip the first newline (because two consecutives are normally a `\par` hence don't match the optional argument's start-token). And depending on the stripping of the first newline in the `f`-argument you need to conditionally add the ignored newline to the verbatim argument if no optional argument is found.

Good point about multiple `^^M`. I suspect space skipping here is not desirable anyway. Perhaps then best to go with my `*` suggestion, which could also be applied to before `v`?

josephwright:
> > I'd maybe drop the warning on `+f` and make `f` just behave correctly. There is no good reason not to, imho, it's also not `+b` for multiline bodies of environments.
> 
> Thinking here is that reading the arg. spec. should be clear: collecting an environment will allow newlines, so should be `+f` not just `f`. That's why I've currently made it a warning - at a technical level, it's not an issue.

One might well argue that both `b` and `f` should act the same way, `+` or not - an easy adjustment if that's the overall agreement. If we do that, I'd normalise to one agreed form.

josephwright:
> > Good point about multiple `^^M`. I suspect space skipping here is not desirable anyway. Perhaps then best to go with my `*` suggestion, which could also be applied to before `v`?
> 
> Yes, applying `*` to `v` would then also be desirable. Though `ltcmd` could check for that itself I guess it's much less complicated to let the user decide and not let `ltcmd` peek ahead for the structure of (possibly all) following arguments.

I think I'll implement the suggestion, we can look at it then decide if we need automation - that would mainly sit in a different part of the code so would still require the underlying `*` code.

base/doc/usrguide.tex
```diff
@@ -858,6 +858,42 @@ \subsection{Typesetting verbatim-like material}
...
+  {\begin{center} #1 #2\end{center}} {}
+\begin{MyVerbatim}[\ttfamily\itshape]
+  % Some code is show here
+  $y = mx + c$
+\end{MyVerbatim}
+
+As grabbing the entire contents verbatim, newlines are always permitted: if you
+use \texttt{f} rather than \texttt{+f}, a warning will be issued and \LaTeX{}
+will assume you meant to include the \texttt{+}. As for the \texttt{v}
+specification, new lines are stored as \cs{ObeyedLine}. In a similar fashion to
```
lvjr, line 887: In `ltnews41.tex` it is called `\obeyedline`, why it is called `\ObeyedLine` here?

base/ltcmd.dtx
```diff
@@ -1331,28 +1331,47 @@
 % \end{macro}
 %
 % \begin{macro}{\@@_normalize_type_b:w}
-%   This argument type is not allowed for commands.  This is only
+% \changes{v1.3a}{2024-12-11}{Extend to cover \texttt{f}-type grabbing}
+% \begin{macro}{\@@_normalize_type_f:w}
+% \changes{v1.3a}{2024-12-11}{New function}
+% \begin{macro}{\@@_normalize_type_b_or_f:nn}
+% \changes{v1.3a}{2024-12-11}{New function}
+%   These argument type is not allowed for commands.  They is only
```
muzimuzhi, line 1339:
```suggestion
%   These argument types are not allowed for commands.  They are only
```

base/ltcmd.dtx
```diff
@@ -1658,16 +1677,27 @@
 % \end{macro}
 %
 % \begin{macro}{\@@_add_type_b:w}
+% \changes{v1.3a}{2024-12-11}{Extend to cover \texttt{f}-type grabbing}
+% \begin{macro}{\@@_add_type_f:w}
+% \changes{v1.3a}{2024-12-11}{New function}
+% \begin{macro}{\@@_add_type_b_or_c:N}
```
muzimuzhi, line 1683:
```suggestion
% \begin{macro}{\@@_add_type_b_or_f:N}
```

muzimuzhi (commented):
Some typos. Only read the non-code part.

josephwright:
I've made two adjustments here:

- `f`-type are always 'long'
- Use @skillmon's suggestion to grab optional arguments 'verb safe'

The second idea is partly-done. At present, you have to turn this on using a `*` specifier. If the mechanism looks OK, I can automate this - but probably only for an immediately-preceeding `D`-type argument.  I'll do that, and document, once we have agreement this looks OK. See the test file for demos.

josephwright:
We could make life a little easier e.g. by saying that optional args before `f`-type have to use 'other' chars, or using a `charcode` peek test (more risky). But on balance, I'd rather go with an automated version of what I already have -

josephwright:
In other news - do we want rollback or rather roll-forward? I'm not looking forward to it ...

josephwright:
A couple more notes

- I suspect we might want to name `\@@_grab_f_loop:w` as something 'friendly' for the case where the user messes up end-of-env: good suggestions welcome
- I've got a simple `\ERROR` at the moment for tokens before/after `\end{env}` on the final line: we need to decide what is meant to work here and what's an error

Skillmon:
For the tokens in front of `\end{env}`: `enverb` uses their count for its auto gobbling feature which makes it automatically work in `.dtx` files or in natural indentation. I'm not suggesting to automatically do this for `f`-type, but making something similar possible would be nice (maybe by putting the contents into a documented `str`-variable, or via an argument processor).

Regarding the tokens after it: There are two possibilities, imho, either throw an error or throw a warning (I think `listings` does that).

base/testfiles-ltcmd/ltcmd009.lvt
```diff
@@ -0,0 +1,106 @@
...
+  { \TYPE { \tl_to_str:n {|#1|#2|} } }
+  { }
+
+\ExplSyntaxOff
+\begin{env3}%
+
+  Content
+  &\^
+\end{env3}
+\begin{env3}[...]
```
Skillmon, line 80: Test `\begin{env3} [...]` as well (with a space).

base/testfiles-ltcmd/ltcmd009.lvt
```diff
@@ -0,0 +1,106 @@
...
+
+  Content
+  &\^
+\end{env3}
+\begin{env4}%
+
+  Content
+  &\^
+\end{env4}
+\begin{env4}[...]
```
Skillmon, line 90: Test `\begin{env4} [...]` as well (with a space).

base/testfiles-ltcmd/ltcmd009.lvt
```diff
@@ -0,0 +1,106 @@
...
+
+  Content
+  &\^
+\end{env4}
+\begin{env5}%
+
+  Content
+  &\^
+\end{env5}
+\begin{env5}$...$
```
Skillmon, line 100: Test all the category codes needing special treatment, not only math-shift.

base/ltcmd.dtx
```diff
@@ -3073,12 +3263,48 @@
...
+      {
+        char_set_catcode_
+        \if_catcode:w \c_math_toggle_token #1 math_toggle \else:
+        \if_catcode:w ^ #1 math_subscript \else:
+        \if_catcode:w \c_math_subscript_token #1 math_superscript \else:
+        \if_catcode:w A #1 letter \else:
+        other \fi: \fi: \fi: \fi:
+        :N
+      }
+        #1
```
Skillmon, line 3307: I *think* we need this treatment for both the start and end token of the delimited optional argument, don't we?

Skillmon:
Also, we currently have no special handling of tokens in the same line as `\begin{env}` here. Many (most?) of the established packages/solutions for verbatim environments do throw an error or a warning for these (and `filecontents` throws a `Missing \begin{document}`).

base/ltcmd.dtx
josephwright: No. For the opening token we are using `\futurelet`, so tokenization is an issue. Once we have the opening token, we _know_ there should be a closing one - that is therefore handled using a delimited argument. The closing token therefore never looks ahead, and we are safe leaving catcodes alone.

josephwright:
> For the tokens in front of `\end{env}`: `enverb` uses their count for its auto gobbling feature which makes it automatically work in `.dtx` files or in natural indentation. I'm not suggesting to automatically do this for `f`-type, but making something similar possible would be nice (maybe by putting the contents into a documented `str`-variable, or via an argument processor).

Based on the general `ltcmd` model, I'd expect the (start of the) last line to simply form part of `#1` if passed through. But as that would not end in a `\obeyedline`, it should be easy enough to pick up if required. If more structure is required, I'd expect it to be _within_ `#1` using some marker token: `\finalline` perhap?

> Regarding the tokens after it: There are two possibilities, imho, either throw an error or throw a warning (I think `listings` does that).

I think it's reasonable to exclude stuff after `\end{foo}` from `#1`, the question is whether simply to rescan it or to throw it away, and then as you say if that's an error or simply a warning/info situation. If it's say info, a consumer could redirect to an error if they want.

josephwright:
> Also, we currently have no special handling of tokens in the same line as `\begin{env}` here. Many (most?) of the established packages/solutions for verbatim environments do throw an error or a warning for these (and `filecontents` throws a `Missing \begin{document}`).

Like the end-of-env case, we get a choice here. We could special-case line one, in which case I would be minded to say that optional arguments don't need to worry about spaces (`\begin{Verbatim} [...]` for example works fine). On the other hand, we could simply collect everything and leave to the consumer to make a decision: here we are talking tokens before the first `\obeyedline` so easy to pick up.

I can see an argument for special-casing both the first and last line, but want to be sure we have consensus - all of this is doable, as I say.

base/ltcmd.dtx
```diff
@@ -3073,12 +3263,48 @@
...
+%   are realistic are included.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_grab_D_verb_safe:N #1
   {
-    \@@_grab_D_aux:NNnNN #1#2 {#3} #4 #6
-    #5 #1
-      { \@@_grab_D_call:Nw #1 }
-      { \@@_add_arg:o \c_novalue_tl }
+    \cs_set_eq:NN \do \char_set_catcode_other:N
+    \dospecials
```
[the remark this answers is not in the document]
josephwright: Yes, I added that in a later commit.

base/doc/usrguide.tex
```diff
@@ -858,6 +858,41 @@ \subsection{Typesetting verbatim-like material}
...
+newlines are always permitted: there is no need for a |+| here. As for the
+\texttt{v} specification, newlines are stored as \cs{obeyedline}. In a similar
+fashion to the \texttt{b}~specification, by default \emph{newlines} are trimmed
+at both ends of the body. Putting the prefix |!| before \texttt{f} suppresses
+space-trimming.
+
+Notice that for technical reasons, we recommend that an optional argument
+coming immediately before an \texttt{f} specification should not allow any
+spaces, achieved by adding the \texttt{!} as showing in the example. However,
+this is left as a choice for the user.
```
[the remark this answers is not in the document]
josephwright: Yes, but at present `*` is simply a hack - I want to get agreement on the plan then will document whichever way it comes out :)

base/ltcmd.dtx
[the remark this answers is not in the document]
lnzspraj: The current solution would lose consecutive spaces in front of the first non-space token if no optional argument is found. Is that ok? If so needs documentation.

davidcarlisle:
> > For the tokens in front of `\end{env}`: `enverb` uses their count for its auto gobbling feature which makes it automatically work in `.dtx` files or in natural indentation. I'm not suggesting to automatically do this for `f`-type, but making something similar possible would be nice (maybe by putting the contents into a documented `str`-variable, or via an argument processor).
> 
> Based on the general `ltcmd` model, I'd expect the (start of the) last line to simply form part of `#1` if passed through. But as that would not end in a `\obeyedline`, it should be easy enough to pick up if required. If more structure is required, I'd expect it to be _within_ `#1` using some marker token: `\finalline` perhap?

If I understand @Skillmon's comment correctly the command handling the grabbed f argument would need to know the last line before processing of the returned argument starts (so the prefix could be stripped from all lines) so returning an inline markup would be usable but not especially convenient

base/ltcmd.dtx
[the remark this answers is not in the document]
josephwright: Yes, I'd thought of that - but I think we should first address the question of what we want to do with the first line anyway, as if we decide nothing is allowed there bar any optional arg (and spaces?), then this isn't an issue.

base/ltcmd.dtx
```diff
@@ -3002,68 +3067,193 @@
...
+      \cs_set_eq:NN \do \char_set_catcode_other:N
+      \dospecials
+      \tl_set:Ne \l_@@_tmpa_tl
+        {
+          \c_backslash_str end
+            \c_left_brace_str \@currenvir \c_right_brace_str
+        }
+      \use:e
+        {
+          \cs_set_protected:Npn \@@_grab_f_check:w
```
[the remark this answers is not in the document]
josephwright: True but I don't think we need that gain here - my feeling unless there is strong evidence is we go for more 'standard' idiom.

base/ltcmd.dtx
```diff
@@ -3002,68 +3067,193 @@
...
+          \l_@@_v_arg_tl
+      }
+    \exp_args:NV \end \@currenvir
   }
-\cs_new_protected:Npn \@@_grab_D_no_strip:w
-  #1#2#3 \@@_run_code:
+%    \end{macrocode}
+%   Look for line markers at each end and tidy up if required.
+%    \begin{macrocode}
+\cs_new:Npn \@@_grab_f_end:n #1
```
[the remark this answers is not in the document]
josephwright: I'm not sure what you are getting at?

base/ltcmd.dtx
```diff
@@ -3073,12 +3263,48 @@
...
+  \char_set_catcode_other:N \^^M
+  \cs_new_protected:Npn \@@_grab_D_aux:NNnNNNN #1#2#3#4#5#6#7
+    {
+      \@@_grab_D_aux:NNnNN #1#2 {#3} #4 #6
+      #7
+        {
+          \group_begin:
+            \@@_grab_D_verb_safe:N #1
+        }
+      #5 #1
```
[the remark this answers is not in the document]
josephwright: Again, I think we need a decision on first line handling before adjusting here, so will leave open for the record.

base/ltcmd.dtx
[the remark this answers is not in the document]
Skillmon: Your current `\obeyedline`-stripping code tests for a leading `\obeyedline` then removes it, tests for a trailing `\obeyedline` then removes it. This here just removes the leading `\obeyedline` (and if there is none it removes the one after the second `\q_nil`, then removes the trailing one (and if there is none it removes the one in front of the third `\q_nil`, and then removes the markers and the remainder of the fallback. Reads much clearer and is faster than your current code to strip them. This suggestion should replace all of your `\@@_grab_f_end:n` code not only that single line (I could not mark the suggestion to cover the whole range because there were deleted lines interspersed in the diff and GH doesn't like that).

Skillmon:
Just an idea: Special handling of the first and last line could also be implemented as an argument processor and then split the `f`-argument in three arguments (the contents on the first line, the body minus the two special lines, the contents on the last line). If the processor isn't used you just get the entire contents in a single argument.

Not sure how the policy is on argument processors only really usable on a single argument type.

base/ltcmd.dtx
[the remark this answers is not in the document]
josephwright: Changed

josephwright:
> Just an idea: Special handling of the first and last line could also be implemented as an argument processor and then split the `f`-argument in three arguments (the contents on the first line, the body minus the two special lines, the contents on the last line). If the processor isn't used you just get the entire contents in a single argument.

Yes, that is certainly a possible- although if the feeling is there should be a 'kernel position' on what syntax makes sense, then this would not quite be the way.

> Not sure how the policy is on argument processors only really usable on a single argument type.

I think provided we can find a suitable non-clashing name, that's not a major concern.

josephwright:
My feeling on the various questions I've left open

- I'd be happy to say that the first line cannot have 'content', only optional argument(s) - would make some of the parsing easier
- I would go with allowing tokens before but not after the `\end{foo}` marker: any tokens after the end should I think be a kernel-level error (i.e. go for consistency)
- As a result, I'd not support `!` - with the two positions above, it's not needed
- I would automate adding my current `*` hack - but only for the immediately preceding `D`-type: I can see how to do this automatically (track the last-added argument type and allow unwinding) - I would not allow it to be set manually as something like `*d^^ *d__` would be a nightmare to support

but I am happy to listen to alternative views.

base/ltcmd.dtx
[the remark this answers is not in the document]
FrankMittelbach: What exactly is the definition of "the first line"? The line on which `\begin`appears? In a document it is easy to say no tokens on that line are supported, but if the environment is placed into another definition (as we have seen in one example, there is no "line ending" within the definition.

base/ltcmd.dtx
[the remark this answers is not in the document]
FrankMittelbach: Might be enough to document the technical issue, but perhaps with a bit more information than just saying "for technical reasons". More on the line  `o f` and the like is not a good idea, because it means ... in such and such ... situations, so better always use ...

FrankMittelbach:
> My feeling on the various questions I've left open

I don't have any strong views on that (and haven't thought about it much to be honest)

>     * I'd be happy to say that the first line cannot have 'content', only optional argument(s) - would make some of the parsing easier

might be ok (to keep it simple), but it might need explaining that if "content" is delivered as part of a macro definition it is always on the "first line".  I think the starting point was exactly such a scenario wasn't it? (with the environment being inside some other definition and som intro text inside) -- if so what would be the documented advice?

>     * I would go with allowing tokens before but not after the `\end{foo}` marker: any tokens after the end should I think be a kernel-level error (i.e. go for consistency)

You have to allow for tokens and treat it as the last line of the content, but more importantly I think you need to allow for tokens after the `\end{foo}`and handle them.

Again, my argument is usage in an outer environment definitions where you may reasonably end up with `\end{foo}<more code>` and you then have that `<more code>` automatically on the same line as far as TeX is concerned 

>     * As a result, I'd not support `!` - with the two positions above, it's not needed

guess that still applies

>     * I would automate adding my current `*` hack - but only for the immediately preceding `D`-type: I can see how to do this automatically (track the last-added argument type and allow unwinding) - I would not allow it to be set manually as something like `*d^^ *d__` would be a nightmare to support

Write up the user documentation and see if this is nicely explainable (and then decide if this syntax addition is really a good idea).

josephwright:
> > ```
> > * I'd be happy to say that the first line cannot have 'content', only optional argument(s) - would make some of the parsing easier
> > ```
> 
> might be ok (to keep it simple), but it might need explaining that if "content" is delivered as part of a macro definition it is always on the "first line". I think the starting point was exactly such a scenario wasn't it? (with the environment being inside some other definition and som intro text inside) -- if so what would be the documented advice?

As the idea here is '`filecontents`-like', the putative `f`-type envs can't go into a macro args - we _need_ line ends. That should be documented, I guess.

> > ```
> > * I would go with allowing tokens before but not after the `\end{foo}` marker: any tokens after the end should I think be a kernel-level error (i.e. go for consistency)
> > ```
> 
> You have to allow for tokens and treat it as the last line of the content, but more importantly I think you need to allow for tokens after the `\end{foo}`and handle them.
> 
> Again, my argument is usage in an outer environment definitions where you may reasonably end up with `\end{foo}<more code>` and you then have that `<more code>` automatically on the same line as far as TeX is concerned

Same point as above: you can't hide verb-like stuff in macro args.

FrankMittelbach:
> Same point as above: you can't hide verb-like stuff in macro args.

I'm not  talking about verb in argument, I'm talking about a case like this:
```
\documentclass{article}

\NewDocumentEnvironment{foo}{+b}{{\ttfamily #1}}{}

\NewDocumentEnvironment{baz}{}{\begin{foo} AAA}{xxx\end{foo} more code}

\begin{document}

\begin{foo}
 abc
\end{foo}

\begin{baz}
  xyz
\end{baz}
\end{document}
```

josephwright:
> > Same point as above: you can't hide verb-like stuff in macro args.
> 
> I'm not talking about verb in argument, I'm talking about a case like this:
> 
> ```
> \documentclass{article}
> 
> \NewDocumentEnvironment{foo}{+b}{{\ttfamily #1}}{}
> 
> \NewDocumentEnvironment{baz}{}{\begin{foo} AAA}{xxx\end{foo} more code}
> 
> \begin{document}
> 
> \begin{foo}
>  abc
> \end{foo}
> 
> \begin{baz}
>   xyz
> \end{baz}
> \end{document}
> ```

Try that with `filecontents`( or `verbatim`) ;)

FrankMittelbach:
> Try that with `filecontents`( or `verbatim`) ;)

there is no expectation that filecontents can be used or should be used for this (that is a simple packaging tool) and verbatim works (with the verbatim package) if you use the syntax without `\begin`, eg from the doc:
```
\newenvironment{myverbatim}%
{\endgraf\noindent MYVERBATIM:%
 \endgraf\verbatim}%
 {\endverbatim}
```
This is a very comon use case and it handles tokens after \endverbatim.

josephwright:
> > Try that with `filecontents`( or `verbatim`) ;)
> 
> there is no expectation that filecontents can be used or should be used for this (that is a simple packaging tool) and verbatim works (with the verbatim package) if you use the syntax without `\begin`, eg from the doc:
> 
> ```
> \newenvironment{myverbatim}%
> {\endgraf\noindent MYVERBATIM:%
>  \endgraf\verbatim}%
>  {\endverbatim}
> ```
> 
> This is a very comon use case and it handles tokens after \endverbatim inside the definition.

I guess we need to sort this first then :)

I modelled the current code on `filecontents` or `verbatim` from the kernel, which both grab entire lines and require `^^M` to be present. That is I think the same as `fancyverb` or `listings`, both of which don't allow hiding the env. in another: you are expected to use a 'generator' to make new verb-like envs.

I've not looked at how the `verbatim` package works - is this done by using a delimited arg.? I'm not super-keen on suggesing `\foo` ... `\endfoo`.

Skillmon:
> > Try that with `filecontents`( or `verbatim`) ;)
> 
> there is no expectation that filecontents can be used or should be used for this (that is a simple packaging tool) and verbatim works (with the verbatim package) if you use the syntax without `\begin`, eg from the doc:
> 
> ```
> \newenvironment{myverbatim}%
> {\endgraf\noindent MYVERBATIM:%
>  \endgraf\verbatim}%
>  {\endverbatim}
> ```
> 
> This is a very comon use case and it handles tokens after \endverbatim inside the definition.

That's not the same situation at all. The `\endverbatim` there doesn't end the collection, the `\end{myverbatim}` will, and putting code on the same line as it will still not work:

```tex
\documentclass{article}

\usepackage{verbatim}

\newenvironment{myverbatim}%
{\endgraf\noindent MYVERBATIM:%
 \endgraf\verbatim}%
 {\endverbatim stuff after it}

\begin{document}
\begin{myverbatim}
This is my verbatim % stuff
\end{myverbatim}watch out
\end{document}
```

Will report

```none
LaTeX Warning: Characters dropped after `\end{myverbatim}' on input line 13.
```

and "watch out" will have been dropped.

Skillmon:
>You have to allow for tokens and treat it as the last line of the content, but more importantly I think you need to allow for tokens after the `\end{foo}` and handle them.
>
>Again, my argument is usage in an outer environment definitions where you may reasonably end up with \end{foo}<more code> and you then have that <more code> automatically on the same line as far as TeX is concerned 

The grabbing of a nested `f`-type (if supported) would have to be done using a non-environment form (like for `verbatim`) because otherwise it's not distinguishable from `\begin{minipage}{\linewidth}\begin{myverbatim}...` and wouldn't know for which `\end` to look. That said it doesn't need to support things following the matching `\end`, the nested usage is something completely different from it (as the things will already be tokenised and not grabbed during parsing in the verbatim category code regime).

josephwright:
I've squashed and rebased the work so far: I plan to add some new commits to adjust behaviour so they can be reviewed with a clear starting point :)

josephwright:
Other than the basic question 'is this the right letter', I think we are good-to-go here now.

base/doc/usrguide.tex
```diff
@@ -860,6 +860,57 @@ \subsection{Typesetting verbatim-like material}
...
+Special handling is applied to a \texttt{o}, \texttt{O}, \texttt{d} or
+\texttt{D}~specification argument immediately before an
+\texttt{f}~specification. This means that when the optional argument is absent,
+the first character of the next line will be read properly. Issues may arise if
+\emph{multiple} optional arguments are used before an \texttt{f}~specification,
+and are therefore strongly discouraged.
+
+Notice that for technical reasons, we recommend that an optional argument
+coming immediately before an \texttt{f} specification should not allow any
+spaces, achieved by adding the \texttt{!} as showing in the example. However,
```
FrankMittelbach, line 911: not clear which spaces are meant (suggest to be explicit and say ignore spaces while scannign for the optional argument, or something like that).

FrankMittelbach, line 911: not clear which spaces are meant (suggest to be explicit and say ignore spaces while scannign for the optional argument, or something like that).

base/changes.txt
```diff
@@ -6,6 +6,10 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2025-01-22  Joseph Wright  <Joseph.Wright@latex-project.org>
+	* ltcmd.dtx, usrguide.tex
+	New "f"-type argument
```
FrankMittelbach, line 11: I still don't like "f", it doesn't resonate with me. Have we ruled out "c" for collect, if so why? Could then perhaps one day extended to `C{<setup>}` with some special setup for the collection.

base/ltcmd.dtx
```diff
@@ -97,6 +97,10 @@
 %<latexrelease>                 {Document~command~parser}%
 %    \end{macrocode}
 %
+%    \begin{macrocode}
+\cs_generate_variant:Nn \tl_head:n { e }
+%    \end{macrocode}
```
FrankMittelbach, line 102: shouldn't this go into l3kernel instead?

base/ltcmd.dtx
```diff
@@ -3300,6 +3405,266 @@
...
+% \begin{macro}{\@@_grab_f_auxiv:}
+% \begin{macro}{\@@_grab_f_auxv:}
+% \begin{macro}{\@@_grab_f_auxvi:N}
+% \begin{macro}{\@@_grab_f_auxvii:}
+% \begin{macro}{\@@_grab_f_auxviii:}
+% \begin{macro}{\@@_grab_f_end:w}
+% \begin{macro}{\@@_grab_f_end:n}
+% \begin{macro}{\@@_grab_f_end_auxi:w}
+% \begin{macro}{\@@_grab_f_end_auxii:w}
+% \begin{macro}{\@@_grab_f_end_auxiii:w}
```
FrankMittelbach, line 3426: might actually be helpful in the documented version of the envs are put closer to the actual definitions (and small groups perhaps in some cases but not like this big lumb

FrankMittelbach (commented):
only a very shallow review. But I'm still not happy about using "f" as I wrote. Otherwise only minor comments

base/changes.txt
josephwright: It is a 'working title' - I am not wedded to it. I'm quite happy if we prefer `c` or one of the other suggestions. Probably needs a 'vote'!

base/ltcmd.dtx
josephwright: Yes, and it has today - but I need to allow for older-`expl3`-newer-kernel.

base/changes.txt
FrankMittelbach: artistic license @Skillmon :-) but yes you are probably right, but that doesn't rule out "c" rather than "f"

base/doc/usrguide.tex
FrankMittelbach: now I dea why this showed up twice ... must have been a hickup with my connection

base/changes.txt
josephwright: @Skillmon  In the end it's a guideline of course - we could decide that the utility trumps the consistency argument. But I wonder if instead we might want some more generic catcode setup - would apply to `v` as well, e.g. `*{<catcodes>}`?

lvjr: In my opinion, letter `c` ('code') is better than `f` ('filecontents').

base/ltcmd.dtx
josephwright: I guess I favour the approach of 'keep all the `aux` functions together'. We can of course change - suggestions on a split?

base/ltcmd.dtx
```diff
@@ -3002,68 +3076,320 @@
...
+                        { \exp_not:N \use:n }
+                        { \exp_not:N \use_none:n }
+                   }
+               }
+          }
+      }
+  }
+%    \end{macrocode}
+% \begin{macro}{\@@_grab_D_aux:NNnNNNN}
+% \changes{v1.3a}{2025-01-222}{Extend to support \texttt{f}-type grabbing}
```
lvjr, line 3390: Typos in the date and `\texttt{f}`.

base/ltcmd.dtx
```diff
@@ -1842,6 +1904,7 @@
 %   ignored when looking for that optional argument.
 % \changes{v1.0g}{2021/08/07}
 %         {Replicate argument processors for all embellishments (gh/639)}
+% \changes{v1.3a}{2025-01-22}{Extend to support \texttt{f}-type grabbing}
```
lvjr, line 1907: An typo in `\texttt{f}`.

base/ltcmd.dtx
FrankMittelbach: I don't mind much either way, but such long lists might actually create issues at page breaks and, you may end up with the margin note on one page but the actualy text way later, but up to you.

Skillmon:
I'll continue collecting typos here (because here I conveniently see what was added in this PR).

Skillmon (commented):
Only some more minor typos.

car222222 (commented):
Note that there are a few somewhat, or very, old comments (not by me) here and it is unclear whether they are now resolved. 

I have added mostly just a a few agreement emojis, and the "usual grammar corrections/suggestions". 

I note that this has, as usual, already been approved so I cannot make any strong requests in this review.  Ho-hum!

car222222:
Neither @davidcarlisle nor @muzimuzhi seem to have reviewed this yet.
Not sure how important this is.

josephwright:
> Neither @davidcarlisle nor @muzimuzhi seem to have reviewed this yet. Not sure how important this is.

I pinged a group of people to get some feedback - we'd taken a while and I felt most ideas were resolved, but needed to be sure about the implementation. So I was not expected all reviews to be 'in' before merged - we have time before a release for adjustment.

base/doc/usrguide.tex
```diff
@@ -860,6 +860,57 @@ \subsection{Typesetting verbatim-like material}
...
+above |\end{MyVerbatim}|) cannot have any text \emph{after} the end of the
+environment. Text \emph{before} the end of environment is treated normally,
+but note that there is no trailing \cs{obeyedline} added if there is text here.
+Other than optional arguments, no text is allowed on the opening line of the
+environment.
+
+Special handling is applied to a \texttt{o}, \texttt{O}, \texttt{d} or
+\texttt{D}~specification argument immediately before an
+\texttt{c}~specification. This means that when the optional argument is absent,
+the first character of the next line will be read properly. Issues may arise if
```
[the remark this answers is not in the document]
Skillmon: It should mean "with the correctly applied verbatim category code".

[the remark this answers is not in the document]
car222222: Thanks! 

@josephwright  So, please add this.

PR #1533, opened 2024-11-04
Add tab char to \dospecials

josephwright:
See latex3/latex3#762

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added

No rollback here: I'm hoping that's reasonable - but if not, let me know.


FrankMittelbach (approved):

base/doc/ltnews41.tex
```diff
@@ -149,6 +149,14 @@ \section{New or improved commands}
 
 \section{Code improvements}
 
+\subsection{Tab character as a special}
+
+In \LaTeX{} News~38, we described the extension of \cs{verb}, etc., to cover
+the tab character as equivalent to a space. We have now added tabs to the
+standard list of characters covered by \cs{dospecials}. This allows them to
+used in for example a \texttt{v}~specification document command without
```
muzimuzhi, line 157: to be used...

```suggestion
be used in for example a \texttt{v}~specification document command without
```

muzimuzhi (commented):
minor wording improvement

josephwright:
@davidcarlisle All OK here?

josephwright:
@davidcarlisle Ping ...

davidcarlisle (approved):

user202729:
Note that this breaks Asymptote verbatim environments indented by tab character.

Previous to this change, [`ProcessAsymptote` command](https://github.com/vectorgraphics/asymptote/blob/master/doc/asy-latex.dtx#L405) works when the `\end{asy}` is indented by tabs, because they have catcode 10 therefore silently ignored. Now they have catcode 12, so they're included into `\@tempa`.

A fix would be to add ``\catcode`\^^I=10 `` after `\dospecials`, but many more environments likely will encounter the same issue e.g. `comment` environment in `comment.sty` might.

josephwright:
@user202729 True bu relying on tabs being accidentality left as spaces is not really great: either verb-like envs allow _nothing_ before `\end{foo}` at the start of the line or they look for the first line containing `\end{foo}` and retain the tokens before that.

PR #1537, opened 2024-11-04
making the rotating package tagging compatible

u-fischer:
Resolves issues latex3/tagging-project#751 and latex3/tagging-project#112

## Status of pull request

- Ready to merge

- [x Test file(s) added
- [ ] Rollback provided (if necessary)?  needed??


required/graphics/rotating.dtx
```diff
@@ -350,9 +360,13 @@
 %    \end{macrocode}
 % Set the float contents in
 % a box of width \cs{textheight} instead of \cs{columnwidth}.
+% \changes{v2.16e}{2024/11/01}{correct tagging of the minipage}
 %    \begin{macrocode}
   \begin{lrbox}\rot@float@box
+  \SuspendTagging{\minipage}%
   \begin{minipage}\textheight
+  \ResumeTagging{\minipage}%
```
FrankMittelbach, line 368: I think I can guess what happens here, but from the outside the suspensions look odd. I wonder if that case happens elsewhere too and should get some sort of an interface, or at least perhaps a sentence or two explaining what goes on.

FrankMittelbach: I think the question to ask is under which circumstances do you not want minipage tagging? Because of it being inside an lrbox/mbox? What should happen if you have an lrbox with 2 minipages inside?

FrankMittelbach: point is use of lrbox is core LaTeX so the combination with minipage should work. without the need to disable and reanable tagging by the user

u-fischer: I do not want the minipage tag (the Div) because I'm in a float and this minipage has no semantic meaning. I do not want the \tagmcend\tagstructend before the minipage (which normally closes a P-tag) and the reopening after the minipage as the minipage is in a lrbox and fills it up. Two minipages (without other material) should probably use a Div around them.

u-fischer: > point is use of lrbox is core LaTeX so the combination with minipage should work. without the need to disable and reanable tagging by the user

Well the problem is that `xxx minipage yyy` in a lrbox needs different tagging to `minipage` alone. (At least with pdflatex which has to insert correctly nested MC-literals.)

davidcarlisle: this looks OK to me modulo the discussion re minipage tagging when I'd defer to whatever @FrankMittelbach and @u-fischer  decide.

FrankMittelbach: Well, eventually a minipage inside an lrbox has to work as this is official std LaTeX usage and should not require any action by a user if that is used in a document. However, that isn't soemthing we need to solve now. But what we should do is making an issue for it in the tagging-project repo, so that it gets address at some point.

So with that I think this PR can be merged

FrankMittelbach (approved):
approved but we should open an issue for minipage inside lrboxes

PR #1541, opened 2024-11-05
revert to 066ad1dd3 and re-apply 3c68c5656

davidcarlisle:
Hotfix PR against main to revert (just) amsmath.dtx tthe June 2024 release but applying the contributed documentation fixes from 
066ad1dd3 and 3c68c5656


josephwright:
I see you've gone backward in version: shouldn't it still go up but with a `\changes` saying 'revert XXX'?

davidcarlisle:
@josephwright  I thought it's clearer to literally revert the git so it's just not making the changes rather than making them then changing them back,  Although now you mention it, if we just had git log I'd say that was cleaner but it means the changes.txt is out of sync,  so for that I suppose it may be better to bump the date and add a changes.txt entr to log this. I could do that in an additional commit now?

required/amsmath/amsmath.dtx
```diff
@@ -81,11 +81,11 @@ Bug reports can be opened (category \texttt{#1}) at\\%
 %    \begin{macrocode}
 \DeclareRelease{}{1994-06-01}{amsmath-2018-12-01.sty}
 \DeclareRelease{}{2018-12-01}{amsmath-2018-12-01.sty}
-\DeclareCurrentRelease{}{2024-10-30}
+\DeclareCurrentRelease{}{2019-04-01}
 %    \end{macrocode}
 %
 %    \begin{macrocode}
-\ProvidesPackage{amsmath}[2024/10/30 v2.17s AMS math features]
+\ProvidesPackage{amsmath}[2024/10/05 v2.17q AMS math features]
```
josephwright, line 88: I still think we need to go _up_ here, so
```suggestion
\ProvidesPackage{amsmath}[2024/11/05 v2.17t AMS math features]
```

FrankMittelbach: yes, we have to ... the version is already on ctan and it is a code change (even if it is a reversal)

PR #1543, opened 2024-11-05
revert to 066ad1dd3 and re-apply 3c68c5656

davidcarlisle:
Reverting (just) amsmath.dtx on develop to match the change made to main branch


required/amsmath/amsmath.dtx
```diff
@@ -81,11 +81,11 @@ Bug reports can be opened (category \texttt{#1}) at\\%
 %    \begin{macrocode}
 \DeclareRelease{}{1994-06-01}{amsmath-2018-12-01.sty}
 \DeclareRelease{}{2018-12-01}{amsmath-2018-12-01.sty}
-\DeclareCurrentRelease{}{2024-10-30}
+\DeclareCurrentRelease{}{2019-04-01}
 %    \end{macrocode}
 %
 %    \begin{macrocode}
-\ProvidesPackage{amsmath}[2024/10/30 v2.17s AMS math features]
+\ProvidesPackage{amsmath}[2024/10/05 v2.17q AMS math features]
```
josephwright, line 88:
```suggestion
\ProvidesPackage{amsmath}[2024/11/05 v2.17t AMS math features]
```

required/amsmath/amsmath.dtx
```diff
@@ -85,7 +85,7 @@ Bug reports can be opened (category \texttt{#1}) at\\%
 %    \end{macrocode}
 %
 %    \begin{macrocode}
-\ProvidesPackage{amsmath}[2024/10/05 v2.17q AMS math features]
+\ProvidesPackage{amsmath}[2024/11/05 v2.17t AMS math features]
```
mrpiggi, line 88: `\DeclareCurrentRelease` was reverted as well. On purpose?

davidcarlisle: @mrpiggi  the original intention was simply to wind back the git history before the change and issue amsmath 2.17q with the date of that change. However that made changes.txt a bit weird so I added today's date and a changes.txt entry.

mrpiggi: I just wondered about`\DeclareCurrentRelease{}{2019-04-01}` because the parent commit of 066ad1d was made in May 2024 more than 5 years later. But if this is intentional then all is fine

davidcarlisle: "intentional" may be putting it too strongly as, as I say, first intention was simply to roll back git, but then we put today's date on anyway for a more consistent `changes.txt`.  Joseph may want to adjust a bit before releasing but I wouldn't over-think this, it's just an emergency roll back so `\dots` isn't broken just to give a bit of time to allow testing the feature we intended to release of supporting `\protect`-ed macros,

PR #1547, opened 2024-11-11
Re-work \MakeTitlecase implementation

josephwright:
Fixes latex3/latex3#1316

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [x] Test file(s) added

I've gone without rollback here: I think we have been adjusting this code a bit and can live with telling users that.

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,6 +6,10 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2024-11-11  Joseph Wright  <Joseph.Wright@latex-project.org>
+	* ltfinal.dtx (subsection{Case changing}):
+	Use \text_titlecase_first:nn for titlecasing
+
 2024-11-06  Joseph Wright  <Joseph.Wright@latex-project.org>
 	* ltoutenc.dtx
 	Add support for \={i} to OT1 and T1
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,9 +6,12 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
-2024-11-11  Joseph Wright  <Joseph.Wright@latex-project.org>
+2024-11-12  Joseph Wright  <Joseph.Wright@latex-project.org>
 	* ltfinal.dtx (subsection{Case changing}):
-	Use \text_titlecase_first:nn for titlecasing
+	Use \text_titlecase_...:nn for titlecasing
+	Introduce option to titlecase first or all words
+	* usrguide.tex:
+	Document new key for titlecasing
 
 2024-11-06  Joseph Wright  <Joseph.Wright@latex-project.org>
 	* ltoutenc.dtx
--- a/base/doc/ltnews41.tex
+++ b/base/doc/ltnews41.tex
@@ -149,6 +149,26 @@
 
 \section{Code improvements}
 
+\subsection{Refinement of \cs{MakeTitlecase}}
+
+We introduced \cs{MakeTitlecase} as a late addition to the June
+2022 release, making use of the improved case code in
+\pkg{expl3}. Unlike upper and lowercasing, making text
+titlecased is more tricky to get right: for example, do you just
+change the first character and leave the rest alone.
+
+A subtle issue was reported against the \pkg{expl3} repository
+(\url{https://github.com/latex3/latex3/issues/1316}) which links
+to how we deal with the question of case changing
+\enquote{words} but shows up if you titlecase text stored in a
+command.
+
+We have looked again at how to implement \cs{MakeTitlecase} to
+be as predictable as possible, and have made a change in this
+release. The command no longer tries to lowercase text before
+applying titlecasing, and gives the correct result for text
+stored in commands.
+
 \section{Bug fixes}
 
 %\subsection{A fix}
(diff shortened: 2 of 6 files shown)


base/doc/ltnews41.tex
```diff
@@ -149,6 +149,26 @@ \section{New or improved commands}
 
 \section{Code improvements}
 
+\subsection{Refinement of \cs{MakeTitlecase}}
+
+We introduced \cs{MakeTitlecase} as a late addition to the June
+2022 release, making use of the improved case code in
+\pkg{expl3}. Unlike upper and lowercasing, making text
+titlecased is more tricky to get right: for example, do you just
+change the first character and leave the rest alone.
```
FrankMittelbach, line 158: or what? the way written doesn't seem like an example

base/changes.txt
```diff
@@ -6,6 +6,10 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2024-11-11  Joseph Wright  <Joseph.Wright@latex-project.org>
+	* ltfinal.dtx (subsection{Case changing}):
+	Use \text_titlecase_first:nn for titlecasing
```
FrankMittelbach, line 11: I tought, that there was some plan to support both "first" and "all" via the optional argument?

josephwright: Plan was to make a change, see if it looks OK, then worry about an option :) I'm not sure what key name to use, honestly.

FrankMittelbach: words = first | all  was proposed wasn't it? I think that would be fine

base/doc/ltnews41.tex
```diff
@@ -154,8 +154,8 @@ \subsection{Refinement of \cs{MakeTitlecase}}
 We introduced \cs{MakeTitlecase} as a late addition to the June
 2022 release, making use of the improved case code in
 \pkg{expl3}. Unlike upper and lowercasing, making text
-titlecased is more tricky to get right: for example, do you just
-change the first character and leave the rest alone.
+titlecased is more tricky to get right: this can apply either to
+the whole text or one a word-by-word basis.
```
mrpiggi, line 158: "one" > "on"

base/doc/ltnews41.tex
josephwright: I've reworded

base/changes.txt
josephwright: Added the new keyword and docs

base/doc/ltnews41.tex
josephwright: Fixed thanks

FrankMittelbach (approved):
I'm ok with it. I do wonder, however, if the "all" key would perhaps need an exception list in which one can specify words that should not be titlecased --- this is really a foreign concept for me as it makes zero sense in German, but my understanding is that words like "of"  are often left alone even if all words are to be titlecased. (would need to be applied to all but the first word I guess so a bit tricky and probably nothing that should be squeezed in this PR).

car222222:
Whilst this selective capitalization is often done (in some typesetting traditions for English at least), coming up with 
a generally used list of such words would be difficult, and 
maybe controversial. 

Thus, I would state that 
"the details of such capitalization are normally the work of an author or editor, rather than of such an automated process", 
and simply provide the "all" option in case it is sometimes useful!

FrankMittelbach:
> Whilst this selective capitalization is often done (in some typesetting traditions for English at least), coming up with a generally used list of such words would be difficult, and maybe controversial.
> 
> Thus, I would state that "the details of such capitalization are normally the work of an author or editor, rather than of such an automated process", and simply provide the "all" option in case it is sometimes useful!

coming up with a list at our end would not work, I agree, but I was just suggesting to offer the possibility to specify such a list. Then a journal or a biblatex style or ... could specify what they want fora specific journal and this way allowing the automation to work.

josephwright:
@FrankMittelbach  As @car222222 says, this is very tricky and there tends to be an editorial aspect - different people regard different words as 'small'. So whilst we could try to extend to cover the typical expectation of titlecasing in English, I think that might be a stretch too far - at least at present.

car222222:
OK, understood.  

Whilst this is certainly doable, it is not clear whether it will be used much by "LaTeX editors" etc.

josephwright:
> coming up with a list at our end would not work, I agree, but I was just suggesting to offer the possibility to specify such a list. Then a journal or a biblatex style or ... could specify what they want fora specific journal and this way allowing the automation to work.

We could look at this as a feature request at the `expl3` layer - that would then be available to the document command - we could have some additional key option to choose between them.

car222222:
I hope it is clear that my previous response was to @FrankMittelbach re his further explanation of what he is suggesting.

FrankMittelbach:
> > 
> 
> We could look at this as a feature request at the `expl3` layer - that would then be available to the document command - we could have some additional key option to choose between them.

that is roughly what I was suggesting. Basically that doesn't need a key but simply a single clist macro to which one can specify the "small" words that should not be titlecased. That could then be used in a class and would by default be empty, e.g., `\SetTitlecaseExceptions{of,the,and,or}` or something similar as an interface.

josephwright:
@FrankMittelbach  I'll add it to the to-do list :) But I think it's out-of-scope for the PR

FrankMittelbach:
> @FrankMittelbach I'll add it to the to-do list :) But I think it's out-of-scope for the PR

that's what I said, didn't I? more something to consider separately

josephwright:
We good to go here?

FrankMittelbach:
from my point yes, (modula resolving conflicts)

PR #1590, opened 2024-12-21
testphase value "latest" 

u-fischer:
This adds the "latest" value to the testphase key. It also updates the PDF management to the newest version 0.96o.

# Internal housekeeping

## Status of pull request
- Ready to merge

- [x] Test file(s) added


required/latex-lab/documentmetadata-support.dtx
```diff
@@ -137,37 +137,39 @@
...
-%      XMP-metadata. In version 0.95z support for |UA-2| has been added but
-%      note that |UA-2| hasn't been released yet. It should be used only together
-%      with pdf version 2.0.
+%
+%      Beside the A-standards it is also possible to use the values
+%      |X-4|, |X-4p|, |X-5g|, |X-5n|, |X-5pg|, |X-6|, |X-6n|, |X-6p| for
+%      a PDF/X and |UA-1|  and |UA-2| for PDF/UA standard.
+%      |UA-2| should only be used together with PDF 2.0.
+%      These keys currently set mostly \emph{only} the relevant
+%      XMP-metadata.
```
FrankMittelbach, line 146: a) not really a sentence I think
b) not sure what this means with respect to UA-1/2

required/latex-lab/documentmetadata-support.dtx
```diff
@@ -137,37 +137,39 @@
...
 %
 %      |pdfstandard| can be used more than once to set overlapping standards, e.g:\\
 %      |pdfstandard=A-2b,pdfstandard=X-4,pdfstandard=UA-1|
-%      
+%
 %      If XMP-metadata are added (see the following key \texttt{xmp})
 %      the needed conformance marker for the standards are set.
```
FrankMittelbach, line 152: needed -> necessary

required/latex-lab/testfiles-OR-luatex/footmisc-003.tlg
```diff
@@ -69,7 +69,7 @@ Completed box being shipped out [1]
 .....\penalty 10000
 .....\pdfliteral page <lua data reference ...>
 .....\pdfliteral page <lua data reference ...>
-.....\pdfstartlink(*+*)x* attr{/Contents (ref) /Border [0 0 1] /H /I /C [0.701176 0.4 0.414118] /StructParent 2 } action struct name{footnote*.5} goto name{footnote*.5}
+.....\pdfstartlink(*+*)x* attr{/Contents <FEFF0047006F00200074006F002000640065007300740069006E006100740069006F006E00200066006F006F0074006E006F00740065002A002E0035> /Border [0 0 1] /H /I /C [0.701176 0.4 0.414118] /StructParent 2 } action struct name{footnote*.5} goto name{footnote*.5}
```
FrankMittelbach, line 72: why does this PR changes all kind of tlgs? is that pdfmanagement change that shows through hre?

FrankMittelbach (approved):
approved, assuming all the test file changes are unrelated and due to pdfmanagement or some other external change

required/latex-lab/documentmetadata-support.dtx
u-fischer: a) not sure how to say it differently b) it means e.g. that setting UA-1/2 doesn't automatically activate tagging or load testphase modules. All the standard keys declare the intended standard in the XMP, sometimes they trigger warnings, sometimes they actually do something (like adding a colorprofile), but only few requirements are really enforced and you have to use a validator to check.

required/latex-lab/testfiles-OR-luatex/footmisc-003.tlg
u-fischer: yes. I extended/corrected/adjusted the handling of internal links and so now the `/Contents` key has a real content and not only a generic `ref`.

required/latex-lab/documentmetadata-support.dtx
FrankMittelbach: a)

I think many people would expect that setting, say UA-2 actually tries to prpduce a UA-2 file not just "claims" that the file is one, so I think this needs to be spelled out. But the fact that they are sometime to more, seems to me a bit of a interface issue to be looked at at some point

b) something like this?

These keys declare the intended standard in the XMP-metadata, but do not initiate other actions such as tagging (which is currently activated through the testphase key) to provide the conformance that is claimed by setting the value. In some cases they trigger warnings if the combinations of settings are obviously wrong and in a few cases they doe some extra work in the background (by adding a necessary color profile, for example) but on the whole you need a validator to check that the resulting PDF actually conforms to the claims that are made with the specified settings.

u-fischer: well that is already there, in the paragraph above.

FrankMittelbach: that wasn't visible in the diff

car222222: What does the "only" mean here?
The sentence needs revision!

required/latex-lab/documentmetadata-support.dtx
car222222: Also: markers . . . are 
Or: marker . . . is

required/latex-lab/documentmetadata-support.dtx
car222222: I see this has been fixed by now.  Good!

PR #1594, opened 2024-12-22
use xhtml and mathml namespaces when writing mathml file

davidcarlisle:
If the matching PR in the tagging project https://github.com/latex3/tagging-project/pull/775 is merged, we will want  to use namespaced elements in the MathML files. This PR adds the matching namespaces when writing out dummy files.


davidcarlisle:
some of the test results will need updating  will do that if we decide to use namespaces and merge this.

u-fischer (approved):

PR #1605, opened 2025-01-02
Gh1604

FrankMittelbach:
# Internal housekeeping

## Status of pull request

 Ready to merge

- [x] Test file(s) added

### DIFF
--- a/base/ltclass.dtx
+++ b/base/ltclass.dtx
@@ -2809,7 +2809,7 @@
 %    \end{macrocode}
 %
 %    In its initial implementation (not using the hook system)
-%    \cs{AtBeginDocument} was made \cs{@preambleonly} because using it
+%    \cs{AtBeginDocument} was made \cs{@onlypreamble} because using it
 %    later had no effect whatsoever, thus was most certainly an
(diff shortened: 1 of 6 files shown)


davidcarlisle (approved):

base/ltclass.dtx
```diff
@@ -2808,8 +2808,20 @@
 %<*2ekernel>
 %    \end{macrocode}
 %
-%    \begin{macrocode}
-\@onlypreamble\AtBeginDocument
+%    In its initial implementation (not using the hook system)
+%    \cs{AtBeginDocument} was made \cs{@preambleonly} because using it
```
mrpiggi, line 2812: Probably `\cs{@onlypreamble}`

muzimuzhi:
Need a ltnews entry?

FrankMittelbach:
> Need a ltnews entry?

I thought it isn't really necessary since this is rather an odd cleanup and not something to recommend. But we could add one if people think so.

base/ltclass.dtx
FrankMittelbach: right you are, thanks

PR #1606, opened 2025-01-03
Suppress "load-only" warning for global options

josephwright:


josephwright:
@davidcarlisle Ping

base/testfiles/github-1606.lvt
```diff
@@ -0,0 +1,18 @@
+\begin{filecontents*}[overwrite]{testpackage.sty}
+\ProvidesExplPackage{testpackage}{2025-02-04}{}{}
+\keys_define:nn { testpackage }
+  { testoption .code:n = { \def \testmacro { seen } } }
```
davidcarlisle, line 4: this doesn't test the load-only code, and if I add `testoption .usage=load` then the duplicate load still produces a warning:

```

LaTeX Warning: Package "testpackage" has already been loaded: ignoring
(LaTeX)        load-time option "testoption".
```

davidcarlisle (changes requested):

base/testfiles/github-1606.lvt
josephwright: Thanks - fixed

base/ltkeys.dtx
```diff
@@ -576,20 +576,27 @@
...
+% \changes{v1.0p}{2025-01-12}{Suppress load-only error for global options}
 %   Indicates that the load-time options for a package have been processed:
 %   once this has happened, make them unavailable either with a warning or
-%   an error.
+%   an error. We suppress the warning where the option is in the global list:
+%   there is a bit of gymnastics with the catcodes.
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_options_loaded:n #1
   {
+    \tl_set:Ne \l_@@_tmpb_tl { \tl_to_str:N \@classoptionslist }
```
davidcarlisle, line 587: using classoptionlist here rather than one of the versions that's been filtered to drop the values (eg for unused option warnings) seems a bit suspect.  If I change the 1606 test file to have 

```
\documentclass[testoption=?]{article}
```

Then the warning comes back for duplicate loads

```

LaTeX Warning: Package "testpackage" has already been loaded: ignoring
(LaTeX)        load-time option "testoption".
```

which doesn't seem quite right?

davidcarlisle (changes requested):

u-fischer:
If I test in this PR

~~~~
\begin{filecontents*}[overwrite]{testpackage.sty}
\ProvidesExplPackage{testpackage}{2025-02-04}{}{}
\keys_define:nn { testpackage }
  {
    testoption .code:n = { \def \testmacro { #1 } } ,
    testoption .usage:n = { load }
  }
\ProcessKeyOptions [ testpackage ]
\endinput
\end{filecontents*}

\input{test2e}

\documentclass[testoption]{article}
\usepackage[testoption=false]{testpackage}
\usepackage[testoption=true]{testpackage}
\show\testmacro
\END 
~~~~

Then I get no warning about an ignored load option. But with 
`\documentclass[testoption=false]{article}` or `\documentclass[testoption=true]{article}` 
I get *two* warnings (also in the current implementation).

josephwright:
OK, I'll think of a second generation plan

PR #1607, opened 2025-01-05
Tagging issues 777 and 778

u-fischer:
## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [n/a ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated


u-fischer:
@FrankMittelbach  @davidcarlisle I have now changed the taggging of presentation tables to use "normal" tags and an ARIA attribute. This works better with derivation. OK?

davidcarlisle:
Looks good to me

required/latex-lab/latex-lab-sec.dtx
```diff
@@ -359,11 +359,12 @@
...
+% \changes{v0.84g}{2025/01/05}{ensure that paragraph is ended (tagging/777)}
 %    \begin{macrocode}
-\AddToHook{cmd/frontmatter/before}{\@@_sec_end:n{-10}}
-\AddToHook{cmd/mainmatter/before} {\@@_sec_end:n{-10}}
-\AddToHook{cmd/backmatter/before} {\@@_sec_end:n{-10}}
-\AddToHook{cmd/appendix/before}   {\@@_sec_end:n{-10}}
+\AddToHook{cmd/frontmatter/before}{\par\@@_sec_end:n{-10}}
+\AddToHook{cmd/mainmatter/before} {\par\@@_sec_end:n{-10}}
+\AddToHook{cmd/backmatter/before} {\par\@@_sec_end:n{-10}}
+\AddToHook{cmd/appendix/before}   {\par\@@_sec_end:n{-10}}
```
FrankMittelbach, line 367: Probably better would be to move the \par command into those comands eventually (to ensure that they always behave identically (with or without tagging).

Also (not checked) but isn't `\@@_sec_end:n{-10}` a related command and should be in a socket rather than a hook? 

Doesn't have to be done now, but there should be a TODO if this is something we should cleanup at some point.

required/latex-lab/latex-lab-table.dtx
```diff
@@ -1015,26 +1043,26 @@
...
-       \tl_set:Nn\l_@@_tabletag_tl {Div} 
-       \clist_clear:N \l_@@_header_rows_clist
-       \clist_clear:N \l_@@_header_columns_clist  
-     },  
+      {
+        \__tag_tbl_enable:
+        \AssignSocketPlug{tagsupport/tbl/hmode/begin}{LayoutTable}
+        \AssignSocketPlug{tagsupport/tbl/vmode/begin}{LayoutTable}
+        \clist_clear:N \l_@@_header_rows_clist
+        \clist_clear:N \l_@@_header_columns_clist
```
FrankMittelbach, line 1064: If I go from presentation back to table/tagging=on it should probably reset the plugs or not? If so the plugs have to be set also above, as this this might happen on the same group level.

If my assumption is correct, then this also needs a test case.

FrankMittelbach (changes requested):
see question about plug settings

required/latex-lab/latex-lab-sec.dtx
u-fischer: I'm not sure if a \par inside `\@@_sec_end:n` is always wanted that's why I added if here explicitly (In the other cases the command is currently is used inside the sectioning commands in vmode, so a \par inside the command is not needed).

I'm not sure what you mean by sockets. \mainmatter etc are defined by the classes, so we can not simply redefine them to add a socket but have to use a generic hook. The problem was that the hook is before the \clearpage and so not necessarly in vmode.

required/latex-lab/latex-lab-table.dtx
u-fischer: `\__tag_tbl_enable:`  sets all sockets, including the begin sockets to the defaults. So the presentation mode only has to overwrite the two.

u-fischer: and there is a testfile (table-015-presentation in testfiles-table-luatex) which tests the switch from presentation back to normal tagging in one group.

u-fischer:
@FrankMittelbach, @davidcarlisle I now added both options, (aria-role and div) so I think that is good to merge?

required/latex-lab/latex-lab-table.dtx
```diff
@@ -138,17 +138,23 @@
...
 % The first option disables the table tagging code and the content of the tabular
 % is then treated more or less like running text. This works ok for simple tables using
 % only hmode-cells (l/c/r) with normal text content, but fails if the table uses vmode-cells
-% (p/m/b). In such cases the second option works better: it keeps the tagging code active
-% but changes the tag names to \texttt{Div} and the grouping structure \texttt{NonStruct}. It also
-% (re)sets the \texttt{table/header-rows} key to empty. 
-% The key should currently only be used in a group as there is no key (yet) to reset to the default 
-% tag names.
+% (p/m/b). In such cases the second option works better: it keeps the tagging code active,
+% but adds an ARIA-role as attribute. When deriving to html this gives the best result
```
FrankMittelbach, line 149: ... as attribute to indicate that this is a presentation table.

required/latex-lab/latex-lab-table.dtx
```diff
@@ -228,7 +234,7 @@
 % \item Not every table should be tagged as a Table structure, often they are
 % only used as layout help, e.g. to align authors in a title pages. In such uses
 % the tagging of the table must be deactivated with \verb|\tagpdfsetup{table/tagging=false}|
-% or \verb|\tagpdfsetup{table/tagging=presentation}|.
+% or changed with \verb|\tagpdfsetup{table/tagging=presentation}|.
```
FrankMittelbach, line 237: ... to become a presentation table.

what about the DIV version?

required/latex-lab/latex-lab-table.dtx
```diff
@@ -1015,26 +1049,58 @@
...
+        \clist_clear:N \l_@@_header_columns_clist  
+      },  
+    table/tagging / div .code:n = 
+      {
+        \__tag_tbl_enable:
+        \@@_init_tagnames_presentation:
+        \AssignSocketPlug{tagsupport/tbl/hmode/begin}{LayoutTable}
+        \AssignSocketPlug{tagsupport/tbl/vmode/begin}{LayoutTable}
+        \clist_clear:N \l_@@_header_rows_clist
+        \clist_clear:N \l_@@_header_columns_clist
```
FrankMittelbach, line 1102: am I missing something? This looks 100% like the previous case

FrankMittelbach (changes requested):
some suggested changes and a question

required/latex-lab/latex-lab-table.dtx
u-fischer: the difference is `\@@_init_tagnames_default:` and `\@@_init_tagnames_presentation:`.

FrankMittelbach: hmm, yes, overlooked that, but isn't that a bit odd then? I mean in case of DIV it is called _presentation and in the other case it is called _default? I would expect the names and the case to match

u-fischer: hm right, the name was from an earlier version where presentation mapped to div, I will change the name to `\@@_init_tagnames_div:`.

FrankMittelbach (approved):

PR #1614, opened 2025-01-08
Set template keys docu

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [n/a] Test file(s) added

One thing to consider: change name from `\SetTemplateKeys` to  `\OverwriteTemplateKeys`, which seems to me sits better with `\AssignTemplateKeys`(as it is theny clear what the order would be). It is not used in texmf (other than  in latex-lab) so that could be a breaking change without problems.


FrankMittelbach:
more exactly there is one usage in `keytheorems` by @mbertucci47 and I think we can arrange a smooth transition if we decide on a name change.

josephwright:
`\SetTemplateKeys` was modelled ultimately on `\setkeys` (which is why we have `\keys_set:nn`, etc.) - but if the overall feeling is an alternative name is better, I see no problem changing. If we do, there needs to be an adjustment in `xtemplate` (which is 'frozen' so does need to keep the existing name available).

base/lttemplates.dtx
```diff
@@ -351,12 +351,41 @@
...
+%    also overwrite the current template key value settings by using the command
+%    \cs{SetTemplateKeys}, i.e., after the template default
+%    values and the values from the instance have been assigned.
+
+%    The main use case for this command is the situation where there
+%    is an argument (normally \texttt{\#1}) to to the template in which
+%    a key/value list can be specified that overwrites the normal
+%    settings. In that case one could use something like
+%\begin{quote}
+%  \verb/\tl_if_empty:nF {#1} { \SetTemplateKeys/\Arg{type}\Arg{template}\verb/{#1} }/
```
josephwright, line 377: Why the test here? You could just apply `\SetTemplateKeys` directly.

base/lttemplates.dtx
```diff
@@ -351,12 +351,41 @@
...
+%\end{quote}
+%    to process this key/value list. The outer test here, is only for
+%    speeding up the processing in case the list is empty.
```
josephwright, line 380: There isn't a lot in it - not sure the complexity is worth it!

josephwright (approved):

base/lttemplates.dtx
FrankMittelbach: yes one could, but

base/lttemplates.dtx
FrankMittelbach: your and my understanding of the word "not much" clearly differs :-) Calling `\SetTemplateKeys` on an empty argument generates more than 500 lines of tracing compared to 11 for the test.

davidcarlisle: I suspect the model of passing on a document level optional argument via `\SetTemplateKeys...{#1}` will be relatively common which means that calling this with an empty set might be more common than if you were directly setting keys "by hand". In which case, rather than suggesting an explicit test here we could consider optimising the empty argument case to reduce those 500 lines?

FrankMittelbach: yes, I agree with you both that it would be better to put it inside; I was documenting what we currently have and there this a) is the main/only real use case, b) it will appear in many/most document level commands/environments over time, and c) that argument will most of the time be empty

So yes, let's move it inside and simplify document and code.

FrankMittelbach:
> `\SetTemplateKeys` was modelled ultimately on `\setkeys` (which is why we have `\keys_set:nn`, etc.) - but if the overall feeling is an alternative name is better, I see no problem changing. If we do, there needs to be an adjustment in `xtemplate` (which is 'frozen' so does need to keep the existing name available).

Maybe it should be called `\SetKnownTemplateKeys` then, because that's what it is and then a confusion with `\AssignTemplateKeys`is less likely (imo).

base/lttemplates.dtx
car222222: So should I wait for this change before reviewing?

josephwright:
> Maybe it should be called `\SetKnownTemplateKeys` then, because that's what it is and then a confusion with `\AssignTemplateKeys`is less likely (imo).

I thought that the fact this says 'template keys' suggested it would only handle known ones - but again no objection to a change.

base/lttemplates.dtx
FrankMittelbach: yes

FrankMittelbach:
> > Maybe it should be called `\SetKnownTemplateKeys` then, because that's what it is and then a confusion with `\AssignTemplateKeys`is less likely (imo).
> 
> I thought that the fact this says 'template keys' suggested it would only handle known ones - but again no objection to a change.

why would it? It could be a wrapper for `\keys_set:nn` as you wrote (and the name currently implies) and that would complain about unkown keys, wouldn't it?

u-fischer:
I don't quite like `\OverwriteTemplateKeys`. It doesn't clarify for me that the command should be used inside a template better than `\SetTemplateKeys`. Actually "overwrite" sounds for me more like a global forced change and so is more confusing than the \Set-command. 
`\SetKnownTemplateKeys` is an option, but it mainly tells me that it won't error if unknown keys are passed to the command.

FrankMittelbach:
> I don't quite like `\OverwriteTemplateKeys`. It doesn't clarify for me that the command should be used inside a template better than `\SetTemplateKeys`. Actually "overwrite" sounds for me more like a global forced change and so is more confusing than the \Set-command. `\SetKnownTemplateKeys` is an option, but it mainly tells me that it won't error if unknown keys are passed to the command.

Nothing really tells that it should only be used in the template code  and it would be hard to get that into the name (which is why that is explicitly documented now) but using "Known" helps when reading code where it is in the right place, then you immediately see why it could get keys that do not fit for that particular template without ding harm and it fits with the underlying command which is `\keys_set_known:nnN`. But I'm not insisting on a change, just think that it might be result in a slightly better code reading.

u-fischer:
> but using "Known" helps when reading code where it is in the right place

yes how unknown keys are handled is certainly a question that would come to my mind, and it would fit better the definition as the underlying command is `\keys_set_known:nnN`.   Side question: should the list of unknown, unprocessed keys be made publicitly available somehow?

FrankMittelbach:
> > but using "Known" helps when reading code where it is in the right place
> 
> yes how unknown keys are handled is certainly a question that would come to my mind, and it would fit better the definition as the underlying command is `\keys_set_known:nnN`. Side question: should the list of unknown, unprocessed keys be made publicitly available somehow?

It would be nice, but rather hard, if even possible, to implement. For example, the key/vals to `\begin{enumerate}` are passed on to 3 or 4 templates making up the list, but also to the item template for the individual items in the list. Such scenarios make it imho infeasable to try to keep track of this. But if anybody sees a good way ...?

FrankMittelbach:
I have now updated and extended the documentation as well as moved the empty test inside the command, so I think it is ready for another review. What remains open is the csname. Are we good with `\SetKnownTemplateKeys` or should we keep the old name?

base/lttemplates.dtx
FrankMittelbach: @car222222 would be ready now I think

u-fischer:
> It would be nice, but rather hard, if even possible, to implement. For example, the key/vals to `\begin{enumerate}` are passed on to 3 or 4 templates making up the list, but also to the item template for the individual items in the list. Such scenarios make it imho infeasable to try to keep track of this. But if anybody sees a good way ...

I didn't mean a general tracking. `\SetTemplateKeys` does a `\keys_set_known:nnN { template / #1 / #2 } {#3} \l_@@_tmp_clist`, and I wondered if this could be a public clist instead of an internal one, e.g. for checks in cases one knows that only known keys should be passed or if one want to pass the rest to an internal template. Or perhaps `\SetKnownTemplateKeys`  could use four argument and the last one stores the remaining keys while `\SetTemplateKeys` stays at is?

FrankMittelbach:
> > It would be nice, but rather hard, if even possible, to implement. For example, the key/vals to `\begin{enumerate}` are passed on to 3 or 4 templates making up the list, but also to the item template for the individual items in the list. Such scenarios make it imho infeasable to try to keep track of this. But if anybody sees a good way ...
> 
> I didn't mean a general tracking. `\SetTemplateKeys` does a `\keys_set_known:nnN { template / #1 / #2 } {#3} \l_@@_tmp_clist`, and I wondered if this could be a public clist instead of an internal one, e.g. for checks in cases one knows that only known keys should be passed or if one want to pass the rest to an internal template. Or perhaps `\SetKnownTemplateKeys` could use four argument and the last one stores the remaining keys while `\SetTemplateKeys` stays at is?

ok, back to the drawing board then. what about this?
```
\ExplSyntaxOn

% error on unknown keys
\cs_set_protected:Npn \SetTemplateKeys #1#2#3
  {
    \tl_if_empty:oF {#3}
      {
        \keys_set:no { template / #1 / #2 } {#3}
      }
  }
  
 % supports unknown keys and returns them for further processing 
\cs_new_protected:Npn \SetKnownTemplateKeys #1#2#3#4
  {
    \tl_if_empty:oF {#3}
      {
        \keys_set_known:noN { template / #1 / #2 } {#3} #4
      }
  }

\SetKnownTemplateKeys {block}{display} {a=a, b=b, c=c} \foo
\show\foo
\SetKnownTemplateKeys {list}{display} {\foo} \bar
\show\bar

% process remaining keys (should be empty after that
% Or to be used when you know that you shouldn't have unknown ones
\SetTemplateKeys {para}{std} {\bar}
```

It would be for sure a breaking change, but I guess not a bad one.

u-fischer:
one would have to check the block code and adapt the uses there but beside this this looks fine imho.

FrankMittelbach:
> one would have to check the block code and adapt the uses there but beside this this looks fine imho.

sure, the block code is already adapted as part of the PR (but it needs a corresponding change then)

mbertucci47:
@FrankMittelbach I've already removed the usage in keytheorems and use `\EditInstance` instead

josephwright (approved):

FrankMittelbach:
> @FrankMittelbach I've already removed the usage in keytheorems and use `\EditInstance` instead

ah, so much the better, then there is  no usage of the current command other than in latex-lab, makes life easy. Btw, what to you think of the new docs and the suggested change/extension?

mbertucci47:
I like the documentation changes. I'm not sure I fully understand the discussion about known/unknown keys, but I'm not a fan of the lack of error when an unknown key is passed to, say, `\begin{enumerate}`. It's too easy to misremember `leftmargin` as `left-margin` or `left margin` and be confused why nothing is changing.

josephwright:
> I like the documentation changes. I'm not sure I fully understand the discussion about known/unknown keys, but I'm not a fan of the lack of error when an unknown key is passed to, say, `\begin{enumerate}`. It's too easy to misremember `leftmargin` as `left-margin` or `left margin` and be confused why nothing is changing.

The problem is that as @FrankMittelbach  has indicated, you might need several templates for a complex case like this. So how do you decide where to issue a warning? We really don't want users to have to work out which template is which for setting keys - we looked at prefixing all keys, and it doesn't make a good interface.

Skillmon:
> Nothing really tells that it should only be used in the template code and it would be hard to get that into the name (which is why that is explicitly documented now) but using "Known" helps when reading code where it is in the right place, then you immediately see why it could get keys that do not fit for that particular template without ding harm and it fits with the underlying command which is `\keys_set_known:nnN`. But I'm not insisting on a change, just think that it might be result in a slightly better code reading.

Not that I have much to say about this, but if this should only be used inside template code how about something like `\SetThisTemplatesKeys` or `\SetTemplateInstanceKeys`? (Please ignore if this sounds stupid)

FrankMittelbach:
> > I like the documentation changes. I'm not sure I fully understand the discussion about known/unknown keys, but I'm not a fan of the lack of error when an unknown key is passed to, say, `\begin{enumerate}`. It's too easy to misremember `leftmargin` as `left-margin` or `left margin` and be confused why nothing is changing.
> 
> The problem is that as @FrankMittelbach has indicated, you might need several templates for a complex case like this. So how do you decide where to issue a warning? We really don't want users to have to work out which template is which for setting keys - we looked at prefixing all keys, and it doesn't make a good interface.

I have now a first implementation for checkingg all user-supplied keys to block environments.

a) it can be handled (but not yet cleanly, I'm missing some support interfaces)
b) it showed me that you probably don't want `\SetTemplateKeys` but only `\SetKnownTemplateKeys` because the  error handling of the former is inferior and exposes the internal template structure and one can do much better to do it manually at the end..
c) current implementation is still WIP

car222222:
@FrankMittelbach @josephwright 
It seems that this is still ongoing, so I shall not review it until it is finalised.

FrankMittelbach:
> Not that I have much to say about this, but if this should only be used inside template code how about something like `\SetThisTemplatesKeys` or `\SetTemplateInstanceKeys`? (Please ignore if this sounds stupid)

Using "This" in the name would help with identifying where it should be, but assuming it is in the right place, the important info for somebody reading the code is whether it only deals with known, keys (and ignores the rest) or whether it handles everything in its argument (and complains if it finds some unknown ones). and putting in "Known" as well makes the name really horrible in my eyes.

Thus I think it is better to find a name that it good for somebody reading the code and document where to put it for those writing the code, hence my current names.

FrankMittelbach:
> @FrankMittelbach @josephwright It seems that this is still ongoing, so I shall not review it until it is finalised.

I think it is ready enough for your review (otherwise you might find it got merged)

The only open question for me is whether or not top fully dop `\SetTemplateKeys`which would mean a slight rewrite and dropping a footnote.

josephwright:
I think we could probably drop `\SetTemplateKeys` here as we have a clearer name

base/lttemplates.dtx
```diff
@@ -351,12 +351,89 @@
...
+
+%    For example, a list, such as \env{enumerate}, is made up from a
+%    \texttt{blockenv}, \texttt{block}, \texttt{list}, and a
+%    \texttt{para} template and in the single user-supplied optional
+%    argument of \env{enumerate} key/values for either of
+%    the templates might get specified. In fact, in that particular example,
+%    that key/value list is also saved and then applied to each
+%    \cs{item} which is implemented through an \texttt{item}
+%    template. This way, one can specify one-off settings to all items
+%    of a single whole list (on the environment level), as well as to
```
car222222, line 414: Maybe: 
in the whole of a single list . . . 

The parenthetical remark is not understandable.

base/lttemplates.dtx
```diff
@@ -351,12 +351,89 @@
...
+%    error message.}
+
+%    For example, a list, such as \env{enumerate}, is made up from a
+%    \texttt{blockenv}, \texttt{block}, \texttt{list}, and a
+%    \texttt{para} template and in the single user-supplied optional
+%    argument of \env{enumerate} key/values for either of
+%    the templates might get specified. In fact, in that particular example,
+%    that key/value list is also saved and then applied to each
+%    \cs{item} which is implemented through an \texttt{item}
+%    template. This way, one can specify one-off settings to all items
```
car222222, line 413: Maybe:
. . . for all the the items

base/lttemplates.dtx
```diff
@@ -351,12 +351,89 @@
...
+%    of a single whole list (on the environment level), as well as to
+%    individual items within a list (by specifying them in the
```
car222222, line 415: Betterment
. . . within that list . . .

base/lttemplates.dtx
```diff
@@ -351,12 +351,89 @@
...
+%    optional argument of an \cs{item}).  With
+%    \cs{SetKnownTemplateKeys} and \cs{SetTemplateKeys} working
+%    together, it is possible to provide this flexibility and still
+%    alert the user if one of their keys is misspelled.
```
car222222, line 419: Maybe: 
if ==> when

base/lttemplates.dtx
```diff
@@ -351,12 +351,89 @@
...
+%
+%    On the other hand you may want to allow for
+%    \enquote{misspellings} without generating an error or a
+%    warning. For example, if you define an template that only accepts
```
car222222, line 423: a template (not an) 

Better:
. . . that accepts only a few keys

base/lttemplates.dtx
```diff
@@ -351,12 +351,89 @@
...
+%    few keys, you might just want to ignore anything specified in the
+%    source if you use this template in place of a different one,
```
car222222, line 425: if ==> when

base/lttemplates.dtx
```diff
@@ -351,12 +351,89 @@
...
+%    without the need to alter the document source. Or you might just
+%    generate a warning message, which is easy, given that the unused
+%    key/values are available in the \cs{UnusedTemplateKeys} variable.
+%
+%    The \cs{SetKnownTemplateKeys} and \cs{SetTemplateKeys} commands
+%    are only supported within the code of a template; using them
+%    elsewhere has unpredictable results.  If they are used together
+%    with \cs{AssignTemplateKeys} then this command should come first
```
car222222, line 433: Unclear: 
Which should come first?

car222222 (commented):
Quite a lot here by now.

car222222:
Thanks for waiting!

car222222:
Not sure what was meant here:

whether or not top fully dop \SetTemplateKeys

Maybe it should read : 
whether to simply drop \SetTemplateKeys

u-fischer (approved):

davidcarlisle (approved):
Having both `\SetKnownTemplateKeys` and  `\SetTemplateKeys`is a little awkward but as discussion in this PR shows the use cases aren't totally clear until more experience with using these in the wild emerges. Keeping both seems wise for now. approving with no change

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1201,7 +1200,15 @@
...
               { \l_@@_block_instance_tl - \int_use:N
                 \g_block_nesting_depth_int }
-              {#1}
+              \UnusedTemplateKeys
+%    \end{macrocode}
+%    After this instance has been processed any remaining unused keys
+%    are in \cs{UnusedTemplateKeys} and we can continue with that as
+%    long as we do not call an instance that does unused key
+%    processing but we don't pass our keys on, so have to same its
+%    current value for now.
```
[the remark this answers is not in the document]
FrankMittelbach: neither can I when rereading. Tried to rewrite it.

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1233,7 +1240,25 @@
...
+%    \end{macrocode}
+%    Again the instance may has processed a few keys
+%    \begin{macrocode}
+      \tl_set_eq:NN \l_@@_unused_blockenv_keys_tl \UnusedTemplateKeys
+    }
+%    \end{macrocode}
+%    At this point, the \cs{l_@@_unused_blockenv_keys_tl} token list
+%    should either be empty or it should only contain key suitable for
+%    the item template, but right now I don't have code to test for
+%    the latter, guess it would help if we have an interface fore
```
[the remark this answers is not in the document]
FrankMittelbach: > What is "the item template"??

A template that implements the \item command (with its optional argument). It is defined elsewhere in the module

> Who is "I" ??? Why personalise this description text?? Try: there is no (available/known/??) code that can test . . .

as far as I know that's me (as the author of that v2 version of the module), but I have now problems to depersonalize
> 
> latter, ==> latter; it would . . . (delete "guess")

not sure why you mind the personalization, it is my personal guess that it would be useful, and but that isn't necessarily a fact that this would be better, but changed anyway ...

> 
> fore ==> for

base/lttemplates.dtx
```diff
@@ -351,12 +351,89 @@
...
+%   \begin{syntax}
+%   \cs{SetKnownTemplateKeys} \Arg{type} \Arg{template} \Arg{keyvals}
+%   \cs{SetTemplateKeys}      \Arg{type} \Arg{template} \Arg{keyvals}
+%   \end{syntax}
+%
+%    In the final argument of \cs{DeclareTemplateCode} one can also
+%    overwrite (some of) the current template key value settings by
+%    using the command \cs{SetKnownTemplateKeys} or
+%    \cs{SetTemplateKeys}, i.e., after the template default values and
+%    the values from the instance have been assigned.
```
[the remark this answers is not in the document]
FrankMittelbach: no idea what you mean by "further ..."
the "i.e.," explains what is meant by overwrite".
I have slightly rewritten, hope that helps.

base/lttemplates.dtx
```diff
@@ -351,12 +351,89 @@
...
+%
+%    The main use case for these commands is the situation where there
+%    is an argument (normally \texttt{\#1}) to the template in which
+%    a key/value list can be specified that overwrites the normal
+%    settings. In that case one could use
```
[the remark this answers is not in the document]
FrankMittelbach: I don't know, why you think this is weird or where you have problems to understand it.

Take the example of an enumerate environment
- this is implemented through an instance (actually several but we can ignore that here) which specify values for certain keys (like how much space is used, how the label looks like, etc).
- So if you use \begin{enumerate} this instance is used and the values are applied
- More exactly, the values applied are either the template default value (if not overwritten in the instance) or those specified in the instance.

Now if you use \begin{enumerate}[leftmargin=2cm]  you want for this particular enumerate to have a different value for leftmargin (compared to the normal one specified by the instance) .

- This optional argument is passed as the first mandatory argument to the instance.
- There it is handled by \SetKnownTemplateKeys and does the overwriting
- As a result this particular  environment in the document gets a different formatting.

base/lttemplates.dtx
```diff
@@ -351,12 +351,89 @@
...
+% \end{function}
+%
+%    If the \meta{keyvals} argument contains keys not known to the
+%    \meta{template} they are simply ignored and stored in the
+%    tokenlist \cs{UnusedTemplateKeys}
+%    without generating an error. This way it is possible to
+%    apply the same key/val list specified by the user on a
+%    document-level command or environment to several templates, which
+%    is useful, if their combination makes up that command or
+%    environment. As a variation of that, you can use this key/value
```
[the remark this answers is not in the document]
FrankMittelbach: slightly rewritten

base/lttemplates.dtx
```diff
@@ -351,12 +351,89 @@
...
+%    list the first time, and for the next template use what remains in
+%    \cs{UnusedTemplateKeys}. The final processing step could then be
```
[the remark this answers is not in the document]
FrankMittelbach: Not sure what you think needs to be explained. \SetKnownTemplateKeys gets a key/val list in ist argument. \UnusedTemplateKeys is such a key/val list, so you use it just like any other and pass it to that command

base/lttemplates.dtx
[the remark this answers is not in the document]
FrankMittelbach: > I cannot follow this bit: . . . which is implemented through . . .

The \item command is implement using its own template (through = is implemented using its own template and not using direct coding, for example)

base/lttemplates.dtx
FrankMittelbach: dropped "whole" but can't see anything wrong with the (...)
and "to all items" -> "for all items"

base/lttemplates.dtx
FrankMittelbach: ah you suggested that one as well

base/lttemplates.dtx
FrankMittelbach: the latter command?   I want to avoid writing \AssignTemplateKeys twice

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1954,7 +2013,19 @@
 %            to make use of user keys on the list level}
 %    \begin{macrocode}
     \@@_evaluate_saved_user_keys:nn {item}{std} 
-    \tl_if_empty:nF{#1}{ \SetTemplateKeys{item}{std}{#1} }
+%    \end{macrocode}
+%    We don't care if not all of the user key from the list level have
+%    been applied, but those explicitly set on the \cs{item} command
+%    should fit, so we generate an error if it doesn't
```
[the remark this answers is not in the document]
FrankMittelbach: rewritten

FrankMittelbach:
> Quite a lot here by now.

@car222222 I think I address them all (with one or the other question)

FrankMittelbach:
One could call it \SetAllTemplateKeys perhaps but I think the distinction is basically clear: one sets what it understands the other tries all and complains of it doesn't understand some keys

required/latex-lab/latex-lab-block.dtx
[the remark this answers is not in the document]
car222222: In context, such personalisation is probably OK.
But I have not seen such usage elsewhere in our documents (or in other similar documentation), so I suggested using a more common style for such documentation.
I would need to read the whole document to see if such usage makes sense in this case, and whether it is clear who the "I" actually is: did you sign it:-)?

base/lttemplates.dtx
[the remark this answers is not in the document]
car222222: Sure: but this explanation just makes me even more sceptical about the current template set-up, with its arbitrary distinctions between the use of arguments to the instance and the setting of key values. 
But I guess it works for your uses, and it lets Joseph play more "clever tricks with key-vals".
One basic problem is that two templates are distinguished by having only somewhat arbitrary syntax differences, rather than whether they are "semantically identical".
I would need to analyse all your current uses in order to further explain these conceptual problems with the current setup: but you probably do not want that done?

base/lttemplates.dtx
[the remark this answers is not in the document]
car222222: That is a good explanation: so please use it in the documentation.

base/lttemplates.dtx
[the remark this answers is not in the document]
car222222: Not sure what you mean here!
Whatever, please add some such explanation.

car222222: There is danger here of confusing people with the different potential hat least) uses of the term "item":
doex it always here refer to a \item construct in a "list environment", o4 is it also used in the more common and general sense (since many things around here consist, for example, of a "collection of items" that are key-vals, or just values: etc., etc.

base/lttemplates.dtx
car222222: Oops! Where did that word appear from: too fancy an "editorial assistant" now installed??

base/lttemplates.dtx
car222222: Yes, better to be more precise than just "this".

car222222:
I have not yet looked at the latest version: do you want me to check the most recent changes? 
I see nothing here to suggest that there even is now such a new version: the last edits recorded here are these from about a week ago:
"Make a proper error message and document . . .".

base/lttemplates.dtx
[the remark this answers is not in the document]
FrankMittelbach: Not as part of this PR, no. Maybe you should read the full latex-lab-block document and then see if what Bruno and I have implemented makes more sense to you then. None of this is really anything new conceptually and deeply based on work that you David and I did already long time ago.

base/lttemplates.dtx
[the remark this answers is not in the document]
FrankMittelbach: sorry no idea what I should write in addition. I wonder if the problem is that you look at isolated snippets (because unfortunately that's what the typical review interface is offering you).

required/latex-lab/latex-lab-block.dtx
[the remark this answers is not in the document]
FrankMittelbach: the document has a title and that gives my name and states that the code is based on earlier work by Bruno

base/lttemplates.dtx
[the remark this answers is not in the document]
FrankMittelbach: tried to do that (though I think it was already there).

u-fischer:
In a new example in the tagging project I came across this example:

~~~~
\DocumentMetadata{uncompress,testphase=latest}
\documentclass{article}
\usepackage{paralist}
\usepackage{multicol}

\begin{document}
\begin{enumerate}[\bfseries 1.]
 \begin{multicols}{4}
  \item $s = t^2$
  \item $s = 9.8t^2$
  \item $s = -16t^2 + 2t$
  \item $s = t^3$
 \end{multicols}
 \item By equation
 \end{enumerate}
\end{document}
~~~~

This compiles in main and develop, compared to the original the numbers are not in bold but the layout and the tagging looks  ok).

In this branch this now errors:

~~~~
! Package block Error: Some keys specified on the enumerate environment are
(block)                unknown.
~~~~

base/lttemplates.dtx
```diff
@@ -359,10 +359,12 @@
 % \end{function}
 %
 
-% \begin{function}{\SetKnownTemplateKeys,\SetTemplateKeys}
+% \begin{function}{\SetKnownTemplateKeys,\SetTemplateKeys,\UnusedTemplateKeys}
 %   \begin{syntax}
 %   \cs{SetKnownTemplateKeys} \Arg{type} \Arg{template} \Arg{keyvals}
 %   \cs{SetTemplateKeys}      \Arg{type} \Arg{template} \Arg{keyvals}
+%
+%   \cs{UnknownTemplateKeys} \% all \meta{keyvals} unused by previous \cs{SetKnownTemplateKeys}
```
u-fischer, line 367: is it Unknown or UnusedTemplateKeys??

FrankMittelbach:
> In a new example in the tagging project I came across this example:
> 
> ```
> \DocumentMetadata{uncompress,testphase=latest}
> \documentclass{article}
> \usepackage{paralist}
> \usepackage{multicol}
> 
> \begin{document}
> \begin{enumerate}[\bfseries 1.]
>  \begin{multicols}{4}
>   \item $s = t^2$
>   \item $s = 9.8t^2$
>   \item $s = -16t^2 + 2t$
>   \item $s = t^3$
>  \end{multicols}
>  \item By equation
>  \end{enumerate}
> \end{document}
> ```
> 
> This compiles in main and develop, compared to the original the numbers are not in bold but the layout and the tagging looks ok).
> 
> In this branch this now errors:
> 
> ```
> ! Package block Error: Some keys specified on the enumerate environment are
> (block)                unknown.
> ```

well, paralist overwrites enumerate, thus even privously it only worked by chance (tagging because evenually it called the list environment I guess).  And according to the compatibility table paralist is incompatible (and now this shows a bit clearer).

Besides: I don't think that putting multicols in the middle can beconsidered correct syntax, but that isn't the problem here.

base/lttemplates.dtx
FrankMittelbach: typo :-(

u-fischer:
> well, paralist overwrites enumerate, thus even privously it only worked by chance (tagging because evenually it called the list environment I guess). And according to the compatibility table paralist is incompatible (and now this shows a bit clearer).

Yes, but should it already be an error? Or should we only warn at the current state that these keys have not been processed? 
 
> Besides: I don't think that putting multicols in the middle can beconsidered correct syntax, but that isn't the problem here.

It's creepy ;-)  But it seems to work as intended.

car222222:
Ok, now I have found the "8 commits 7 days ago".  
This surprisingly contains stuff from a lot more than 7 days ago and only 2/3 days ago . . . very strange.

car222222:
I added some replies: 
see the "6 files changed" above. 

I can still,see many problems with the explanations, in both linguistic details and more fundamentally. 

But since the code is now as agreed, next to release this now.

I anyway need to look again at how this whole area has evolved, in the light of the complexities of current usages.  

. . . one day!

FrankMittelbach:
> It's creepy ;-) But it seems to work as intended.

aus Falschem folgt Beliebiges

FrankMittelbach:
> Ok, now I have found the "8 commits 7 days ago". This surprisingly contains stuff from a lot more than 7 days ago and only 2/3 days ago . . . very strange.

I only wanted to point out that there have been commits. One of them is bringing the banch up to speed with unrelated development already in the -dev branch, so if you look at the individual commits you will see a lot more.

FrankMittelbach:
> I added some replies: see the "6 files changed" above.

sorry don't see any new replies from you (and I also got no email notices, which I normally get)

> I can still,see many problems with the explanations, in both linguistic details and more fundamentally.
> 
> But since the code is now as agreed, next to release this now.
> 
> I anyway need to look again at how this whole area has evolved, in the light of the complexities of current usages.
> 
> . . . one day!

well, one day look at the whole thing, it seems  that on the level of review of snippets that is just not workable (as you then do not see the explanations that are nearby (which may or may not be sufficient, but that can't be assest from then bits that have changed.).

car222222:
I agree with @FrankMittelbach on this:

"one day look at the whole thing, it seems that on the level of review of snippets that is just not workable"

But the "whole thing" should include also the mandatory (normal) arguments of template use, for example: why is it so important that changing an argument from being a key-value to a #n form completely changes the "template type" despite there being absolutely no change in what the template does.  

A plea for more mutability within a template type!

PR #1618, opened 2025-01-11
correct doc about list handler (ltluatex)

Udi-Fogiel:
The documentation states that of all the functions in the callback returns true then the callback returns true as well, but the following example (and the code) shows that the return value in this case is the head passed to the callback.

```tex
\directlua{
luatexbase.create_callback('foo', 'list', false)
luatexbase.add_to_callback('foo', function(head) print('bar') return true end, 'baz') 
print(luatexbase.call_callback('foo', token.scan_list()))
}\hbox{foo}

\stop
```

# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -6,6 +6,12 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2025-01-10  Udi Fogiel  <udi@udifogiel.com>
+
+	* ltluatex.dtx (subsubsection{Handlers}):
+	Correct documentation about the return value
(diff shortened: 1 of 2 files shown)


davidcarlisle (approved):
Yes this seems OK. The documentation was true when written but the code changed in 2021

```
2021-11-17  Marcel Krüger  <Marcel.Krueger@latex-project.org>

	* ltluatex.dtx:
	Classify [hv]pack_quality callbacks as exclusive
	Never pass true to the engine for LuaTeX list callbacks
```

PR #1627, opened 2025-01-18
adding \invisibletimes and potentially other uses of intent

davidcarlisle:
# Internal housekeeping

## Status of pull request

- Feedback wanted 

- [X] Test file(s) added

This adds a new latex-lab file for handling `intent`, currently just used to define `\invisibletimes` and `\functionapplication` which, if `luamml` is loaded, inserts
`<mo intent="times">U+2062</mo>` into the generated MathML.


FrankMittelbach (commented):
I don't feel that confident to approve or disapprove here, but I wonder about the names.
Is it really useful to have basically random names, or would irt be better to consider providing some uniformity, like
```
\intent{times}
\intent{applied to}
```
or \mathintent.

davidcarlisle:
@FrankMittelbach my thought was that the current names are more adding operator names (which could be in amsmath or unicode-math) with the fact that they use intent more or less hidden in the tagging code.  the names match the html/mathml names eg `\invisibletimes`  corresponds to `&InvisibleTimes;` 

Longer term I'd like something like `\intent{...}`  or `\sum[intent=formal power series]` or some such syntax that annotates an existing math atom with an additional attribute, but here `\invisibletimes` isn't adding an attribute to something that was there: it's just inserting a new operator `<mo intent="times"></mo>` in the same way that `\times` inserts `<mo>×</mo>`

required/latex-lab/latex-lab-mathintent.dtx
```diff
@@ -0,0 +1,109 @@
...
+%
+%    \begin{macrocode}
+\protected\def\invisibletimes{
+\if_cs_exist:N\luamml_annotate:en
+  \luamml_annotate:en {
+   core = {[0] = 'mo', intent="times","^^^^2062"},
+}{
+ \latelua{}
+}
+\fi:
```
josephwright, line 85: Shouldn't these be `\cs_new_protected`? Also, can't this be done late enough that they are expanded to a single definition rather than re-checking every time they are used?

davidcarlisle: in the end, probably yes (but in the end they wouldn't be in latex-lab either)  the timing is a bit tricky as luamml is (normally) only implicitly loaded if suitable tagging options are set up and unicode-math has  been loaded by begin document, so you'd have to delay the definition until then, which is possible of course, although the main concern here was finding something that works, ie passing luamml information without affecting the typeset result. I started with `\protected\def` as  I was going to suggest it as an addition for amsmath so it was always available (as a no-op   (but then I needeed to test for the l3 luamml any anyway so it's a bit hybrid.

u-fischer (approved):

required/latex-lab/testfiles-math-luatex/mathml-luamml-5.lvt
```diff
@@ -0,0 +1,25 @@
+\ExplSyntaxOn
+\sys_gset_rand_seed:n{1000}
+\ExplSyntaxOff
+\DocumentMetadata{testphase={phase-III,math}, pdfversion=2.0,
```
FrankMittelbach, line 4: not really important but we should start using testphase=latest in my opinion, that is helping in the furture if new functionaly is breaking stuff, that might otherwise be masked by loading only a portion of the code.

davidcarlisle: done

Issue #1632, opened 2025-01-21
global use of counter commands such as  `\alph*` to access the current counter.


davidcarlisle:
## Brief outline of the enhancement

`enumitem` has popularised using `\arabic*` and similar in item labels to access the current list counter. This feature isn't yet available in the `latex-lab` list module but the addition of `\@currentcounter` to the format back in 2020 makes it possible for these commands to be globally available, not just in item labels. So for example `\roman*` (really `\roman{*}`) would access the current counter as used by `\label` and typeset it in roman. 

It could have a trivial definition, as shown below,

```latex
\documentclass{article}
\makeatletter
\protected\ExpandArgs{c}\def{c@*}{\value\@currentcounter}
\makeatother

\begin{document}

\section{aaa}
\arabic*

\subsection{bb}
\subsection{bbb}

\arabic*


\begin{enumerate}
\item \arabic*
\end{enumerate}
\end{document}
```

PR #1637, opened 2025-01-22
code for #1632

davidcarlisle:
- Ready to merge

- [X] Test file(s) added

I haven't added rollback `\c@*` is a new feature so only rollback would be to undefined, and `\stepcounter ` already has recent rollback for pre-socket versions

see #1632 for the enhancement


base/ltcounts.dtx
```diff
@@ -72,13 +75,13 @@
 % To set the value of counters.
 %
 % \DescribeMacro{\addtocounter}
-% Increase the counter |#1| by the number |#2|.
+% Increase the \marg{counter} |#1| by the number |#2|.
 %
 % \DescribeMacro{\stepcounter}
-% Increase a counter by one.
+% Increase th \marg{counter} by one.
```
muzimuzhi, line 81:
```suggestion
% Increase the \marg{counter} by one.
```

base/ltxref.dtx
```diff
@@ -465,7 +466,7 @@
 \def\refstepcounter#1{%
   \UseSocket{refstepcounter}{%
     \stepcounter{#1}%
-    \edef\@currentcounter{#1}%
+    \edef\@currentcounter{\ifx*#1\@currentcounter\else#1\fi}%
```
muzimuzhi, line 469: What if `#1` contains multiple tokens and starts with `*`? For example `\refstepcounter{*ctr}`.

Do we want to check that if `#1` is a single token here?

base/ltcounts.dtx
```diff
@@ -246,7 +249,19 @@
 %    \end{macrocode}
 % \end{macro}
 %
-% \begin{macro}{\@newctr}
+% \begin{macro}{\c@*}
+% \changes{v1.2a}{2025/01/22}{\cs{c@*} added gh/1632}
```
muzimuzhi, line 253:
```suggestion
% \changes{v1.2a}{2025/01/22}{\cs{c@*} added (gh/1632)}
```

base/doc/ltnews41.tex
```diff
@@ -244,6 +244,14 @@ \subsection{Socket and plug conditionals}
 %
 \githubissue{1577}
 
+\subsection{Accessing the current counter}
+Counter commands such as \cs{alph}, \cs{stepcounter}, may now have
+argument \texttt{*} to denote the current counter (as used by
```
u-fischer, line 249: have the argument

base/changes.txt
```diff
@@ -6,6 +6,12 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2025-01-22  David Carlisle  <David.Carlisle@latex-project.org>
+
+	* ltcounts.dtx, ltxref.dtx:
+	Allow * to denote the current counter (as used by \label) in counter
+	related commands, \alph*  \stepcounter* etc. gh/1632
```
muzimuzhi, line 13:
```suggestion
	related commands, \alph*, \stepcounter* etc. (gh/1632)
```

base/doc/ltnews41.tex
```diff
@@ -244,6 +244,14 @@ \subsection{Socket and plug conditionals}
...
+\cs{label}).  This is compatible with the \pkg{enumitem} use of
```
u-fischer, line 250: with the package \pkg{enumitem}

base/ltcounts.dtx
```diff
@@ -56,14 +56,17 @@
...
 %
 % \changes{v1.0c}{1994/03/29}
 %     {Create file from parts of ltmiscen and ltherest.}
 % \changes{v1.1a}{1994/05/19}{Extracted file from ltcntlen.}
 % \changes{v1.1d}{1996/04/23}{Documentation improvements}
 %
 % \section{Counters and Lengths}
-% Commands for defining and using counters. This file defines:
+% Commands for defining and using counters.
+% This file definef the following commands. In each case \marg{counter} may be
```
u-fischer, line 68: defines

base/ltcounts.dtx
```diff
@@ -56,14 +56,17 @@
...
+% |*| denoting the current counter as set by a previous |\refstepcouter|.
```
u-fischer, line 69: |\refstepcounter|

base/ltcounts.dtx
u-fischer, line 81: th -> the

base/ltcounts.dtx
```diff
@@ -246,7 +249,19 @@
 %    \end{macrocode}
 % \end{macro}
 %
-% \begin{macro}{\@newctr}
+% \begin{macro}{\c@*}
+% \changes{v1.2a}{2025/01/22}{\cs{c@*} added gh/1632}
+% Make the current counter available as a \LaTeX\ counter
+% with name |*|, so `\alph*` returns the current list
```
u-fischer: why *list* counter?

muzimuzhi (changes requested):
One typo correction, two styling suggestions, and a question on usages like `\refstepcounter{*ctr}`.

u-fischer (commented):
a few typos. And we probably should check if some packages set \@currentcounter and so perhaps need a guard too.

base/ltxref.dtx
FrankMittelbach: I think one has to (such "*..." counter exist)

base/changes.txt
FrankMittelbach: but then also comma in front of "etc."

base/ltcounts.dtx
FrankMittelbach: because it started out like that :-) but it is now broader, so should be changed.

base/ltxref.dtx
davidcarlisle: I suppose so, I wanted to avoid doing too much overhead but i guess I should make this safer

base/doc/ltnews41.tex
davidcarlisle: I guess so, it's probably ok as it is as well, but ...

base/doc/ltnews41.tex
davidcarlisle: I _keep_ doing that (reading `pkg` as package so not adding it)

base/ltxref.dtx
davidcarlisle: also `\@kernel@refstepcounter` needs the same

davidcarlisle:
```
\def\ltx@star@counter{*}
\def\refstepcounter#1{%
  \UseSocket{refstepcounter}{%
    \stepcounter{#1}%
    \edef\reserved@a{#1}%
    \ifx\reserved@a\ltx@star@counter\else
      \edef\@currentcounter{#1}%
    \fi
 .....
```

?

davidcarlisle: oh I meant to do `\let\@currentcounter\reserved@a` inside the if

davidcarlisle: resolved by 4cc48dfef I hope

FrankMittelbach: I think that'll do

FrankMittelbach:
> a few typos. And we probably should check if some packages set @currentcounter and so perhaps need a guard too.

As far as I can see, the only files that set `\@currentcounter`to something like `#1` and not to an explicit string like "postnote" are our own, so I think we should be okay

base/ltxref.dtx
```diff
@@ -462,10 +463,14 @@
 %    \end{macrocode}
 % \end{plugdecl}
 %    \begin{macrocode}
+\def\ltx@star@counter{*}
```
FrankMittelbach, line 466: somewhere there should be a `\begin{macro}{\ltx@star@counter}`.

perhaps `\ltx@currentcounter@alias`is a better name? (maybe not)

FrankMittelbach (approved):
approved with a minor comment

base/ltxref.dtx
davidcarlisle: I added macro env,left name as it is

base/ltcounts.dtx
```diff
@@ -56,14 +56,17 @@
...
 % \changes{v1.0c}{1994/03/29}
 %     {Create file from parts of ltmiscen and ltherest.}
 % \changes{v1.1a}{1994/05/19}{Extracted file from ltcntlen.}
 % \changes{v1.1d}{1996/04/23}{Documentation improvements}
 %
 % \section{Counters and Lengths}
-% Commands for defining and using counters. This file defines:
+% Commands for defining and using counters.
+% This file defines the following commands. In each case \marg{counter} may be
+% |*| denoting the current counter as set by a previous |\refstepcounter|.
```
Skillmon, line 69: That's a bad idea for `\counterwithin`/`\counterwithout`, isn't it?

davidcarlisle: @Skillmon  er probably

```

\show\cl@subsection
\counterwithin{equation}{*}
\show\cl@subsection
\expandafter\show\csname cl@*\endcsname

\show\cl@equation
\counterwithin{*}{equation}
\show\cl@equation

```

but what would you have it do in either case, be an error or act as if the current counter name had been used at that point?

davidcarlisle: or not be an error, and do what it does now, which is put the `*` in the reset list so the reset works on whatever the current counter is at that time, which is a bit weird but sort of what the user asks for.

FrankMittelbach: I think I would change `\@ifbothcounters` and generate an error "*-alias for current counter not supported", or you could expand there any * to the current counter and then do what it currently does, but I think if anything it should generate an error (or just a documentation that with those you get what you get :-))

FrankMittelbach: I agree, but that could be handled by just documenting it, rather then runtime processing for something nobody (sane) would even try. But I leave this up to @davidcarlisle

davidcarlisle: could do something like

```
\def\@ifbothcounters#1#2#3{%
  \begingroup\let\c@*\@undefined
  \@ifundefined{c@#1}{\endgroup\@nocounterr{#1}}%
    {% else counter is defined
     \@ifundefined{c@#2}{\endgroup\@nocounterr{#2}}%
       {\endgroup% else both counter and within  are defined
        #3}}}
```

But I'm tempted just to adjust the documentation and say don't use `*` it's not like the `\refstepcounter` case which puts you in a loop so really need a guard, here it just does what it does.

FrankMittelbach: either or would be fine imho, but I think  you can move the \endgroup to the outside since counter defs are fully global.

davidcarlisle: ah it's good someone knows this code:-) OK  I'll do that, it looks less weird.

davidcarlisle: hopefully resolved by f9f49c376

FrankMittelbach: maybe one should make a remark somewhere (ltnews41?) that the * convention is not allowed on all commands.

davidcarlisle: @FrankMittelbach  added to ltnews ddd622501

Skillmon (approved):

muzimuzhi:
> I haven't added rollback `\c@*` is a new feature so only rollback would be to undefined, and `\stepcounter ` already has recent rollback for pre-socket versions

The last-minute commit https://github.com/latex3/latex2e/pull/1637/commits/f9f49c3764b77a1d4dad6a6e6ad42f370dbc7138 modified `\@ifbothcounters`, which may need a rollback version.

PR #1659, opened 2025-02-12
OR consolidation

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development

- [n/a] Test file(s) added


FrankMittelbach:
At this point in time I'm just looking for feedback on code and names for the sockets and hooks. If those are ok, then the next step is integration and consolidation of the documentation (and the a final review).

davidcarlisle:
>  since the next level is always "page" or "column"

Ah.  `build` works quite well then

u-fischer:
Do packages like manyfoot or reledmac need more sockets or hooks for additional footnote systems?

FrankMittelbach:
> Do packages like manyfoot or reledmac need more sockets or hooks for additional footnote systems?

- with its current implementation manyfoot does not (it simply uses the column/before hook)
- reledmac overwrites `\@makecol` so will work in that respect, but if it could be implemented using the existing hooks or needs further ones remains to be seen.

See `changing-@makecol-dependencies.md` for package analysis

josephwright (approved):

FrankMittelbach:
sorry, merged by mistake already to develp. Could you @car222222 , @u-fischer, @davidcarlisle, @josephwright nevertheless add your comments even though in the end you can't any longer "approve" please?

u-fischer:
If only phase-I is used (so tagging is activated but new-or-2 is not loaded) then the footer/header disappear:

~~~~
\DocumentMetadata{uncompress,testphase={phase-I}}
\documentclass{article}
\pagestyle{headings}

\begin{document}
\section{blub}
abc
\end{document}
~~~~

FrankMittelbach:
I have made no provisions to keep phase-I functional.  I think it is a waste of time to try downgrading new functionality, not that code from phase-II is in the kernel.

I was expecting that we simply take it out or make it euiv to phase-II or make it load new-or-2 or ...

but new-or-2 should also vanish once tagpdf has absorbed the declarations still in there.

FrankMittelbach:
if you want to keep phase-I functional I guess all that is really necessary is to move the plug assignments  in new-or-2 to tagpdf or to any other place where they are executed when when tagging is enabled.

u-fischer:
I don't mind to remove/disable phase-I, but then it should be either disabled fully, or perhaps be upgraded to phase-II.

FrankMittelbach:
I think that this phasing is separate of this PR. I didn't want to introduce more complications here. Basically

 - new-or-1 should vanish (file and module)
 - new-or-2 should vanish (remaining code moved elsewhere - to be decided)
 
For the vanishing headers when tagging is activated but nothing else: this is due to the sockets are actually noop rather than returning `#2`, I would expecting that get a suitable plug (from new-or-2) and noop is never used.

But I could add in lttagging:
```
\NewSocketPlug{tagsupport/build/page/header}{transparent}{#2}
\AssignSocketPlug{tagsupport/build/page/header}{transparent}

\NewSocketPlug{tagsupport/build/page/footer}{transparent}{#2}
\AssignSocketPlug{tagsupport/build/page/footer}{transparent}
```
Or even make noop return `#2` for such tagging sockets

u-fischer:
The socket definition in lttagging is wrong. As soon as a socket is used outside the latex-lab code you need to ensure that the default plugs are correct. So you should assign to a two-argument socket a default plug so that it passes `#2` as identity. .

~~~~
\NewSocketPlug{tagsupport/build/page/header}{default}{#2}
\AssignSocketPlug{tagsupport/build/page/header}{default}
\NewSocketPlug{tagsupport/build/page/footer}{default}{#2}
\AssignSocketPlug{tagsupport/build/page/footer}{default}
~~~~

Similar 1-argument sockets needs then

~~~~
\AssignSocketPlug{tagsupport/xxxx}{noop}
~~~~

FrankMittelbach:
> The socket definition in lttagging is wrong. As soon as a socket is used outside the latex-lab code you need to ensure that the default plugs are correct. So you should assign to a two-argument socket a default plug so that it passes `#2` as identity.

But they are only used with \UseTaggingSocket and that does the right thing  automatically whether tagging is on or off. The problem with phase-I is only that that enables tagging but then there is no suitable plug because this is only in new-or-2 currently.

So it doesn't need the definition you suggest unless you intend to use the socket with `\UseSocket`. None of the other sockets in lttagging have plugs set up or assigned, e.g., `\NewSocket{tagsupport/tbl/colspan}{}` and they are in precisely the same situation.

We can, of course, set up "no-tag-use" plugs for those that have 1 or 2 arguments, but as long as they have  asuitable tagging plug and are only used with `\UseTaggingSocket`that shouldn't be necessary.

u-fischer:
> But they are only used with \UseTaggingSocket and that does the right thing automatically whether tagging is on or off. 

No it does not. Without the correct default plug, it eats its arguments:

~~~~
\DocumentMetadata{uncompress,testphase={phase-I}}
\documentclass{article}
\NewSocket{tagsupport/test}{2}
\begin{document}
xxxx 

\UseSocket{tagsupport/test}{abc}{cde}

\UseTaggingSocket{test}{abc}{cde}
\end{document}
~~~~

FrankMittelbach:
Same happens here with phase-I if you don't add "table" module, then the necessary tagging plug is missing and the default is wrong:
```
\DocumentMetadata{uncompress,testphase={phase-I}}
\documentclass{article}

\usepackage{array}

\begin{document}

\begin{tabular}{cccc}
a & b & c & d \\
\multicolumn{3}{c}{X} & d
\end{tabular}

\end{document}
```

and until now we always said: don't use document elements if the necessary module (in this case table) is not loaded.
With the integration of the OR code in the kernel, the module for OR needs to be loaded, which is not the case with the current phase-I.

As I said, we can add (rather unnecessary) defaults for the situation that tagging is enabled but the tagging plugs are not assigned. Maybe it is a bit safer, but it somehow seems to me mainly dead code, because when tagging gets enabled all the tagging sockets that are part of the kernel should have tagging code plugs assigned.

u-fischer:
I don't think that it is good, if code in ltoutput.dtx relies on the loading of an experimental latex-lab module to correctly declare the tagging plugs. If we use tagging sockets in base then we should ensure that either the final tagging plugs or at least sensible fallbacks are defined and assigned there too.

FrankMittelbach:
But if we do this generally (and there are a dozen cases already in lttagging that are missing such plug settings) we might want to consider inventing `\NewTaggingSocket` so that we can say
```
\NewTaggingSocket{tlb/colspan}{1}
```
and that then makes the right default assignment to noop (or transparent for 2 arguments) automatically behind the scenes.

FrankMittelbach:
Something like
```
%  \begin{macro}{\NewTaggingSocket}
%    
%    
% \changes{v1.0n}{2025/02/16}{Initialize tagging sockets with 1 or 2 arguments}
%    \begin{macrocode}
\cs_new_protected:Npn \NewTaggingSocket #1 #2 {
  \socket_new:nn { tagsupport / #1 } { #2 }
  \int_case:nnF
      { \int_use:c { #2 } }
         {
           0 \prg_do_nothing:
           1 { \socket_assign_plug:nn { tagsupport /  #1 } { noop } }
%    \end{macrocode}
%    Tagging sockets with two arguments have an identity plug that
%    just passes the second argument. Its already assigned so we only
%    have to alter it.
%    \begin{macrocode}
           2 { \socket_set_plug:nnn { tagsupport / #1 } { identity } { ##2 } }
%    \end{macrocode}
%
%    \begin{macrocode}
         }
         \ERRORnewtaggingsocket  % that should get a proper error message
}
%    \end{macrocode}
%  \end{macro}
```
but then we probably also want `\NewTaggingSocketPlug` and `\AssignTaggingSocketPlug` to have a consistent interface.

base/doc/ltnews41.tex
```diff
@@ -186,13 +188,125 @@ \section{Replacement for the legacy mark mechanism}
...
+a largely hardwired algorithm with a limited number of configuration
+possibilities.  Packages that attempted to alter one or the other
+aspect of the process had to overwrite the internals with the usual
+problems: incompatibilities and out of date code whenever something
+was changed in \LaTeX{}.
+
+To improve this situation and to support the production of accessible
+PDF documents we started to refactor the output routine and added a
+number of hooks and sockets, so that packages that want to adjust the
+output routine can do so safely with out the dangers associated with
```
davidcarlisle, line 204: < with out
> without

base/doc/ltnews41.tex
```diff
@@ -186,13 +188,125 @@ \section{Replacement for the legacy mark mechanism}
...
+   \cs{newpage} or \cs{clearpage}.
+   While this is clearly a bug, it was the case since the days
+   of \LaTeX~2.09; thus for
+   compatibility we continue to support this behavior.
+\end{description}
+There are more configuration possibilities, mainly for class
+developers; more documentation on those can be found in
+\cite[\S54 ltoutput.dtx]{41:source2e}.
+
+
```
davidcarlisle, line 301: Not necessssarily now but at some point we'll probably need to put this kind of information in `clsguide` it will be hard to find in ltnews after a few releases.

base/ltoutput.dtx
```diff
@@ -1213,6 +1232,240 @@
...
+%   
+%      \end{description}
+%
+%      \task{fmi}{Decide on names for these helper commands. We could keep
+%        them similar to the above (because they are only supposed to be
+%        used by a few packages). However, we could alternatively use
+%        CamelCase names, under the assumption that classes may also
+%        directly define a plug for the socket, instead of loading a
+%        support package such as \pkg{footmisc} (or in addition
+%        to). --- decide}
```
davidcarlisle, line 1359: If we are going with the current names it's probably best to delete this `\task` as suggesting the names will change will discourage use.

josephwright (commented):
Looks good to me

base/doc/ltnews41.tex
FrankMittelbach: I agree, at least to the level that I described it in ltnews, the cite here is really to the more nitty gritty details, e.g., how to make yet another plug (which 0% people will ever do).  I could what I have placed into ltnews also place into clsguide

base/ltoutput.dtx
FrankMittelbach: Well, I'm just not sure if the assumption is really correct. and there will be any usage outside of things like footmisc or ftnright and there the nams with `@` are probably fine. The task is there mainly to suggest that people think if there are use cases.

PR #1661, opened 2025-02-14
Tagging of section numbers

u-fischer:
* Fix issue [latex3/tagging-project#787](https://github.com/latex3/tagging-project/issues/787)
* Change the tagging of section numbers to `section-number`  role-mapped to Span as this gives a better reading experience:

image

image

# Internal housekeeping
- Ready to merge

- [x] Test file(s) added (updated)


davidcarlisle (approved):
yes this gives a much better reading.
With `Lbl`, section headings are announced twice: a heading like  "2 Introduction" read as "section heading level 1, 2; section heading level 1, introduction".
With `Span` it is read as "section heading level 1, 2 introduction"

required/latex-lab/latex-lab-sec.dtx
```diff
@@ -68,7 +68,9 @@
 % \begin{itemize}
 %  \item The heading/title text of the section should be surrounded by a 
 %  heading tag, typically \texttt{Hn} with some value of \texttt{n}.
-%  The number of the section command can optionally be put in a \texttt{Lbl}. 
+%  Theoretically the number of the section command could be 
+%  optionally be put in a \texttt{Lbl}. But AT doesn't handle this well, so
```
FrankMittelbach, line 72: In theory, one could put the number of the section command in an \texttt{Lbl}. However, current AT doesn't ...

required/latex-lab/latex-lab-sec.dtx
```diff
@@ -637,11 +639,20 @@
...
-% A simple internal command. There is no need for something public,
+% \begin{macro}{\@hyp@section@target@nnn,\@hyp@section@target@n}
+% Two simple internal commands. There is no need for something public,
 % as packages defining their own version of \cs{@startsection} will 
 % probably need something slightly different based on \cs{MakeLinkTarget}. 
+% \changes{v0.84i}{2025/02/13}{Use only an external wrapper with one argument, 
+% to allow to hide the argument from \cs{MakeUppercase},
+% tagging-project issue \#787}
 %    \begin{macrocode}
+\cs_new_protected:Npn \@hyp@section@target@n #1 %
```
FrankMittelbach, line 650: can the argument contains \par (in optarg)?

required/latex-lab/latex-lab-sec.dtx
```diff
@@ -651,6 +662,7 @@
        \MakeLinkTarget#1{#2}
      }
   }
+\AddToNoCaseChangeList{\@hyp@section@target@n}
```
FrankMittelbach, line 665: Just a thought: I don't know how long this no case change list already is, but wouldn't it be faster (and perhaps even better readable, if there is a general purpose \donotchangecase{...} with one argument that is used around material that should not change case?

FrankMittelbach (approved):
approved with some queries and some textual suggestions

required/latex-lab/latex-lab-sec.dtx
josephwright: Do you mean `\NoCaseChange`?

required/latex-lab/latex-lab-sec.dtx
u-fischer: No. This is a prefix for a destination name (so that you get xyz.1.2. instead of say section.1.2 in the destination).

required/latex-lab/latex-lab-sec.dtx
u-fischer: I could instead put everywhere \NoCaseChange around the command if that is is faster, not sure.

davidcarlisle: it probably is faster as it has no effect on MakeUppercase in general whereas adding to the list means that every use of MakeUppercase has to check for `\@hyp@section@target@n`

josephwright: If we need more speed, I an always rework how `\AddToNoCaseChangeList` works

FrankMittelbach: Yes I meant that, I wasn't sure that surived in the final implementation. My guess is that it would be faster and and also clearer  codewise. what's your option @josephwright ?

FrankMittelbach: > If we need more speed, I an always rework how `\AddToNoCaseChangeList` works

maybe, but I don't see that it helps to have a long list if the task is always identical and understandably covered by a single command

u-fischer: I switched to NoCaseChange

PR #1668, opened 2025-02-19
Gh1667

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added

additional plugs as suggested aren't yet included (probably after next -dev release)


user227621:
I think something got confused here. You have deleted the plug (wrongly) named `floats-footnotes-space` (which actually implements the sequence `floats-space-footnotes`). This means that there is no longer a plug that implements the sequence `floats-space-footnotes`. The sequence for the `bottom` option has been changed to `space-floats-footnotes`, but that should actually be `floats-space-footnotes`.

I think the correct thing to do would be to simply change the name of the plug `floats-footnotes-space` to `floats-space-footnotes` and correct its description.

FrankMittelbach:
> > I think something got confused here. You have deleted the plug (wrongly) named `floats-footnotes-space` (which actually implements the sequence `floats-space-footnotes`). This means that there is no longer a plug that implements the sequence `floats-space-footnotes`.
> 
> The plug `floats-space-footnotes` was already in the file (with the correct implementation). The other one (now deleted) was simply a copy with a different name.

ok looks like I have to many versions open ... guess you are right here, there is a mistake on my part, so forget the first part of my reply.

u-fischer:
*  `\usepackage[bottom]{footmisc}` should use the (missing) `floats-space-footnotes`. That is what I get without the new or. 

* The date of the footmisc replacement should be updated, currently it says
~~~~
latex-lab-footmisc 2022/03/08 v6.0d a miscellany of footnote facilitie
s -- latex-lab version
~~~~

* Typo in ltoutput:  defaultfor

FrankMittelbach:
> > The sequence for the `bottom` option has been changed to `space-floats-footnotes`, but that should actually be `floats-space-footnotes`.
> 
> That is not true either (I think).

you were right here too ... not my day ...

user227621:
> you were right here too ... not my day ...

Well, maybe today is really not your day ;-). Now the descriptions in clsguide and ltnews are correct, but the actual plug definition (starting on line 2468) still uses the wrong name `floats-footnotes-space` (instead of `floats-space-footnotes`) and has the wrong description:

```latex
%  \begin{plugdecl}{floats-footnotes-space}
%     As before but the \cs{vfill} is at the bottom (page is ragged
%    bottom).
%    \begin{macrocode}
\NewSocketPlug {build/column/outputbox}{floats-footnotes-space} {%
  \@outputbox@attachfloats
  \@if@footnotes@TF
      {\@outputbox@append{\vfill}}%
      {\@outputbox@reinsertbskip}%
  \@outputbox@appendfootnotes
}
%    \end{macrocode}
%  \end{plugdecl}
```

The changes entry could also be reworded:

```latex
2025-02-19  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>

	* ltoutput.dtx:
	Dropped plug floats-footnoteotes-space that was added by mistake
```

No plug was dropped, and the wrong name was not so badly wrong that it contained the word `footnoteotes` ;-).

Issue #1675, opened 2025-02-24
\counterwithin does not expand its arguments enough


FrankMittelbach:
## Brief outline of the bug

`\counterwithin` in contrast to `\numberwithin` does not expand its definitions far enough so that one ends up with and undexpanded counter argument, e.g.,
```
> \thesection=macro:
->\thechapter .\arabic {\blub }.
```
While that is typically harmless it is nonetheless a bug that should be corrected.

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{book}
\usepackage{amsmath}
\def\blub{section}
\numberwithin{\blub}{chapter}
\show\thesection
\counterwithin{\blub}{chapter}
\show\thesection

\begin{document}
abc
\end{document}
```

Issue #1673, opened 2025-02-24
amsmath numberwithin error message


nbeisert:
## Brief outline of the bug

The amsmath macro `\numberwithin` reports a missing parent counter with an unspecific error message:

> ! LaTeX Error: No counter '?' defined.

rather than a specific error message (as used for a missing child counter):

> ! LaTeX Error: No counter 'secion' defined.

This is due to the use of `\@nocnterr{#3}` in `\numberwithin` rather than `\@nocounterr{#3}`. 

Moreover `\@nocnterr{#3}` has `{#3}` as an unwarrented argument which triggers another error message:

> ! LaTeX Error: Missing \begin{document}.

The suggested fix is to change `\@nocnterr{#3}` in the second line of `\numberwithin` to `\@nocounterr{#3}`. 

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}
\documentclass{article}
\usepackage{amsmath}

\numberwithin{equation}{section}

\numberwithin{eqation}{section}

\numberwithin{equation}{secion}

\makeatletter
\DeclareRobustCommand{\numberwithin}[3][\arabic]{%
  \@ifundefined{c@#2}{\@nocounterr{#2}}{%
    \@ifundefined{c@#3}{\@nocounterr{#3}}{%
      \@addtoreset{#2}{#3}%
      \@xp\xdef\csname the#2\endcsname{%
        \@xp\@nx\csname the#3\endcsname .\@nx#1{#2}}}}%
}

\numberwithin{equation}{secion}

\begin{document}
\section{section}
\begin{equation} EQ1 \end{equation}
\begin{equation} EQ2 \end{equation}
\end{document}
```

davidcarlisle:
Thanks for reporting this, it does appear to be as you say. (Another possible fix we could consider would be to define this to be an alias for `\counterwithin` . We decided against that at the time `\counterwithin` was added, but if we need to change the defintion of `\numberwithin` anyway it is a possibility.

FrankMittelbach:
`\counterwithin` was introduced in 2018, so I think it is okay to assume it is available. If somebody tries to install a current amsmath into an even older format I think that would fail anyway.

u-fischer:
Not sure if it matters, but `\numberwithin` and `\counterwithin` have a slightly different expansion behaviour:

~~~~
\documentclass{book}
\usepackage{amsmath}
\def\blub{section}
\numberwithin{\blub}{chapter}
\show\thesection
[...]

PR #1676, opened 2025-02-25
fix for #1675

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [n/a] `ltnewsX.tex` (and/or `latexchanges.tex`) updated  - don't think this is necessary


base/changes.txt
```diff
@@ -6,6 +6,12 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2025-02-25  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
+
+	* ltcounts.dtx (subsection{Environment Counter Macros}):
+	\counterwithin needs to expands its arguments in case
+	the contain macros (gh/1675)
+
```
mbertucci47, line 14: expands -> expand and the -> they

base/ltcounts.dtx
```diff
@@ -478,27 +478,18 @@
...
 %
-%  \begin{macro}{\counterwithout,\counterwithin}
+%  \begin{macro}{\counterwithin}
 %     New implementation using xparse and supporting an optional
 %    format argument.
-%    \changes{v1.1m}{2021/07/08}{New implementation for
-%          \cs{counterwithout} and \cs{counterwithin}}
+%  \changes{v1.1m}{2021/07/08}{New implementation for
+%           \cs{counterwithin}}
+% \changes{v1.2b}{2025/02/22}{\cs{counterwithin} needs to expands its arguments
```
mbertucci47, line 486: expands -> expand

josephwright (approved):

davidcarlisle (approved):
Code looks fine should there be a combined ltnews entry for this and the amsmath change?

FrankMittelbach:
> Code looks fine should there be a combined ltnews entry for this and the amsmath change?

maybe I wasn't sure what to write ... care to suggest/add one to the pr?

base/ltcounts.dtx
```diff
@@ -565,6 +559,61 @@
...
+%<latexrelease>                 {\counterwithout}{counter without}%
+%    \end{macrocode}
+%
+%    \begin{macrocode}
+\NewDocumentCommand \counterwithout {sO{\arabic}mm}{%
+  \@ifbothcounters{#3}{#4}{%
+    \@removefromreset{#3}{#4}%
+    \IfBooleanF #1%
+        {\expandafter
+         \gdef\csname the#3\endcsname {#2{#3}}}%
```
muzimuzhi, line 580: Does `\counterwithout` need a similar expansion on `#3` in defining `\the#3`?

```suggestion
        {\expandafter
         \xdef\csname the#3\endcsname {\unexpanded{#2}{#3}}}%
```

If so the rollbacks for `\counterwith(in|out)` need not be split.

Currently
```tex
\documentclass{book}
\usepackage{amsmath}
\def\blub{section}

\counterwithout{\blub}{chapter}
\show\thesection
% > \thesection=macro:
% ->\arabic {\blub }.

\begin{document}
\end{document}
```

base/ltcounts.dtx
```diff
@@ -507,9 +498,9 @@
     \@addtoreset{#3}{#4}%
     \IfBooleanF #1%
        {\expandafter
-         \gdef\csname the#3\expandafter\endcsname
-         \expandafter
-             {\csname the#4\endcsname .#2{#3}}}%
+         \xdef\csname the#3\expandafter\endcsname
+            {\expandafter\noexpand\csname the#4\endcsname
+             .\noexpand#2{#3}}}%
```
muzimuzhi, line 503: Is `#2` forced to be a single token, usually a command which takes one mandatory argument? Do we need `\unexpanded{#2}` here?

base/testfiles/github-1675.tlg
```diff
@@ -0,0 +1,17 @@
+This is a generated file for the LaTeX2e validation system.
+Don't change this file in any respect.
+> \thesection=\long macro:
+->\thechapter .\@arabic \c@section .
+l. ...\show\thesection
+> \thesection=macro:
+->\thechapter .\arabic {section}.
+l. ...\show\thesection
+> \thesection=macro:
+->\arabic {\blub }.
```
u-fischer, line 10: as @muzimuzhi  commented: I think counterwithout should expand in a similar way, and not create an `\arabic{\blub}`.

base/ltcounts.dtx
FrankMittelbach: yes it idoes need it, in fact I even made a test for it but forgot to do it then (the test shows the wrong output, sigh)

base/ltcounts.dtx
FrankMittelbach: It is documented to be a single token (format-cmd) but I guess you are right, it doesn't hurt to use unexpanded in this place.

base/ltcounts.dtx
FrankMittelbach: You are right that now the rollbacks could be combined again, but I'm not going to do this it is much easier to handle individual rollbacks (if there are changes in the future) and it doesn't really hurt much if the rollback is minimally slower. We have combined ones only because initially that is much easier to set up.

FrankMittelbach:
> > Code looks fine should there be a combined ltnews entry for this and the amsmath change?
> 
> maybe I wasn't sure what to write ... care to suggest/add one to the pr?

I added one (or rather two) for this and #1673. Please have a look @davidcarlisle

muzimuzhi (approved):

base/changes.txt
```diff
@@ -6,6 +6,12 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2025-03-01  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
+
+	* ltcounts.dtx (subsection{Environment Counter Macros}):
```
muzimuzhi, line 11: This subsection title is odd. There is no "environment counter" now. Could be resolved in a separate PR.

https://github.com/latex3/latex2e/blob/4f1d77ce355109250987fcbe128a692d612ba834/base/ltcounts.dtx#L139-L173

u-fischer (approved):

base/changes.txt
FrankMittelbach: I changed things a bit, but these files really would need a more general docu update one day, by the look of it (as it is still a mix of old 2.09 docu with much of it no longer correct byx now)

PR #1679, opened 2025-02-26
Add \expandableinput

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [n/a] Test file(s) added (straight copy of `expl3` function, so covered there)


base/changes.txt
```diff
@@ -6,6 +6,10 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2025-01-26  Joseph Wright  <Joseph.Wright@latex-project.org>
+	* ltexpl.dtx, usrguide.tex
+	Add \expandableinput
```
FrankMittelbach, line 11: There was the suggestion to use \ExpandableInput, not sure which is better or if there should/could be adifferent name

base/doc/ltnews41.tex
```diff
@@ -545,6 +545,20 @@ \subsection{Improving the handling of \cs{label}, \cs{index}, and \cs{glossary}}
...
+\subsection{A version of \cs{input} for expansion contexts}
+
+The \LaTeX{} definition of \cs{input} cannot be used in places where \TeX{} is
+performing expansion: the classic example is at the start of a tabular cell.
+There are a number of reasons for this: the key ones are that \cs{input}
+records which files are read and provides pre- and post-file hooks.
+
+To support the need to carry out file input in expansion contexts, we have now
+added \cs{expandableinput}: this skips recording the file name and any file
+hooks, but otherwise behaves like \cs{input}. In particular, it sill uses
```
FrankMittelbach, line 557: maybe sill(y) but :-)

base/doc/ltnews41.tex
```diff
@@ -545,6 +545,20 @@ \subsection{Improving the handling of \cs{label}, \cs{index}, and \cs{glossary}}
...
+\cs{input@path} when doing file lookup (contrasting with the \TeX{} primitive,
+saved in \LaTeX as \cs{@@input}).
```
FrankMittelbach, line 559: which is internally available for programmers as \cs{@@input}).

base/doc/usrguide.tex
```diff
@@ -1312,6 +1312,21 @@ \section{Expandable floating point (and other) calculations}
...
+\end{decl}
+The \LaTeX{} definition of \cs{input} cannot be used in places where \TeX{} is
+performing expansion: the classic example is at the start of a tabular cell.
+There are a number of reasons for this: the key ones are that \cs{input}
+records which files are read and provides pre- and post-file hooks.
+
+To support the need to carry out file input in expansion contexts, the command
+\cs{expandableinput} is available: this skips recording the file name and any
+file hooks, but otherwise behaves like \cs{input}. In particular, it sill uses
+\cs{input@path} when doing file lookup.
```
FrankMittelbach, line 1328: see above

base/ltexpl.dtx
```diff
@@ -549,4 +549,26 @@
 %<latexrelease>\EndIncludeInRelease
 %    \end{macrocode}
 %
+% \begin{macro}{\expandableinput}
+% \changes{v1.3i}{2025/02/26}{Added document level name for \cs{file_input_raw:n} (gh/514)}
```
FrankMittelbach, line 553: if it is called raw here then perhaps \rawinput might be a better choice on the document level.

FrankMittelbach (approved):
approved with some editorial comments + question on the name

base/changes.txt
josephwright: I wasn't sure if the name suggestion was all-lower or CamelCase - as this feels like a document command ...

base/ltexpl.dtx
josephwright: Was my initial thinking, but I wanted to see how it looked to others - and we had said 'expandable input' as a no (capitalisation not specified!)

base/changes.txt
car222222: Are "document commands" the CamelCase ones?

base/doc/usrguide.tex
```diff
@@ -1312,6 +1312,21 @@ \section{Expandable floating point (and other) calculations}
...
+  |\expandableinput| \arg{filename} \\
+\end{decl}
+The \LaTeX{} definition of \cs{input} cannot be used in places where \TeX{} is
+performing expansion: the classic example is at the start of a tabular cell.
+There are a number of reasons for this: the key ones are that \cs{input}
+records which files are read and provides pre- and post-file hooks.
+
+To support the need to carry out file input in expansion contexts, the command
+\cs{expandableinput} is available: this skips recording the file name and any
+file hooks, but otherwise behaves like \cs{input}. In particular, it still uses
```
car222222, line 1327: Does this wording mean/imply that "file hooks are recorded". 
If so, what does this mean? 
If not, then this needs rewording.

base/ltexpl.dtx
car222222: "expandable" seems to make more sense than "rawl, does it not?

car222222 (approved):
A few additional thoughts/questions.

base/ltexpl.dtx
car222222: Oops: should be "raw" !

josephwright: As we discussed in the team meeting, getting general users to understand when they might need something expandable is hard - so I don't think there is a perfect name. I can live with `\expandableinput`, `\ExpandableInput` or `\rawinput` (or some others - things like `\untrackedinput` or `\lowlevelinput` would also make some sense). Basically we have to pick _something_ :)

davidcarlisle: I think I'd prefer `\ExpandableInput` (camel case like `\InputIfFileExists`)  `\UntrackedInput` would also work.

davidcarlisle (approved):
Approving modulo ongoing discussion on the name. I left a comment already on that, although it's only a mild preference in any case.

base/ltexpl.dtx
car222222: I tend to agree with David: but using such a name name needs a very clear explanation of what it does, and does not, do! Including a reference to the name!

josephwright: @car222222 I added something I hope is clear to `usrguide`

PR #1691, opened 2025-03-04
Correct handling of lost chars in boxes with LuaTeX

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/doc/ltnews41.tex
+++ b/base/doc/ltnews41.tex
@@ -581,17 +581,14 @@ redefinitions in these places accept an extended syntax.
 
 \subsection{Tracing lost characters}
 
-The \eTeX{} primitive \cs{tracinglostchars} allows the user to obtain a warning
-or error if a character is not available. The default value used in \LaTeX{} is
-set so that lost character information is written to the log and terminal.
-Users may wish to make this into an error, in which case they should set it to
-one higher. This is a \TeX{} primitive so the syntax for this is:
-\begin{verbatim} \advance\tracinglostchars by 1\relax \end{verbatim}
+In LaTeX News 33~\cite{41:ltnews33} we announced that \cs{tracingall} changes
+\cs{tracinglostchars} to an error condition. This change has been reverted and
+\cs{tracingall} and \cs{tracingnone} no longer alter \cs{tracinglostchars} but
+retain its current setting.
 
-The reason we recommend setting the value relative to the default is that to
-get the same behavior across all engines, a different absolute value of
-\cs{tracinglostchars} is needed. This relates to how lost characters inside
-boxes are handled, where Lua\TeX{} takes a different approach to other engines.
+The default value used in \LaTeX{} is set so that lost character information is
+written to the log and terminal Users may wish to make this into an error, in
+which case they should set the value to~5.
 
 \githubissue{1687}
 
--- a/base/doc/ltnews41.tex
+++ b/base/doc/ltnews41.tex
@@ -390,8 +390,6 @@ be added later when the formula processing has already finished.
 %
 \taggingissue{762}
 
-
-
 \section{New or improved commands}
 
 \subsection{Socket and plug conditionals}
@@ -578,8 +576,24 @@ because there, they were redefined to expect just a mandatory argument
 and then do nothing. We have now changed that behavior so that the
 redefinitions in these places accept an extended syntax.
 %
+
 \githubissue{311}
 
+\subsection{Tracing lost characters}
+
+The \eTeX{} primitive \cs{tracinglostchars} allows the user to obtain a warning
+or error if a character is not available. The default value used in \LaTeX{} is
+set so that lost character information is written to the log and terminal.
+Users may wish to make this into an error, in which case they should set it to
+one higher. This is a \TeX{} primitive so the syntax for this is:
+\begin{verbatim} \advance\tracinglostchars by 1\relax \end{verbatim}
+
+The reason we recommend setting the value relative to the default is that to
+get the same behavior across all engines, a different absolute value of
+\cs{tracinglostchars} is needed. This relates to how lost characters inside
+boxes are handled, where Lua\TeX{} takes a different approach to other engines.
+
+\githubissue{1687}
 
 \section{Bug fixes}
 
--- a/base/doc/ltnews41.tex
+++ b/base/doc/ltnews41.tex
@@ -587,8 +587,8 @@ In LaTeX News 33~\cite{41:ltnews33} we announced that \cs{tracingall} changes
 retain its current setting.
 
 The default value used in \LaTeX{} is set so that lost character information is
-written to the log and terminal Users may wish to make this into an error, in
-which case they should set the value to~5.
+written to the log and terminal. Users may wish to make this into an error, in
+which case they should set the value to~5 (not~3); this works in all engines.
 
 \githubissue{1687}
 
--- a/base/ltfssbas.dtx
+++ b/base/ltfssbas.dtx
@@ -35,7 +35,7 @@
 %
 %
 \ProvidesFile{ltfssbas.dtx}
-             [2025/03/06 v3.2q LaTeX Kernel (NFSS Basic Macros)]
+             [2025/03/05 v3.2q LaTeX Kernel (NFSS Basic Macros)]
 % \iffalse
 \documentclass{ltxdoc}
 \begin{document}
--- a/base/ltfssbas.dtx
+++ b/base/ltfssbas.dtx
@@ -35,7 +35,7 @@
 %
 %
 \ProvidesFile{ltfssbas.dtx}
-             [2025/02/20 v3.2p LaTeX Kernel (NFSS Basic Macros)]
+             [2025/03/06 v3.2q LaTeX Kernel (NFSS Basic Macros)]
 % \iffalse
 \documentclass{ltxdoc}
 \begin{document}
@@ -2333,6 +2333,8 @@
 %    {Add version of \cs{showhyphens} that works with Xe\TeX.}
 % \changes{v3.2m}{2024/06/17}
 %    {set \cs{tracinglostchars} to 0 in \cs{showhyphens}}
+% \changes{v3.2q}{2025/03/05}
+%    {set \cs{tracinglostchars} to 0 in \cs{showhyphens} for all engines}
 %    The |\showhyphens| command must be redefined since the version in
 %    \texttt{plain.tex} uses |\tenrm|.  We have also made some further
 %    adjustments for its use in \LaTeX.
@@ -2348,6 +2350,7 @@
 \DeclareRobustCommand\showhyphens[1]{%
   \setbox0\vbox{%
     \color@begingroup
+    \tracinglostchars\z@
     \everypar{}%
     \parfillskip\z@skip\hsize\maxdimen
     \normalfont
--- a/base/ltplain.dtx
+++ b/base/ltplain.dtx
@@ -32,7 +32,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{ltplain.dtx}
-             [2024/11/02 v2.3k LaTeX Kernel (Plain TeX)]
+             [2025/03/05 v2.3l LaTeX Kernel (Plain TeX)]
 % \iffalse
 \documentclass{ltxdoc}
 \GetFileInfo{ltplain.dtx}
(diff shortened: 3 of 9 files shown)


FrankMittelbach (commented):
That is unfortunately not enough for the issue at hand. LaTeX sets \tracinglostchars in several places, i.e., in \tracingall and \tracingnone and those need to be adjusted as well.

By the way what are the differences between the different values 2/3/4 ?

josephwright:
OK, so perhaps plan B is `\chardef` the 'set on' value
```latex
\ifdefined\directlua
  \chardef\tracinglostchars@on=4
\else
  \chardef\tracinglostchars@on=3
\fi
```
then use this in the definition of `\loggingall`, etc.?

josephwright:
> By the way what are the differences between the different values 2/3/4 ?

For 4/5 instead of 2/3 -  'Hans agreed to match the classical behaviour when prodded'

josephwright:
Plan B would presumably go in `ltplain`, btw.

davidcarlisle:
> OK, so perhaps plan B is `\chardef` the 'set on' value
> 
> ```
> \ifdefined\directlua
>   \chardef\tracinglostchars@on=4
> \else
>   \chardef\tracinglostchars@on=3
> \fi
> ```
> 
> then use this in the definition of `\loggingall`, etc.?

we could do, or alternatively as `\loggingall` is already set in an edef with some engine level tests could do

`\tracinglostchars\ifx\directlua\undefined\thr@@\else 4 \fi`

shouldn't it be 3+2=5 ?

davidcarlisle:
> shouldn't it be 3+2=5 ?

or more exactly shouldn't the tracing values be 2 and 4, was it intended to make  `\tracingall` turn this in to an error condition?

davidcarlisle:
so

```
\tracinglostchars\ifx\directlua\undefined\tw@\else 4 \fi
```

FrankMittelbach:
> > shouldn't it be 3+2=5 ?
> 
> or more exactly shouldn't the tracing values be 2 and 4, was it intended to make `\tracingall` turn this in to an error condition?

I guess it was the intention back then, even though ltnews is a bit vague on the subject. However I'm okay with lkeeping it always just a warning which would also resolve the still open issue that `\tracingnone`sets it incorrectly back to 1. However that means
- `\tracinglostchars` setting should be taken out of both `\tracingall`  and `\tracingnone`
- There should be an ltnews entry that `\tracinglostchars` is recommented to be set to 5 in luatex (not 3) because that is what we propagate in ltnews up to now

Or alternatively, we go `\let\@@tracinglostchars=\tracinglostchars` and redefine it to be a macro which does different things when in luatex (i.e., undo Hans's special design to give the warning always in LaTeX by default, i.e., somehting like
```
\def\tracinglostchars{\afterassignment\@tracinglostchars\@tmpcnta}
\def\@tracinglostchars{%
  \ifcase\@tmpcnta ...
}
```
I think I'm in favor of the latter solution so that in LaTeX \tracinglostchars has the usual meaning for 0,1,2,3 for all engines.

davidcarlisle:
@FrankMittelbach  yes it ocurred to me that that would mean we could/should remove it from \tracingall/tracingnone.

I'm not sure I like the macro setter, it would work well for setting but I think you'd need to support `\showthe\tracinglostchars` and that may be tricky with any kind of wrapper.

FrankMittelbach:
> @FrankMittelbach yes it ocurred to me that that would mean we could/should remove it from \tracingall/tracingnone.
> 
> I'm not sure I like the macro setter, it would work well for setting but I think you'd need to support `\showthe\tracinglostchars` and that may be tricky with any kind of wrapper.

true, that is a catch and not worth trying to find a solution for (even though there probably is one given that it is only needed in LuaTeX). But if not we need an ltnews entry that explain that you need to do =5 not =3 with LaTeX to get missing glyph errors.

josephwright:
> true, that is a catch and not worth trying to find a solution for (even though there probably is one given that it is only needed in LuaTeX). But if not we need an ltnews entry that explain that you need to do =5 not =3 with LaTeX to get missing glyph errors.

No, it's not that - you'll get errors if you set to 3, but only when the text is used. The new higher values are needed to get a missing char error when characters are typeset in a box, before the box is used.

base/ltplain.dtx
```diff
@@ -1718,14 +1729,25 @@
...
+%  in recent Lua\TeX{}}
 %    \begin{macrocode}
 %<latexrelease>\IncludeInRelease{2021/06/01}{\loggingall}
 %<latexrelease>                 {\tracingstacklevels and \tracinglostchars=3}%
 %<*2ekernel|latexrelease>
 \edef\loggingall{%
   \tracingstats\tw@
   \tracingpages\@ne
-  \tracinglostchars\thr@@
+  \tracinglostchars
```
davidcarlisle, line 1741: as discussed in the comments, 3/5 is consistent with the change making it 3 back in 2021 but I think that was in error and over-compensenation for the fact that we increased the default from 1 to 2 at that time.

I don't think it's expected that `\tracingall` makes non-error conditions into errors, so I would make this 2 and 4 or (more simply) remove the setting from here and from `\tracingnone` so the generic tracing on/off commands leave this at its default level.

davidcarlisle (changes requested):
would also need ltnews entry telling people to use 4 in luatex

FrankMittelbach:
> > true, that is a catch and not worth trying to find a solution for (even though there probably is one given that it is only needed in LuaTeX). But if not we need an ltnews entry that explain that you need to do =5 not =3 with LaTeX to get missing glyph errors.
> 
> No, it's not that - you'll get errors if you set to 3, but only when the text is used. The new higher values are needed to get a missing char error when characters are typeset in a box, before the box is used.

but that is my point; yes you still get errors with a setting of 3 but not necessarily always (in contrast to pdftex, say) so to get the same behavior as with the other engines you need to set it to 5 not 3 in luatex

base/ltplain.dtx
josephwright: If you want to set to 2 all of the time, does it even need to be in `\tracingall`, etc.? The standard setting is now 2 anyway. (Or are we worrying about places that set to zero/one not being reset?)

FrankMittelbach: I agree \tracingall should not make it an error (it should simply leave the setting alone) but the recommendation for users in ltenews is to set it to 3 (to always make it an error) and that sensible and thus the new recommendation for LuaTeX would need to be to set it to 5 not 4.

josephwright:
> but that is my point; yes you still get errors with a setting of 3 but not necessarily always (in contrast to pdftex, say) so to get the same behavior as with the other engines you need to set it to 5 not 3 in luatex

Yes, true - I guess I wonder if that's our job to document, as this is not something we control, it's an engine feature.

davidcarlisle:
> > would also need ltnews entry telling people to use 4 in luatex
> 
> Again, only if they want identical behaviour to pdfTeX, etc.: setting to 2 or 3 still works, it's only an issue for boxes ...

3 only works if you use the box (and I think this is where this came up) it doesn't error at all if you don't use it, so 

```
\settowidth\foo{abcdefghijk}
```

won't error if the font has no `f`  if you just use `\foo` and don't typset an f somewhere.

base/ltplain.dtx
davidcarlisle: @josephwright  "does it even need to bein `tracingall`" N it doesn't, that's what @FrankMittelbach  and I suggested: "or (more simply) remove the setting from here "

eg9:
I believe that LuaTeX has always been wrong in the treatment of `\setbox`. Not that we can do anything about this, except recommending to set `\tracinglostchars` to 5 in LuaLaTeX (error when a character is missing).

Likely `\tracingall` should set it to 2 or 4, so no error is raised. No need to restore back to the previous value: whoever uses `\tracingall` knows how to read its output.

Or, in the long run, we could ask the maintainers of pdftex to make 4 act the same as 2 and 5 the same as 3.

base/ltplain.dtx
josephwright: (Do we need rollback for that? I'm guessing no)

FrankMittelbach: It doesn't need rollback but the whole setting of \tracinglostchars shoulw be removed including the next 9 lines setting it to 3 or 5. And the same with \tracingnone. It should stay transparent to whatever LaTeX has set or the user has changed to.

If there is a setting it should first save the current value (to return to that in \tracingnone) and then set it to 2/4 but I think it is over-engineering and we should take out the set as suggested by @davidcarlisle

base/ltplain.dtx
```diff
@@ -1820,7 +1842,16 @@
...
+    \ifdefined\luatexversion
+      \ifnum\luatexversion>120 %
+        4 %
+      \else
+        \tw@
+      \fi
+    \else
+      \tw@
+    \fi
```
FrankMittelbach, line 1854: take out ...

josephwright: I'm on it :)

base/ltfssbas.dtx
```diff
@@ -35,7 +35,7 @@
 %
 %
 \ProvidesFile{ltfssbas.dtx}
-             [2025/02/20 v3.2p LaTeX Kernel (NFSS Basic Macros)]
+             [2025/03/06 v3.2q LaTeX Kernel (NFSS Basic Macros)]
```
muzimuzhi, line 38: Perhaps this should still be 2025/03/05, not 06. Could be changed at last minute.
```suggestion
             [2025/03/05 v3.2q LaTeX Kernel (NFSS Basic Macros)]
```

base/doc/ltnews41.tex
```diff
@@ -578,8 +576,24 @@ \subsection{Improving the handling of \cs{label}, \cs{index}, and \cs{glossary}}
...
 \githubissue{311}
 
+\subsection{Tracing lost characters}
+
+The \eTeX{} primitive \cs{tracinglostchars} allows the user to obtain a warning
+or error if a character is not available. The default value used in \LaTeX{} is
+set so that lost character information is written to the log and terminal.
+Users may wish to make this into an error, in which case they should set it to
+one higher. This is a \TeX{} primitive so the syntax for this is:
+\begin{verbatim} \advance\tracinglostchars by 1\relax \end{verbatim}
```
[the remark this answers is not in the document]
josephwright: One for @FrankMittelbach  to decide about :)

base/testfiles/github-0524.xetex.tlg
```diff
@@ -0,0 +1,34 @@
...
+{\tracingassigns}
+{changing \tracingassigns=1}
+{\tracingrestores}
+{\tracingonline}
+{\tracingcommands}
+{horizontal mode: \tracingrestores}
+{\tracingassigns}
+{into \tracingassigns=1}
+{the letter h}
+Missing character: There is no h ("68) in font nullfont!
```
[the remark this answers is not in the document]
josephwright: Yeah, doubt it's deliberate - but I'm not sure there is much to be done

base/doc/ltnews41.tex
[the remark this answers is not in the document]
davidcarlisle: Rather than advance, couldn't we just tell to use 5 which gives an error on all engines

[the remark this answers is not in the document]
FrankMittelbach: > Rather than advance, couldn't we just tell to use 5 which gives an error on all engines

I agree that make the reading an explanation simpler.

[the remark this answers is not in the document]
josephwright: > Rather than advance, couldn't we just tell to use 5 which gives an error on all engines

OK, although I'm a bit wary that formally the behaviour is formally undefined for engines other than LuaTeX here.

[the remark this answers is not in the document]
davidcarlisle: In general all the tracing parameters are defined for all integer values,  with all values larger acting like  the largest defined value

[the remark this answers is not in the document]
FrankMittelbach: TeXbook 212 but its only explicitly says so for \tracingcommands there. Anyway, I think David is right high values give you the largest defined one by convention in all engines

FrankMittelbach (approved):

davidcarlisle (approved):

PR #1693, opened 2025-03-06
Footnote tests for latex-lab

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


FrankMittelbach:
> * for some reason a folder "testfiles-paris" with one file appeared?

incorrect upload

>     * I do think that some tests should test also with hyperref (you removed that everywhere), as a large portion of the footnote code tries to handle links from the marker to the footnote and the tagging should be tested too.

my bad; at midnight two days ago I thought that our tagging automatically loads hyperref. So yes, I should do that.
So what should I do: enable hyperref in every second file or in all?

>     * Should the testfiles in testfiles-OR and testfiles-OR-luatex be cleaned up?

probably, but not in this pr. All I wanted to do is to see how my test files for footmisc behave when using tagging, so this is basically a dump of them with tagging added and then check that there are no surprised

>     * unrelated but github-003b errors as you actually can not use `\MakePerPage[5]{footnote}`  after loading footmisc with the perpage option, as this already calls `\MakePerPage{footnote}`  and the perpage package doesn't like a second call.

hmm, missed that then. Wonder what that was supposed to test or if that did work with the old footmisc. Have to check.

FrankMittelbach:
> > my bad; at midnight two days ago I thought that our tagging automatically loads hyperref. So yes, I should do that. So what should I do: enable hyperref in every second file or in all?
> 
> Hm, not sure. But the most interesting parts are if there is e.g. a \footref or a \footnotemark/footnotetext combination.
> 

probably, but not as part of this PR. All I wanted to do is to add tests for footmisc to see if that still behaves as before if replaced with a tagged version

> > > * unrelated but github-003b errors as you actually can not use `\MakePerPage[5]{footnote}`  after loading footmisc with the perpage option, as this already calls `\MakePerPage{footnote}`  and the perpage package doesn't like a second call.

> It never worked imho.

so it seems. I just added a comment to the lvt

u-fischer (approved):

u-fischer:
sorry I forgot: I think the config-file is missing in the build.lua.

PR #1692, opened 2025-03-05
Gh1689

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

## Checklist of required changes before merge will be approved
- [x] Test file(s) added
- [x] Version and date string updated in changed source files
- [x] Relevant `\changes` entries in source included
- [x] Relevant `changes.txt` updated
- [x] Rollback provided (if necessary)?
- [x] `ltnewsX.tex` (and/or `latexchanges.tex`) updated

PR #1701, opened 2025-03-21
Footmisc bug 19

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added

This fixes a bug reported for footmisc

https://github.com/FrankMittelbach/fmitex-footmisc/issues/19

 but is also present in the kernel code


base/changes.txt
```diff
@@ -1,3 +1,21 @@
+2025-03-21  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
+
+	* ltoutput.dtx
+	(subsection{The output routine configuration components}):
+	Back up by \dp\@outputbox in the footnotes-floats-legacy plug if there
+	are no footnotes (footmisc/19)
+
+	(subsection{The output routine configuration components}):
+	Add experimantal socket for baseline attachment
```
mbertucci47, line 9: experimantal -> experimental

base/changes.txt
```diff
@@ -1,3 +1,21 @@
...
+	of footnotes -- subject to change (footmisc/19)
+
+	(subsubsection{Output routine commands}):
+	Remove a `fil' bottom skip even if we  are in ragged bottom
```
mbertucci47, line 13: extra space after "we"

base/ltoutput.dtx
```diff
@@ -2350,9 +2403,18 @@
 %    Then the footnotes are appended:
 %    \begin{macrocode}
      \@outputbox@append{%
+%    \end{macrocode}
+%    First we call a socket which may or may not back up by the depth
+%    of the \cs{@outputbox} so that we start adding from the baseline
+%    of the last line.
+%    Then we add the skip above the footnotes.
+%  \changes{v1.4q}{2025/03/20}{Add experimantal socket for baseline
```
mbertucci47, line 2411: experimantal -> experimental

base/ltoutput.dtx
```diff
@@ -2628,6 +2740,43 @@
 %  \end{socketdecl}
 %
 %
+%  \begin{socketdecl}{build/column/footnotes/baselineattach}
+%
+%  \changes{v1.4q}{2025/03/20}{Add experimantal socket for baseline
```
mbertucci47, line 2745: experimantal -> experimental

base/ltoutput.dtx
```diff
@@ -2628,6 +2740,43 @@
...
+       \ifdim\@outputbox@depth>\maxdepth
+          \maxdepth
+       \else
+         \@outputbox@depth
+       \fi
+  \fi
+}
+%    \end{macrocode}
+%    This is not turned on by default, and as mentioned, is so far
+%    only experimental and might change/valish again.
```
mbertucci47, line 2773: valish -> vanish

FrankMittelbach: sounds like a nice word though :-)

base/ltoutput.dtx
```diff
@@ -1070,22 +1069,71 @@
...
+
+%    Thus this backing up should only happen if there are no
+%    footnotes. The problem is that is it not known to \LaTeX{} at the
+%    point when \cs{newpage} is called if there are footnotes or not;
+%    that is only determined in the output routine.  We have therefore
+%    moved the code in the newly introduced output routine sockets.
+%  \changes{v1.4q}{2025/03/20}{Remove skip added in 2.4a (footmisc/19)}
+%    \begin{macrocode}
+%  \par
+%  \vskip \glueexpr \@flushglue
```
CastleStar14654, line 1092: Why `\@flushglue` is used here in this commentted line?

FrankMittelbach: it is a slightly different formulation of what was there before (combining 1fil with the negative of \prevdepth in a single vskip). I should take it out again  and replace it with exactly what there before, because it is just a comment by now --- or take out all commented out lines.

base/ltoutput.dtx
```diff
@@ -2628,6 +2739,43 @@
...
+%    \begin{macrocode}
+  \ifdim\@outputbox@depth>\z@
+    \vskip -%
+       \ifdim\@outputbox@depth>\maxdepth
+          \maxdepth
+       \else
+         \@outputbox@depth
+       \fi
+  \fi
+}
```
CastleStar14654, line 2769: For `build/column/outputbox` plugs where floats occur directly above footnotes, e.g. `space-floats-footnotes` and `floats-footnotes`, this socket plug would possibly reduce the glue between bottom floats and footnotes by `\@outputbox@depth`.

It would be necessary to update the `\@outputbox@depth` to the depth of bottom float boxes also in `\@cflb` (or, for conssitency and simplicity, in `\@outputbox@attachfloats`)?

CastleStar14654: Also, a similar plug for floats (such as `build/column//baselineattach`) could possibly be added. (A weird feature, though)

base/ltoutput.dtx
```diff
@@ -2592,12 +2653,62 @@
...
+%    end of the \cs{@outputbox} but we limit the skip to be at most
+%    \cs{maxdepth}. Thus very large depths will not be fully canceled.
+%    \begin{macrocode}
+           \vskip -%
+             \ifdim\@outputbox@depth>\maxdepth
+               \maxdepth
+             \else
+               \@outputbox@depth
+             \fi
+          \fi
```
CastleStar14654, line 2707: This part is actually necessary in front of all other `\@outputbox@reinsertbskip`.

Is it possible to wrap this "baseline removal subroutine" into a macro, which could also be shared by `build/column/footnotes/baselineattach`?

base/ltoutput.dtx
```diff
@@ -2224,49 +2282,43 @@
...
 %    (needed, for example, in \pkg{ftnright}). 
 %    \begin{macrocode}
-    \ifvoid \@outputbox
-      \global\let\@outputbox@reinsertbskip\relax
-    \else
+   \ifvoid \@outputbox
+     \global\let\@outputbox@reinsertbskip\relax
+   \else
 %    \end{macrocode}
 %    If not, we drop the final skip at the end of \cs{@outputbox}
```
car222222, line 2300: ?? : 

If it is not void then we drop . . .

base/ltoutput.dtx
```diff
@@ -2224,49 +2282,43 @@
...
-          \unskip
+     \@outputbox@append{%
+       \@tempskipa\lastskip
+       \ifnum \gluestretchorder\@tempskipa>\z@
+         \unskip
 %    \end{macrocode}
 %        
 %  \begin{macro}{\@outputbox@reinsertbskip}
-%    We also record the value so that it can be reinserted
+%    We also record its value so that it can be reinserted
```
car222222, line 2311: More precision needed: 
  what does "its" refer to??

base/ltoutput.dtx
```diff
@@ -2350,9 +2402,18 @@
 %    Then the footnotes are appended:
 %    \begin{macrocode}
      \@outputbox@append{%
+%    \end{macrocode}
+%    First we call a socket which may or may not back up by the depth
```
car222222, line 2406: "may or may not" is confusing here! 

Maybe:
. . . socket which controls whether or not we backup by . . .

base/ltoutput.dtx
```diff
@@ -2350,9 +2402,18 @@
...
+%    of the \cs{@outputbox} so that we start adding from the baseline
```
car222222, line 2407: "start adding" what? 

Maybe better:
. . . so that the added skip starts at the baseline of the last line.

base/ltoutput.dtx
```diff
@@ -2350,9 +2402,18 @@
...
+%    of the last line.
+%    Then we add the skip above the footnotes.
```
car222222, line 2409: Does this make sense in context? 
What skip is "the skip" that we add? 
Maybe start with: 
  Following this we also add . . .

base/ltoutput.dtx
```diff
@@ -2592,12 +2653,62 @@
 \NewSocketPlug {build/column/outputbox}{footnotes-floats-legacy} {%
 %    \end{macrocode}
 %    In the legacy case we don't really want to take out the bottom
-%    skip, but rather then altering \cs{@outputbox@removebskip} in
+%    skip, but rather than altering \cs{@outputbox@removebskip} in
 %    \cs{makecol} to do nothing, which we would then have to undo in
 %    each other layout, we simply reinsert the dropped skip
```
car222222, line 2658: each  ==>  every 

And Better:
  . . . layout, we immediately reinsert the dropped skip.

base/ltoutput.dtx
```diff
@@ -2592,12 +2653,62 @@
...
 %    immediately again.
+%
+%    In the legacy case there is, however, one other thing that needs
+%    doing: if there was a \cs{vfil} at the bottom of the page (e.g.,
+%    from \cs{newpage}) and there are no footnotes, then we need to back up
+%    by the depth of the last line in the page box. Otherwise you
+%    might end up with incorrect alignment if there is a higher-order
+%    fill on the page (e.g., \cs{vfill}) that pushes the text material
+%    to the bottom because then the depth will not be taken into
```
car222222, line 2667: . . . bottom, since in such cases the depth . . .

base/ltoutput.dtx
```diff
@@ -2592,12 +2653,62 @@
...
+%    account by \TeX{}.
+%    
+%    Given, that a bottom skip was taken out earlier, we can look at
```
car222222, line 2670: taken out  ==>  removed

base/ltoutput.dtx
```diff
@@ -2592,12 +2653,62 @@
...
+%    the \cs{@outputbox@depth} value for this and back up by that
```
car222222, line 2671: What is "this"? 
And why can you "look at . . . for this"?? 
Maybe? : 
we find "this?" in . . .

base/ltoutput.dtx
```diff
@@ -2592,12 +2653,62 @@
...
 %    \begin{macrocode}
-  \@outputbox@reinsertbskip
+  \ifx\@outputbox@reinsertbskip\relax
+  \else
+%    \end{macrocode}
+%    Otherwise, if we have footnotes we also do nothing.
+%    \begin{macrocode}
+  \@if@footnotes@TF{}%
+%    \end{macrocode}
+%    Otherwise, if the depth is greater than zero we back up. If it is
```
car222222, line 2686: back up by how much?

base/ltoutput.dtx
```diff
@@ -2592,12 +2653,62 @@
...
+%    zero, there is no point doing anything and if the depth is
+%    negative it must have been deliberately set to this value in
+%    which case we aren't altering it either.
```
car222222, line 2689: . . . to this value, so in this case we also do not change it.

base/ltoutput.dtx
```diff
@@ -2592,12 +2653,62 @@
...
+%    The test has to be made inside \cs{@outputbx@append} because only
+%    then \cs{@outputbox@dp} has a correct value.
```
car222222, line 2691: there does . . . have the correct value.

base/ltoutput.dtx
```diff
@@ -2592,12 +2653,62 @@
...
+%    \begin{macrocode}
+      {%
+        \@outputbox@append{%
+          \ifdim\@outputbox@depth>\z@
+%    \end{macrocode}
+%    If the depth is positive we back up by appending a skip at the
+%    end of the \cs{@outputbox} but we limit the skip to be at most
```
car222222, line 2698: . . . ; but . . .

base/ltoutput.dtx
```diff
@@ -2628,6 +2739,43 @@
 %  \end{socketdecl}
 %
 %
+%  \begin{socketdecl}{build/column/footnotes/baselineattach}
+%
+%  \changes{v1.4q}{2025/03/20}{Add experimental socket for baseline
+%    attachment of footnotes -- subject to change (footmisc/19)}
+%
+%    Experimental socket allowing to back up by the depth of the
```
car222222, line 2747: . . . socket that allows backing up by . . .

base/ltoutput.dtx
```diff
@@ -2628,6 +2739,43 @@
...
+%    outputbox (but not more than \cs{maxdepth}  before attaching footnotes.
```
car222222, line 2748: . . . attaching any/the footnotes.

base/ltoutput.dtx
```diff
@@ -2628,6 +2739,43 @@
...
+%    \begin{macrocode}
+\NewSocket{build/column/footnotes/baselineattach}{0}
+%    \end{macrocode}
+%  \end{socketdecl}
+%
+%  \begin{plugdecl}{on}
+%    \begin{macrocode}
+\NewSocketPlug {build/column/footnotes/baselineattach}{on} {%
+%    \end{macrocode}
+%    We do nothing if the depth is negative and we also never back up
```
car222222, line 2758: . . . negative, and we never backup by more than . . .

base/ltoutput.dtx
```diff
@@ -2628,6 +2739,43 @@
...
+    \vskip -%
+       \ifdim\@outputbox@depth>\maxdepth
+          \maxdepth
+       \else
+         \@outputbox@depth
+       \fi
+  \fi
+}
+%    \end{macrocode}
+%    This is not turned on by default, and as mentioned, is so far
```
car222222, line 2771: "as mentioned" where?? 

Maybe: 
. . default, and it is currently

car222222 (approved):
Lots of ideas/comments, and a few question marks.

base/changes.txt
car222222: Also, "Add an . . . "

base/ltoutput.dtx
FrankMittelbach: As you correctly observe `\@cfb` always results in an `\@outputbox` with zero depth so reducing by that depth is pointless. 

Floats have always a depth of zero as well (due to the way `\@endfloatbox` is defined -- it adds
```
      \par\vskip\z@skip      %% \par\vskip\z@ added 15 Dec 87
```
at the end of each float) and therefore using `\unskip`and then trying to back up by `\@outputbox@depth` would make no differences either.

Changing that would affect sequences of floats in the top or bottom area and therefore would either require surgery in several other areas and/or would most likely then result not only in adjustments  of the layout of a single page but also alter the pagination of the whole document as it might mean that the content in the float areas change size when constructed and taken into account. One could probably avoid that but only with even more areas to deal with possible side effects that need to be taken care of.

Bottom line I'm not sure I want to go there :-) which is why I said elsewhere  I'm unsure or or how that configuration possibility  could/should be generalized

CastleStar14654: Agree. Fair enough too keep the depth of float boxes as 0. :-)

Meanwhile, what I proposed above for `build/column/floats/baselineattach` is about how bottom floats are attached to the main text (or the footnotes above): from the baseline or the bottom of the last line.

And, `\@outputbox@depth` should be conssitently updated (to 0) if bottom floats are attached in `\@outputbox@attachfloats`.

base/ltoutput.dtx
```diff
@@ -1461,6 +1508,17 @@
 %       If (short) footnotes are run as a paragraph this socket gets
 %         the plug \plug{para} assigned which is defined elsewhere. 
 %
+%    \item[\socket{build/column/footnotes/baselineattach} (0 arguments)]
+%
+%       By default footnotes are attached a distance of
```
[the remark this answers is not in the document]
FrankMittelbach: changed to:
```
%       By default, footnotes are attached after the last line of text
%       by adding a \cs{skip}\cs{footins}, i.e., the space is measured
%       from the bottom of \cs{@outputbox} (plug \plug{noop}).
```
better?

base/ltoutput.dtx
```diff
@@ -1461,6 +1508,17 @@
...
+%       \cs{skip}\cs{footins} (plug \plug{noop}). The socket can be
+%       changed to \plug{on}, in which case the skip starts at the
```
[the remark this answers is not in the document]
FrankMittelbach: yes not quite the right wording

base/ltoutput.dtx
```diff
@@ -2224,49 +2282,43 @@
...
 %    \end{macrocode}
-%    We first test if we are in a \cs{raggedbottom} layout. If not we
-%    do nothing, but we don't disable the code because
-%    \cs{raggedbottom} may get used only for some parts of the
-%    document.
-%    \begin{macrocode}
-  \ifx\@textbottom\relax \else
-%    \end{macrocode}
 %    In some special circumstances the \cs{@outputbox} might be
 %    void. As \cs{@outputbox@append} below would change this, we
```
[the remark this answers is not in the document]
FrankMittelbach: sorry, don't understand the suggestion

base/ltoutput.dtx
FrankMittelbach: rewritten

base/ltoutput.dtx
FrankMittelbach: changed

FrankMittelbach:
> Lots of ideas/comments, and a few question marks.

thanks. Made the changes, with 2-3 things to to look at if it is now better/clearer and one item I didn't understand the suggestion.

base/ltoutput.dtx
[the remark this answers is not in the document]
car222222: Neither do I. 
Probably meant for elsewhere??

base/ltoutput.dtx
```diff
@@ -2224,49 +2283,43 @@
...
-%    better handle this case and leave it unchanged in this situation,
-%    because otherwise we can't detect that situation any longer
+%    better handle this case by leaving it unchanged in this situation,
```
car222222, line 2293: Here:
we better  ==>  we had better

Better! ?

car222222 (approved):
Fixed one misplaced comment.

base/ltoutput.dtx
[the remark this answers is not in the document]
car222222: Yes, a great improvement!

base/ltoutput.dtx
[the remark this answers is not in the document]
FrankMittelbach: it might be on and off rather than on and noop (noop is automatically generated and it does the right thing, so I left it as that for now) and yes I'm not happy about the socket name either, but as I wrote right now it is just experimental and might change in functionality and name

base/ltoutput.dtx
FrankMittelbach: we had better handle ????  I think correct it is just  "we handle this case..."

car222222: OK: 
   we handle 
is good.

base/ltoutput.dtx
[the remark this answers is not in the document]
car222222: Understood.  OK!

base/ltoutput.dtx
CastleStar14654: Two simple examples, one in legacy mode and one in new default (`footnotes-floats`). I compile them using the latest format file generated from the branch of this PR.

<details>

<summary>Code for lagacy-mode example.</summary>

```latex
\documentclass{article}
\usepackage[pass,showframe]{geometry}

\begin{document}
Some texts.

Added a \verb|\vfill| below.
\vfill
Added a \verb|\vfill| above.

The \textbf{baseline} of this final line is aligned with the bottom of the \verb|\@outputbox| as expected.
\end{document}
```

</details>

image

<details>

<summary>Code for non-legacy example.</summary>

```latex
\documentclass{article}
\usepackage[pass,showframe]{geometry}

\AssignSocketPlug{build/column/outputbox}{footnotes-floats}

\begin{document}
Some texts.

Added a \verb|\vfill| below.
\vfill
Added a \verb|\vfill| above.

In this file, \verb|build/column/outputbox| socket \verb|footnotes-floats| is used.

The \textbf{bottom} of this final line is aligned with the bottom of the \verb|\@outputbox|, different from the expected behaviour.
\end{document}
```

</details>

image

What happens in the non-legacy version is exactly what the 2017/04/15 update tried to fix.

base/ltoutput.dtx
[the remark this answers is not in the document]
FrankMittelbach: :-)

base/changes.txt
FrankMittelbach: @car222222, if you add a comment/suggestion to a comment by somebody else whihc I have already handled and marked as "resolved" it helps a lot of you then also "unresolve" it, because then it is much easier to find.

Vice versa, if you are ok with a change made to an earlier comment from you, makring it "resolved" helps me to not spend a lot of time.

(not sure your interface has these buttons, but if it has ...) thanks

base/ltoutput.dtx
FrankMittelbach: can't really argue with those examples :-(

But if it is necessary to do this backup each time when `\@outputbox@reinsertbskip` is called and not `\relax` we might as well put it into the definition of that, I guess

CastleStar14654: Indeed.

Perhaps something like this in `\@outputbox@removebskip`:
```latex
\xdef\@outputbox@reinsertbskip
  {\noexpand\@outputbox@append{%
    \noexpand\@backup@outputbox@depth%
    \vskip\the\@tempskipa}}
```

(I add the `\noexpand` because I suppose that we would always want to remove the **current** `\@outputbox@depth`)

And define `\@outputbox@append` somewhere like
```latex
\def\@outputbox@append{
  \ifdim\@outputbox@depth>\z@
    \vskip -%
      \ifdim\@outputbox@depth>\maxdepth
        \maxdepth
      \else
        \@outputbox@depth
      \fi
  \fi
}
```

So that `\@outputbox@append` can also be invoked by `build/column/footnotes/baselineattach`.

FrankMittelbach: but in your MWE you aren't using the legacy plug; yes the others work with the redefinition but footnotes-floats-legacy would break imho

FrankMittelbach: the OR is a can of worms :-)  There is more wrong. `\maxdepth` can be zero (when a bottom float has been attached) This is why there is also `\@maxdepth` so the test and the skip should be against that. But unless I'm mistaken that setting of `\maxdepth`  when bottom floats are around never made sense (or if so only in the early days of 2.09).

CastleStar14654: Oh, so this `\maxdepth` of zero would influence all `\@backup@outputbox@depth` or footnotes below the bottom float... Indeed a mess.

As you mentioned, the float boxes would always have a depth of zero. It thus seems that the `\global \maxdepth \z@` line in `\@addtobot` is really unnecessary - the depth is already zero (if the float is really at the bottom of page).

FrankMittelbach: > Oh, so this `\maxdepth` of zero would influence all `\@backup@outputbox@depth` or footnotes below the bottom float... Indeed a mess.

I nearly overlooked it among all the tests that change because  they have to change. But I noticed the surprins `\glue 0.0` after all. Yes, eventually one should test against `\maxdepth` but for now I use `\@maxdepth` and update all test files (and verify them) and afterwards alter the `\maxdepth` and then there should be no change. This is really the most tricky part of LaTeX.

I think the  new logic will then be that `\maxdepth` will be the value used, but at the end of the OR it is reset to `\@maxdepth` as before. This way, one can alter maxdepth for a single page (inside a page hook)  to affect only one page

CastleStar14654: All of the latest changes works for me, except when `build/column/footnotes/baselineattach` is `on` in `footnotes-floats-legacy`, in the case of "manual `\newpage` with `\vfill` mid-page".

<details>
<summary>MWE using latest ltoutput without customized definitions; see page 5</summary>

```latex
\documentclass{article}
\usepackage[pass,showframe]{geometry}
\usepackage{lipsum,xcolor}

\newif\ifbotfl \botfltrue
% \botflfalse

\AssignSocketPlug{build/column/footnotes/baselineattach}{on}

% \AssignSocketPlug{build/column/outputbox}{space-footnotes-floats}
% \AssignSocketPlug{build/column/outputbox}{footnotes-space-floats}
% \AssignSocketPlug{build/column/outputbox}{floats-space-footnotes}
% \AssignSocketPlug{build/column/outputbox}{space-floats-footnotes}
% \AssignSocketPlug{build/column/outputbox}{floats-footnotes}
% \AssignSocketPlug{build/column/outputbox}{footnotes-floats}
\AssignSocketPlug{build/column/outputbox}{footnotes-floats-legacy}


% this rule indicates the starting location of \skip\footins
\newbox\fnhtbox
\newlength\fndp
\fndp=\dimexpr\footnotesep/7*3\relax
\setbox\fnhtbox=\hbox{%
  \color{red}
  \rule{1em}{\dimexpr\footnotesep+\skip\footins}%
  \rule[-\fndp]{1em}{\fndp}
}
\dp\fnhtbox=0pt
\ht\fnhtbox=0pt

% this rule indicates the range of \textfloatsep
\newlength\floatheight
\setlength{\floatheight}{30pt}
\newbox\floathtbox
\setbox\floathtbox=\hbox{\rule[-\textfloatsep]{1em}{\dimexpr\floatheight+2\textfloatsep}}
\dp\floathtbox=0pt
\ht\floathtbox=0pt

\newcommand\addfn{\footnote{\usebox\fnhtbox Sample footnote texts.}}
\fboxsep=-0.4pt
\fboxrule=0.4pt
\newcommand\addbotfl{
  \ifbotfl
    \begin{figure}[b]
      \hfil
      \usebox\floathtbox\framebox[0.8\linewidth]{\rule{0pt}{\floatheight}}
      \hfil
    \end{figure}
  \fi
}

\begin{document}
\section{Page 1: the normal page break}
\lipsum[1-3]\addfn
\addbotfl

\lipsum[4-5]

\section{Page 2: manual \texttt{\string\pagebreak}}
\lipsum[6]\addfn
\addbotfl

\lipsum[7] gggg\pagebreak

\section{Page 3: manual \texttt{\string\pagebreak} with \texttt{\string\vfill} mid-page}
\lipsum[6]\addfn
\addbotfl
\vfill
\lipsum[7] gggg\pagebreak

\section{Page 4: manual \texttt{\string\newpage}}
\lipsum[8]\addfn
\addbotfl

\lipsum[9] gggg\newpage

\section{Page 5: manual \texttt{\string\newpage} with \texttt{\string\vfill} mid-page}
\lipsum[8]\addfn
\addbotfl
\vfill
\lipsum[9] gggg\newpage

\end{document}
```
</details>

image

The reason might be that the `\@output@depth` is not updated to the latest value after `\@outputbox@removebskip`, but remains as 0 (due to the `\vfil` by `\newpage`).

(this is not a problem for `floats-footnotes` or `footnotes-floats`, because there `\@outputbox@reinsertbskip` happens at the end.)

Possibly update `\@output@depth` at the end of `\@outputbox@append`? (As well as the end of `\@outputbox@attachfloats`.)

base/ltoutput.dtx
```diff
@@ -1070,22 +1069,70 @@
...
-  \fi
+%  
+%    In 2017 this was fixed by doing the backing-up here. However, the
+%    problem with that is that if the page contains footnotes the
+%    \cs{skip}\cs{\footins} is measured from the baseline of the last
+%    line and not (as on all other pages) from the bottom of the last
+%    line. One can argue that measuring from the baseline is better,
+%    but if that is done it should be done everywhere not just on some
+%    pages.
+
```
CastleStar14654, line 1083: Is this blank line without `%` intended? This line implicitly adds a `\par` to `\newpage`, but just seceral lines above, a `\par` is explicitly removed.

base/ltoutput.dtx
CastleStar14654: By the way, in my local installation, the new `footnotes-floats` mode is not automatically invoked when I use `\DocumentMetadata`. I don't know whether this is because I do not update some other necessary packages to the dev version.

base/ltoutput.dtx
FrankMittelbach: no not intended, thanks

base/ltoutput.dtx
```diff
@@ -1046,7 +1046,7 @@
 %    would appear that this would be done by a |\leavevmode|.
 %    \begin{macrocode}
 %</2ekernel|fltrace>
-%<latexrelease>\IncludeInRelease{2017/04/15}%
+%<latexrelease>\IncludeInRelease{2025/06/01}%
```
CastleStar14654, line 1049: There is another possibly unintended blank line (`\par`) above in  `\clearpage` introduced recently in #1692 (Line 986) (though within the vertical mode, this `par` does no harm)

(cannot comment on that line within this PR)

base/ltoutput.dtx
FrankMittelbach: > The reason might be that the `\@output@depth` is not updated to the latest value after `\@outputbox@removebskip`, but remains as 0 (due to the `\vfil` by `\newpage`).

don't think that is the issue. What shuld happen in my opinion is that in \@outputbox@removebskip we should have:
```
                 \noexpand\@if@footnotes@TF
                     {\UseSocket{build/column/footnotes/baselineattach}}%
                     {\noexpand\@backup@outputbox@depth}%
```
to support that approach.

FrankMittelbach: > By the way, in my local installation, the new `footnotes-floats` mode is not automatically invoked when I use `\DocumentMetadata`. I don't know whether this is because I do not update some other necessary packages to the dev version.

works for me (with testphase=latest).  With just `\DocumentMetadata{}` it is not changing the socket plug. I guess it probably should. @u-fischer ?

CastleStar14654: I see.

You are using the fact that (1) after `\@outputbox@appendfootnotes` is called, `\@if@footnotes@TF` is always false, and the fact that (2) currently in all `build/column/outputbox` plugs (except for `footnotes-floats-legacy`), `@outputbox@removebskip` is only called when `\@if@footnotes@TF` is false.

In my opinion, this logic is a bit tricky, and `\@outputbox@depth` is not always reflecting the current depth of `\@outputbox`.

But at least, all the plugs provided in the kernel right now work perfectly, and I really cannot come up with any weird cases where your method won't work! :-)

CastleStar14654: (The following comment is not very relevant to the current PR. Just for your consideration)

I wonder why using `\vfill` rather than `\vfil` in these `build/column/outputbox` plugs? I briefly tested using [d307f28](https://github.com/latex3/latex2e/pull/1701/commits/d307f2888306e488e396d40a14b167a77ba2cc79) version, and it seems that they work as expected when `\vfill` exists in the main text.

(But `build/column/footnotes/baselineattach=on` breaks when there is `space` before `footnotes`)

FrankMittelbach: It would make sense to use `\vfil` rather than `\vfill` as you suggest as that gives the user more options to adjust a page. Not sure why I did that, perhaps it was this way in footmisc or I didn't think about it properly. But I guess it shuld be done in a separate PR and a separate issue going with it. I probably means that again a few hundred test files change :-(

FrankMittelbach: > You are using the fact that (1) after `\@outputbox@appendfootnotes` is called, `\@if@footnotes@TF` is always false, and the fact that (2) currently in all `build/column/outputbox` plugs (except for `footnotes-floats-legacy`), `@outputbox@removebskip` is only called when `\@if@footnotes@TF` is false.

yes

> In my opinion, this logic is a bit tricky, and `\@outputbox@depth` is not always reflecting the current depth of `\@outputbox`.

is it not? it is within `\@outputbox@append` at least that was the plan. Outside of that no, but there it is hard to maintain which is why I consider it being unspecified there (should document that properly I guess). If I understand your original suggestion correctly then that would be incorrect some several edge cases too. For example, after `\@outputbox@reinsertskip` the box depth is indeed zero and to pretend otherwise is fudging things to achieve a specific result.

But I think my last suggestion could be improved by moving the Socket into the -legacy plug

> But at least, all the plugs provided in the kernel right now work perfectly, and I really cannot come up with any weird cases where your method won't work! :-)

as I said all of this is a bigger brain twister than most people realize and it is easy to overlook somehting. But I think we are now where we want to be more or less

FrankMittelbach:
If you happen to look at the last commit @CastleStar14654 please ignore it, I must have been already asleep when I wrote that nonsense. But now it is too late and I wil fix/revert it tomorrow

base/ltoutput.dtx
CastleStar14654: ## Continuing the discussion on `\vfil` and `\vfill`

I made more tests, and conclude that changing all `\vfill` to `\vfil` in `build/column/outputbox` plugs works perfectly well, as long as the true branch of `\@if@footnotes@TF` is changed to `\@outputbox@append{\UseSocket{build/column/footnotes/baselineattach}\vfil}`.

For example:
```latex
\NewSocketPlug {build/column/outputbox}{space-footnotes-floats} {%
  \@if@footnotes@TF
      {\@outputbox@append{%
% NEW THING
        \UseSocket{build/column/footnotes/baselineattach}%
% NEW THING
        \vfil}}%
      {\@if@bottomfloats@TF
        {\@outputbox@append{\vfil}}%
        {\@outputbox@reinsertbskip}%
      }%
  \@outputbox@appendfootnotes
  \@outputbox@attachfloats
}
```

<details>
<summary>Attached is my test file</summary>

```latex
\documentclass{report}
\usepackage[pass,showframe]{geometry}
\usepackage{lipsum,xcolor,hyperref}

\pagestyle{headings}

% this rule indicates the starting location of \skip\footins
\newbox\fnhtbox
\newlength\fndp
\fndp=\dimexpr\footnotesep/7*3\relax
\setbox\fnhtbox=\hbox{%
  \color{red}
  \rule{1em}{\dimexpr\footnotesep+\skip\footins}%
  \rule[-\fndp]{1em}{\fndp}
}
\dp\fnhtbox=0pt
\ht\fnhtbox=0pt

% this rule indicates the range of \textfloatsep
\newlength\floatheight
\setlength{\floatheight}{30pt}
\newbox\floathtbox
\setbox\floathtbox=\hbox{\rule[-\textfloatsep]{1em}{\dimexpr\floatheight+2\textfloatsep}}
\dp\floathtbox=0pt
\ht\floathtbox=0pt

\newcommand\addfn{\ifaddfn
    \footnote{\usebox\fnhtbox Sample footnote texts.}
  \fi}
\fboxsep=-0.4pt
\fboxrule=0.4pt
\newcommand\addbotfl{
  \ifbotfl
    \begin{figure}[b]
      \hfil
      \usebox\floathtbox\framebox[0.8\linewidth]{\rule{0pt}{\floatheight}
        \vbox to \floatheight{
          \vfil
          This is a bottom float
          \vfil
        }}
      \hfil
    \end{figure}
  \fi
}

\newif\ifbotfl \botfltrue
\newif\ifaddfn \addfntrue

\newcommand{\testsuite}{
  \subsection{Page 1: the normal page break}
  \lipsum[1-3]\addfn
  \addbotfl

  \lipsum[4-5]

  \subsection{Page 2: manual \texttt{\string\pagebreak}}
  \lipsum[6]\addfn
  \addbotfl

  \lipsum[7] gggg\pagebreak

  \subsection{Page 3: manual \texttt{\string\pagebreak} with \texttt{\string\vfill} mid-page}
  \lipsum[6]\addfn
  \addbotfl
  \vfill
  \lipsum[7] gggg\pagebreak

  \subsection{Page 4: manual \texttt{\string\newpage}}
  \lipsum[8]\addfn
  \addbotfl

  \lipsum[9] gggg\newpage

  \subsection{Page 5: manual \texttt{\string\newpage} with \texttt{\string\vfill} mid-page}
  \lipsum[8]\addfn
  \addbotfl
  \vfill
  \lipsum[9] gggg\newpage
}

\newcommand{\largetestsuite}[1]{
  \chapter{#1}
  \AssignSocketPlug{build/column/outputbox}{#1}
  \newpage

  \section{Footnotes + bottom floats}
  \addfntrue\botfltrue
  \testsuite

  \section{Footnotes}
  \addfntrue\botflfalse
  \testsuite

  \section{bottom floats}
  \addfnfalse\botfltrue
  \testsuite

  \section{Nothing}
  \addfnfalse\botflfalse
  \testsuite
}

\begin{document}
  \part{baselineattach off}
  \largetestsuite{space-footnotes-floats}
  \largetestsuite{footnotes-space-floats}
  \largetestsuite{floats-space-footnotes}
  \largetestsuite{space-floats-footnotes}
  \largetestsuite{floats-footnotes}
  \largetestsuite{footnotes-floats}
  \largetestsuite{footnotes-floats-legacy}

  \AssignSocketPlug{build/column/footnotes/baselineattach}{on}
  \part{baselineattach on}
  \largetestsuite{space-footnotes-floats}
  \largetestsuite{footnotes-space-floats}
  \largetestsuite{floats-space-footnotes}
  \largetestsuite{space-floats-footnotes}
  \largetestsuite{floats-footnotes}
  \largetestsuite{footnotes-floats}
  \largetestsuite{footnotes-floats-legacy}

\end{document}
```
</details>

base/ltoutput.dtx
```diff
@@ -2592,11 +2716,41 @@
...
+%    even if the same code is executed a second time (in
+%    \cs{@outputbox@reinsertbskip} or in
+%    \cs{@outputbox@appendfootnotes}) it will not alter the box further.
+%  
+%  \changes{v1.4r}{2025/03/25}{Add support for attaching footnotes
+%    with a skip starting at the baseline of the last text line (footmisc/19)}
+%    \begin{macrocode}
+  \@outputbox@append{%
+      \UseSocket{build/column/footnotes/baselineattach}}%
   \@outputbox@reinsertbskip
```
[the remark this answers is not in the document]
CastleStar14654: The `\@if@bottomfloats@TF{\maxdepth\z@}{}` might not be necessary, if it is decided that "no `\maxdepth\z@`" is the expected behaviour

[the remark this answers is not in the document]
FrankMittelbach: > Just curious. Is `footnotes-floats-legacy` expected to behaviour the same as the previous LaTeX release?

not its subtle bugs I would say.
> 
> I just found that the latest version _cannot_ reproduce the true legacy version (I compare with the output of TeX Live 2024 on overleaf). One reason is that `\maxdepth\z@` done by bottom floats are no loger the case.

The result of page 4 in your test file is a bug in the old legacy code , i.e., the footnote separation should start from the bottom not the baseline and page 5 is a bug in the new code and should get fixed (you are really or for finding all the edge cases :-) thanks.)

> The correct `footnotes-floats-legacy` would be:

I don't think so, the legacy should not reintroduce the \maxdepth hack of 2.09.

I think what should happen is that `\@outputbox@reinsertbskip` becomes even smarter and does
```
               \noexpand\@outputbox@append{%
                 \noexpand\@if@footnotes@TF
                     {}%
                     {%
                      \noexpand\@if@bottomfloats@TF
                         {}%
                         {\noexpand\@backup@outputbox@depth}%
                     }%
                 \vskip\the\@tempskipa}}%
```
at least that's my current thinking.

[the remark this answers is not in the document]
FrankMittelbach: Actually it needs a little more, namely in the -legacy plug:
```
  \@if@footnotes@TF
    {\@outputbox@append{%
        \UseSocket{build/column/footnotes/baselineattach}}%
    }{}%
  \@outputbox@reinsertbskip
  \@outputbox@appendfootnotes
  \@outputbox@attachfloats
```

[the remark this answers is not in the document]
CastleStar14654: Makes sense to me

base/ltoutput.dtx
```diff
@@ -2592,11 +2722,43 @@
...
+%  
+%    Note that after this code has backed up, the depth of
+%    \cs{@outputbox} will be zero as there is a skip at its bottom. So
+%    even if the same code is executed a second time (in
+%    \cs{@outputbox@reinsertbskip} or in
+%    \cs{@outputbox@appendfootnotes}) it will not alter the box further.
+%  
+%  \changes{v1.4r}{2025/03/25}{Add support for attaching footnotes
+%    with a skip starting at the baseline of the last text line (footmisc/19)}
+  %    \begin{macrocode}
```
CastleStar14654, line 2757: This `\begin{macrocode}` line should be un-indented

base/ltoutput.dtx
FrankMittelbach: Come to think about it, this is even necessary if we stay with `\vfill` because there is still the possibility that somebody would do `\vspace{0pt plus 1filll}`, though granted that is not very likely.

Makes me wonder if the logic should change  completely and we provide just `build/column/baselineattach`and apply it to both footnotes and floats.

CastleStar14654: Or, also providing a `build/column/botfloats/baselineattach` for larger customization possibilities.

(Though it seems impossible that there would be anyone wanting to do baselineattach for footnotes only or bottom floates only.)

FrankMittelbach: as you say offering to separately set the baselineattach is kind of weird and the code is full of subtle traps then. In contrast switching always between  baseline attach and non baseline attach, kind of makes sense and also makes the logic much much simpler, e.g., through a `\@outputbox@appendspace`.

base/ltoutput.dtx
```diff
@@ -2667,7 +2667,7 @@
 %    \begin{macrocode}
 \NewSocketPlug {build/column/outputbox}{space-floats-footnotes} {%
   \@if@footnotes@TF
-      {\@outputbox@append{\vfill}}%
+      {\@outputbox@append{\UseSocket{build/column/footnotes/baselineattach}\vfill}}%
       {\@if@bottomfloats@TF
         {\@outputbox@append{\vfill}}%
         {\@outputbox@reinsertbskip}}%
```
CastleStar14654, line 2673: Here, it is necessary to test for bottom floats first, then footnotes.

(But this does not matter anyway, if `build/column/footnotes/baselineattach` will be changed to `build/column/baselineattach` later)

FrankMittelbach: right, the logic is wrong :-( thanks for spotting

base/ltoutput.dtx
```diff
@@ -1061,7 +1062,6 @@
     \global \@inlabelfalse
   \fi
   \if@nobreak \@nobreakfalse \everypar{}\fi
-  \par
```
CastleStar14654, line 1064: I know that this `\par` does nothing, because there is a `\vfil` following, but perhaps it is better to retain it? So that the `2025/06/01` version of `\newpage` is exactly the same as the `0000/00/00` version?

CastleStar14654: I see, thanks!

base/doc/ltnews41.tex
```diff
@@ -294,12 +295,31 @@ \section{Configurable output routine}
...
+
+For this reason there are now the sockets
+\socket{build/column/footnotes/baselineattach} and
+\socket{build/column/floats/baselineattach}. The can be assigned the
+plug \plug{on} in which case the attachment of footnotes/floats is
+measured from the baseline of the last text line. To mimic the
+previous behavior, they are by default both \plug{off}.  One probably
+wants to set them both to \plug{on} or both to \plug{off}, but for
+maximum flexibility the two can be individually changed.
+
```
davidcarlisle, line 317: I (think) I'd lean towards combining these.

davidcarlisle (approved):
looks good, but still possibly merge the baseline attach sockets

base/doc/ltnews41.tex
FrankMittelbach: it sure is possible to merge the too and in fact I'm also leaning towards that since everything else seems artifical, but without feedback I thought I stay with both.

PR #1707, opened 2025-04-03
Draft: Support replacing font families for scriptfont and scriptscriptfont

zauguin:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development

- [ ] Test file(s) added

### DIFF
--- a/base/ltfsstrc.dtx
+++ b/base/ltfsstrc.dtx
@@ -1122,6 +1122,7 @@
 % \changes{v1.0o}{1990/06/24}{\cs{Adding} tracing code.}
 % \changes{v1.0q}{1990/07/07}{Group number added to tracing.}
 % \changes{v2.1a}{1992/09/22}{Introduced \cs{tf@size} for math size.}
+% \changes{v?.?a}{2025/04/03}{Introduced \cs{transform@scriptfont} for script fonts.}
 %    \begin{macrocode}
 %<*trace>
   \ifnum \tracingfonts>\tw@
@@ -1145,11 +1146,13 @@
 %    \begin{macrocode}
   \pickup@font \let\textfont@name\font@name
 %    \end{macrocode}
-%    Same game for |\scriptfont| and |\scriptscriptfont|:
+%    Same game for |\scriptfont| and |\scriptscriptfont|, but here we additionally
+%    transform the fontname through |\transform@scriptfont| to allow different
+%    family names in math scripts.
 %    \begin{macrocode}
-  \xdef\font@name{\csname \string#2/\sf@size\endcsname}%
+  \xdef\font@name{\csname \transform@scriptfont {sf}{#2}/\sf@size\endcsname}%
   \pickup@font \let\scriptfont@name\font@name
-  \xdef\font@name{\csname \string#2/\ssf@size\endcsname}%
+  \xdef\font@name{\csname \transform@scriptfont {ssf}{#2}/\ssf@size\endcsname}%
   \pickup@font
 %    \end{macrocode}
 %    Then we append the new |\textfont|\ldots{} assignments to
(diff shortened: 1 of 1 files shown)


davidcarlisle (changes requested):
Looks good, obviously `v?.?` would need fixing. probably also a short note in ltnews.

davidcarlisle (approved):

base/ltfsstrc.dtx
```diff
@@ -1165,6 +1168,45 @@
...
+% \end{macro}
+%
+% \begin{macro}{\declare@scriptfont@mapping}
+%    |\declare@scriptfont@mapping| can be used to define replacement font families which will be used
+%    for math subscripts for subsubscripts when a specific font family is used for the main font.
+%    For this |\declare@scriptfont@mapping| takes five arguments: The first argument should be \texttt{sf}
+%    \texttt{ssf} to indicate if the replacement should be used for the scriptfont or the scriptscriptfont.
+%    Then  it expects the encoding and family name of the base font for which the mapping should be applied
+%    and the encoding and family of the font which will be used for the scriptfont.
+%    \begin{macrocode}
```
FrankMittelbach: not sure why this should be an internal 2e command. It seems to me rather a font setup declaration as any of the other ones, so should be \DeclareScriptfontMapping or something like that (one would need to check what others are around to make sure it fits.

Where should these declarations live? In font support packages? In the kernel? ...?

FrankMittelbach (approved):
somewhere between approve and request for changes; see comment

base/ltfsstrc.dtx
zauguin: Currently all expected uses are in `unicode-math` (or `lua-unicode-math`) but in theory they could also live in font support packages,

FrankMittelbach: ok, asked differently: suppose somebody comes up with a new set of math fonts. Who would need to provide the necessary mapping then and where should that live?

zauguin: The mappings should be created by the font author in their font support package.

zauguin: `\DeclareScriptfontMapping` would be fine for me. I couldn't find any existing commands describing a mapping between different font families and no mention of script fonts in command names either, so there is not too much concern about consistency. Existing commands dealing with script fonts refer to MathSizes, but that wouldn't fit here since it explicitly is not about sizes.

FrankMittelbach: Its the consistency with other NFSS related commands .. but too late for me to review that tonight

FrankMittelbach: > The mappings should be created by the font author in their font support package.

then it should be CamelCase I would say

FrankMittelbach: > Its the consistency with other NFSS related commands .. but too late for me to review that tonight

it belongs next to \DeclareMathSizes I would say, so `\DeclareMath<something>`

FrankMittelbach: My suggestion would be
```
\DeclareMathScriptfontMapping { base-encoding } { base-family } {sf or ssf}  { replace-encoding } {replace family}
```
or, assuming we always want both sf and ssf anyway, then perhaps better simply:
```
\DeclareMathScriptfontMapping { base-encoding } { base-family } 
         { sf-replace-encoding } {sf-replace family}
         { ssf-replace-encoding } {ssf-replace family}
```

Whatever, the final decision of the name, the command would need documentation in fntguide and a mentioning in the ltnews entry

zauguin: I changed the name and interface as suggested and added to fntguide.

base/doc/fntguide.tex
```diff
@@ -1039,6 +1039,28 @@ \subsection{Declaring math sizes}
...
+\end{decl}
+
+Declares that the font family \m{s-family} in encoding \m{s-enc} should
+be used in `script' position and \m{ss-family} in encoding \m{ss-enc}
+in `scriptscript' encoding should be used in `scriptscript' positions
+whenever \arg{t-family} is selected in encoding \m{t-enc} as a math font.
+
+If no such declaration is given for a \arg{t-enc}/\arg{t-family} pair then
+different sizes of the same font are used in all `text', `script', and
+`scriptscript' positions.
```
FrankMittelbach, line 1057: maybe add:

positions as specified through a \DeclareMathSizes declaration.

base/doc/ltnews42.tex
```diff
@@ -155,6 +155,23 @@ \subsection{Normalizing key names for block environments}
 
 \section{New or improved commands}
 
+\subsection{Support separate font families for script fonts}
+
+In \TeX's math processing separate fonts can be selected for text, script and
+scriptscript sizes. \LaTeX's NFSS traditionally uses this to load three sizes
+of the same font, handling adjustment needed for making fonts appear better in
+a script location through the use of optical sizes. This works great for traditional
```
FrankMittelbach: \LaTeX's NFSS traditionally uses the same font family at different sizes, handling ...

base/doc/ltnews42.tex
```diff
@@ -155,6 +155,23 @@ \subsection{Normalizing key names for block environments}
...
+\TeX{} fonts, but for OpenType fonts this leads to issues. OpenType MATH assumes
+the font in a script location has separate features set and therefore received specific
+adjustments.
+
+To support this without relying on heuristics based on the font size, a new command
+\verb+\DeclareMathScriptfontMapping+ has been added. It can be called with three
+encodings and faily names to indicate that whenever the first family is selected as a
+math font, the second and thirs should be used as a script respectively scriptscript font.
```
FrankMittelbach: It takes 3 pairs of encoding/family arguments to indicate that for the first pair when used as the math main font the second and the third should be used as the script and scriptscript font, respectively.

base/doc/fntguide.tex
zauguin: I'm a bit concerned that this might make it sound too much like `\DeclareMathSizes` is *only* relevant if `\DeclareMathScriptfontMapping` is not used. While it actually selects the sizes either way.

base/doc/fntguide.tex
```diff
@@ -1039,6 +1039,30 @@ \subsection{Declaring math sizes}
...
+In either case, the sizes are specified through a |\DeclareMathSizes|
+declaration.
```
zauguin, line 1059: Maybe adding this makes it clearer?

base/doc/fntguide.tex
FrankMittelbach: yes, forget it. you other suggestion is better.

base/doc/fntguide.tex
FrankMittelbach: yes

PR #1714, opened 2025-04-12
Gh1712: add missing \par to \newpage

u-fischer:
I think this is simply a correction of the previous PR in develop and so didn't change version/date of  ltoutput.dtx is not change and added not changes entry, but could do it if wanted ...

## Status of pull request
- Ready to merge

- [x] Test file(s) added
- [ ] Version and date string updated in changed source files: no, see above
- [ ] Relevant `\changes` entries in source included: ditto
- [ ] Relevant `changes.txt` updated: ditto


base/ltoutput.dtx
```diff
@@ -1089,7 +1089,7 @@
 %    moved this code to the newly introduced output routine sockets.
 %  \changes{v1.4q}{2025/03/20}{Remove skip added in 2.4a (footmisc/19)}
 %    \begin{macrocode}
-%  \par
```
FrankMittelbach, line 1092: This needs to have a clear explanation why it is necessary even if we are already in vmode because that is not at all obvious with the normal meaning of \par.

FrankMittelbach: Also needs changes and v/date change imo

FrankMittelbach (changes requested):

base/ltoutput.dtx
car222222: Agreed that it needs explanation.

u-fischer: done.

FrankMittelbach (approved):

PR #1719, opened 2025-04-25
Gh1717

davidcarlisle:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [X] Test file(s) added


u-fischer:
The testfile show page 2 instead of 1:

~~~~
macro:->{}{2}{}{}{}
~~~~

A real label would probably have to substract 1?

davidcarlisle:
> The testfile show page 2 instead of 1:
> 
> ```
> macro:->{}{2}{}{}{}
> ```
> 
> A real label would probably have to substract 1?

hmm, you only asked for a label, not one with a correct value:-)  I should probably manually adjust the test file to 1, then look tomorrow where it would be safe t adjust the page count.

u-fischer:
> hmm, you only asked for a label, not one with a correct value:-)

Sorry ;-) I saw the number an wondered. Something like this seems to work:

~~~
\AddToHook{enddocument/afterlastpage}
 {\addtocounter{page}{-1}\RecordProperties{blub}{abspage,page}\label{lastpage}}
~~~

davidcarlisle:
> > hmm, you only asked for a label, not one with a correct value:-)
> 
> Sorry ;-) I saw the number and wondered. Something like this seems to work:
> 
> ```
> \AddToHook{enddocument/afterlastpage}
>  {\addtocounter{page}{-1}\RecordProperties{blub}{abspage,page}\label{lastpage}}
> ```

I don't think the number should be reduced every time it is accessed, I think at the point the `\write` is switched to `\immediate\write` in the OR, we need to adjust the value of the page counter

davidcarlisle:
we could decrement the page counter somewhere around the `%<<<<` comments but it depends a bit exactly
how we document the `afterlastpage`  and `enddocument/info` hooks, what value we expect there.

```
\def\enddocument{%
   \@kernel@before@enddocument
   \UseOneTimeHook{enddocument}%
   \@kernel@after@enddocument
   \@checkend{document}%
   \clearpage                                %<<<<  page counter is one more than last page from here on.
   \@kernel@before@enddocument@afterlastpage %<<<<  we switch to immediate writes here
   \UseOneTimeHook{enddocument/afterlastpage}%
   \@kernel@after@enddocument@afterlastpage
   \begingroup
     \if@filesw
       \immediate\closeout\@mainaux
       \let\@setckpt\@gobbletwo
       \let\@newl@bel\@testdef
       \let\new@label@record\@kernel@new@label@record@testdef
       \@tempswafalse
       \makeatletter \@@input\jobname.aux
     \fi
     \UseOneTimeHook{enddocument/afteraux}%
     \UseOneTimeHook{enddocument/info}%
   \endgroup
   \UseOneTimeHook{enddocument/end}%
   \deadcycles\z@\@@end}
```

base/ltshipout.dtx
```diff
@@ -1789,11 +1789,13 @@
 %    that condition and also check if
 %    \cs{@kernel@after@shipout@lastpage} contains any code. If both
 %    are empty we omit the page generation.
+%    \changes{v1.0p}{2025/04/27}{Increment page counter gh/1717}
 %    \begin{macrocode}
           \bool_lazy_and:nnF
             { \hook_if_empty_p:n {shipout/lastpage} }
             { \tl_if_empty_p:N \@kernel@after@shipout@lastpage }
             {
+              \advance\c@page\@ne
```
FrankMittelbach, line 1798: Might be correct to change locally here but if so it probably warrants a comment

FrankMittelbach (commented):
Too difficult to review with just a phone at hand so only a comment

u-fischer (approved):

PR #1583, opened 2024-12-13
Series changes (a version of #1396)

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

## Checklist of required changes before merge will be approved
- [n/a] Test file(s) added
- [x] Version and date string updated in changed source files
- [x] Relevant `\changes` entries in source included
- [x] Relevant `changes.txt` updated
- [x] Rollback provided (if necessary)?
- [x] `ltnewsX.tex` (and/or `latexchanges.tex`) updated

This is a slightly adjusted version of #1396 because that pull request would be difficult to merge after all several changes to ltfssaxes.dtx. It contains the pull request #1582 as the two are somewhat interwined.

PR #1727, opened 2025-05-12
Minor followup to #1583

user227621:
This PR is a minor followup to #1583:
- Correct a wrong GitHub number in two `\changes` entries
- Split the entries for `m?` and `?m` into individual `macrocode` blocks, analogous to the other entries
- Move one block for `m` weight rules between the blocks for `sl` weight rules and `sb` weight rules
- A rewording
- Unification of spaces

# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/ltfssaxes.dtx
+++ b/base/ltfssaxes.dtx
@@ -1841,6 +1841,9 @@
 \DeclareFontSeriesChangeRule {ulx}{m?}  {x}   {}
 \DeclareFontSeriesChangeRule {ulex}{m?} {ex}  {}
 \DeclareFontSeriesChangeRule {ulux}{m?} {ux}  {}
+%    \end{macrocode}
+%
+%    \begin{macrocode}
 \DeclareFontSeriesChangeRule {eluc}{m?} {uc}  {}
 \DeclareFontSeriesChangeRule {elec}{m?} {ec}  {}
 \DeclareFontSeriesChangeRule {elc}{m?}  {c}   {}
@@ -1850,6 +1853,9 @@
 \DeclareFontSeriesChangeRule {elx}{m?}  {x}   {}
 \DeclareFontSeriesChangeRule {elex}{m?} {ex}  {}
 \DeclareFontSeriesChangeRule {elux}{m?} {ux}  {}
+%    \end{macrocode}
+%
+%    \begin{macrocode}
 \DeclareFontSeriesChangeRule {luc}{m?}  {uc}  {}
 \DeclareFontSeriesChangeRule {lec}{m?}  {ec}  {}
 \DeclareFontSeriesChangeRule {lc}{m?}   {c}   {}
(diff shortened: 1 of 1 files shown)


user227621:
Looking at the rules for `m?` and `?m`, I wonder if we should specify `m` as the fourth argument for all rules where this is not already the result series. Otherwise, if the result series does not exist, `m?` would be tried as the series value, which cannot work. This can already happen with Computer Modern:

```latex
\documentclass{article}

\begin{document}

\sffamily

\fontseries{sbc}\selectfont Text \fontseries{m?}\selectfont Text

\end{document}
```

gives strange warnings:

```
LaTeX Font Warning: Font shape `OT1/cmss/c/n' undefined
(Font)              using `OT1/cmss/m?/n' instead on input line 7.


LaTeX Font Warning: Font shape `OT1/cmss/m?/n' undefined
(Font)              using `OT1/cmss/m/n' instead on input line 7.
```

FrankMittelbach:
I agree with the idea of adding "m" in the forth argument; m? or ?m does resolve to m with the normal settings but it is certainly  inefficient to make an additional test when that one can't work, only to come to m in the end.

But I'm not going to merge this PR prior to the June release because we just had our last -dev release and this isn't a hot fix due to a bug that needs fixing.

base/ltfssaxes.dtx
```diff
@@ -261,10 +261,10 @@
 % \end{itemize}
 %
 % \changes{1.0k}{2024/12/13}{Minor modifications to a few
-%   \cs{DeclareFontSeriesChangeRule} entries (gh/1396)}
+%   \cs{DeclareFontSeriesChangeRule} entries (gh/1583)}
```
FrankMittelbach, line 264: please don't change existing change entries but provide additional ones with the right version and date (because they reflect what was the situation in 2024 and now)

one or two more below as well.

user227621: This `\changes` entry was incorrect because #1396 is the number of a PR that was not merged. The actual PR was #1583. It's okay to correct an incorrect `\changes` entry, isn't it?

I have added `\changes` entries for the changes made with this PR.

user227621:
@FrankMittelbach I have implemented the addition of `m` as fourth argument for `m?`/`?m` rules where `m` is not already the result series (according to [your comment](https://github.com/latex3/latex2e/pull/1727#issuecomment-2872616910)) and added changes entries.

Is an ltnews entry necessary? Most changes only affect the documentation, i.e. `source2e.pdf`, but not the format itself, and the additional `m` in the forth argument of some rules prevents some strange warnings about font substitutions but does not (with the normal settings) affect typesetting. If so, it seems I cannot (easily) add such an entry, because there is no file `ltnews42.tex` in my branch (I started that branch before the latest release in June).

FrankMittelbach:
> Is an ltnews entry necessary? Most changes only affect the documentation, i.e. `source2e.pdf`, but not the format itself, and the additional `m` in the forth argument of some rules prevents some strange warnings about font substitutions but does not (with the normal settings) affect typesetting. If so, it seems I cannot (easily) add such an entry, because there is no file `ltnews42.tex` in my branch (I started that branch before the latest release in June).

I think an ltnews entry would be good, but I can do that if it is too complicated for you. Though, you can simply merge current develop into your branch and ltnews42  should be there for you to add to.

base/ltfssaxes.dtx
FrankMittelbach: ah, ok. I thought it was an PR that was merged.

FrankMittelbach:
please mark me for (re-)review once you think you are done

user227621:
@FrankMittelbach Should a rollback have been provided (for the additional alternative font series values)?

FrankMittelbach:
> @FrankMittelbach Should a rollback have been provided (for the additional alternative font series values)?

I would say no. They add missing rules that otherwise lead to wrong processing. If somebody previously added their own rules those rules would still work and overwrite what is now in place.

PR #1757, opened 2025-06-04
Mark 'historical' packages in tools

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/doc/ltnews42.tex
+++ b/base/doc/ltnews42.tex
@@ -167,13 +167,13 @@
 
 The \pkg{tools} bundle contains a range of packages with different usage
 profiles. Some of these were necessary in the transition from \LaTeX{}~2.09 to
-\LaTeXe{}, while others are very widely used in up-to-date documents (for
-example \pkg{array}). We have marked a small number of packages in \pkg{tools}
-as retained for historical and stability reasons, and where relevant pointed
-to more up-to-date alternatives:
+\LaTeXe{}, while others are very widely used in current documents (for example
+\pkg{array}). We have therefore marked a small number of packages in
+\pkg{tools} as \emph{retained only for historical and stability reasons}, and
+where relevant pointed to more up-to-date alternatives; the list is:
 \begin{itemize}
-  \item \pkg{rawfonts}: retained as part of \LaTeX{}~2.09 support
   \item \pkg{enumerate}: use \pkg{enumitem} instead
+  \item \pkg{rawfonts}: retained as part of \LaTeX{}~2.09 support
   \item \pkg{somedefs}: retained as part of \LaTeX{}~2.09 support
   \item \pkg{theorem}: use \pkg{amsthm} instead
   \item \pkg{verbatim}: use \pkg{fancyvrb} instead
--- a/base/doc/ltnews42.tex
+++ b/base/doc/ltnews42.tex
@@ -161,7 +161,23 @@
 
 %\section{Changes to packages in the \pkg{graphics} category}
 
-%\section{Changes to packages in the \pkg{tools} category}
+\section{Changes to packages in the \pkg{tools} category}
+
+\subsection{Updating the status of some components}
+
+The \pkg{tools} bundle contains a range of packages with different usage
+profiles. Some of these were necessary in the transition from \LaTeX{}~2.09 to
+\LaTeXe{}, while others are very widely used in up-to-date documents (for
+example \pkg{array}). We have marked a small number of packages in \pkg{tools}
+as retained for historical and stability reasons, and where relevant pointed
+to more up-to-date alternatives:
+\begin{itemize}
+  \item \pkg{rawfonts}: retained as part of \LaTeX{}~2.09 support
+  \item \pkg{enumerate}: use \pkg{enumitem} instead
+  \item \pkg{somedefs}: retained as part of \LaTeX{}~2.09 support
+  \item \pkg{theorem}: use \pkg{amsthm} instead
+  \item \pkg{verbatim}: use \pkg{fancyvrb} instead
+\end{itemize}
 
 %\section{Changes to files in the \pkg{cyrillic} category}
 
--- a/required/tools/changes.txt
+++ b/required/tools/changes.txt
@@ -8,7 +8,7 @@ are not part of the distribution.
 
 2025-06-04  Joseph Wright  <Joseph.Wright@latex-project.org>
 
-	* enumerate.dtx, rawfonts.dtx, theorem.dtx, verbatim.dtx
+	* enumerate.dtx, rawfonts.dtx, somedefs.dtx, theorem.dtx, verbatim.dtx
 	Add 'historical' notice
 	* tools-overview.tex
 	Add 'historical' markers
--- a/required/tools/changes.txt
+++ b/required/tools/changes.txt
@@ -5,6 +5,14 @@ completeness or accuracy and it contains some references to files that
 are not part of the distribution.
 =======================================================================
 
+
+2025-06-04  Joseph Wright  <Joseph.Wright@latex-project.org>
+
+	* enumerate.dtx, rawfonts.dtx, theorem.dtx, verbatim.dtx
+	Add 'historical' notice
+	* tools-overview.tex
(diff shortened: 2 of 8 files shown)


davidcarlisle:
I think this will need a news item (and `\pkg` needs defining:-)

required/tools/enumerate.dtx
```diff
@@ -62,6 +62,16 @@
 % \changes{v3.00}{1999/03/05}{Add extension hook for tools/2916}
 % \changes{v3.00}{2015/07/23}{Documentation fixes tools/2916}
 %
+% \changes{v3.00}{2025-06-04}{Add notice pointing to \pkg{enumitem}}
+% \begin{flushleft}
+%   \bfseries
+%   This package is retained in the \LaTeX{} \pkg{tools} bundle for
+%   stability reasons. Whilst bug fixes will be applied to \pkg{enumerate},
+%   no \emph{new} features will be considered. For new material, the
+%   \LaTeX{} team recommend that \pkg{enumitem} is used in place of
```
FrankMittelbach: while a team is usually more than one person I still think of it as singular ... is that different in English?

You are correct, . . . BUT . . . see below,
Chris

required/tools/tools-overview.tex
```diff
@@ -142,7 +142,8 @@
   Adds \emph{large delimiters} around arrays. Requires array.
 }%
 \entry{enumerate}{%
-  Extended version of the enumerate environment.
+  Extended version of the enumerate environment; historical, use
+  \pkg{enumtem} instead.
```
FrankMittelbach: enumitem

FrankMittelbach (changes requested):

required/tools/enumerate.dtx
car222222: Interesting point!
Yes, "team" is singular but . . . 
Expressions like this are very common: 
"The team all agree with you!"
Very strict (even more fussy than I am :-) 
grammar rules would state that this should be:
"All the members of the team agree with you!"

Without the "all" there, as in the actual text here, I think that treating it as singular is better:
"The team agrees with you!"

josephwright: I guess I think of it as 'LaTeX team' = 'a collection of individuals' but 'LaTeX Project' = 'a corporate author', which is why I phrased as I did. But if we want to change, I'll sort.

josephwright:
> I think this will need a news item (and `\pkg` needs defining:-)

What would you want in the news? I don't think this is anything new as such.

FrankMittelbach:
> What would you want in the news? I don't think this is anything new as such.

I think it wouldn't hurt to explicitly call out the package names that we suggest not to use in new documents (and what to use instead). Alter Kaffee, yes, but how many people know? And if there is a news entry there is something to point to (and a number of people read it).

required/tools/enumerate.dtx
FrankMittelbach: one team recommends two teams recommend (that is what my gut feeling says :-) regardless of consisting of individuals

car222222: Yes, as I tried to,explain above!
If @josephwright wishes to imply that the team is unanimous and we all, then better to be clearer, by using the format: ". . . team all recommend that . . ."

base/doc/ltnews42.tex
```diff
@@ -161,7 +161,23 @@ \section{Bug fixes}
 
 %\section{Changes to packages in the \pkg{graphics} category}
 
-%\section{Changes to packages in the \pkg{tools} category}
+\section{Changes to packages in the \pkg{tools} category}
+
+\seubsection{Updating the status of some components}
```
davidcarlisle:
```suggestion
\seubsection{Updating the status of some components}
```
```suggestion
\subsection{Updating the status of some components}
```

davidcarlisle (approved):
approve modulo fixes in doc

base/doc/ltnews42.tex
josephwright: Already fixed :)

FrankMittelbach (approved):
looks fine to me

josephwright:
@car222222 You happy with the news text here?

base/doc/ltnews42.tex
```diff
@@ -161,7 +161,23 @@ \section{Bug fixes}
...
-%\section{Changes to packages in the \pkg{tools} category}
+\section{Changes to packages in the \pkg{tools} category}
+
+\subsection{Updating the status of some components}
+
+The \pkg{tools} bundle contains a range of packages with different usage
+profiles. Some of these were necessary in the transition from \LaTeX{}~2.09 to
+\LaTeXe{}, while others are very widely used in up-to-date documents (for
+example \pkg{array}). We have marked a small number of packages in \pkg{tools}
+as retained for historical and stability reasons, and where relevant pointed
```
car222222: Probably better:
as \emph{retained only for . . . reasons}
(Or some such distinguisher!)

base/doc/ltnews42.tex
```diff
@@ -161,7 +161,23 @@ \section{Bug fixes}
...
 %\section{Changes to packages in the \pkg{graphics} category}
 
-%\section{Changes to packages in the \pkg{tools} category}
+\section{Changes to packages in the \pkg{tools} category}
+
+\subsection{Updating the status of some components}
+
+The \pkg{tools} bundle contains a range of packages with different usage
+profiles. Some of these were necessary in the transition from \LaTeX{}~2.09 to
+\LaTeXe{}, while others are very widely used in up-to-date documents (for
```
car222222: Do you mean? : 
. . . but are no longer needed, while . . . 

And, maybe:
up-to-date  ==>  current

base/doc/ltnews42.tex
```diff
@@ -161,7 +161,23 @@ \section{Bug fixes}
...
+example \pkg{array}). We have marked a small number of packages in \pkg{tools}
+as retained for historical and stability reasons, and where relevant pointed
+to more up-to-date alternatives:
+\begin{itemize}
+  \item \pkg{rawfonts}: retained as part of \LaTeX{}~2.09 support
+  \item \pkg{enumerate}: use \pkg{enumitem} instead
```
car222222: Maybe use alphabetical order here too?

required/tools/changes.txt
```diff
@@ -5,6 +5,14 @@ completeness or accuracy and it contains some references to files that
 are not part of the distribution.
 =======================================================================
 
+
+2025-06-04  Joseph Wright  <Joseph.Wright@latex-project.org>
+
+	* enumerate.dtx, rawfonts.dtx, theorem.dtx, verbatim.dtx
```
car222222: This list should contain also  "somedefs.dtx"!

base/doc/ltnews42.tex
```diff
@@ -161,7 +161,23 @@ \section{Bug fixes}
...
 
-%\section{Changes to packages in the \pkg{tools} category}
+\section{Changes to packages in the \pkg{tools} category}
+
+\subsection{Updating the status of some components}
+
+The \pkg{tools} bundle contains a range of packages with different usage
+profiles. Some of these were necessary in the transition from \LaTeX{}~2.09 to
+\LaTeXe{}, while others are very widely used in up-to-date documents (for
+example \pkg{array}). We have marked a small number of packages in \pkg{tools}
```
car222222: Better:
We have therefore marked ...

base/doc/ltnews42.tex
```diff
@@ -161,7 +161,23 @@ \section{Bug fixes}
...
+as retained for historical and stability reasons, and where relevant pointed
+to more up-to-date alternatives:
```
car222222: Add something like this: 
. . . alternatives. The list is:

car222222 (approved):
Several small things, some more important than others (probably).

base/doc/ltnews42.tex
josephwright: I've done 'current', but I'm not sure about 'no longer needed' - they are still needed for 2.09 sources.

PR #1771, opened 2025-06-11
modify dated amsmath file so it works when rolled back from current formats

davidcarlisle:
## Status of pull request

- Ready to merge

- [X] Test file(s) added

I modified the banner string but left date (needed for rollback) and version number (for historical accuracy) in the dated file.


required/amsmath/amsmath-2018-12-01.sty
```diff
@@ -27,7 +27,7 @@
 %%
 \NeedsTeXFormat{LaTeX2e}% LaTeX 2.09 can't be used (nor non-LaTeX)
 [1994/12/01]% LaTeX date must be December 1994 or later
-\ProvidesPackage{amsmath}[2018/12/01 v2.17b AMS math features]
+\ProvidesPackage{amsmath}[2018/12/01 v2.17b AMS math features (updated for rollback)]
```
josephwright, line 30: I'd have preferred a version change: would help in the tests!

josephwright (approved):

required/amsmath/amsmath-2018-12-01.sty
FrankMittelbach: In this particular case on could change the version (but not the date) because amsmath goes
```
\DeclareCurrentRelease{}{2019-04-01}
```
i.e., it does not allow you to roll back to a version but only to a date. However, I don't think it is making a good example if we do. In other packages there will be both a version string as well as a date and so both are really part of the rollback syntax and you would affect rollback in documents if you change either of them. (same reason why can't bump the \CurrentVersion date up).

josephwright: Awkward but I see the point - otherwise we good to go here?

davidcarlisle: > I'd have preferred a version change: would help in the tests!

it would help us perhaps but I think it's confusing to roll back to a version that didn't previously exist (even if that's what this is, really)

FrankMittelbach: I don't see that we are rolling back to a different version or date here. We are rolling back to the version current on that date (only we have corrected how this is happening). If You install a TL from that date/year it would say 2.17b and all we did is to adjust the rollback so the version as it was then comes back to life.

so in my opinion it is as it should be and the comment in the \ProvidesPackage is really the correct way to handle it.

required/amsmath/amsmath-2018-12-01.sty
```diff
@@ -166,9 +166,11 @@ For additional information on amsmath, use the \lq ?\rq\space option.%
 \let\negthinspace\!
 \renewcommand{\:}{\tmspace+\medmuskip{.2222em}}
 \let\medspace\:
+\let\negmedspace\@undefined % allow rollback (gh/1770)
 \newcommand{\negmedspace}{\tmspace-\medmuskip{.2222em}}
 \renewcommand{\;}{\tmspace+\thickmuskip{.2777em}}
 \let\thickspace\;
+\let\negthickspace\@undefined % allow rollback (gh/1770)
```
FrankMittelbach, line 177: perhaps it would be clearer if all these are moved up front with a comment there, or, given that the change in lines 2680 can't really be move just add at the top off the file something like
```
%%
%% Minor adjustments for the original file to allow rollback -- search for (gh/1770)
%% 
```

davidcarlisle: done

josephwright:
Ah, don't we need to rebase onto `main`?

FrankMittelbach:
> Ah, don't we need to rebase onto `main`?

yes we do, as it was branched of dev. but I thought you said you do that?

PR #1772, opened 2025-06-13
Fix amsmath timing for some packages in latex-lab-mathpkg

mbertucci47:
Updates latex-lab-mathpkg to load amsmath before packages that error otherwise, except in the case of `grmath` where the package loses its functionality if amsmath is loaded after. Even for packages that aren't math-specific like cleveref, I think it makes sense to do this here since latex-lab-mathpkg is only loaded when math tagging is loaded, which is when the issue occurs since it loads amsmath at begindocument.

Note this is already done for the `cases` package.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added


mbertucci47:
Changes entry added

PR #1796, opened 2025-06-25
Extend \ProcessKeyOptions to support multiple families

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [x] Test file(s) added

### DIFF
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -10,6 +10,10 @@ not part of the distribution.
 	* ltcmd.dtx, lthooks.dtx
 	Support \showstream (gh/1062)
 
+2025-06-25  Joseph Wright  <Joseph.Wright@latex-project.org>
+	* ltkeys.tex, clsguide.tex
+	Support for multiple key families in \ProcessKeyOptions (gh/1756)
+
 2025-06-24  Joseph Wright  <Joseph.Wright@latex-project.org>
 	* usrguide.tex
 	Add c-type arg. to overview (gh/1761)
--- a/base/changes.txt
+++ b/base/changes.txt
@@ -11,7 +11,7 @@ not part of the distribution.
 	Support \showstream (gh/1062)
 
 2025-06-25  Joseph Wright  <Joseph.Wright@latex-project.org>
-	* ltkeys.tex, clsguide.tex
+	* ltkeys.dtx, clsguide.tex
 	Support for multiple key families in \ProcessKeyOptions (gh/1756)
 
 2025-06-24  Joseph Wright  <Joseph.Wright@latex-project.org>
--- a/base/ltkeys.dtx
+++ b/base/ltkeys.dtx
@@ -62,8 +62,10 @@
 %     \cs{DeclareKeys} \oarg{family} \marg{declarations}
 %   \end{syntax}
 %   Creates a series of options from a comma-separated \meta{declarations} list.
-%   The \meta{family} is a namespace for the keys.
-%   Each entry in this list is a key--value pair, with the \meta{key} having one
+%   The \meta{family} is a namespace for the keys; if not given, the basename of
+%   the current class or package is used.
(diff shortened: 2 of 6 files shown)


josephwright:
Views on whether a news entry is needed welcome

base/changes.txt
```diff
@@ -6,6 +6,10 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2025-06-25  Joseph Wright  <Joseph.Wright@latex-project.org>
+	* ltkeys.tex, clsguide.tex
```
FrankMittelbach: ltkeys.dtx

base/ltkeys.dtx
```diff
@@ -62,6 +62,7 @@
 %     \cs{DeclareKeys} \oarg{family} \marg{declarations}
 %   \end{syntax}
 %   Creates a series of options from a comma-separated \meta{declarations} list.
+%   The \meta{family} is a namespace for the keys.
 %   Each entry in this list is a key--value pair, with the \meta{key} having one
```
FrankMittelbach: family is a name space and it is talking about "a list" but not saying to what this refers to

base/testfiles/github-1756.lvt
```diff
@@ -0,0 +1,25 @@
...
+\ProvidesExplPackage{testpackage}{2025-06-25}{}{}
+\keys_define:nn { testpackage-a }
+  {
+    testoption-a .code:n = { \def \testmacroa { seen } } ,
+    testoption-b .code:n = { \def \testmacrob { seen } }
+  }
+\keys_define:nn { testpackage-b }
+  {
+    testoption-a .code:n = { \def \testmacroa { seen } } ,
+    testoption-c .code:n = { \def \testmacroc { seen } }
```
FrankMittelbach: maybe better use typeout and different text per option so that one sees which options are actually executed, ie. -a twice or only one of them etc.

FrankMittelbach (changes requested):
some suggestions and yes I think an ltnews entry is good

base/ltkeys.dtx
josephwright: I've revised a bit - hope this helps.

base/testfiles/github-1756.lvt
FrankMittelbach: "resolved" because you think it is not helpful or why?

FrankMittelbach: ah you did, stange why is it not showing here

base/ltkeys.dtx
```diff
@@ -120,12 +123,12 @@
...
-%     \cs{ProcessKeyOptions} \oarg{family}
+%     \cs{ProcessKeyOptions} \oarg{families}
 %   \end{syntax}
 %   The \cs{ProcessKeyOptions} function is used to check the current
-%   option list against the keys defined for \meta{family}. Global (class)
-%   options and local (package) options are checked when this function
-%   is called in a package.
+%   option list against the keys defined for \meta{families} (a comma-separated
+%   list). Global (class) options and local (package) options are checked when
+%   this function is called in a package.
```
davidcarlisle: probably should say what happens if the key is defined in more than one family, I think it is just passing on the unused keys to test against the later families so the first family that defines the key wins?

josephwright: I'll update the comment - each family gets all the keys it knows, so the last one will 'win' (Assuming of course they even do similar things - they could all successfully process a common key.)

davidcarlisle: ah so you process them all, so I did misread the code:-) definitely that should be explicitly stated

josephwright: See if you are happy with the edit

PR #1798, opened 2025-06-25
Support \showstream

josephwright:
# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/ltcmd.dtx
+++ b/base/ltcmd.dtx
@@ -34,8 +34,8 @@
 %%% From File: ltcmd.dtx
 %
 %    \begin{macrocode}
-\def\ltcmdversion{v1.3d}
-\def\ltcmddate{2025-04-21}
+\def\ltcmdversion{v1.33}
+\def\ltcmddate{2025-06-26}
 %    \end{macrocode}
 %
 %<*driver>
@@ -2798,6 +2798,7 @@
 %         {Make \cs{ShowCommand} stop for interaction}
 % \changes{v1.1b}{2022/11/30}
 %         {Don't stop for the \cs{begin} part of an environment}
+% \changes{v1.3e}{2025-06-26}{Support \cs{showstream} (gh/1062)}
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_show_environment:N #1
   {
@@ -2819,8 +2820,13 @@
           {#2}
       }
   }
-\cs_new_protected:Npn \@@_show:x #1
-  { \iow_term:x { > ~ #1 . \iow_newline: } }
+\cs_new_protected:Npe \@@_show:x #1
+  {
+    \cs_if_exist:NTF \iow_show:e
+      { \iow_show:e }
+      { \iow_term:x }
+        { > ~ #1 . \exp_not:N \iow_newline: }
+  }
 \cs_new_protected:Npn \@@_show_environment_end:N #1
   {
     \exp_args:NNx \@@_check_end:Nn \l_@@_tmpa_tl { \cs_to_str:N #1 }
--- a/base/ltcmd.dtx
+++ b/base/ltcmd.dtx
@@ -35,7 +35,7 @@
 %
 %    \begin{macrocode}
 \def\ltcmdversion{v1.33}
-\def\ltcmddate{2025-06-26}
+\def\ltcmddate{2025-06-27}
 %    \end{macrocode}
 %
 %<*driver>
@@ -2798,7 +2798,7 @@
 %         {Make \cs{ShowCommand} stop for interaction}
 % \changes{v1.1b}{2022/11/30}
 %         {Don't stop for the \cs{begin} part of an environment}
-% \changes{v1.3e}{2025-06-26}{Support \cs{showstream} (gh/1062)}
+% \changes{v1.3e}{2025-06-27}{Support \cs{showstream} (gh/1062)}
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_show_environment:N #1
   {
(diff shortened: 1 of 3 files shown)


muzimuzhi:
Need a test (setting `\showstream` and loading the show-output file using `\SHOWFILE`)?

Also need to check all other modules which has tracing support. There are
- `\ShowCommand` from `ltcmd` module,
- `\ShowSocket`/`\socket_show:n` from `ltsockets` module,
- `\ShowInstanceValues` and friends from `lttemplates` module,
- `\ShowMarksAt` from `ltmarks` module,
- `\ShowFloat` from `ltoutput` module.

josephwright:
> Also need to check all other modules which has tracing support. There are
> 
>     * `\ShowCommand` from `ltcmd` module,
> 
>     * `\ShowSocket`/`\socket_show:n` from `ltsockets` module,
> 
>     * `\ShowInstanceValues` and friends from `lttemplates` module,
> 
>     * `\ShowMarksAt` from `ltmarks` module,
> 
>     * `\ShowFloat` from `ltoutput` module.

Of these, only `\ShowCommand` has the mix of `show` and `term` usage that would be an issue - I've updated. Other places either use the `l3msg` code or are debugging code which is exclusively `\iow_term:n` or `\typeout`.

josephwright:
> Need a test (setting `\showstream` and loading the show-output file using `\SHOWFILE`)?

I was rather hoping this is so esoteric that we don't need to worry ....

base/ltcmd.dtx
```diff
@@ -2819,8 +2820,13 @@
...
   }
-\cs_new_protected:Npn \@@_show:x #1
-  { \iow_term:x { > ~ #1 . \iow_newline: } }
+\cs_new_protected:Npe \@@_show:x #1
+  {
+    \cs_if_exist:NTF \iow_show:e
+      { \iow_show:e }
+      { \iow_term:x }
+        { > ~ #1 . \exp_not:N \iow_newline: }
+  }
```
FrankMittelbach: I guess this is just personal preferences, but I would normally prefer to not do definitions using :Npe but instead using two separate definitions, because this way the resulting definition is actually on record rather than being produced during the definition (also avoiding cluttering things with `\exp_not:N` inside), i.e. doing
```
\cs_if_exist:NTF \iow_show:e
{
  \cs_new_protected:Npn \@@_show:x #1
     { \iow_term:e { > ~ #1 . \iow_newline: } }
}
{
  \cs_new_protected:Npn \@@_show:x #1
     { \iow_term:x { > ~ #1 . \iow_newline: } }
}
```

base/lthooks.dtx
```diff
@@ -6175,10 +6176,12 @@
...
   }
-\cs_new_protected:Npn \hook_show:n #1
+\cs_new_protected:Npx \hook_show:n #1
   {
-    \cs_set_eq:NN \@@_log_cmd:x \iow_term:x
-    \@@_normalize_hook_args:Nn \@@_log:nN {#1} \tl_show:x
+    \cs_if_exist:NTF \iow_show:e
+      { \cs_set_eq:NN \exp_not:N \@@_log_cmd:x \iow_show:e }
+      { \cs_set_eq:NN \exp_not:N \@@_log_cmd:x \iow_term:x }
+    \exp_not:N \@@_normalize_hook_args:Nn \exp_not:N \@@_log:nN {#1} \tl_show:x
```
FrankMittelbach: same here

muzimuzhi:
> Of these, only `\ShowCommand` has the mix of `show` and `term` usage that would be an issue - I've updated.

Maybe provide a common command (in a more general module), for example `\__kernel_show:(n|e)`?

base/ltcmd.dtx
josephwright: My approach is to look at how much repetition I get: if there's a lot, then I go for one definition and expansion, if it's not much, I go for two. I can change of course but here the bulk of the tokens are the same in both paths ...

josephwright:
> > Of these, only `\ShowCommand` has the mix of `show` and `term` usage that would be an issue - I've updated.
> 
> Maybe provide a common command (in a more general module), for example `\__kernel_show:(n|e)`?

I'm hoping we don't have to keep the definitions as-is for too long: `\iow_show:n` is new so there is a period we have to allow for transition, but as these are all 'debug' features, I hope after a couple of release cycles we can 'tidy up'. (I have in mind more generally to sort `x`-type expansion, so this will not be forgotten.)

base/ltcmd.dtx
FrankMittelbach: as  I said, really a question of style I guess. My point is that I find the split of behavior on the outer side much easier to read in the source (and thus clearer) but I'm not insisting :-)

josephwright: Changed :)

FrankMittelbach (approved):

base/ltcmd.dtx
```diff
@@ -34,8 +34,8 @@
 %%% From File: ltcmd.dtx
 %
 %    \begin{macrocode}
-\def\ltcmdversion{v1.3d}
-\def\ltcmddate{2025-04-21}
+\def\ltcmdversion{v1.33}
```
muzimuzhi: Bump d -> e, rather than d -> 3?

```suggestion
\def\ltcmdversion{v1.3e}
```

base/lthooks.dtx
```diff
@@ -6163,6 +6163,7 @@
 %
 %
 % \begin{macro}{\hook_show:n,\hook_log:n}
+% \changes{v1.1l}{2025/06/27}{Support \cs{showstream} (gh/1062)}
```
muzimuzhi: The file version is bumped from 1.1l to 1.1m in this PR.

```suggestion
% \changes{v1.1m}{2025/06/27}{Support \cs{showstream} (gh/1062)}
```

muzimuzhi (changes requested):
Typos among version strings.

base/lthooks.dtx
muzimuzhi: This is marked resolved while not applied?

This PR bumps the version of `base/lthooks.dtx` from 1.1l to 1.1m, so the newly added `\changes` entry should use 1.1m too.

josephwright: Should now be sorted

muzimuzhi (approved):

PR #1799, opened 2025-06-26
Allow for active chars in \DeclareRobustCommand

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [x] Test file(s) added

### DIFF
--- a/base/ltdefns.dtx
+++ b/base/ltdefns.dtx
@@ -1088,6 +1088,8 @@
 % \changes{v1.5v}{2025-06-26}{New macro}
 % \begin{macro}{\declare@robustcommand@auxii}
 % \changes{v1.5v}{2025-06-26}{New macro}
+% \begin{macro}{\declare@robustcommand@auxiii}
+% \changes{v1.5v}{2025-06-26}{New macro}
 %    This is a package-writers command, which has the same syntax as
 %    |\newcommand|, but which declares a protected command.  It does
 %    this by having\\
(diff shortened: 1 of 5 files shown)


base/ltdefns.dtx
```diff
@@ -1111,7 +1116,10 @@
 %    Then if |\protect| is |\@typeset@protect| (normally |\relax|)
 %    then we just perform |\_<space>|, and otherwise
 %    |\@x@protect@| gobbles everything up and expands to
-%    |\protect\_|.
+%    |\protect\_|. None of that works with an active char (as we have only one
+%    character token to play with), so we resort to using the engine mechanism
+%    there. (The braces in passing |#1| there are defensive in case someone has
+%    passed more than one token!)
```
davidcarlisle, line 1124: see comments below

base/ltdefns.dtx
```diff
@@ -1132,6 +1140,23 @@
    \ifx#1\@undefined\else\ifx#1\relax\else
       \@latex@info{Redefining \string#1}%
    \fi\fi
+  \ifcat\noexpand#1\noexpand~%
```
davidcarlisle: probably safer to have `\ifcat\noexpand~\noexpand#1` if we are worrying about multiple tokens,

base/ltdefns.dtx
```diff
@@ -1132,6 +1140,23 @@
    \ifx#1\@undefined\else\ifx#1\relax\else
```
davidcarlisle: probably safer to have `\ifx\@undefined` and `\ifx\relax` if we are worrying about multiple tokens,

davidcarlisle (commented):
Just directly using `\protected` misses the protection for csname or filenames that for example `~` has.

Currently

```
\documentclass{article}

\catcode`\:=\active
\DeclareRobustCommand{:}{(colon)}

\typeout{a:b}
\show:

\begin{document}

\input{a:b}

\end{document}
```

inputs `a:b.tex`  but after the PR it gives

```
! LaTeX Error: File `a(colon)b.tex' not found.
```

(assuming there is a local `a:b.tex` )

base/ltdefns.dtx
FrankMittelbach: yes otherwise one might get spurious matches

base/ltdefns.dtx
FrankMittelbach: Maybe really add a test below with incorrectly adding multiple tokens

josephwright:
> \documentclass{article}
> 
> \catcode`\:=\active
> \DeclareRobustCommand{:}{(colon)}
> 
> \typeout{a:b}
> \show:
> 
> \begin{document}
> 
> \input{a:b}
> 
> \end{document}

Fair point: sorely tempting to define `\DeclareRobustCommand` using `\NewDocumentCommand`, grab all the args, then you can just do `\ifincsname\expandafter\@firstoftwo\else\expandafter\@secondoftwo\fi{\string<char>}{<def>}`. I'll think of a more '2e' solution.

base/ltdefns.dtx
josephwright: Well in `\NewDocumentCommand`, etc., we do explicitly test .. but that's well outside of the 2e tradition.

josephwright:
Current test errors are due to an issue in a support file that has `\DeclareRobustCommand{<not-one-token>}` :)

u-fischer:
> Current test errors are due to an issue in a support file that has \DeclareRobustCommand{<not-one-token>} :)

I can make an update ;-)

PR #1823, opened 2025-07-10
Block context adjustments

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [stub] `ltnewsX.tex` (and/or `latexchanges.tex`) updated

Intended to be a prototype for wider testing, so needs to be in -dev even if it will probably evolve between now and November.


required/latex-lab/latex-lab-context.md
```diff
@@ -0,0 +1,212 @@
...
+- `tiny`         denotes that we are typesetting in `\tiny` font size
+- `scriptsize`  
+- `footnotesize`  
+- `small`  
+- `large`  
+- `Large`  
+- `LARGE`  
+- `huge`  
+- `Huge`  
+
```
dbitouze, line 66: I don't know if it's relevant here but, IMO, these old good names are unfortunately not the best choice:

- they are a mix between:
  - “abstract” ones (“tiny”, “small”, “large”, “huge”),
  - “not abstract” ones (“scriptsize”, “footnotesize”),
- it is implicit but not absolutely clear that “large” < “Large” < “LARGE” and that “huge” < “Huge”.

Why not something (say `\smaller[i]` and `\bigger[i]`) in the spirit of the `relsize` package but being relative to the normal size, not to the current size?

FrankMittelbach: This is not about designing fontsize commands,  they are the way they are out there in million of documents and despite their perhaps somewhat unfortunate names one has to live with them in my opinion. And the "secondary context names" here are simply a mirror of the existing convention and for the purpose of design (which deals with final real values) relative names without a fixed starting point are not helpful at all. 
In documents they may, but it is kind of interesting that relsize never caught any  noticeable number of followers, which seems to indicate that most people find it easier to grasp what `\LARGE` means compared to `\bigger[3]`.
 
You design, say, a title for a certain size not for somehting that is bigger than a starting point you don't know about. And in the case here you design a list for use in an area where all text is of a certain size -- and design is not linear, i.e. when your text grows by 20% that doesn't mean that all your spacing should grow by the same factor (or grow at all. Or vice versa, a list in a footnote might turn all vertical spacing to zero, even if the textsize shrinks only by a few percent.

dbitouze: Speaking of `relsize`, I think its lack of popularity comes from the fact that the font size change is relative to the current size, not the normal size.

About the non linearity of the design, indeed, but `\smaller[i]` and `\bigger[i]` I had in mind would just be alias for what the current command `\tiny` to `\Huge` (and the weird `\Tiny` and `\TINY` provided by `beamer`) do.

But I understand this is not the right place for such a discussion. I just hope that if a new set of classes is created (I remember someone started to create such a set with `L3`), a better set of command names will be provided.

Thanks for answering!

base/doc/ltnews42.tex
```diff
@@ -207,6 +207,11 @@ \subsection{Normalizing key names for block environments}
 \emph{to write}
 
 
+\subsection{Contexts in typesetting}
+
+\emph{to write if latex-lab-context makes it to the next stage before 2025/11/01}
```
josephwright, line 212: :-)

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1311,10 +1318,11 @@
...
 %
 %  \begin{macro}{\DebugBlocksOn,\DebugBlocksOff}
-%    
+%    If we are debugging blocks we also want to know about template
+%    instances, so we turn the debugging for templates as well (for now).
 %    \begin{macrocode}
-\cs_new_protected:Npn \DebugBlocksOn  { \block_debug_on:  }
-\cs_new_protected:Npn \DebugBlocksOff { \block_debug_off: }
+\cs_new_protected:Npn \DebugBlocksOn  { \block_debug_on: \template_debug_on: }
+\cs_new_protected:Npn \DebugBlocksOff { \block_debug_off: \template_debug_off:}
```
josephwright:
```suggestion
\cs_new_protected:Npn \DebugBlocksOff { \block_debug_off: \template_debug_off: }
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1455,21 +1468,22 @@
   \l_@@_setup_code_tl
 %    \end{macrocode}
 %    Next call a block instance at the appropriate level passing it
-%    any key/value list provided in the optional argument (keys that
-%    are not recognized are ignored---currently with an
-%    error).
+%    any remaining key/value from the optional document-level argument
+%    (i.e., those now stored in \cs{UnusedTemplateKeys}.
```
josephwright:
```suggestion
%    (i.e., those now stored in \cs{UnusedTemplateKeys}).
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1455,21 +1468,22 @@
...
+%
+%
+%  \changes{v0.9j}{2025/07/11}{Expand arguments before passing them to
+%    \cs{Useinstance} to get better error messages}
 %    \begin{macrocode}
-  \@@_debug_typeout:n{use~ 'block'~ instance:~
-           \l_@@_block_instance_tl - \int_use:N \g_block_nesting_depth_int }
-  \UseInstance{block}
+  \exp_args:Nee \UseInstance{block}
```
josephwright, line 1478: I'd favour `\ExpandArgs` with a design-level interface - but it's not important.

required/latex-lab/latex-lab-context.dtx
```diff
@@ -0,0 +1,799 @@
...
+%   (i.e., the default \enquote{primary context} is the main galley
+%   and by default no \enquote{secondary context} is specified).
+%   
+% \subsection{The \enquote{primary context}}
+%   
+%   The \enquote{primary context} is described with a fixed (but
+%   extensible?) set of names:
+%   \begin{itemize}
+%
+%   \item \meta{empty} denotes that we are in "galley" context
```
josephwright:
```suggestion
%   \item \meta{empty} denotes that we are in \enquote{galley} context
```

required/latex-lab/latex-lab-context.dtx
```diff
@@ -0,0 +1,799 @@
...
+%   This declaration defines the rule that when the current primary
+%   context is \meta{current} and \meta{new} is requested as the new
+%   context then instead \meta{result} is made the new
+%   \enquote{primary context}.
+%
+%   If \texttt{*} is used in the first argument then this indicates
+%   any current context and thus \meta{new} is always replaced by
+%   \meta{result}. This can be written shorter as
+%   \begin{quote}
+%     \cs{DeclarePrimaryContextRule*} \Arg{new}\Arg{result}
```
josephwright, line 181: I suspect that we should document `\DeclarePrimaryContextRule*` as a separate command - this is the normal take on stars - but not for this initial release.

required/latex-lab/latex-lab-context.dtx
```diff
@@ -0,0 +1,799 @@
...
+%   This declaration defines the rule that when the current secondary
+%   context is \meta{current} and \meta{new} is requested as the new
+%   context then instead \meta{result} is made the new
+%   \enquote{secondary context}.
+%
+%   If \texttt{*} is used in the first argument then this indicates
+%   any current context and thus \meta{new} is always replaced by
+%   \meta{result}. This can be written shorter as
+%   \begin{quote}
+%     \cs{DeclareSecondaryContextRule*} \Arg{new}\Arg{result}
```
josephwright, line 224: As above

required/latex-lab/latex-lab-context.dtx
```diff
@@ -0,0 +1,799 @@
...
+%   \item
+%     If not then \meta{inst-name}\texttt{:}\meta{primary context} is
+%   tried next.
+%   \item
+%     If that doesn't exist either then \meta{inst-name} is used as usual.
+%   \end{itemize}
+% 
+%   This means it becomes trivial to alter the behavior of
+%   instances if they appear in a special typesetting context. For
+%   example, in \LaTeXe{} display blocks, e.g. lists, center, etc all
```
josephwright:
```suggestion
%   example, in \LaTeXe{} display blocks, e.g., lists, center, etc., all
```

required/latex-lab/latex-lab-context.dtx
```diff
@@ -0,0 +1,799 @@
...
+%   when instances are selected could then (perhaps) be a function of
+%   the template type but for a heading type looking at
+%   front/main/back matter would be more approriate.\footnote{Then, of
+%   course, a primary context change would not reet any secondary
+%   context.}
+%
+%   Bottom line, what is here is nothing more than a first prototype
+%   and likely to change a lot.
+%
+% \section{Currently mainly internal interfaces}
```
josephwright, line 331: If they are internal, they should be `__`

required/latex-lab/latex-lab-context.dtx
```diff
@@ -0,0 +1,799 @@
...
+%    \begin{macrocode}
+\ProvidesExplPackage {latex-lab-testphase-context}
+                     {\ltlabcontextdate}
+                     {\ltlabcontextversion}
+  {Providing context for instance, etc.}
+%    \end{macrocode}
+%
+%    \begin{macrocode}
+%<*package>
+%<@@=context>
```
josephwright, line 396: Add to the prefix database?

required/latex-lab/latex-lab-context.dtx
```diff
@@ -0,0 +1,799 @@
...
+%
+%
+%
+%  \begin{macro}{\SetPrimaryContext}
+%    Change the context: if the target context is empty, just do
+%   it. Otherwise if we have  rule for current context and new context
+%   use that; otherwise if there is a star rule aply that; otherwise
+%   set the new value as requested,
+%    \begin{macrocode}
+\newcommand \SetPrimaryContext [1] {
```
josephwright: Why `\newcommand` here?

josephwright: (And the others below)

required/latex-lab/latex-lab-context.dtx
```diff
@@ -0,0 +1,799 @@
...
+%
+%
+% \subsection{Supporting font size changes as a secondary context}
+%
+%
+%
+%  \begin{macro}{\@setfontsize}
+%    We try to be careful with setting the name as we don't know if
+%   \cs{string} returns a backslash or not.
+%    \begin{macrocode}
```
josephwright, line 567: Could use `\cs_to_str:N`

required/latex-lab/latex-lab-context.dtx
```diff
@@ -0,0 +1,799 @@
...
+%   if we are in the context of the main galley).
+% \end{variable}
+%
+% \begin{variable}{\l_context_secondary_tl}
+%   This variable holds the current secondary context name (or is
+%   empty if we are typesetting in \cs{normalsize}). It is by default
+%   automatically set by all fontsize commands in core \LaTeX{}. If
+%   additional secondary contexts are used then the variable may not
+%   be empty even if we are typesetting in \cs{normalsize}
+% \end{variable}
```
josephwright: Given the restrictions on content, I'd make these `str` data types.

josephwright (approved):
Comments are all minor tweaks at this stage: can mop in during integration

required/latex-lab/latex-lab-context.dtx
FrankMittelbach: yes, all depends on whether or not we tweak this part of the interface

required/latex-lab/latex-lab-context.dtx
FrankMittelbach: not soooo internal but not yet final :-)  we need them across modules, so can't be really __ but we might provide someting totally different in the end (that is why I called them "internal"

required/latex-lab/latex-lab-context.dtx
FrankMittelbach: yes please. We need it in the end

required/latex-lab/latex-lab-context.dtx
FrankMittelbach: because it is still faster :-) and I'm an old timer, but I'm not complaining if that gets changed eventually

required/latex-lab/latex-lab-context.dtx
FrankMittelbach: when updating 2e commands I prefer staying with 2e coding and not mixing stuff that's my rule of thumb here. In the end this would be moved back to its original place  and there everything is (at the moment 2e)

required/latex-lab/latex-lab-context.dtx
josephwright: Not faster than `\cs_set:Npn` (and `\cs_set_protected:Npn` is very very close) :) And really, it's one expansion if you use `\NewDocumentCommand` as this will be optimised

required/latex-lab/latex-lab-context.dtx
josephwright: Sure, I just meant there is a way - and of course LuaTeX has `\csstring` ;-)

required/latex-lab/latex-lab-context.dtx
FrankMittelbach: done and hopefully correct this time around

PR #1834, opened 2025-08-02
Uf extend para

u-fischer:
* switches to sockets for the para/main (Part) structure, so that they can be disabled by assigning the noop plug.
* fixes (https://github.com/latex3/tagging-project/issues/835
* switches to tagging sockets for theorem like environments (in block and firstaid)
* fixes https://github.com/latex3/tagging-project/issues/950  (missing space in message)
* add hooks to Gin-keys alt, actualtext and artifact (for https://github.com/latex3/tagging-project/issues/957)

## Status of pull request

- Ready to merge

- [x] Test file(s) added


base/doc/ltnews42.tex
```diff
@@ -201,6 +201,31 @@ \subsection{Requiring or testing for the Tagging support code}
...
 \cs{IfPDFManagementActiveTF} is provided.
  
+\subsection{Moving paragraph tagging into sockets}
+
+Paragraphs in \LaTeX{} can be nested, e.g., you can have a paragraph
+containing a display quote, which in turn consists of more than one
+(sub)paragraph, followed by some more text which all belongs to the
+same outer paragraph.
+
+To model such \enquote{big} paragraphs \LaTeX{} uses a structure named \texttt{text-unit} and use \texttt{text} (rollmapped to \texttt{P}) only for (portions of) the actual paragraph text\LaTeX{}.
```
FrankMittelbach: Maybe add a footnote:  The name text-unit is likely to change in the future it is currently under review.

base/doc/ltnews42.tex
```diff
@@ -201,6 +201,31 @@ \subsection{Requiring or testing for the Tagging support code}
...
+
+Paragraphs in \LaTeX{} can be nested, e.g., you can have a paragraph
+containing a display quote, which in turn consists of more than one
+(sub)paragraph, followed by some more text which all belongs to the
+same outer paragraph.
+
+To model such \enquote{big} paragraphs \LaTeX{} uses a structure named \texttt{text-unit} and use \texttt{text} (rollmapped to \texttt{P}) only for (portions of) the actual paragraph text\LaTeX{}. 
+
+This is semantically clear and allows processors who care to identify
+the complete paragraphs by looking for \texttt{text-unit} tags. But we got also request for an option to disable the tagging of the \enquote{big} paragraphs, so with this release we put the relevant tagging code into sockets. The \enquote{big} paragraphs can now be disabled by assigning to these socket the \texttt{noop} plug:
```
FrankMittelbach: we moved the relvant tagging code into sockets

I would use \enquote{big paragraphs}

(please remember putting in newlines for better diffs)

base/doc/ltnews42.tex
FrankMittelbach: In fact, we might change now so that for 2025/11 the code contains a new name already. My vote would be for "paragraph" despite the fact that LaTeX has a \paragraph heading, but this should probably be discussed elsewhere.

base/doc/ltnews42.tex
```diff
@@ -201,6 +201,31 @@ \subsection{Requiring or testing for the Tagging support code}
...
+\begin{verbatim}
+\AssignTaggingSocketPlug{para/begin/main}{noop}
+\AssignTaggingSocketPlug{para/end/main}{noop}
+\end{verbatim}
+
+\subsection{Hooks for \cs{includegraphics} keys}
+The three keys definitions \texttt{alt}, \texttt{actualtext} and \texttt{artifact} used by \cs{includegraphics} contain now hooks, named \texttt{Gin/alt},
+\texttt{Gin/actualtext}, and \texttt{Gin/artifact}. The first two are hooks with one argument and take as argument the value of the key. The hooks are processed even if tagging is not activated.
+With them, it is for example possible to store the alternative text:
```
FrankMittelbach: With them it is, for example, possible  % I think

...store the alternative text for reuse elsewhere, e.g., in a message:

base/lttagging.dtx
```diff
@@ -344,6 +379,18 @@
 %    \end{macrocode}
 %  \end{macro}
 %
+%\begin{taggingsocketdecl}{paraOn,paraOff}
+% These sockets allow to enable/disable paratagging.
+%    \begin{macrocode}
+\NewTaggingSocket{paraOn}{0}
+\NewTaggingSocketPlug{paraOn}{kernel}{\bool_set_true:N \l__tag_para_bool}
+\NewTaggingSocket{paraOff}{0}
```
FrankMittelbach: not sure about the name. Maybe `para/on` and `para/off` instead. Seems to fit also better with `para/restore`

base/lttagging.dtx
```diff
@@ -369,11 +416,135 @@
 % \end{taggingplugdecl}
 %
 % \begin{taggingsocketdecl}{para/begin,para/end}
-% These sockets are currently defined in tagpdf. They overwrite
-% definitions in the latex-lab-block code. There is also a simpler
-% definition that probably should be a general socket too.
-% TODO: move this into lttagging.
+% These sockets were previously defined in tagpdf. 
+% There are the main sockets to handle the paragraph tagging.
```
FrankMittelbach: tagging when it is active (i.e., when `para/on` or `para/restore`was executed earlier).

base/lttagging.dtx
```diff
@@ -369,11 +416,135 @@
...
+%    \begin{macrocode}
+\NewTaggingSocketPlug{para/end/main}{kernel}
+ {
+   #1
+   \__tag_gincr_para_main_end_int:
+   \tag_struct_end:
+ }
+%    \end{macrocode}
+% \end{taggingplugdecl}
+% And now we assign this plugs
```
FrankMittelbach: plugs:

base/lttagging.dtx
```diff
@@ -369,11 +416,135 @@
...
+\NewTaggingSocketPlug{para/begin}{plain}{}
+\NewTaggingSocketPlug{para/end}{plain}{}
+\NewTaggingSocketPlug{para/begin}{block}{}
+\NewTaggingSocketPlug{para/end}{block}{}
+
+%    \end{macrocode}
+
+% \end{taggingsocketdecl}
+% 
+% \begin{taggingsocketdecl}{para/begin/main,para/end/main}
```
FrankMittelbach: Not sure about the name "main". Given that you allow to leave out this "main" it seems to be the wrong name.
"/outer" perhaps?  Not sure what covers the concept best.

base/lttagging.dtx
```diff
@@ -369,11 +416,135 @@
...
+        \bool_if:NF \l__tag_para_flattened_bool
+         {
+           \UseTaggingSocket{para/end/main}{}
+         }
+      }
+  }
+%    \end{macrocode}
+% \end{taggingplugdecl}
+% As tagpdf is currently assigning the sockets after the block code we assign
+% if even later
```
FrankMittelbach: it even later (for now):

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3798,31 +3789,36 @@
 %    We do not want potential tagging from \tn{textbf} here,
 %    so we use \tn{bfseries} to set the font. The commands set also the link targets
 %    which should be inside the main structure.
-%    \begin{macrocode}
+%    \begin{macrocode}  
+\NewTaggingSocket{block/theorem/caption}{2}
+\NewTaggingSocketPlug{block/theorem/caption}{kernel}
+ {
+   \tag_struct_begin:n{tag=Caption}
```
FrankMittelbach, line 3796: There was the request to not produce caption tags, wasn't there? If so that could now be handld by using a different plug for this socket. Perhaps we can anticipate that arelady now? by providing an alternative plug in addition to kernel?

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3798,31 +3789,36 @@
...
-    \tag_struct_end:          
-    \group_end:
-  \tag_struct_end:
-  \tagpdfparaOn
+  \group_begin:
+  \bfseries
+  \tag_socket_use:nnn {block/theorem/caption} {}
+   { 
+     \tag_socket_use:nnn {mc} {} {#1\ }
+     \tag_socket_use:nnn {struct-mc} {tag=Lbl} {#2}
```
FrankMittelbach, line 3818: Given that we don't want "Lbl"  directly in the tag structure. Shouldn't that be indirect here (and elsewhere)? I.e. a name in the latex namespace that is rolemapped?

required/latex-lab/latex-lab-graphic.dtx
```diff
@@ -433,10 +433,36 @@
 %   \DocumentMetadata{tagging=on,debug=BBox}
 %  \end{verbatim}
 %  
+%  \subsection{Hooks}
+%  The three key definitions \texttt{alt}, \texttt{actualtext} and \texttt{artifact} used by
+%  \cs{includegraphics} contain hooks, named \texttt{Gin/alt},
+%  \texttt{Gin/actualtext}, and \texttt{Gin/artifact}. The first two take as argument
```
FrankMittelbach: and \texttt{Gin/artifact}, respectively.

required/latex-lab/latex-lab-graphic.dtx
```diff
@@ -433,10 +433,36 @@
...
+%  \includegraphics[alt=Hello World]{example-image-A}
+%  The alt text of the graphic was \myalttext.
+%  \end{verbatim}
+%  
+%  Please note
+%  \begin{itemize}
+%   \item The hooks are also executed if the keys are set with \verb+\setkeys{Gin}+.
+%   \item The arguments are given to the hooks without any processing or purifying. 
+%   \item The hooks are not executed in the picture environment or in tikz pictures
+%   as they use different keys.
```
FrankMittelbach, line 455: are we plan to have similar hooks there? If so should the hooks be called differently and reused in both places? I mean if you want to store the alt text you may want to store it in all cases. But perhaps it is better to use separate hooks and just repeat the code chunks.

required/latex-lab/latex-lab-graphic.dtx
```diff
@@ -433,10 +433,36 @@
...
+%  
+%  Please note
+%  \begin{itemize}
+%   \item The hooks are also executed if the keys are set with \verb+\setkeys{Gin}+.
+%   \item The arguments are given to the hooks without any processing or purifying. 
+%   \item The hooks are not executed in the picture environment or in tikz pictures
+%   as they use different keys. 
+%   \item The hooks are obviously not executed if the keys are not used. 
+%   That means they can't be used directly to issue a warning or an error 
+%   that an alt text is missing. For this some additional logic is needed, e.g.
```
FrankMittelbach: e.g.,

required/latex-lab/latex-lab-graphic.dtx
```diff
@@ -433,10 +433,36 @@
...
+%  \begin{verbatim}
+%  \AddToHookWithArguments{Gin/alt}{\gdef\myalttext{#1}}
+%  \includegraphics[alt=Hello World]{example-image-A}
+%  The alt text of the graphic was \myalttext.
+%  \end{verbatim}
+%  
+%  Please note
+%  \begin{itemize}
+%   \item The hooks are also executed if the keys are set with \verb+\setkeys{Gin}+.
+%   \item The arguments are given to the hooks without any processing or purifying.
```
FrankMittelbach, line 453: why is that the case? Why not first purify and pass the resulting string to the hook? Might make more sense since that is what is going to be used. Or perhaps provide both to the hook: the purified version and the plan version, i.e. a hook with 2 arguments
```
         \tl_if_empty:nTF{#1}
           { \tl_clear:N\l_@@_graphic_alt_tl}
           { \tl_set:Ne\l_@@_graphic_alt_tl{\text_purify:n{#1}}}
        \exp_arg:Nnno  \hook_use:nnw {Gin/alt}{2} \l_@@_graphic_alt_tl {#1}       
```

required/latex-lab/testfiles-minipage/minipage-002-todo.luatex.tlg
```diff
@@ -166,8 +166,6 @@ Package tagpdf Info: Parent-Child 'P' --> 'MC'.
 (tagpdf)             Relation is 1 (='0..n') on line ...
 tagpdf DEBUG Info: Tagging suspended
 (tagpdf DEBUG)     level: 0 ==> 1, label: \tikzpicture  [on line ...]
-tagpdf DEBUG Info: Struct 12 begin ignored with options: tag=\l__tag_para_main_tag_tl ,attribute-class=\l__tag_para_main_attr_class_tl , 
-(tagpdf DEBUG)     [on line ...]
```
FrankMittelbach, line 170: why are those no longer showing?

required/latex-lab/testfiles-table-luatex/table-004-tabularx.tlg
```diff
@@ -3,25 +3,20 @@ Don't change this file in any respect.
 [Sockets] ==> Socket 'tagsupport/kernel/endpe/vmode' containing plug 'default' used.
 tagpdf DEBUG Info: Tagging suspended
 (tagpdf DEBUG)     level: 0 ==> 1, label: tabularx [on line ...]
-[Sockets] ==> Socket 'tagsupport/para/begin' containing plug 'block' used.
-tagpdf DEBUG Info: Struct 2 begin ignored with options: tag=\l__tag_para_main_tag_tl ,attribute-class=\l__tag_para_main_attr_class_tl , 
-(tagpdf DEBUG)     [on line ...]
+[Sockets] ==> Socket 'tagsupport/para/begin' containing plug 'kernel' used.
```
FrankMittelbach, line 6: why was this "block" and is now "kernel"?

FrankMittelbach (commented):
quite a mouthful ... so for now only comments

base/doc/ltnews42.tex
davidcarlisle:
```suggestion
To model such \enquote{big} paragraphs \LaTeX{} uses a structure named \texttt{text-unit} and use \texttt{text} (rolemapped to \texttt{P}) only for (portions of) the actual paragraph text\LaTeX{}. 
```

base/doc/ltnews42.tex
davidcarlisle:
```suggestion
the complete paragraphs by looking for \texttt{text-unit} tags. But we got also request for an option to disable the tagging of the \enquote{big} paragraphs, so with this release we put the relevant tagging code into sockets. The \enquote{big} paragraphs can now be disabled by assigning the \texttt{noop} plug to these sockets:
```

base/lttagging.dtx
```diff
@@ -344,6 +379,18 @@
 %    \end{macrocode}
 %  \end{macro}
 %
+%\begin{taggingsocketdecl}{paraOn,paraOff}
+% These sockets allow to enable/disable paratagging.
```
davidcarlisle:
```suggestion
% These sockets allow paragraph tagging to be enabled/disabled.
```

davidcarlisle (commented):
just some English comments on the doc, so far

base/lttagging.dtx
u-fischer: Well the variable is called `\l__tag_para_main_tag_tl`, so I think outer would be confusing.

required/latex-lab/latex-lab-block.dtx
u-fischer: yes, but we should decide first how we organize these names across all the latex-lab modules.

required/latex-lab/latex-lab-graphic.dtx
u-fischer: I'm not sure ;-(  I have a certain inclination towards separate hooks, but they are arguments for both.

required/latex-lab/latex-lab-graphic.dtx
u-fischer: I didn't want to remove the option to handle the original argument. But passing both is naturally an option.

required/latex-lab/testfiles-minipage/minipage-002-todo.luatex.tlg
u-fischer: because the \tagstruct-commands are now hidden in tagging sockets and so no longer executed and so they no longer can give this message.

required/latex-lab/testfiles-table-luatex/table-004-tabularx.tlg
u-fischer: it was block because at some time we had two versions ('default' and  'block' if the block code was loaded). But now that we have only one version that is used always, and so 'block' imho no longer fits and kernel aligns better with the other plug names.

required/latex-lab/latex-lab-block.dtx
u-fischer: Imho that is a rather special wish, and it is also rather unsemantically so it doesn't feel like something that we want to promote.

required/latex-lab/testfiles-table-luatex/table-004-tabularx.tlg
FrankMittelbach: I thought that might be the reason, but wasn't 100% sure last night

required/latex-lab/latex-lab-graphic.dtx
FrankMittelbach: I think it would be better. In most cases you want to present the same string as it is used by the algorithm but sometimes you may want to use what the user has typed.

required/latex-lab/latex-lab-graphic.dtx
FrankMittelbach: I wasn't sure either. Perhaps separate hooks are indeed better.

base/lttagging.dtx
FrankMittelbach: well, you could rename the internal variable as well. My point is from the programmers point of view "main" doesn't seem to be something that you can simply drop if for some reason you don't like it. "outer" (or some other name) on the other hand is something that is more understandable as being an extra structure that may be left out.

u-fischer: We do not think that it should be dropped as it removes the semantic of the "real", main paragraph. So why should we give it a name that looks as if it is a good idea?  (changing the name of the variable is not straightforward as it is used in tagpdf too, so needs some coordination).

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: ok, let's keep this separate from the PR but we should attend to this between now and November I would say.

base/lttagging.dtx
FrankMittelbach: I don't think that "outer" means that it is not a good idea. But what we do is an extension of what, say HTML (or PDF for that matter) encodes as "P". We call that "text", ie short for "text-chunk" which is really what it is. What we currently call "text-unit" is the real paragraph or "paragraph as perceived by a human" but there is nothing "main" about it. Now we need a word for indicating that we put a structure at this point and I'm not saying that "outer" for that is much better. Better would be perhaps `para/begin/pstruct` or `parastruct`, i.e., the structure put around a paragraph as perceived by a human reader.

base/doc/ltnews42.tex
```diff
@@ -201,10 +201,51 @@ \subsection{Requiring or testing for the Tagging support code}
...
 
+ 
+\subsection{Moving paragraph tagging into sockets}
+
+Paragraphs in \LaTeX{} can be nested, e.g., you can have a paragraph
+containing a display quote, which in turn consists of more than one
+(sub)paragraph, followed by some more text which all belongs to the
+same outer paragraph.
+
+To model such \enquote{big paragraphs} \LaTeX{} uses a structure named
```
FrankMittelbach: I like the name "semantic paragraph" that you used elsewhere. That might be what we have been looking for and would also, allow to use para/semantic  (instead of para/main) which I think is better than using "main"

base/doc/ltnews42.tex
```diff
@@ -201,10 +201,51 @@ \subsection{Requiring or testing for the Tagging support code}
...
+assigning the \texttt{noop} plug to these sockets:
+\begin{verbatim}
+\AssignTaggingSocketPlug{para/begin/main}{noop}
+\AssignTaggingSocketPlug{para/end/main}{noop}
+\end{verbatim}
+
+\subsection{Hooks for \cs{includegraphics} keys}
+The three keys definitions \texttt{alt}, \texttt{actualtext} and 
+\texttt{artifact} used by \cs{includegraphics} contain now hooks, 
+named \texttt{Gin/alt}, \texttt{Gin/actualtext}, and \texttt{Gin/artifact}.
```
FrankMittelbach: I know that Gin and Rot have a long tradition inside graphics drivers, but does it really make sense to expose these names on a higher level? I mean, other than oldtimers who can make sense of "Gin"? Even if longer, wouldn't `includegraphics/alt` be better?

base/doc/ltnews42.tex
```diff
@@ -201,10 +201,51 @@ \subsection{Requiring or testing for the Tagging support code}
...
+
+This is semantically clear and allows processors who care to identify
+the complete paragraphs by looking for \texttt{text-unit} tags. 
+But we got also request for an option to disable the tagging of the 
+\enquote{big paragraphs}, so with this release we moved the relevant tagging 
+code into sockets. The \enquote{big paragraphs} can now be disabled by 
+assigning the \texttt{noop} plug to these sockets:
+\begin{verbatim}
+\AssignTaggingSocketPlug{para/begin/main}{noop}
+\AssignTaggingSocketPlug{para/end/main}{noop}
```
FrankMittelbach: here you have para/main but below it is para/paragraph

leftover?

or with my new suggestion para/semantic

base/lttagging.dtx
```diff
@@ -409,11 +457,136 @@
...
+% \end{taggingsocketdecl}
+% 
+% \begin{taggingsocketdecl}{paragraph/begin,paragraph/end}
+% These sockets handle the \UseStructureName{para/paragraph} begin and end structure.
+% These are stored in sockets of their own to be able to disable
+% them globally.
+% 
+%    \begin{macrocode}
+\NewTaggingSocket{paragraph/begin}{1}
+\NewTaggingSocket{paragraph/end}{1}
```
FrankMittelbach: with the idea of "semantic" that could then become para/semantic/begin

FrankMittelbach (commented):
sorry for coming up with another idea ...

required/latex-lab/latex-lab-graphic.dtx
FrankMittelbach: looking over these old comments and the discussion around Gin (and your changes there) I still have the impression that calling them key/alt with the meaning of receiving the  value of an alt key in purified and raw  version) which then could be reused on any all key implementation would perhaps be more appropriate than to force different alt implementation to do the same thing in different hooks, e.g., storing the alt data.

FrankMittelbach (approved):
I'm okay, with it assuming everybody is okay with different hooks getting the alt key data for different implementation (or no hooks, as it is currently the case for alt keys on envs/cmds that are not based on graphicx). Still feel that it makes sense to use the same hook also for tlks pictures and the like, but it is not a really problem if that is not the case I guess.

PR #1862, opened 2025-09-22
Metafamily support

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


car222222:
Only one comment/question at this stage:
Where is it explained what a "meta family" actually is?

base/doc/ltnews42.tex
```diff
@@ -329,6 +329,32 @@ \subsection{Support separate font families for script fonts}
 \githubissue{1707}
 
 
+\subsection[Programming support for meta font families]
+           {Programming support for \LaTeX's meta font families}
+
+\LaTeX{} knows three main document font families: \cs{rmfamily} for
+the document's serifed font family, \cs{sffamily} for its sans serif
+font family, and \cs{ttfamily} for its monspaced font family. In
+addition, other font families can be used by the user ot in a document
```
josephwright: 'ot'?

josephwright (commented):
Is there a test of this functionality somewhere? I don't see one being added.

base/ltfssini.dtx
```diff
@@ -592,6 +635,7 @@
 %
 %
 %  \begin{macro}{\update@series@target@value}
+%
 %    In this macro used in the look you basically find the nested
```
davidcarlisle: look -> hook ?

davidcarlisle (approved):
looks OK, although as Joseph says a test file would probably help

FrankMittelbach:
> Only one comment/question at this stage: Where is it explained what a "meta family" actually is?

documented in ltfssini.dtx and in ltnews42

base/ltfssini.dtx
FrankMittelbach: no loop, I guess. But the text is garbled ...

car222222:
Well maybe: but my not understanding your news 42 entry is what made me wonder why this concept is used!

car222222:
But there is nothing in ltfssini.dtx that explains why a "word" called meta is ever used, or what it means! 
So now I am even more confused.
(There is a lot of other very strange language in this file! Where did all this come from?)

car222222:
Is it Meta family or Metafamily ??

FrankMittelbach:
> Is there a test of this functionality somewhere? I don't see one being added.

it is not easy to make one (without loading a lot of fonts into the test suite). But I added something now --- which of course caught an issue ...

u-fischer:
As `\@currentmetafamily` is meant as a public interface it should imho be documented in fntguide.tex too and not only in ltnews.

Not directly related to the PR, but with fontspec using `\addfontfeatures` resets the meta family, so 

~~~~
\sffamily  \@currentmetafamily 123 \addfontfeatures{Numbers=OldStyle}  \@currentmetafamily 123
~~~~

gives `sf123 ??123` which is probably not what users would expect.

FrankMittelbach:
> As `\@currentmetafamily` is meant as a public interface it should imho be documented in fntguide.tex too and not only in ltnews.

agreed.


> Not directly related to the PR, but with fontspec using `\addfontfeatures` resets the meta family, so
> 
> ```
> \sffamily  \@currentmetafamily 123 \addfontfeatures{Numbers=OldStyle}  \@currentmetafamily 123
> ```
> 
> gives `sf123 ??123` which is probably not what users would expect.

I can't replicate that. Can you provide me with a full test file? (or perhaps you haven't used the latest version)

u-fischer:
Try with 
~~~~
\documentclass{article}

\usepackage{fontspec}
\setsansfont{TeXGyreHeros}
\begin{document}

\makeatletter
\sffamily \@currentmetafamily 123 \addfontfeatures{Numbers=OldStyle} 
 \@currentmetafamily 123

\end{document}
~~~~

FrankMittelbach:
> Try with
> 
> ```
> \documentclass{article}
> 
> \usepackage{fontspec}
> \setsansfont{TeXGyreHeros}
> \begin{document}
> 
> \makeatletter
> \sffamily \@currentmetafamily 123 \addfontfeatures{Numbers=OldStyle} 
>  \@currentmetafamily 123
> 
> \end{document}
> ```

Ah that's why I don't see it. But in that case after \addfontfeatures your font is no longer equal to the one selected with \sffamily and just calling \sffamily again will remove the old style numbers again (which proves it). Basically you go from `TeXGyreHeros(0)` to `TeXGyreHeros(1)`  in your example. So I think this is consistent and explainable. You have to set the feature on the \setsansfont level if it should apply to the font selected by \sffamily.

u-fischer:
>  But in that case after \addfontfeatures your font is no longer equal to the one selected with \sffamily

Yes I know that. 

> So I think this is consistent and explainable. 

On the technical side it is explainable, but imho a user would view both font variants as part of the sf metafamily.  

> You have to set the feature on the \setsansfont level if it should apply to the font selected by \sffamily. 

The one question is if \addfontfeature should retain the metafamily and if yes how. The other question is what user should do who want to setup two font variants, e.g. proportional numbers normally but lining numbers in a tabular.

FrankMittelbach:
> > But in that case after \addfontfeatures your font is no longer equal to the one selected with \sffamily
> 
> Yes I know that.
> 
> > So I think this is consistent and explainable.
> 
> On the technical side it is explainable, but imho a user would view both font variants as part of the sf metafamily.
> 
> > You have to set the feature on the \setsansfont level if it should apply to the font selected by \sffamily.

conceptually, I agree but that would be a fontspec bug/misfeature and should probably be tracked there (@wspr )

> The one question is if \addfontfeature should retain the metafamily and if yes how. 

If you define fontfamily and give it an NFSSfamily name, e.g.,
```
\newfontfamily\tgtermes{TeX Gyre Termes}[NFSSFamily=tgtermes] 
```
and then stick that `tgtermes` into `\rmdefault`, then it looks as if fontspec does change the font feature without generating internally a new font family but instead alters the existing one linked to the NFSSfamily name.

> The other question is what user should do who want to setup two font variants, e.g. proportional numbers normally but lining numbers in a tabular.

That is in my opinion a different question and there would be different ways to model that, e.g., as a sub axis on the shape level which is what I would do I guess. But at the moment there is no support for this kind of extension.

wspr:
In hindsight fontspec shouldn't have offered \addfontfeatures -- instead, the interface should be a way to create a new font/family based on a previous one plus extra features.

I wonder if it would break backwards compatibility terribly if \addfontfeatures wrote over the previous NFSS definition. Maybe it wouldn't even be possible, I expect grouping is problematic.

u-fischer:
> In hindsight fontspec shouldn't have offered \addfontfeatures -- instead, the interface should be a way to create a new font/family based on a previous one plus extra features.

I agree ;-). It is slow and if it is (mis)used to often that can be a real problem. I saw an example somewhere. But removing the command does not resolve the general problem how to handle fonts that practically belong to the same meta font family, e.g. `sf`  but with lining numbers instead of oldstyle, but technically belong to some other family as they are defined with `\newfontfamily` or created with `\addfontfeatures`.

car222222:
@u-fischer This is perhaps confusing "family" and "metafamily"? ... they are conceptually distinct.

Issue #1864, opened 2025-09-23
Setting \partokencontext


davidcarlisle:
## Brief outline of the enhancement

In newer engines LaTeX could set `\partokencontext=2 ` so that an explicit `\par` token is added if TeX is in hmode at the end of a vbox and similar constructs.  This would help tagging and similar code that needs to detect the end of paragraphs.
```latex
%\partokencontext=2
\RequirePackage{latexbug}       % <--should be always the first line (see CONTRIBUTING)!
\documentclass{article}

\AddToHook{para/end}{!END OF PARA}

\begin{document}

AAA

BBB

\vbox{CCC}

\end{document}
```

## Minimal example showing the desired new behaviour

Uncomment `\partokencontext=2` and the end of the paragraph CCC will be detected.

eg9:
This fix breaks other constructs:

https://tex.stackexchange.com/q/762692/4427

```
% Source - https://tex.stackexchange.com/q/762692
% Posted by Klaus
% Retrieved 2026-05-11, License - CC BY-SA 4.0

\documentclass{article}
\begin{document}
\[\begin{array}{l}
foo

\end{array}\]
\end{document}
```

davidcarlisle:
we added a guard for currentgrouptype=6 (alignment groups) but that only works in simple cases, and not at all, in `array` where it's already a math group. (probably should open a new issue rather than comments on the closed one)

perhaps instead

```
\let\par\@empty\partokencontext=\z@
```

and in `\@arrayparbxrestore`

```
\def\@arrayparboxrestore{%
  \let\if@nobreak\iffalse
  \let\if@noskipsec\iffalse
  \let\par\@@par
  \partokencontext=\tw@
```

so that `\partokencontext` is only set when `\par` is not a no-op`

FrankMittelbach:
yes, I guess this is a safer approach (and you could just reopen this one rather than making a new one)

PR #1865, opened 2025-09-23
Partokencontext

davidcarlisle:
# Internal housekeeping

setting `\partokencontext=2` for #1864

## Status of pull request

- Ready to merge

- [X] Test file(s) added


base/doc/ltnews42.tex
```diff
@@ -430,6 +430,18 @@ \subsection{Support word exclusion in case changing}
...
 
+\subsection{Automatic insertion of \cs{par} tokens}
+Since 2022 the major \TeX\ engines have a parameter, \cs{partokencontext},
+that controls whether a \cs{par} token is added if \TeX\ is in horizontal mode
+at the end of \cs{vbox} and similar contexts. This gives more control
+than the classical behaviour where the internal \textit{end paragraph} routine
+is invoked with no explicit token being added.
+
+\LaTeX\ now sets this parameter to 2 by default to enable automatic insertion
+of \cs{par}.
```
josephwright, line 446: Do we want to say explicitly this allows para hooks and tagging to work more smoothly?

base/ltfinal.dtx
```diff
@@ -1304,6 +1304,31 @@
...
+% is invoked with no explicit token being added.
+%
+% Setting \cs{partokencontext} to 2 ensures a literal \cs{par} is
+% used in all contexts, setting to 0 reverts to the classic \TeX\ behavior.
+%    \begin{macrocode}
+%</2ekernel>
+%<*2ekernel|latexrelease>
+%<latexrelease>\IncludeInRelease{2025/11/01}%
+%<latexrelease>                 {\partokencontext}{automatic par insertion}%
+\ifx\partokencontext\@undefined\else \partokencontext=\tw@ \fi
```
josephwright:
```suggestion
\ifdefined\partokencontext\partokencontext=\tw@\fi
```

base/ltfinal.dtx
```diff
@@ -1304,6 +1304,31 @@
...
+%</2ekernel|latexrelease>
+%<latexrelease>\EndIncludeInRelease
+%<latexrelease>\IncludeInRelease{0000/00/00}%
+%<latexrelease>                 {\partokencontext}{automatic par insertion}%
+%<latexrelease>\ifx\partokencontext\@undefined\else \partokencontext=\z@ \fi
```
josephwright:
```suggestion
%<latexrelease>\ifdefined\partokencontext\partokencontext=\z@\fi
```

base/doc/ltnews42.tex
davidcarlisle: I wondered about that but it mostly improves things that have a direct use of vbox which shouldn't really be used anyway,. any documented vmode like parbox or minipage is already adding \par we could of course adjust the wording anyway. I see some tests are failing so I'll need to do a second pass anyway

josephwright: I was thinking more for the impact on packages, which an end user might be using with no insight into a raw `\vbox` being used. But I do see that this is all arguably 'not really user facing'.

base/ltfinal.dtx
davidcarlisle: I wondered if anyone would comment:-) I was thinking `\ifx` was more the latex.ltx style but I see we already have quite a few `\ifdefined`

base/doc/ltnews42.tex
FrankMittelbach: it improves the compatibility of older packages/classes with tagging code and in that respect it is worth making a note I think (or even put it in the section on tagging project)

base/testfiles/github-1864.lvt
```diff
@@ -0,0 +1,20 @@
...
+
+\begin{document}
+
+\START
+
+AAA
+
+BBB
+
+\vbox{CCC}
```
FrankMittelbach, line 20: you might as well test the other places like \vadjust and what have you here. Doesn't cost anything

base/doc/ltnews42.tex
davidcarlisle: I didn't put it in the tagging project section as it could affect documents not using that, but I'll add a note something like

> This allows the paragraph hooks to detect the end of paragraph even in contexts such as at the end of a \vbox, where traditionally package code has had to be modified to add an explicit \par. This is expected to improve compatibility of existing packages with the tagging code

base/ltfinal.dtx
FrankMittelbach: we sure have, but perhaps it is about time to clean these up while we go along

base/testfiles/github-1864.lvt
davidcarlisle: Oh this testing idea is a bad idea, I don't get what I expected

davidcarlisle: added guards to tabular version of `\par`  to avoid looping in `\noalign`

PR #1866, opened 2025-09-23
Skip hash doubling if not required in hook code

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/lthooks.dtx
+++ b/base/lthooks.dtx
@@ -32,7 +32,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{lthooks.dtx}
-             [2025/09/23 v1.1m LaTeX Kernel (hooks)]
+             [2025/09/24 v1.1n LaTeX Kernel (hooks)]
 % \iffalse
 %
 \documentclass{l3doc}
--- a/base/lthooks.dtx
+++ b/base/lthooks.dtx
@@ -3643,6 +3643,10 @@
 %     \@@_prop_gput_labeled_cleanup:nnn,
 %     \@@_prop_gput_labeled_do:Nnnn
 %    }
+% \begin{macro}[EXP]{\@@_hash_check:nTF}
+% \changes{v1.1n}{2025/09/23}{New function}
+% \begin{macro}[EXP]{\@@_hash_check_aux:w}
+% \changes{v1.1n}{2025/09/23}{New function}
 %
 %    With \cs{hook_gput_code:nnn}\Arg{hook}\Arg{label}\Arg{code} a
 %    chunk of \meta{code} is added to an existing \meta{hook} labeled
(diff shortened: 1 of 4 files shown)


josephwright:
> Does this need additional unit tests?

We have a lot of use of hooks, so I hope this is already covered. But if people want an explicit test its easy enough to add.

base/lthooks.dtx
```diff
@@ -3777,21 +3781,41 @@
...
+    \exp_not:N \exp_after:wN \exp_not:N \@@_hash_check_aux:w
+      \exp_not:N \tl_to_str:n {#1} \c_hash_str \c_hash_str
+      \exp_not:N \q_stop
+  }
+\use:x
+  {
+    \cs_new:Npn \exp_not:N \@@_hash_check_aux:w
+      ##1 \c_hash_str ##2 \c_hash_str ##3 \exp_not:N \q_stop
+  }
+  { \tl_if_blank:nTF {#3} }
```
FrankMittelbach: is this if_blank or rather if_empty (assuming without checking that the latter is slightly faster)?

FrankMittelbach (approved):

base/lthooks.dtx
```diff
@@ -3777,21 +3781,41 @@
...
         \exp_args:Nx \@@_prop_gput_labeled_cleanup:nnn
           {
             \@@_if_replacing_args:TF
               { \exp_not:n }
-              { \@@_double_hashes:n }
+              {
+                \@@_hash_check:nTF {#3}
+                  { \exp_not:n }
+                  { \@@_double_hashes:n }
+              }
```
muzimuzhi, line 3801: Just curious: do other uses of `\@@_double_hashes:n` need a similar filter?

base/lthooks.dtx
```diff
@@ -3777,21 +3781,41 @@
...
                 {#3}
           }
           {#1} {#2}
       }
   }
+\cs_new:Npx \@@_hash_check:nTF #1
```
muzimuzhi, line 3807: Is the `x`-type expansion preferable than `e`-type, in LaTeX2e base?

base/lthooks.dtx
josephwright: This is the only place where it is apply to the 'no arguments' branch - all other cases is where there (probably) is an argument to handle so doubling is needed.

base/lthooks.dtx
josephwright: We will at some stage want to switch from `x`- to `e`-type, but that will break any usage with older `expl3`. The ChangeLog for `l3kernel` has the switch marked as 2023-10-10, so we are still <2 years. I think we have to wait a while yet (likely the typical 5-year Linux distro cycle).

josephwright: I've had the same in `siunitx`, and decided it was still too early to switch - but there I can perhaps be more agresive -  two years is perhaps enough.

FrankMittelbach: Given that we are in kernel mode and the kernel can expect the commands to be around, I see no reason not to. But I guess it doesn't really matter either way here.

josephwright: @FrankMittelbach At the moment, the kernel doesn't _rely_ on anything in `expl3` with `:e` expansion - I'd be happy to change, but probably we do that globally (as I have done inside `l3kernel` itself). Perhaps for 2026-06-01?

davidcarlisle: @josephwright `\@onefilewithoptions` uses `\expanded` which is same consideration, more or less

josephwright: @davidcarlisle  Not exactly: we said we require `\expanded` some time ago, but something like `\cs_new:Npe` has only been available at all since October 2023.

muzimuzhi: > but something like `\cs_new:Npe` has only been available at all since October 2023.

Oh I missed that. This time, time passed so slowly.

PR #1868, opened 2025-09-25
double @@ for l3doc

davidcarlisle:
`\@@par` needs to be written as `\@@@@par` in yesterday's update, as it's within a `%<@@=tbl` region.


josephwright (approved):

required/tools/array.dtx
```diff
@@ -2527,7 +2527,7 @@ Bug reports can be opened (category \texttt{#1}) at\\%
 %    \changes{v2.6m}{2025/09/24}{Check \cs{currentgrouptype} in \cs{par} (gh1864)}
```
josephwright, line 2527: Should we have a `\changes` here?

FrankMittelbach: too embarrassing :-)

davidcarlisle: I'll add one

PR #1869, opened 2025-09-25
Tagging 994

u-fischer:
Disable \tagpdfsetup is tagging=off is used https://github.com/latex3/tagging-project/issues/994

## Status of pull request
- Ready to merge

- [x] Test file(s) added
- [n/a ] Rollback provided (if necessary)?
- [n/a ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated

### DIFF
--- a/required/latex-lab/documentmetadata-support.dtx
+++ b/required/latex-lab/documentmetadata-support.dtx
@@ -607,7 +607,7 @@
 % We disable the setup command, as various keys assume that the tagging structure is there,
 % see tagging-project issue \#994.
 %    \begin{macrocode}
-           \cs_set_eq:NN\tagpdfsetup\use_none:n
+           \cs_set_protected:Npn\tagpdfsetup##1{}
 %    \end{macrocode}
 %
 %    perhaps, need to test:
--- a/required/latex-lab/documentmetadata-support.dtx
+++ b/required/latex-lab/documentmetadata-support.dtx
@@ -18,8 +18,8 @@
 % for those people who are interested or want to report an issue.
 %
 %    \begin{macrocode}
-\def\documentmetadatasupportversion{1.0s}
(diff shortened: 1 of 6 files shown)


required/latex-lab/documentmetadata-support.dtx
```diff
@@ -603,6 +603,13 @@
 %    \begin{macrocode}
            \tag_suspend:n {global}
 %    \end{macrocode}
+% \changes{v1.0t}{2025/09/25}{Disable \cs{tagpdfsetup}}
+% We disable the setup command, as various keys assume that the tagging structure is there,
+% see tagging-project issue \#994.
+%    \begin{macrocode}
+           \cs_set_eq:NN\tagpdfsetup\use_none:n
```
josephwright: I'd prefer a `protected` version as this keeps the expansion behaviour of the command unchanged.

FrankMittelbach (approved):

required/latex-lab/documentmetadata-support.dtx
u-fischer: hm, yes perhaps. I have a tendency to prefer to get rid of the command as soon as possible but that naturally changes the behaviour e.g. in a tabular. @FrankMittelbach and thoughts?

davidcarlisle: probably protected would be better here, the example in the issue was with multirow but with multicolumn, `\tagpdfsetup{...}\multicolumn...` should probably fail with `tagging=off` so you find out early not just when you have `tagging=on`

FrankMittelbach: normally I'm also a fan of vanishing without trace, but given that we have a trace if tagging=on I think it should stop expansion also when it is off.

PR #1879, opened 2025-10-09
Block revisited

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [n/a] Test file(s) added


required/latex-lab/latex-lab-block.dtx
```diff
@@ -172,6 +174,33 @@
 %
 % \subsection{Template types}
 % 
+%
+% \begin{TemplateInterfaceDescription}{blockenv}
+%    \TemplateArgument{1}{key/value list to alter the default parameters of the template
+%       instances used by the particular blockenv environment}
+%    \TemplateArgument{2}{Caption text in cases this environment
```
davidcarlisle, line 181:
```suggestion
%    \TemplateArgument{2}{Caption text in case this environment
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1529,6 +1570,7 @@
                        { - \int_use:N \l_@@_inner_level_counter_tl }
                   }
                   \l_@@_unused_blockenv_keys_tl
+                  { #2 }
 %    \end{macrocode}
 %    Again the instance may has processed a few keys from the sofar
```
davidcarlisle, line 1573:
```suggestion
%    Again the instance may have processed a few keys from the sofar
```

davidcarlisle (approved):
a couple of editorial comments. added inline.

I still wonder a bit about pulling out captions as a distinguished argument, probably Ok but if we end up wanting more heading-like things we may need more or to access this via the kv argument (although that is probably inefficient if most environments don't use it) I think this is OK in this round, see how it turns out in practice

required/latex-lab/latex-lab-block.dtx
car222222: And "so far" is two words, is it not: or is this only for us oldies??

required/latex-lab/latex-lab-block.dtx
```diff
@@ -927,6 +934,41 @@
...
+\cs_new:Npn \__tl_if_NoValue:n #1 {
+  \__tl_if_empty_if:o { \__tl_if_NoValue:w {} #1 {} ? ! \NoValue ? ? ! }
+    \prg_return_true:
+  \else:
+    \prg_return_false:
+  \fi:
+}
+\cs_new:Npn \__tl_if_NoValue:w #1 \NoValue #2 ? #3 ? ! { #1 #2 }
+
+\cs_new_eq:NN \NoValue \c_novalue_tl
```
josephwright: All now in the kernel

josephwright (approved):
I think we do need to see if this 'sticks' (Ulrike's comments elsewhere), but at the moment the only way I think we find out is by trying it - so happy with the changes.

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1129,6 +1129,7 @@
 %   interfaces, and the template code is covered later.
 %    \begin{macrocode}
 \NewTemplateType{blockenv}{2}
+\NewTemplateType{simpleblockenv}{1}
```
u-fischer, line 1132: wouldn't it be easier to leave the original name `blockenv`  and to name the new one e.g. `blockenvcaptioned` or `xblockenv` or something like that`? Then no external packages would need an update ...

required/latex-lab/latex-lab-block.dtx
car222222: Nothing happened it seems! 
No response?

FrankMittelbach: precisely. I'm currently doing an experiment that I'm likely to revert and will add the corrections only after I'm sure in which directions I want to go.

car222222: The term sofar is derived from the words “sound, fixing and ranging.”

Is this acronym that to which you are referring?

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: maybe, but whether or not 3 packages have to change should not really influence the names we chose in the end. Once it is clear that we go in the direction of a split (which I still don't really like) then final names can be decided which may or may not mean packages need to change.

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: I'm refering to the fact that making the edits now and then thowing away these sets of comits means I also undo the edits

car222222: Understood! 
Sorry for the noise: 
I had thought that @davidcarlisle was simply ignoring me, as usual:-)

FrankMittelbach: What has David to do with this? It was my typo that you called out. And really, nobody is ignoring you "as usual"  with or without smiley and certainly not David despite the repeated claims. Please get real Chris.

car222222: Only that David had made some changes/suggestions to that sentence. 
And so: Reasonably real ... maybe.
Nothing to get in such a tizzie about! 
But not worth fussing about it!

FrankMittelbach:
replaced by a new pull

Issue #1099, opened 2023-06-25
generic hook not available for expl3 commands if the replacement text contains :~ 


u-fischer:
## Brief outline of the bug

In the following example one gets and error if one try to use the generic hook of `\testB`:

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}    
\documentclass{article}
\ExplSyntaxOn %required
\def\testA#1{hallo:}
\def\testB#1{hallo:~}

\AddToHook{cmd/testA/before}{\tiny}
\AddToHook{cmd/testB/before}{\tiny}

\ExplSyntaxOff

\begin{document}
blub
\end{document}
```

gives 

~~~~
! LaTeX hooks Error: Generic hooks cannot be added to '\testB'.
~~~~

## Log file (required) and possibly PDF file  

[test-utf8.log](https://github.com/latex3/latex2e/files/11860052/test-utf8.log)

FrankMittelbach:
Well, the original example fails because of the `:` having catcode 11 and later on catcode 12 when the patching happens.

The problem here seems to me that something like `hello:~` is actually wrong as it produces the word `hello:` not `hello` followed by the punctuation `:`  (which is what is intended I would assume if a command contains text to be used later on). And expl3 doesn't offer a solution for this and I'm not sure how it could easily. 

We can't use `\:` because that has a meaning in text. We could perhaps use `\?`  or `\<` but that wouldn't be pretty and furthermore it would likely break some documents that also discovered that these control symbols aren't in use.
We could offer `\textcolon` but that has the disadvantage that you need to write `\textcolon\space` most of the time.

@PhelypeOleinik any thoughts?

If we aren't going in that direction but accept that we end up with catcode 11 colons in text inside `\ExplSyntaxOn` then perhaps the patching could make a second pass before giving up. If it can't patch it could try to patch once more with  `:` having catcode 11.

u-fischer:
> Well, the original example fails because of the `:` having catcode 11 and later on catcode 12 when the patching happens

Yes something like this. What still eludes me, is why patching a definition with replacement text `hallo:world` and `hallo~world`  works, but patching fails, if both are present `hallo~world:earth`.

FrankMittelbach:
as a guess: because the moment the definition contains spaces a different patching code is used?

PhelypeOleinik:
[...]

PR #1881, opened 2025-10-12
Gh1099

davidcarlisle:
reset catcode of space when patching to add hook code for #1099 

# Internal housekeeping

## Status of pull request

- Ready to merge

- [X] Test file(s) added
- [ ] Rollback provided (if necessary)? No

### DIFF
--- a/base/doc/ltnews42.tex
+++ b/base/doc/ltnews42.tex
@@ -712,6 +712,8 @@ Previously attempts to add hooks to commands would fail if the original definiti
 had used  \verb|~| in an \verb|\ExplSyntaxOn| context.
 %
 \githubissue{1099}
+
+
 \section{Bug fixes}
 
 \subsection{Support active characters correctly with \cs{DeclareRobustCommand}}


base/testfiles/github-1099.lvt
```diff
@@ -0,0 +1,25 @@
...
+\ExplSyntaxOn %required
+\def\testA#1{hallo:}
+\def\testB#1{hallo:~}
+
+\START
+
+\AddToHook{cmd/testA/before}{\tiny}
+\AddToHook{cmd/testB/before}{\tiny}
+
+\ExplSyntaxOff
```
FrankMittelbach, line 14: shouldn't that work with \ExplSyntaxOff before doing the \AddToHook?

base/changes.txt
```diff
@@ -6,6 +6,11 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2025-10-12  David Carlisle  <David.Carlisle@latex-project.org>
+
+	* ltcmdhooks.dtx (subsubsection{Patching by retokenization}):
+	Reset the catcode of space (gh/1099)
```
FrankMittelbach: that comment isn't understandable without looking at the code.

.. when patching commands written using L3 prog layer commands.

or something.

davidcarlisle:
> also should have ltnews entry (I think)

done

base/testfiles/github-1099.lvt
FrankMittelbach: a I see, that explicitly tests the case where we are *still* in ExplSyntaxOn context.

base/changes.txt
davidcarlisle: well I had left in the emacs-added subsection reference about patching, but OK:-)

base/doc/ltnews42.tex
```diff
@@ -703,6 +703,15 @@ \subsection{Automatic insertion of \cs{par} tokens}
...
 
+\subsection{Improved access to Generic Hooks}
+The code to add generic hooks such as
+\verb|\AddToHook{cmd/somecmd/before}{...}|
+has been improved so that it is more likely to succeed
+in cases where the command has been defined using \emph{expl3} syntax.
+Previously attempts to add hooks to commands would fail if the original definition
+had used  \verb|~| in an \verb|\ExplSyntaxOn| context.
+%
+\githubissue{1099}
```
FrankMittelbach, line 714: a few empty lines are appreciated  (by me) in places like that

base/testfiles/github-1099.lvt
davidcarlisle: not sure what you mean, do you mean it would have worked before with ExplSyntaxOn or that it should have it now?  It wouldn't have worked previously whether or not ExplSyntaxOn was used here as the space token in the definition would be dropped on the rescan.

base/changes.txt
FrankMittelbach: sure, but it means you actually have to go there to see what is meant (most of the time we actually manage to have the changes.txt entry  written so that you can deduce what i going one more or less)

base/testfiles/github-1099.lvt
FrankMittelbach: it mean I didn't got what was tested :-)

base/doc/ltnews42.tex
davidcarlisle: added

PR #1899, opened 2025-11-03
Hotfix listings microtype

u-fischer:

# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added


.github/tl_packages
```diff
@@ -47,6 +47,10 @@ tugboat
 ulem
 unicode-math
 varwidth
+latex-lab
```
FrankMittelbach: As I said elsewhere I guess it is wrong to make a dependency to latex-lab on that level (mixing released versions with development code). That would be okay in this case, especially on main, but going forward I could see a firstaid on -dev with a dependency on latex-lab also on -dev and then things fail

josephwright: I see what you are getting at but there's no mixing of development and release code: CTAN hold `latex-lab` and `latex-lab-dev`, and the former only gets updated when the kernel does.

FrankMittelbach: for this hotfix on main, there is no difference, but for anything on dev there is: latex-lab from the released version is then added during testing, but latex-lab from the development version may be needed (not here but in general). In that case you would have to add that fix to latex-lab-firstaid and you could not add it to firstaid until the current dev made it into the main release. Could all be done this way, but my feeling is that the dependencies are unclear in that case, but perhaps I see an issue where there is none

FrankMittelbach (approved):

josephwright (approved):

PR #1903, opened 2025-11-05
Use/test \NoValue rather than "-NoValue-"

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/ltcmd.dtx
+++ b/base/ltcmd.dtx
@@ -1971,7 +1971,7 @@
 %   arguments, pad the defaults |#2| with some \cs{NoValue}
 %   until there are as many as embellishments~|#1|.  These functions are
 %   also used when defining expandable commands. The \cs{prg_do_nothing:} here
-%   are removed in \cs{\@@_defaults_def:nnn}: for a user-supplied default the
+%   are removed in \cs{@@_defaults_def:nnn}: for a user-supplied default the
 %   token is injected to avoid brace stripping, and for \cs{NoValue} means
 %   we have a common code path later.
 %    \begin{macrocode}


base/doc/ltnews43.tex
```diff
@@ -158,6 +158,24 @@ \section{New or improved commands}
 
 \section{Code improvements}
 
+\section{Revision of handling of \enquote{no value} concept}
+
+The commands \cs{NewDocumentCommand}, etc., introduce the idea of
+differentiating an absent optional argument from one which is simply empty.
+When an argument is entirely empty, it is given the special \enquote{no value}
```
muzimuzhi:
```suggestion
When an argument is entirely absent, it is given the special \enquote{no value}
```

base/doc/usrguide.tex
```diff
@@ -413,19 +413,13 @@ \subsection{Testing special values}
...
 When two optional arguments follow each other (a syntax we typically
 discourage), it can make sense to allow users of the command to
 specify only the second argument by providing an empty first
 argument.
 \NEWdescription{2022/06/01}
 Rather than testing separately for emptiness and for
-|-NoValue-| it is then best to use the argument type~|O| with an
+\cs{NoValue} it is then best to use the argument type~|O| with an
 empty default value, and then test for emptiness using the
 conditional \cs{IfBlankTF} (described below) instead.
```
FrankMittelbach: not sure I understand that suggestion (not that it is new)

base/ltcmd.dtx
```diff
@@ -1989,7 +1985,7 @@
 \cs_new_protected:Npn \@@_add_default:
   {
     \int_incr:N \l_@@_current_arg_int
-    \tl_put_right:Nn \l_@@_defaults_tl { \c_novalue_tl }
+    \tl_put_right:Nn \l_@@_defaults_tl { { \prg_do_nothing: \NoValue } }
```
FrankMittelbach, line 1991: the do nothing is there because later there is an "o" expansion? In  other words it could be simplified perhaps?

base/doc/ltnews43.tex
```diff
@@ -158,6 +158,24 @@ \section{New or improved commands}
...
+tokens, which are therefore hard to input accidentally.
+
+Whilst this allowed us to easily detect \enquote{no value}, it turns out there
+are places we want to be able to add such a value. This comes up particularly
+in creating templates for some parts of the document structure as part of the
+wider tagging project.
+
+We have therefore changed the approach to use a marker token, \cs{NoValue}, and
+updated the \cs{IfNoValue(TF)} test and relatives. This means you now
+\emph{can} type in an optional argument that is \enquote{not there}, but this
```
FrankMittelbach: that is \enquote{interpreted as not being present}, but ...

perhaps?

base/ltcmd.dtx
```diff
@@ -2978,7 +2976,7 @@
         \@@_split_signature_loop:Nw
       }
   }
-\exp_args:NV \@@_tmp:w \c_novalue_tl
+\@@_tmp:w \NoValue
```
FrankMittelbach, line 2982: why is that needed? wasn't that tmp business only there to get the content of c_novalue_tl into the definition?

ah I see, that later you had removed them, so probably overlooked

josephwright: No, I left this one as we need 9 copies - so in this one case it's clearer (to me) to keep it

base/ltcmd.dtx
josephwright: There is, but it's a long way down the process: see the line a little above where we have the same construct for user-supplied default.

FrankMittelbach: it just feels wrong (or convoluted) but I might be mistaken

josephwright: I haven't traced it all, but I think the `\prg_do_nothing:` is there to allow for the user default being a brace group, to avoid stripping braces. I'll make sure I check on that and add a comment.

base/doc/usrguide.tex
josephwright: We are saying that rather than use `o` and test for `\NoValue` then test for blank, just use `O` and test purely for blank.

FrankMittelbach: but if the spec say "oo" then a user could use `[\NoValue][3pt]` would that be really bad?

base/ltcmd.dtx
FrankMittelbach: If that's the reason then it wouldn't be needed in line 1988 since the input there is fixed to \NoValue, no?

josephwright: No, you still need it otherwise `\NoValue` would be expanded. The expansion takes place as part of `\@@_defaults_def:nnn`, which applies `\exp_not:o` to whatever default is supplied if needed. Where the user supplies the argument, `\prg_do_nothing: { foo }` turns into `{ foo }` and all is good. In the past, `\c_novalue_tl` turned into `-NoValue-` and again all worked. But in switching to `\NoValue` I've tried to avoid using `\c_novalue_tl` as it becomes a distraction. So we need the `\prg_do_nothing:` to avoid `\NoValue` being expanded - we are applying `o`-type not `e`-type here.

base/doc/usrguide.tex
josephwright: Up to now, we've been clear that _any_ optional argument has a 'value', as `-NoValue-` is deliberately not something the user can type in. Allowing a predictable token does mean you can revisit that, but I'd say it's meant to be reserved for programmers - I don't like the idea of `\foo[\NoValue][bong]` claiming that `#1` was not given, not least because no other parser would give that result.

josephwright:
> up to you if you want to make any of the suggested changes, none of them are showstoppers for merging I guess.

I'll wait for @davidcarlisle to take a look too - it's a big change so I would like a couple of people to have given it a once over.

davidcarlisle (approved):
The comment diffs all look good, 

I have looked at the code diffs and they look reasonable but haven't checked every line of `\@@_expandable_grab_R_alt_aux:NNwNNn` and friends:-) 

However the test suite passes, so I think this is good to merge.

PR #1906, opened 2025-11-10
Double\alignmarks in hook code

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [x] Test file(s) added


base/ltcmdhooks.dtx
```diff
@@ -1314,6 +1314,28 @@
...
+% \changes{v1.0m}{2025-11-10}{Support \cs{alignmark}}
+%    \begin{macrocode}
+\cs_new:Npn \@@_double_hashes_output_aux:N #1
+  {
+    \if_catcode:w ## \exp_not:N #1
+      \exp_after:wN \use_ii:nnnn
+    \fi:
+  }
+\sys_if_engine_luatex:T
+  {
```
FrankMittelbach, line 1330: I would welcome a note why luatex needs its own code path here.

josephwright: Done

josephwright:
> I think there should be a test that reflects that \alignmark is now working in this respect, i.e., one that would fail in earlier kernels

At the moment, our tests for hooks are all pdfTeX-only - how do you want me to handle this?

FrankMittelbach:
> > I think there should be a test that reflects that \alignmark is now working in this respect, i.e., one that would fail in earlier kernels
> 
> At the moment, our tests for hooks are all pdfTeX-only - how do you want me to handle this?

what I normally do is to write a test that fails prior to adding a fix so that it then shows as resolved. If the testfiles-lthook... only run pdftex then perhaps you can just stick such a test into the main testfiles dir (as github-whatever-number-it was.lvt)

FrankMittelbach (approved):

josephwright:
@davidcarlisle Ping ...

muzimuzhi:
Late to the Party. Need a ltnews entry?

josephwright:
This is a very niche change: I dont think we need to cover it

PR #1908, opened 2025-11-12
Gh1841

u-fischer:
## Status of pull request

- Ready to merge

- [ ] Test file(s) added (no, a meaning full test would need titlesec to demonstrated that is no longer loops)
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated. No, I don't think it is needed?

### DIFF
--- a/base/testfiles/github-0944.tlg
+++ b/base/testfiles/github-0944.tlg
@@ -22,7 +22,7 @@ LaTeX Warning: Reference `unknown' on page 1 undefined on input line ....
 ] (github-0944.aux)
  ***********
 -formatname- <-formatversion-> pre-release-2 (-release-version-test- branch)
-L3 programming layer <2025-11-06>
+L3 programming layer <2025-11-14>
  ***********
  *File List*
  -show the file list here-
--- a/texmf/tex/latex/l3kernel/expl3-code.tex
+++ b/texmf/tex/latex/l3kernel/expl3-code.tex
@@ -79,7 +79,7 @@
 %% and all files in that bundle must be distributed together.
 %% 
 %% File: expl3.dtx
-\def\ExplFileDate{2025-11-06}%
+\def\ExplFileDate{2025-11-14}%
 \begingroup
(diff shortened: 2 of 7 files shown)


FrankMittelbach:
>     * [ ]  Test file(s) added (no, a meaning full test would need titlesec to demonstrated that is no longer loops)

we do have testfiles-search which can hold such tests (just a thought)

base/ltxref.dtx
```diff
@@ -398,21 +399,35 @@
...
-%<latexrelease>                 {\label}{store five label arguments}%
-\providecommand\@currentlabelname{}
-\providecommand\@currentHref{}
-\providecommand\@kernel@reserved@label@data{}
-\NewHookWithArguments{label}{1}
+%<latexrelease>\IncludeInRelease{2025/11/10}%
+%<latexrelease>                 {\label}{disable commands while writing}%
 \def\label#1{\@bsphack
   \begingroup
   \UseHookWithArguments{label}{1}{#1}%
```
FrankMittelbach, line 406: looking at this I wonder if the group is actually necessary as \protected@write has as group and the \UseHookWith... could therefore perhaps just go into the second argument there --- not sure though.

base/ltxref.dtx
```diff
@@ -398,21 +399,35 @@
...
-  \protected@write\@auxout{}%
+  \protected@write\@auxout{%
+    \let\label\@gobble@om
+    \let\index\@gobble@som
+    \let\glossary\@gobble@om}%
```
FrankMittelbach, line 414: given that a redefinition for \label inside \label appears to be rather odd on first glance, I would appreciate a comment what this solves or why that might be necessary. I a could of years nobody will understand that code otherwise.

u-fischer: I added a comment and a testfile

FrankMittelbach: don't see either. not uploaded?

davidcarlisle (approved):

Issue #877, opened 2022-06-24
\cmd from ltxdoc not compatible with usage in section titles if hyperref is loaded


jfbu:
This is motivated from me seeing #875 and having recently dealt with keeping from ltxdoc only some parts and not using hypdoc.

hypdoc (Version: 2021-11-15 v1.15 where I am currently) does 

```latex
\pdfstringdefDisableCommands{%
%...
  \def\cs#1{\textbackslash #1}%
%...
}
```
which makes `\cs` usable in section titles in presence of hyperref.

But it has no provision for `\cmd` (and `\cmd@to@cs`) injection of `\char`.  As #875 modifies `\cmd` for fix of #874, perhaps it is opportunity to make it hyperref compatible.

FrankMittelbach:
Please always include a test file even for enhancement suggestions (and when you think it is simple), that helps us, thanks

jfbu:
> Please always include a test file even for enhancement suggestions (and when you think it is simple), that helps us, thanks

ok, sorry. I guess my file was something like
```latex
\documentclass{ltxdoc}
\usepackage{hyperref}
\begin{document}
\tableofcontents

\section{\cmd{\foo}}

\end{document}
```

u-fischer:
I don't really know where \cmd is all used. Would it work to base it on `\cs_to_str:N`? That would work in bookmarks to:

~~~~
\documentclass{ltxdoc}

\ExplSyntaxOn
\def\cmd#1{\texttt{\textbackslash\cs_to_str:N#1}}
\ExplSyntaxOff
\usepackage{doc}
\begin{document}

\tableofcontents

\section{\cs{blub} \cmd{\section}}

\cmd{\blub}

\end{document}
~~~~

FrankMittelbach:
We do not use `\textbackslash` deliberately as that has  a bad definition in OT1 (using the math character) and a lot of documentation  out there is in OT1. It might be better to just do
```
\makeatletter
\pdfstringdefDisableCommands{%
  \def\cmd#1{\textbackslash\expandafter\@gobble\string#1}%
}
\makeatother
```

u-fischer:
> We do not use `\textbackslash` deliberately as that has a bad definition in OT1

ah. That is the reason. Then I would suggest something like that

~~~~
\providecommand\texorpdfstring[2]{#1}
\ExplSyntaxOn
[...]

PR #1911, opened 2025-11-18
handle cmd in bookmarks, issue #877

u-fischer:
## Status of pull request

- Ready to merge

- [x] Test file(s) added
- [ ] Rollback provided (if necessary)? imho not needed.
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated imho not needed.

### DIFF
--- a/base/ltxdoc.dtx
+++ b/base/ltxdoc.dtx
@@ -30,7 +30,7 @@
 %
 %<class>\NeedsTeXFormat{LaTeX2e}
 %<class>\ProvidesClass{ltxdoc}
-%<class>         [2024/02/08 v2.1j Standard LaTeX documentation class]
+%<class>         [2025/11/18 v2.1k Standard LaTeX documentation class]
 %
 %<*driver>
 \documentclass{ltxdoc}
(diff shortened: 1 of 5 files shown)


base/ltxdoc.dtx
```diff
@@ -300,10 +300,17 @@
...
 %\DeclareRobustCommand\cs[1]...                       % defined later
 %\def\cmd#1{\cs{\expandafter\cmd@to@cs\string#1}}     % can't use with new \cs 
 \def\cmd#1{\texttt{\char`\\\expandafter\cmd@to@cs\string#1}}
+%    \end{macrocode}
+% hyperref allow to fill the pdfstringdef-hook before the package is loaded:
+%    \begin{macrocode}
+\providecommand\pdfstringdefPreHook{}
+\g@addto@macro\pdfstringdefPreHook{%  
+  \def\cmd#1{\textbackslash\expandafter\@gobble\string#1}}%
 \def\cmd@to@cs#1#2{\char\number`#2\relax}
```
FrankMittelbach, line 314: looks all okay, but wouldn't it be better to simply define cmd always in that way?

u-fischer: öh sorry I don't understand. The issue says that \textbackslash should not be used for typesetting and the code here is literally your suggestion from the issue?

FrankMittelbach: ah, forgot about the fact that we don't use \textbackslash to support OT1 cmtt fonts properly. So ignore the remark.

PR #1916, opened 2025-11-21
Gh1175

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/ltcmdhooks.dtx
+++ b/base/ltcmdhooks.dtx
@@ -274,8 +274,8 @@
 %
 % Command hooks are intended to be added to document commands, which are
 % typically defined globally. As such, adding a hook is a global operation,
-% even if the command was previously only locally-defined. The
-% \emph{content} of hooks respect \TeX{} groupings
+% even if the command was previously only locally-defined. Addition of
+% material to hooks is also global.
 %
 % \subsection{Commands that look ahead}
 % \label{sec:look-ahead}


base/ltcmdhooks.dtx
```diff
@@ -255,6 +255,25 @@
...
+% include the hook in the copy, for example
+% \begin{verbatim}
+%   \NewCommandCopy\baz\foo
+% \end{verbatim}
+% would include the hook in this copied command.
+%
+% \subsection{Grouping}
+%
+% Command hooks are added respecting \TeX{} grouping, such that any hooks
+% defined inside a group will only apply within that group.
```
muzimuzhi: This is wrong. Patching a command (adding `\UseHookWithArguments...` to a command body) _is_ global.

```tex
\documentclass{article}
\begin{document}
\def\foo{foo}
{
  \AddToHook{cmd/foo/after}{[after]}
  \foo % "foo[after]"
}
\foo % "foo[after]"
\end{document}
```

From source code, `\__hook_def_cmd:w` is let to one of `\xdef` or `\gdef`, both act global.
https://github.com/latex3/latex2e/blob/7c825d41be4fab5f0fdd7fd18c6d180febc2d41e/base/ltcmdhooks.dtx#L1026-L1044

muzimuzhi: > Patching a command (adding `\UseHookWithArguments...` to a command body) _is_ global.

This also means that the patching unconditionally makes a command globally defined, which might worth mentioning in the doc.

```tex
\documentclass{article}
\begin{document}
\foo % undefined
{
  \def\foo{foo}%
  \AddToHook{cmd/foo/after}{[after]}% make \foo defined globally
  \foo % "foo[after]"
}
\foo % "foo[after]"
\end{document}
```

base/ltcmdhooks.dtx
```diff
@@ -255,6 +255,27 @@
...
+%   \NewCommandCopy\baz\foo
+% \end{verbatim}
+% would include the hook in this copied command.
+%
+% \subsection{Grouping}
+%
+% Command hooks are intended to be added to document commands, which are
+% typically defined globally. As such, adding a hook is a global operation,
+% even if the command was previously only locally-defined. The
+% \emph{content} of hooks respect \TeX{} groupings
```
muzimuzhi: > The _content_ of hooks respect TeX groupings

Still... ambiguous. `\AddToHook` and `\hook_gput_code:nnn` both operate globally.

FrankMittelbach: and they shouldn't. The addition of the hook itself should be global i.e. the code change to the underlying command, but the addition of the hook code chunk should be local (but your example shows that this doesn'T seem to be the case, which to  me seems to be a bug not really a feature

josephwright: Hmm, we called it `gput` so 'm not sure that follows - at this stage I think document what we have

base/ltcmdhooks.dtx
```diff
@@ -274,8 +274,8 @@
 %
 % Command hooks are intended to be added to document commands, which are
 % typically defined globally. As such, adding a hook is a global operation,
-% even if the command was previously only locally-defined. The
-% \emph{content} of hooks respect \TeX{} groupings
+% even if the command was previously only locally-defined. Addition of
+% material to hooks is also global.
 %
```
mrpiggi, line 279: Just curious whether you want to mention, that it would still be possible to achieve different local behavior within a group by one level of indirection
```latex
\NewDocumentCommand\foo{foo executed}{}
\newcommand\afterfoo{with global hook}
\AddToHook{cmd/foo/after}{\afterfoo}

\foo

\begingroup
\renewcommand\afterfoo{with grouping hook}
\foo
\endgroup
```

base/ltcmdhooks.dtx
FrankMittelbach: yes my mistake, forgot that hooks are different and handled globally

PR #1931, opened 2025-12-08
Evaluate section level before structure names

carhensi:
When using KOMA-Script classes (`scrartcl`) with `\DocumentMetadata` and tagging enabled, internal LaTeX3 variable names like `\l__tag_name_sec/\numexpr 1\relax` appear as structure types in the PDF instead of being mapped to standard PDF tags. This causes PDF/UA-1 validation failures (Clause 7.1, Test 5).

**Root cause:** KOMA-Script passes `\numexpr 1\relax` (unexpanded) as the section level parameter. When `latex-lab-sec.dtx` constructs structure names using `\UseStructureName{sec/#1}`, the `\numexpr` expression becomes part of the variable name without being evaluated.

**Context:** I found this issue while testing with KOMA-Script classes (latex3/tagging-project#1118), which are [not yet officially compatible](https://latex3.github.io/tagging-project/tagging-status/#classes) with tagging. However, this fix hopefully makes the section tagging infrastructure more robust by properly evaluating numeric expressions, which will benefit the codebase when KOMA-Script and other classes become compatible in the future. The underlying issue is not KOMA-specific but affects any document class that passes unexpanded `\numexpr` expressions as section level parameters.

## Solution

Evaluate the section level parameter `#1` using `\int_eval:n {#1}` before passing it to `\UseStructureName`. This ensures that expressions like `\numexpr 1\relax` are converted to `1` before string concatenation.

## Testing

**Minimal test case:**

```
\RequirePackage{latexbug}
\DocumentMetadata{
 pdfstandard = {a-3u, ua-1},
 tagging     = on,
 testphase   = phase-III
}
\documentclass{scrartcl}
\tagpdfsetup{role/map-tags=pdf}
\begin{document}
\section{Test}
Text.
\end{document}
```

**Before fix:** 18 `tag_name_sec` warnings per section  
**After fix:** 0 warnings

**Impact:** No effect on standard classes (`article`, `book`, `report`) which already pass integer values. Makes section tagging robust against unexpanded numeric expressions.

---

- [x] Test file(s) added (`testfiles-sec/koma-section-tagging.lvt`)

### DIFF
--- a/required/latex-lab/changes.txt
+++ b/required/latex-lab/changes.txt
@@ -1,3 +1,7 @@
+2025-12-08  Carsten Hensiek  <carsten@hensiek.com>
+	* latex-lab-sec.dtx: evaluate section level parameter before constructing
+	structure names to fix KOMA-Script compatibility
+
 2025-12-02  Ulrike Fischer  <Ulrike.Fischer@latex-project.org>
 	* latex-lab-math.dtx: add support for the MSFT-attribute
 
--- a/required/latex-lab/latex-lab-sec.dtx
+++ b/required/latex-lab/latex-lab-sec.dtx
@@ -16,8 +16,8 @@
 %
 % for those people who are interested or want to report an issue.
 %
-\def\ltlabsecdate{2025-10-20}
-\def\ltlabsecversion{0.84k}
+\def\ltlabsecdate{2025-12-08}
+\def\ltlabsecversion{0.84l}
 %    \end{macrocode}
 %<*driver>
 \DocumentMetadata{tagging=on,pdfstandard=ua-2}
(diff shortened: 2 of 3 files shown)


u-fischer:
I can not reproduce your problem with a current KOMA class. As it not longer uses `\@startsection` it doesn't see the sockets and structure names and it doesn't tag the sectioning commands at all. 

With an older KOMA I can see the error, but imho a better fix is to replace the `\numexpr` by `\inteval` in `\scr@startsection` and not to add `\int_eval:n` in lots of places.

~~~~
\DocumentMetadata{
  tagging     = on,
}
\documentclass{scrartcl}
\tagpdfsetup{role/map-tags=pdf}
\usepackage{etoolbox}
\makeatletter
\patchcmd\scr@startsection{\def\startsection@secnumdepth{\numexpr #2\relax}}
{\edef\startsection@secnumdepth{\inteval{#2}}}{}{\fail} % or \def, that should KOMA decide
\makeatletter
\begin{document}
\section{Test Section}
Test text.
\subsection{Test Subsection}
More text.
\end{document}
~~~~

Side remark: the testfile in your PR doesn't compile, as latexbug errors and regression-test is not loaded.

FrankMittelbach:
One additional comment: we are currently in the process of replacing the internal section implementation; the new one will be based on templates (and then also used at some point by KOMA classes). This means that changing what is there right now is unfortunately of little value.

PR #1938, opened 2025-12-17
Block revisited version 2

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added

News entry still needs to be provided

### DIFF
--- a/required/latex-lab/latex-lab-block.dtx
+++ b/required/latex-lab/latex-lab-block.dtx
@@ -9,7 +9,7 @@
 %
 %    https://www.latex-project.org/lppl.txt
 %
-\def\ltlabblockdate{2025-12-21}
+\def\ltlabblockdate{2025-12-24}
 \def\ltlabblockversion{0.9l}
 
 %<*driver>
@@ -1252,20 +1252,27 @@
 %
 %
 %  \begin{environment}{verbatim,verbatim*}
-%    Here are the definitions for the verbatim environments which, on this
-%    level, looks like everything  else encountered so far.
+%    Here are the definitions for the verbatim environments They look
+%    somewhat different than others (but this isn't the final
+%    definition).
+%    At the moment we use 2 optional arguments, the second is only
+%    there so that there is yet another scan even if one optional
+%    argument got detected. That then scans away the newline so that
+%    afterwards we can reinsert one via \cs{obeyedline}. A better
+%    solution will be to use a \texttt{c} specifier for grabbing the
+%    body, but that is for another day not Christmas Eve.\fmi{fix}
 %  \changes{v0.9f}{2025/06/21}{Move special verbatim code into a key}
 %    \begin{macrocode}
 \AddToHook{begindocument/before}[./legacy-verbatims]{
-  \RenewDocumentEnvironment{verbatim}{ ={legacy-code} !o }
-    { \SimpleBlockEnv{verbatim} {#1} } { \BlockEnvEnd }
+  \RenewDocumentEnvironment{verbatim}{ ={legacy-code} !o !o }
+    { \SimpleBlockEnv{verbatim} {#1} \obeyedline } { \BlockEnvEnd }
 %    \end{macrocode}
 %    
 %  \changes{v0.9f}{2025/06/21}{Use separate blockenv instances for
 %    verbatim and verbatim*}
 %    \begin{macrocode}
-  \RenewDocumentEnvironment{verbatim*}{ ={legacy-code} !o }
-    { \SimpleBlockEnv{verbatim*} {#1} } { \BlockEnvEnd }
+  \RenewDocumentEnvironment{verbatim*}{ ={legacy-code} !o !o }
+    { \SimpleBlockEnv{verbatim*} {#1}  \obeyedline } { \BlockEnvEnd }
 %    \end{macrocode}
 % \end{environment}
 %
@@ -1275,7 +1282,8 @@
 %    handling where backslash and braces retain their normal
 %    meanings. We also reimplement it using the template approach    
 %    The \env{alltt*} variant didn't exist in the package,
-%    but it is trivial to set it up as well.
+%    but it is trivial to set it up as well.\fmi{The parsing here
+%    should be adjusted as well, eventually.}
 %    \begin{macrocode}
   \NewDocumentEnvironment{alltt}{ ={legacy-code} !o }
     { \SimpleBlockEnv{alltt} {#1} } { \BlockEnvEnd }
@@ -2870,7 +2878,8 @@
 %    e.g., \verb=...\end{itemize} \item ...\par= it can happen that
 %    the socket code is called while
 %    \texttt{@endpe} is already handled and then we should not attempt
-%    to close a \struct{\UseStructureName{para/semantic}} structure). So we need to check for this.
+%    to close a \struct{\UseStructureName{para/semantic}} structure).
+%    So we need to check for this.
 %    \begin{macrocode}
        \legacy_if:nT { @endpe }
          {
@@ -3851,13 +3860,20 @@
 %    \begin{macrocode}
   ,fixed-word-spaces     = \l__par_fixed_word_spaces_bool  % name??
   ,final-hyphen-demerits = \finalhyphendemerits
-  ,newline-cmd           = \\
+  ,newline-cmd           = \l_@@_newline_cmd_tl
   ,para-attr-class       = \l__tag_para_attr_class_tl                          
 }
 {
   \template_debug_typeout:n{~\space template:~ 'std';~
                             argument:~ \exp_not:n{|#1|}}
   \SetTemplateKeys{para}{std}{#1}
+%    \end{macrocode}
+%    Is that the right aproach?\fmi{revisit}
+%    \begin{macrocode}
+  \exp_args:NNo \cs_set_protected:Npn \\ { \l_@@_newline_cmd_tl }
+%    \end{macrocode}
+%
+%    \begin{macrocode}
   \skip_set:Nn \@rightskip \rightskip
 }
 %    \end{macrocode}
@@ -6199,7 +6215,6 @@
 %
 %
 %
-%
 %  \begin{macro}{\@setupverbinvisiblespace}
 %  ^^A
 %    In the \pdfTeX{} engine we need to use \tn{pdffakespace} chars
@@ -7023,7 +7038,7 @@
 %
 %
 %
-% \subsection{Support for \pkg{enumitem} interfaces}
+% \subsection{Support for the \pkg{enumitem} interfaces}
 %
 %    The current implementation  incorporates most features of
 %    \pkg{enumitem}. The plan is that the enumitem interfaces are
@@ -7054,6 +7069,23 @@
 %    \texttt{enumitem.sty} because it is only needed if the package is
 %    explicitly loaded in existing documents.
 %    
+%
+% \subsection{Support for the \pkg{doc} package}
+%
+%    When the \pkg{doc} package is loaded it wants to remove a
+%    \texttt{\%} sign from the start of each verbatim line. For this
+%    it uses \cs{check@percent} which we stick into the
+%    \socket{verbatim/startline} socket.
+%
+% \begin{plugdecl}{doc}
+%    \begin{macrocode}
+\NewSocketPlug {verbatim/startline}{doc}{ \check@percent }
+\AddToHook{package/doc/after}{
+  \AssignSocketPlug{verbatim/startline}{doc}
+}
+%    \end{macrocode}
+% \end{plugdecl}
+%
 %    \begin{macrocode}
 %</package-finish>
 %    \end{macrocode}
--- a/required/latex-lab/testfiles-block/blocks-verbatim-02.luatex.tlg
+++ b/required/latex-lab/testfiles-block/blocks-verbatim-02.luatex.tlg
@@ -220,6 +220,7 @@ Completed box being shipped out [1]
 ....\OT1/cmtt/m/n/10 a
 ....\OT1/cmtt/m/n/10 b
 ....\OT1/cmtt/m/n/10 c
+....\hbox(0.0+0.0)x0.0, direction TLT
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
 ....\glue(\rightskip) 0.0
@@ -279,6 +280,7 @@ Completed box being shipped out [1]
 ....\OT1/cmtt/m/n/10 d
 ....\OT1/cmtt/m/n/10 e
 ....\OT1/cmtt/m/n/10 f
+....\hbox(0.0+0.0)x0.0, direction TLT
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
 ....\glue(\rightskip) 0.0
@@ -352,6 +354,7 @@ Completed box being shipped out [1]
 ....\OT1/cmtt/m/n/10 g
 ....\OT1/cmtt/m/n/10 h
 ....\OT1/cmtt/m/n/10 i
+....\hbox(0.0+0.0)x0.0, direction TLT
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
 ....\glue(\rightskip) 0.0
--- a/required/latex-lab/testfiles-block/blocks-verbatim-02.tlg
+++ b/required/latex-lab/testfiles-block/blocks-verbatim-02.tlg
@@ -217,6 +217,7 @@ Completed box being shipped out [1]
 ....\T1/cmtt/m/n/10 a
 ....\T1/cmtt/m/n/10 b
 ....\T1/cmtt/m/n/10 c
+....\hbox(0.0+0.0)x0.0
 ....\pdfliteral page{EMC}
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
@@ -281,6 +282,7 @@ Completed box being shipped out [1]
 ....\T1/cmtt/m/n/10 d
 ....\T1/cmtt/m/n/10 e
 ....\T1/cmtt/m/n/10 f
+....\hbox(0.0+0.0)x0.0
 ....\pdfliteral page{EMC}
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
@@ -359,6 +361,7 @@ Completed box being shipped out [1]
 ....\T1/cmtt/m/n/10 g
 ....\T1/cmtt/m/n/10 h
 ....\T1/cmtt/m/n/10 i
+....\hbox(0.0+0.0)x0.0
 ....\pdfliteral page{EMC}
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
--- a/required/latex-lab/testfiles-block/blocks-verbatim-03.luatex.tlg
+++ b/required/latex-lab/testfiles-block/blocks-verbatim-03.luatex.tlg
@@ -86,6 +86,7 @@ Completed box being shipped out [1]
 ....\OT1/cmtt/m/n/10 b
 ....\OT1/cmtt/m/n/10 a
 ....\OT1/cmtt/m/n/10 r
+....\hbox(0.0+0.0)x0.0, direction TLT
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
 ....\glue(\rightskip) 0.0
@@ -178,6 +179,7 @@ Completed box being shipped out [1]
 ....\OT1/cmtt/m/n/10 b
 ....\OT1/cmtt/m/n/10 a
 ....\OT1/cmtt/m/n/10 r
+....\hbox(0.0+0.0)x0.0, direction TLT
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
 ....\glue(\rightskip) 0.0
--- a/required/latex-lab/testfiles-block/blocks-verbatim-03.tlg
+++ b/required/latex-lab/testfiles-block/blocks-verbatim-03.tlg
@@ -88,6 +88,7 @@ Completed box being shipped out [1]
 ....\T1/cmtt/m/n/10 b
 ....\T1/cmtt/m/n/10 a
 ....\T1/cmtt/m/n/10 r
+....\hbox(0.0+0.0)x0.0
 ....\pdfliteral page{EMC}
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
@@ -168,6 +169,7 @@ Completed box being shipped out [1]
 ....\T1/cmtt/m/n/10 b
 ....\T1/cmtt/m/n/10 a
 ....\T1/cmtt/m/n/10 r
+....\hbox(0.0+0.0)x0.0
 ....\pdfliteral page{EMC}
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
--- a/required/latex-lab/testfiles-block/blocks-verbatim-nested-gh119.luatex.tlg
+++ b/required/latex-lab/testfiles-block/blocks-verbatim-nested-gh119.luatex.tlg
@@ -92,6 +92,7 @@ Completed box being shipped out [1]
 ......\OT1/cmtt/m/n/10.95 i
 ......\OT1/cmtt/m/n/10.95 n
 ......\OT1/cmtt/m/n/10.95 x
+......\hbox(0.0+0.0)x0.0, direction TLT
 ......\penalty 10000
 ......\glue(\parfillskip) 0.0 plus 1.0fil
 ......\glue(\rightskip) 0.0
@@ -264,6 +265,7 @@ Completed box being shipped out [1]
 ....\OT1/cmtt/m/n/10.95 i
 ....\OT1/cmtt/m/n/10.95 n
 ....\OT1/cmtt/m/n/10.95 x
+....\hbox(0.0+0.0)x0.0, direction TLT
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
 ....\glue(\rightskip) 0.0
@@ -435,6 +437,7 @@ Completed box being shipped out [1]
 ....\OT1/cmtt/m/n/10.95 i
 ....\OT1/cmtt/m/n/10.95 n
 ....\OT1/cmtt/m/n/10.95 x
+....\hbox(0.0+0.0)x0.0, direction TLT
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
 ....\glue(\rightskip) 0.0
--- a/required/latex-lab/testfiles-block/blocks-verbatim-nested-gh119.tlg
+++ b/required/latex-lab/testfiles-block/blocks-verbatim-nested-gh119.tlg
@@ -92,6 +92,7 @@ Completed box being shipped out [1]
 ......\T1/cmtt/m/n/10.95 i
 ......\T1/cmtt/m/n/10.95 n
 ......\T1/cmtt/m/n/10.95 x
+......\hbox(0.0+0.0)x0.0
 ......\pdfliteral page{EMC}
 ......\penalty 10000
 ......\glue(\parfillskip) 0.0 plus 1.0fil
@@ -273,6 +274,7 @@ Completed box being shipped out [1]
 ....\T1/cmtt/m/n/10.95 i
 ....\T1/cmtt/m/n/10.95 n
 ....\T1/cmtt/m/n/10.95 x
+....\hbox(0.0+0.0)x0.0
 ....\pdfliteral page{EMC}
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil
@@ -444,6 +446,7 @@ Completed box being shipped out [1]
 ....\T1/cmtt/m/n/10.95 i
 ....\T1/cmtt/m/n/10.95 n
 ....\T1/cmtt/m/n/10.95 x
+....\hbox(0.0+0.0)x0.0
 ....\pdfliteral page{EMC}
 ....\penalty 10000
 ....\glue(\parfillskip) 0.0 plus 1.0fil


required/latex-lab/changes.txt
```diff
@@ -1,3 +1,28 @@
+2025-12-17  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
+
+	* latex-lab-block.dtx:
+	In November/December the block module was heavily revised with so
+	many changes that it difficult to enumerate them all.
```
mbertucci47: missing "is" after "it"

required/latex-lab/changes.txt
```diff
@@ -1,3 +1,28 @@
...
+	A new template type captionedtext was added.
+
+	Many key names have been adjusted to provide a better set
+	(that may need some further adjustments).
+
+	Support for theorem-like and proof environments was
+	adjusted. It is now possible to automatically get a new line
+	if the body starts with a list.
+
+	\SimpleBlockEnv, \BlockEnv and \BlockEnvEnd was added to
```
mbertucci47: "was" -> "were"

required/latex-lab/latex-lab-firstaid.dtx
```diff
@@ -56,7 +56,8 @@
 % Similar to the main firstaid package the goal is to remove the 
 % patches once the packages have been updated.
 %
-% \changes{v0.85g}{2024/10/04}{Removed firstaid for blindtext. No longer needed with new doenpe-code.}
+% \changes{v0.85g}{2024/10/04}{Removed firstaid for blindtext.
+%   No longer needed with new doenpe-code.}
```
mbertucci47: doenpe -> doendpe

required/latex-lab/latex-lab-block.dtx
```diff
@@ -153,25 +155,66 @@
...
-% that ties them together as necessary.
+% related parameters, and \insttype{item} (for item layouts and handling).
+%
+% To address the independent aspects we have the template type
+% \insttype{blockenv} that ties them together as necessary when we
+% build document level environments.
 %
 % For example, a \env{quote} environment would make use of a (display)
-% \insttype{block} and some \insttype{para} instnce while an standard
+% \insttype{block} and some \insttype{para} instance while an standard
```
mbertucci47: an -> a

required/latex-lab/latex-lab-block.dtx
```diff
@@ -153,25 +155,66 @@
...
+%    \TemplateSemantics
+%    This template type is used to implement document-level
+%    environments. It defines a \insttype{block} instance to handle the
+%    layout at the \enquote{edge} of the environment data, possibly
+%    some paragraph setup through a \insttype{para} instance, potentially an
+%    \enquote{inner} instance for more complicated environments (such
+%    as lists), and possibly some additional setup code for certain
+%    environments.
+%
+%    Arguments 2--4 are passed to instance handling the inner
```
mbertucci47: possibly want "to _an_ instance"?

required/latex-lab/latex-lab-block.dtx
```diff
@@ -198,12 +241,45 @@
...
 %    \TemplateSemantics
 %    Handle the aspects related to list design, e.g., the use and
 %    formatting of counters, etc.
 %
-%    Note that this does not cover block-related aspects, i.e., a list
-%    instance could be used both for a display list or for an inline line.
+%    Standard \LaTeXe{} lists have no heading/caption, so arguments
+%    2--4 are ignored in the standard \insttype{list} template. But
+%    special lists, such as a list of ingredients for a cookbook,
+%    might so they might be other templates that make use of them.
```
mbertucci47: they -> there

required/latex-lab/latex-lab-block.dtx
```diff
@@ -198,12 +241,45 @@
...
+%
+%    Note that this template type does not cover block-related
+%    aspects, i.e., a list instance could be used both for a display
+%    list or for an inline line.
```
mbertucci47: inline list?

required/latex-lab/latex-lab-block.dtx
```diff
@@ -198,12 +241,45 @@
...
+%       not given then \cs{NoValue}}
+%    \TemplateArgument{4}{Sub-caption/heading text in case this environment
+%       supports a caption, otherwise \cs{NoValue}}
+%    \TemplateSemantics
+%    Produces a text block with an associated caption/heading, e.g., a
+%    theorem-like environment. There may not be a user-supplied
+%    caption text---the caption may consist of a fixed text only like ``Lemma''.
+%
+%    Handles the aspects related to the caption design and typically
+%    supports keys for the adjusting the layout of the body text,
```
mbertucci47: remove "the" before "adjusting" I think

required/latex-lab/latex-lab-block.dtx
```diff
@@ -295,19 +375,28 @@
...
+%    reasons) in the \tn{@listdepth} counter and incremented by each
+%    block by one. The starting value at top-level (outside any block)
+%    is zero. A block environment with
+%    \key{transparent-level}\texttt{=true} also increments the level
+%    before it evaluates and sets its parameters but then decrements
+%    it again, just before it starts processing its body.
+%
+%    The template first checks that the block is not too deeply nested.
+%
+%    after the level was increased then corresponding \tn{@list...} macro
```
mbertucci47: Capitalize "after"? The sentence could also be reworded

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1271 +546,3138 @@
...
-%    \instname{quotationblock-\meta{level}}.
+%    The \enquote{unnumbered?} argument (\texttt{\#2}) is ignored, as
+%    proofs aren't numbered.  The template makes use of the caption
+%    argument (\texttt{\#3}) but in contrast to theorem-like
+%    environments this template replaces the \key{title} key value with the content of
+%    this argument (if not \cs{NoValue}).
 %
+%    Typically there is only one layout for proofs so that there is no
+%    need to split the formatting over two templates as done for
+%    theorem-like environment. That's the reason why the template as
```
mbertucci47: as -> has

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1271 +546,3138 @@
...
-% \subsection{The \env{verse} environment}
+% \subsection{The \env{display} and \env{displayflattened} environments}
 %
-%    The \env{verse} environment is currently still implemented as a
-%    list without real items (as in \LaTeXe{}. That needs updating.\fmi{fix}
 %
+%  \begin{environment}{displayblock,displayblockflattened}
+%    There are two basic block environments (\env{displayblock} and
+%    \env{displayblockflattened}) which are similar to \LaTeXe{}'s
+%    \env{trivlist} except that there aren't degenerated lists and
```
mbertucci47: there -> they

required/latex-lab/changes.txt
davidcarlisle: @mbertucci47  if you use the +/- button the change can be accepted/committed directly in this web view (no need now)

mbertucci47: @davidcarlisle oh thanks, didn’t know that

FrankMittelbach: dpending on how much context you have to give it is actually even less work, because you can edit the line(s) this way directly. But I'll deal with your input the way it is now, thanks so far for it

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1271 +546,3138 @@
...
+%
+%
+%
+%  \begin{environment}{proof}
+%    The \env{proof} environment expects one optional argument holding
+%    an alternative title for the proof. We parse this optional
+%    argument as an implicit key/value argument, so that it is
+%    possible to interpret it either as the value for the key
+%    \key{note} or as a key/value list that holds special key settings
+%    for this particular environment instance. the result is analyzed
```
mbertucci47:
```suggestion
%    for this particular environment instance. The result is analyzed
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1271 +546,3138 @@
...
-%  \end{template}
+%  \end{instance}
 %
 %
-%  \begin{macro}{\l_@@_unused_blockenv_keys_tl}
-%    
+%
+%  \begin{instance}{captionedtext proof}
+%    We use a special \insttype{captionedtext} template to set up the
+%    proof because proof are not numbered and the argument to a proof
```
mbertucci47:
```suggestion
%    proof because proofs are not numbered and the argument to a proof
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1271 +546,3138 @@
...
-%  \begin{macro}{\endblockenv}
-%    The code executed when a blockenv ends is 99\% the same for all
-%    blockenvs (at least up to now). Small differences exist, though. They 
-%    are  accounted for first in the conditionals.
 %
-%    We make this a public command so that new block environments can
-%    be set up without the need to resort to L3 layer
-%    programming.\fmi{name is bad}
+%  \begin{instance}{para center}
+%    Centering a paragraph means putting stretchable glue on both sides-
```
mbertucci47:
```suggestion
%    Centering a paragraph means putting stretchable glue on both sides.
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1271 +546,3138 @@
...
+%
+% \subsection{Paragraph tags}
+%
+%  Paragraphs in \LaTeX{} can be nested, e.g., you can have a paragraph
+%  containing a display quote, which in turn consists of more than one
+%  (sub)paragraph, followed by some more text which all belongs to the
+%  same outer paragraph.
+%
+%  In the PDF model and in the HTML model that is not supported --- a
+%  limitation that conflicts with real live, given that such
```
mbertucci47:
```suggestion
%  limitation that conflicts with real life, given that such
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1271 +546,3138 @@
...
+%  constructs are quite normal in spoken and written language.
+%
+%  The approach we take to resolve this is to model such \enquote{big}
+%  paragraphs with a structure named \struct{text-unit} and use \struct{text} (roll-mapped to \struct{P})
```
mbertucci47:
```suggestion
%  paragraphs with a structure named \struct{text-unit} and use \struct{text} (role-mapped to \struct{P})
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1271 +546,3138 @@
...
+%  paragraph the structures
+% \begin{tabbing}
+% \hspace*{1em}\=\hspace*{1em}\=\hspace*{1em}\=\hspace*{1em}\= \kill
+%    \>    \struct{text-unit}\\
+%    \>\>  \struct{text}\\
+%    \>\>\> The paragraph text \ldots\\
+%    \>\>  \struct{/text}\\
+%    \>    \struct{/text-unit}
+% \end{tabbing}
+% The \struct{text-unit} structure is roll-mapped to \struct{Part} or possibly to
```
mbertucci47:
```suggestion
% The \struct{text-unit} structure is role-mapped to \struct{Part} or possibly to
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1262 +547,3129 @@
...
+%  \begin{macro}{\g_block_nesting_depth_int}
+%    \LaTeXe{} already has a counter to record the nesting depth of
+%    blocks, but  we want our own name because it isn't really tied to
+%    \enquote{lists} any more. However, \tn{@listdepth} is really part
+%    of the legacy interface (for example \env{minipage} alters it to
+%    point to a different counter) so that we are stuck with using at
+%    least indirectly for now and the following line makes this look
+%    like an L3 integer variable but internally expands to \tn{@listdepth}:
+%    \begin{macrocode}
+\cs_new:Npn \g_block_nesting_depth_int { \@listdepth }  % a fake int
```
josephwright: As `\@listdepth` is a count anyway, I'd much prefer `\cs_new_eq:NN` here. If you'd rather not do that, then I'd have a function `\block_nesting_depth:` which wraps the value.

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1262 +547,3129 @@
...
+%    differences exist, though. They are accounted for first in the
+%    conditionals.
+%
+% \changes{v0.9l}{2025/11/21}{Renamed \cs{endblockenv} to \cs{BockEnvEnd}}
+%
+%    We make this a public command so that new block environments can
+%    be set up without the need to resort to L3 layer
+%    programming.
+%    \begin{macrocode}
+\cs_new:Npn \BlockEnvEnd {
```
josephwright: Here and elsewhere: this should be `protected` (I know it doesn't make a lot of different in use, but as a general approach it's cleaner).

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1677,969 +3694,1566 @@
...
+%    \begin{macrocode}
+\DeclareTemplateCode{para}{std}{1}
+{
+  ,para-indent           = \parindent
+%    \end{macrocode}
+%    The next parameter needs integrating in the basic paragraph
+%    handling (not done yet) and it should therefore probably a public
+%    name like the rest.\fmi{integrate/fix}
+%    \begin{macrocode}
+  ,begin-hspace          = \l__par_start_skip              % name??
```
josephwright: We will need to look back over these internals from other modules - once we are happy the general approach is correct.

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1677,969 +3694,1566 @@
...
+  ,left-hspace           = \leftskip 
+  ,right-hspace          = \rightskip 
+  ,end-hspace            = \parfillskip
+  ,fixed-word-spaces     = \l__par_fixed_word_spaces_bool  % name??
+  ,final-hyphen-demerits = \finalhyphendemerits
+  ,newline-cmd           = \\
+  ,para-attr-class       = \l__tag_para_attr_class_tl
```
josephwright, line 3864: In particular, tagging internals seem to come up quite a bit - we do need to address that (outside of this PR, just noting for the future).

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1262 +547,3129 @@
...
+%    \begin{macrocode}
+\RequirePackage{latex-lab-testphase-context}
+%    \end{macrocode}
+%
+%    \begin{macrocode}
+\ExplSyntaxOn
+%    \end{macrocode}
+%
+%
+% \subsubsection{Augmented \cs{SetKnownTemplateKeys}}
```
josephwright, line 2707: Shouldn't all of this just go straight into `lttemplates`?

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1262 +547,3129 @@
...
+%
```
josephwright, line 2708: Shouldn't all of this just go straight into `lttemplates`?

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1262 +547,3129 @@
...
+\AddToHook{build/column/before}{%
+  \if@endpe \@endpefalse \aftergroup\@endpetrue \fi
+}  
+%    \end{macrocode}
+%
+%
+%
+%
+%
+% \subsubsection{Other useful \pkg{expl3} commands}
```
josephwright, line 2950: Presumably we should review these (after it's merged) with a view to dding to the core of `expl3` (also perhaps some stuff from `xbox`).

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1262 +547,3129 @@
...
+%
+%
+%  \begin{macro}{\BlockEnv,\SimpleBlockEnv}
+%    To simplify the environment declarations later we provide two
+%    simply commands that invoke a \insttype{blockenv} instance. The
+%    matching counterpart to these commands is \cs{BockEnvEnd}
+%    (defined below) that carries out all necessary action when a
+%    block environment ends.
+%    \begin{macrocode}
+\cs_new_protected:Npn \BlockEnv            % #1#2#3#4 implicit
```
josephwright, line 3615: In general for doc commands we keep the args explicit - I think here it's OK but we should just be clear on that.

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,1262 +547,3129 @@
...
+%    \end{macrocode}
+%  \end{macro}
+%
+%  \begin{macro}{\c@maxblocklevels}
+%    A counter to increase or decrease the number of supported
+%    level. If increased, one needs to supply additional level instances.
+%    \begin{macrocode}
+\newcounter{maxblocklevels}
+\setcounter{maxblocklevels}{6}
+%    \end{macrocode}
```
josephwright, line 3672: Is this best as a LaTeX counter or as an L3 global `int` with `\cs_set_eq:NN` for a 'convenient' 2e interface?

josephwright (changes requested):

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: I did this initially but this doesn't work, because of the reason I give there: minipages changes the meaning of `\@listdepth` and we need to allow for that. And now I don't want a wrapper, because eventually I want to retire the 2e counter and only use an L3 counter.

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: The block module shouldn't  define para templates, it is just that I had no place for it right now. It should eventually be in ltpara. But in either case I guess the para parameters should probably be public as mentioned.

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: sure, eventually that should happen. But I wanted to first see that it works properly and has the right implementation (might as well wait a bit longer with that).

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: yes, but again there is no hurry with that imho, they are clearly marked for this

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: Good question. It is something a class may want to change without having a need to do anything other than declaring  things (ie instances). Which is why I made it a LaTeX counter not a global L3 int.

required/latex-lab/latex-lab-block.dtx
josephwright: In that case, can we have `protected` here? `int` variables do not expand ... (I'm still worried this doesn't really work as a well-behaved `int`)

required/latex-lab/latex-lab-block.dtx
josephwright: I was thinking of the same thing I have in `ltx-talk`, for example, where `pauses` (`\c@pauses`) is also `\g_@@_pauses_int` (and I suspect should be `\g_talk_pauses_int).

josephwright (approved):
Modulo my concern about the fake int ...

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: I can't see why or how it can misbehave, but I don't mind doing that. As I said, eventually, \@listdepth should get removed, but that will take time.

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: With `\c@maxlevels` used in exactly one place in the code I'm not sure that giving it two names is actually an improvement, so for now I think I'll stay with just the one name.

required/latex-lab/changes.txt
```diff
@@ -1,3 +1,28 @@
+2025-12-17  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
+
+	* latex-lab-block.dtx:
+	In November/December the block module was heavily revised with so
+	many changes that it is difficult to enumerate them all.
+	But here is an attempt:
+
+	Block environments now support 4 mandatory argument not just one.
```
davidcarlisle:
```suggestion
	Block environments now support 4 mandatory arguments not just one.
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -213,34 +290,34 @@
...
-%    levels of nesting are supported, etc.
-%
-%    Finally, the template type defines how it appears in a tagged PDF
-%    document, what tag names are used, how they are rolemapped and
-%    whether it adds additional attributes, etc. 
+%    A sub-type used as part of \insttype{captionedtext} when
+%    producing theorem-like environments. It does the bulk of the work
+%    and sets up most of the formatting. It has been separated out
+%    because many theorem-like environments use the same theorem
+%    layout and only differ in in the fixed caption text they
```
u-fischer: in in ->in

required/latex-lab/latex-lab-block.dtx
```diff
@@ -255,16 +332,16 @@
...
 %               executed after legacy defaults (from \tn{@listi},
 %               \tn{@listii}, etc.) are used but before the block instance
 %               is called.}{\meta{empty}}
 %  \TemplateKey{block-instance}{tokenlist}{Part of the name of the
 %               \insttype{block} instance that is called. The full name has
-%               a \texttt{-}\meta{level} appended.}{displayblock}
+%               a \texttt{-}\meta{level} appended.}{std-display}
 %  \TemplateKey{para-instance}{tokenlist}{Paragraph settings to use
 %               within the environment. If \meta{empty} then outer values are
 %               retained. However, the block template resets some values, which
```
u-fischer: What does the "However" mean?  Is that an open TODO to correct something?

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
-%    implementation is
+%    Numbering of the environment is suppressed unconditionally if the
+%    \key{numbered} is set to \keyvalue{false}. Otherwise the
+%    environment is numbered except when \texttt{\#2} is
+%    \cs{BooleanTrue}, i.e., if the star form of the environment was
+%    used.
+%
+%    The caption of the environment can consist of a title, a number,
+%    a punctuation, some spaces and a note. Their order is defined
+%    \key{order}. If a component is specified but has no value, e.g.,
```
u-fischer: defined -> defined by the key

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
+%    no note  or the numbering suppressed on an individual environment,
+%    then the component and any preceding spaces are ignored.
+%
+%    Spaces between elements are uniform (as one can only specify
+%    \texttt{space} in the the \key{order} key, but it is possible to
```
u-fischer: the the -> the

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
+%    This flattens inner paragraphs without any surrounding tag
+%    structure by using the \keyvalue{basic} tagging recipe.
+%    \begin{macrocode}
+\DeclareInstance{blockenv}{displayblockflattened}{std}
+{
+   name                   = displayblockflattened
+  ,tagging-recipe         = basic
+  ,tagging-suppress-paras = true
+  ,transparent-level      = true
+  ,block-instance         = std-display
```
u-fischer: that is the default block-instance, isn't it? so can be deleted.

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
-% \section{The Implementation}
+%  \subsubsection{Their \insttype{blockenv} instances}
 %
+%  \begin{instance}{blockenv quotation}
+%    For the \env{quotation} environment:
+%    \begin{macrocode}
+\DeclareInstance{blockenv}{quotation}{std}
+{
+   name                = quotation
+  ,tag-name            = quotation
```
u-fischer: tag names should be changed to symbolic names (the relevant names must be activated in latex-lab-namespace), so e.g. `tag-name=\UseStructureName{block/quotation}`

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
-%    setting it unconditionally to false (globally), probably roughly the same,
-%    so we set it always:
-%    \begin{macrocode}
-%           \if@domathendpe
-            \@domathendpefalse
-%           \fi
-  }
+\AddToHook{begindocument/before}[./legacy]{
+  \DeclareDocumentEnvironment{verse}{!O{}}
+    { \SimpleBlockEnv{verse} {#1} } { \BlockEnvEnd }
```
u-fischer, line 1219: verse should imho gets its own symbolic name (currently missing in latex-lab-namespace).

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
+%
+%  
+%  \begin{instance}{block verbatim-1,
+%                   block verbatim-2,
+%                   block verbatim-3,
+%                   block verbatim-4,
+%                   block verbatim-5,
+%                   block verbatim-6 }
+%  Verbatim instances have there own levels so that one can specify
+%  specific indentations or vertical separations between line.
```
u-fischer: line -> lines

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
+% document.
+%
+% This new implementation based on templates, is supposed to cover the
+% functionality of \pkg{amsthm} including it declarations so that
+% documents that use \pkg{amsthm} explicitly or implicitly via their
+% class should continue to work seamlessly.
+%
+% For other packages that provide theorem-like environments we have to
+% see if they could be easily remodeled using the new implementation
+% of if there is a need for extended templates.
```
u-fischer: of ->or

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
+%    declare the instances directly (which gives you a bit more
+%    flexibility).  A few such styles are predeclared, matching what
+%    is offered by \pkg{amsthm}. These are shown below.
+%
+%  \begin{instance}{thmstyle plain}
+%    The main style used for many theorem-like environments, i.e., the
+%    one you get if no special \cs{theoremstyle} has been specified.
+%
+%    We do not want potential tagging from \tn{textbf} here, so we use
+%    \tn{bfseries} to set the font.
```
u-fischer: we do not tag `\textbf` specially (only `\emph`}

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
+  ,final-hyphen-demerits = 0
+  ,newline-cmd           = \@centercr
 }
 %    \end{macrocode}
-% \end{template}
+%  \end{instance}
 %
+%    Here are the attribute definitions used in the
+%    \key{para-attr-class} in the above instances:
+%    \begin{macrocode}
```
u-fischer, line 2240: we should imho move these "standard" attributes to a better place. E.g. into latex-lab-namespace (similar for e.g. table attributes).

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
+\DeclareRobustCommand\raggedright{\UseInstance{para}{raggedright}{}}
+\DeclareRobustCommand\justifying {\UseInstance{para}{justify}{}}
+%    \end{macrocode}
+%    \LaTeX{}'s default is to typeset paragraphs justified.
+%    \begin{macrocode}
+\justifying
 %    \end{macrocode}
+%  \end{macro}
 %
+%  \begin{instance}{para raggedleft}
```
u-fischer: para raggedleft -> para verse

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
-  \@@_debug_typeout:n{template:list:std}
+%  In the PDF model and in the HTML model that is not supported --- a
+%  limitation that conflicts with real life, given that such
+%  constructs are quite normal in spoken and written language.
 %
-%    \end{macrocode}
-%    We start by looking at the user supplied keys in \texttt{\#1}. If
-%    there aren't any we reset \cs{@@_evaluate_saved_user_keys:nn} to
+%  The approach we take to resolve this is to model such \enquote{big}
+%  paragraphs with a structure named \struct{text-unit} and use \struct{text} (role-mapped to \struct{P})
```
u-fischer: `\struct{text-unit}`  -> `\struct{\UseStructureName{para/semantic}}`. 
`\struct{text}`  -> `\struct{\UseStructureName{para/textblock}}`.  (Currently it is the same, but if we change it will adapt ...).

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
+%
+%
+% \subsubsection{Implementation of \insttype{captionedtext}
+%                and \insttype{thmstyle} templates}
+%
+%
+% \begin{template}{captionedtext thmlike}
+%    The template for typical theorem-like environments is rather
+%    trivial, just setting keys and then passing used keys and the
+%    arguments to a \insttype{thmstyle} instance to the real work.
```
u-fischer: to -> to do

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
-%    should be aplicable, so we generate an error if that isn't the case:
-    %    \begin{macrocode}
-    \SetKnownTemplateKeys{item}{std}{#1}
-    \tl_if_empty:NF \UnusedTemplateKeys
-       {
-         \msg_error:nnee { block } { unknown-keys }
-             { \noexpand\item command }
-             \UnusedTemplateKeys
-       }
+%    The we check if there are any keys passed to the instance from
```
u-fischer: The -> Then

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
+%    the outside.
+%    \begin{macrocode}
+  \SetKnownTemplateKeys{captionedtext}{thmlike}{#1}
 %    \end{macrocode}
-%    
-%    If no optional argument was given then \cs{l_@@_label_given_tl}
-%    is still equal to \cs{c_novalue_tl} and so we can distinuish
-%    that from \verb=\item[]=.
-%
+%    Finally we we apply the style, which is just an instance of of
```
u-fischer: we we -> we   of of -> of

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
+%  \end{macro}
 %
+%
+%
+% \begin{template}{thmstyle std}
+%    The \insttype{thmstyle} implements the theorem-like
+%    environment and assumes that the fixed part of the caption is
+%    already stored in \cs{l_@@_title_tl}. The reason for this
+%    separation into two templates is that typically the same design
+%    is us for different theorem-like environments only differing in
```
u-fischer: us -> used

required/latex-lab/latex-lab-block.dtx
```diff
@@ -397,2249 +555,4797 @@
...
+%    If we use numbering then we need a link target and increment the counter.
 %    \begin{macrocode}
-\cs_new_protected:Npn \@@_make_label_box:n #1
-  {
-    \hbox_set:Nn \l_@@_one_label_box
+    \bool_if:NT \l_@@_numbered_bool
+       {  
+         \@kernel@refstepcounter{ \l_@@_counter_tl }
+         \MakeLinkTarget{ \l_@@_counter_tl }
+       }
```
u-fischer, line 5177: We need a target also for unnumbered theorems, e.g. for a `\pageref{xxx}` commands.  Also are we sure here, that they can be no page break between the target and the following caption??

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3516,18 +6125,26 @@
 %    \fmi{next should be a socket}    
 %    \begin{macrocode}
   \tl_set:Nn \l__tag_para_tag_tl {codeline}
+%    \end{macrocode}
```
u-fischer, line 6194: This should be replaced by `\AssignStructureRole{para/textblock}{\UseStructureName{block/verbatim/codeline}}`  and `\NewStructureName{block/verbatim/codeline}
\AssignStructureRole{block/verbatim/codeline}{codeline}`  should be activated in latex-lab-namespace

Also the doc package adds `\check@percent`  to the definition to gobble percent chars in dtx files, and we should check how this could be supported here (see issue https://github.com/latex3/tagging-project/issues/1150)

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: good point ... :-(

dw latex-lab-block.dtx 
latex-lab-block.dtx:306: in
latex-lab-block.dtx:744: the
latex-lab-block.dtx:1728: m
latex-lab-block.dtx:4932: the
latex-lab-block.dtx:5030: we
latex-lab-block.dtx:5030: of
latex-lab-block.dtx:6369: the
latex-lab-block.dtx:6397: an
latex-lab-block.dtx:6489: and

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: should it? Shouldn't it be a tagging socket first of all? Also I'm not sure the structure names should all start with block/ ... seems an unnecessary overspecification to me ... definitely something to discuss (but not in relation to this PR)

`\check@percent` now supported via a socket (might need also a hook in the end).

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: How, should that target look like?

As to the break, there should be no page break possible unless the declaration key(s) incurrently add break points

required/latex-lab/latex-lab-block.dtx
u-fischer: you could hide it into a socket, but imho assigning a name is cheap, so you do not gain much by socket. But in any case it should use a symbolic name and avoid internal variables.

required/latex-lab/latex-lab-block.dtx
u-fischer: `\MakeLinkTarget[prefix]{}` then you get destinations `prefix*.N` with some number N. `prefix` can be anything, but I would suggest the env-name.

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: sure, but I wasn't sure where to. I think that kind of thing should be done by a separate PR moving them from all over the place.

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: you mean in `tag-name`?

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: question then is, should it say `\UseStructureName{block/quotation}` (besides that I don't think it should necessarily say `block/` each time) or should the `\UseStructureName` be implicit? Probably should be explicit even if more to type.

But then, probably also  `tag-name:e =` to avoid expanding it over and over again.

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: No, just a fact, but the rest of the text was odd - changed

FrankMittelbach:
> 
> Not needed directly but imho one should start to switch to symbolic names. `\def\LBody{LBody}` etc looks a bit odd.

yes, but I haven't looked at how this is set up and what is necessary to do so (besides, as I said elsewhere I'm not totally convinced that we have the naming convention nailed)

> Shouldn't the support for \cls{amsart} and \cls{amsbook} classes and also the enumitem keys stuff move into latex-lab-firstaid?

maybe, but I thought I first develop everything in one file. But it isn't really firstaid at least in case of enumitem. There the whole package gets a replacement and that doesn't fit the purpose of firstaid in my opinion.

We should also move the para template definitions to its own file/module

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: we don't really know the env name (probably could if \newtheorem is defined differently) so for now I used `theorem-like`.

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: yes all the internal variables should vanish (that is older code really). Maybe a socket is overkill, but perhaps `\AssignStructureRole` also should not do anything if tagging is suspended

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1562,16 +3597,55 @@
...
+\cs_new_protected:Npn \g_block_nesting_depth_int { \@listdepth }  % a fake int
+                                                                  % for now
+%    \end{macrocode}
+%  \end{macro}
+%
+%
+%
 %  \begin{macro}{\l_@@_unused_blockenv_keys_tl}
-%    
+%    The token lists that holds key values we haven't yet used while
```
mbertucci47:
```suggestion
%    The token list that holds key values we haven't yet used while
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1591,26 +3665,30 @@
...
+%    If this block is not a transparent one we have to decrement the level
+%    now again, otherwise that had happened earlier:
 %    \begin{macrocode}
-  \bool_if:NT \l_@@_level_incr_bool
+  \bool_if:NF \l_@@_transparent_level_bool
     { \int_gdecr:N \g_block_nesting_depth_int }
 %    \end{macrocode}
-%    If this block was a list and there are still \tn{item} labels to
-%    be placed we move to horizontal mode to get them typeset.
+%    If this block was captioned or a list or and there are still
```
mbertucci47: I think this should be reworded

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1794,30 +3932,73 @@
...
+%    same line as the following text (using \cs{everypar}) but not if
+%    that text starts with a display block, so we ensure that the
+%    heading gets typeset now.\fmi{This is similar to the standalone case
+%    for block captions so perhaps that can be combined, check}
+%    \begin{macrocode}
+       { \mode_leave_vertical: }
+%    \end{macrocode}
+%    If no such heading is waiting we might have a block caption
+%    waiting to be typeset and this might be requested to be set
+%    \enquote{unchained}. In that case we have also have to ensure
```
mbertucci47:
```suggestion
%    \enquote{unchained}. In that case we also have to ensure
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1930,54 +4100,114 @@
...
+      }      
       {
+        \legacy_if_set_false:n { @noparlist }
+        \legacy_if:nT { @newlist } { \@noitemerr }
         \legacy_if:nTF { @nobreak }
-          { \addvspace{\skip_eval:n{\@outerparskip-\parskip}} }
+          {
+%    \end{macrocode}
+%    We are not resetting \texttt{@nobreak} here as it should apply
+%    also the upcoming item.
```
mbertucci47:
```suggestion
%    We are not resetting \texttt{@nobreak} here as it should also
%    apply to the upcoming item.
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -1987,9 +4217,11 @@
 %
 %
 %
-% \subsubsection{Implementation of list templates \ldots}
-%
+% \subsubsection{Implementation of \insttype{list} templates}
 %
+%   This \insttype{list} is one of the template types that can be used
+%   as an \key{inner-type} in a \insttype{blockenv} the other one
```
mbertucci47:
```suggestion
%   as an \key{inner-type} in a \insttype{blockenv}; the other one
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2183,7 +4417,7 @@
 %    If we encounter horizontal material before the first \tn{item} we
 %    do want a \tn{@noitemerr} straight away, because afterwards we
 %    end up with tagging structure faults whose cause is the
-%    missing \tn{item}. So we setup up \cs{@@_item_everypar:} to test
+%    missing \tn{item}. So we setup up \cs{@@_everypar:} to test
```
mbertucci47:
```suggestion
%    missing \tn{item}. So we set up \cs{@@_everypar:} to test
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2255,28 +4490,40 @@
...
-    label-align     = {
+    ,label-align     = {
       left    = \tl_set:Nn \l_@@_item_align_tl { \relax \hss } ,
       center  = \tl_set:Nn \l_@@_item_align_tl { \hss \hss } ,
       right   = \tl_set:Nn \l_@@_item_align_tl { \hss \relax } ,
       parleft = \NOT_IMPLEMENTED ,
-    } ,
+     }
+%    \end{macrocode}
+%    The key{label-placement} are implemented using two booleans (at
```
mbertucci47:
```suggestion
%    The key{label-placement} is implemented using two booleans (at
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2597,49 +4859,523 @@
...
+%    To avoid unnecessary key/val processing we make a quick check if
+%    there was an optional argument.
+%    \begin{macrocode}
+          \tl_if_novalue:nTF {#1}          % avoids reparsing label={}
+            { \@@_item_instance:n { } }
+            { \@@_item_instance:n {#1} }
+%    \end{macrocode}
+%    The \insttype{item} instance puts the item label into
+%    \cs{g_@@_label_standalone_bool} ready to be placed later. To make
+%    that happen we nee to signal that by  setting the legacy switch
```
mbertucci47:
```suggestion
%    that happen we need to signal that by setting the legacy switch
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2597,49 +4859,523 @@
...
+%   trigger \tn{everypar}. Otherwise we simply close the previous item
+%   with \tn{par} after removing some horizontal space.  Between items,
+%   there is a penalty and some space.
+%    \begin{macrocode}
+\cs_new_protected:Npn \@@_inter_item: {
+  \legacy_if:nT { @inlabel }
+                { \indent \par } % case of \item\item
+%    \end{macrocode}
+%    \tn{par} may have a strange definition and may not get us back to
+%    vertical mode in one go, so we better do not treat the next line
```
mbertucci47:
```suggestion
%    vertical mode in one go, so we better not treat the next line
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2597,49 +4859,523 @@
...
+%    \begin{macrocode}
+\DeclareTemplateCode{captionedtext}{thmlike}{4}
+{
+  ,counter         = \l_@@_counter_tl
+  ,title           = \l_@@_title_tl
+  ,style           = \l_@@_style:nnnn
+}
+{
+%    \end{macrocode}
+%    some debugging info as usual (showing the arguments that are passed):
```
mbertucci47:
```suggestion
%    Some debugging info as usual (showing the arguments that are passed):
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2597,49 +4859,523 @@
...
+     }
+}
 %    \end{macrocode}
-% \end{variable}
-%
+%  \end{macro}
 %
 %
+%  \begin{macro}{\@@_do_number:}
+%    The number (if present) has a similar formatting but it uses an
```
mbertucci47, line 5328:
```suggestion
%    The number (if present) has a similar formatting but it uses a
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2597,49 +4859,523 @@
...
+     }
+     { \@@_drop_spaces: }
+}
+%    \end{macrocode}
+%  \end{macro}
 %
 %
+%  \begin{macro}{\@@_do_punct:}
+%    The punctuation is handled slightly. It unconditionally drops any
+%    dangling spaces not just when it is empty:
```
mbertucci47: I find these two sentences confusing

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2651,97 +5387,101 @@
...
 %    \begin{macrocode}
-              \legacy_if_gset_false:n { @newlist }
-            }
-            { \@@_inter_item: }
+  \template_debug_typeout:n{~\space template:~ 'proof';~
+                            arguments:~ \exp_not:o{\exp_after:wN |#1|#2|#3|#4|}}
 %    \end{macrocode}
-%    To avoid unnecessary key/val processing we make a quick check if
-%    there was an optional argument.
+%    Evaluate document-level key settings. As all given key should be
```
mbertucci47:
```suggestion
%    Evaluate document-level key settings. As all given keys should be
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2759,17 +5499,19 @@
...
+%    \texttt{@endpe} is currently true, because that means we are
+%    right now just after the end of a \texttt{blockenv} and in the
+%    process of looking if we have to end the current
+%    \struct{\UseStructureName{para/semantic}}, i.e., it is already
+%    open.
 %    
-%    The command is mapped to \cs{__kernel_displayblock_beginpar_vmode} in various
-%    tagging recipes.  It is also used in the math code!
+%    The command is mapped to
+%    \cs{__kernel_displayblock_beginpar_vmode} in various tagging
```
mbertucci47:
```suggestion
%    \cs{__kernel_displayblock_beginpar_vmode:} in various tagging
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -2891,9 +5637,9 @@
 %
 %
 % \begin{socketdecl}{tagsupport/kernel/endpe/vmode}
-% A tagging socket which ends a structure. Used in \cs{begin} 
-% and \cs{para_end:}
-% Not yet in lttagging (name and function should be reviewed).
+%    A tagging socket which ends a structure. Used in \cs{begin} and
+%    \cs{para_end:} Not yet in lttagging (name and function should be
```
mbertucci47:
```suggestion
%    \cs{para_end:}. Not yet in lttagging (name and function should be
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3256,22 +6004,27 @@
 %  \begin{macro}{\@@_recipe_standard:}
 %    The \keyvalue{standard} recipe does the following:
 %    \begin{itemize}
-%    \item surround the block with a \struct{text-unit} structure if not already in a
-%       a \struct{text-unit}. In the latter case end the
-%    MC and the \struct{text} but leave the \struct{text-unit} open.
+%    \item surround the block with a
+%       \struct{\UseStructureName{para/semantic}} structure if not
+%       already in a a \struct{\UseStructureName{para/semantic}}. In
```
mbertucci47:
```suggestion
%       already in a \struct{\UseStructureName{para/semantic}}. In
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3685,32 +6347,15 @@
 %
 %
 %  \begin{macro}{\l_@@_legacy_env_params_tl}
-%    
+%    The token list in which the declarations from the second argument
+%    of \env{list} are temporarily stored. This is then used
```
mbertucci47:
```suggestion
%    of \env{list} are temporarily stored. This is then used in
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
-     \tag_socket_use:nnn {struct-mc} {tag=Lbl} {#2}
-   }
-  \group_end:
-  \tag_socket_use:n {para/on}
-%    \end{macrocode}
-%    
+%  \begin{macro}[no-user-doc]{\newtheorem}
+%    We reimplement the extended \pkg{amsthm} version of the declaration which
+%    also supports a star form indicating that this theorem-like
+%    environment should never been numbered.
```
mbertucci47:
```suggestion
%    environment should not be numbered.
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
 %    \begin{macrocode}
-  \tag_socket_use:nn { block/startpara/direct } { \PARALABEL }
+\RenewDocumentCommand \newtheorem { s m o m o } {
 %    \end{macrocode}
-%    
+%    Is the environment definable at all? If not there is no much
```
mbertucci47:
```suggestion
%    Is the environment definable at all? If not there is not much
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
+      }
 %    \end{macrocode}
-%    Here is where \env{verbatim} and \env{verbatim*} technically
-%    differ: in the former we set up spaces to become nobreakable
-%    spaces (if necessary followed by a \tn{pdffakespace} in the
-%    pdf\TeX{} engine)
-%    and in \env{verbatim*} we set it up to generate visible space chars.
-%  \changes{v0.9f}{2025/06/21}{Move special verbatim code into a key}
+%    The \insttype{captionedtext} instance is simple: the counter, if
+%    present is either argument \texttt{\#2} or \texttt{\#3}; the \key{title}
```
mbertucci47:
```suggestion
%    present, is either argument \texttt{\#2} or \texttt{\#3}; the \key{title}
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
-%  own instances (for precisely that reason).
+%  \begin{macro}{\ParseLaTeXeTheoremlike}
+%    The arguments to \cs{ParseLaTeXeTheoremlike} are as follows:
+%    \begin{itemize}
+%    \item[\#1:] instance name to use (of type \enquote{blockenv})
+%    \item[\#2:] unnumbered? boolean normally provided by using the star
+%    form of the environment
+%    \item[\#3:] key/val for layout adjustments settings provided in the
+%       optional argument of the \enquote{theorem-like}
+%       environment. If a note to the them was given in that argument,
```
mbertucci47:
```suggestion
%       environment. If a note to the theorem was given in that argument,
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
-%                   block verbatimblock-2,
-%                   block verbatimblock-3,
-%                   block verbatimblock-4,
-%                   block verbatimblock-5,
-%                   block verbatimblock-6 }
+%  \begin{macro}{\swapnumbers}
+%    Beside declaring theorem-like environments with \cs{newtheorem}
+%    and \cs{theoremstyle}, the \pkg{amsthm} package also introduced
+%    \cs{swapnumbers} to swap title and number in all environments
+%    (because that is a common requirement. The new implementation
```
mbertucci47:
```suggestion
%    (because that is a common requirement). The new implementation
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
-%                   block listblock-3,
-%                   block listblock-4,
-%                   block listblock-5,
-%                   block listblock-6 }
-%    The block instances for the various list environments use the
-%    same underlying instance (well, by default) and nothing 
-%    needs to be set up specifically (because that is already done in
-%    the legacy \tn{list\meta{romannumeral}} unless a
-%    different layout is wanted.
+%    More complicated is argument \texttt{\#9}. It not empty it can
```
mbertucci47:
```suggestion
%    More complicated is argument \texttt{\#9}. If not empty it can
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
-%                   list itemize-2,
-%                   list itemize-3,
-%                   list itemize-4}
-%    For \env{itemize} environments this is all we need to do and we
-%    refer back to the external definitions rather than defining the
-%    \key{item-label} code in the instance to ensure that old
-%    documents still work.
-%
+%    Give special definitions for the commands and then expand
+%    \texttt{\#9} and used the result to edit the instance we defined
```
mbertucci47:
```suggestion
%    \texttt{\#9} and use the result to edit the instance we defined
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
-%    \begin{macrocode}
-\DeclareInstance{item}{basic}{std}
-                { label-align = right }
-%    \end{macrocode}
 %
+% \subsubsection{Supporting QED in proofs}
+%
+%  The \pkg{amsthm} package contains some elaborate code to support
+%  placing a QED symbol into the proof (by default at the end, but
+%  alternatively manually placed with \cs{qedhere}. This code is
```
mbertucci47:
```suggestion
%  alternatively manually placed with \cs{qedhere}). This code is
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
+%    therefore no longer necessary. In fact it has also been extended
+%    by providing \env{alltt*}.
+%    \begin{macrocode}  
+\declare@file@substitution{alltt.sty}{null.tex}
+%    \end{macrocode}
+%
+%
+% \subsection{Replacement for \pkg{amsthm}}
+%
+%    The \pkg{amsthm} is basically supported out of the box (though there
```
mbertucci47:
```suggestion
%    The \pkg{amsthm} package is basically supported out of the box (though there
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
+%    are currently still a few limitation with \cs{newtheoremstyle} and
+%    perhaps also in other places. So this here is a bit premature,
```
mbertucci47:
```suggestion
%    perhaps also in other places). So this here is a bit premature,
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
+  \let \theoremstyle    \amstheoremstyle
+  \let \proof           \amsproof
+  \let \endproof        \amsendproof
 }
 %    \end{macrocode}
-%  \end{instance}
 %
 %
 %
 %
```
mbertucci47, line 7039: What about amsproc?

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
+%
+% \subsection{Support for \pkg{enumitem} interfaces}
+%
+%    The current implementation  incorporates most features of
+%    \pkg{enumitem}. The plan is that the enumitem interfaces are
+%    either natively available or are emulated and mapped to new
+%    interfaces, so that documents using \pkg{enumitem} work
+%    seamlessly.
+%
+%    Some keys have gotten new news, so there the task is to map old
```
mbertucci47:
```suggestion
%    Some keys have gotten new names, so there the task is to map old
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -3720,958 +6365,684 @@
...
+}
 %    \end{macrocode}
-%  \end{macro}
-%
+%    This mapping needs further work to be comprehensive.
 %
+%    The \pkg{enumitem} package also provides declarations to build
+%    out new lists and adjust the layout of existing list. Here we
+%    need to develop emulations that edit existing instances, define
+%    new ones and also provide a mechanisms that define new document
```
mbertucci47:
```suggestion
%    new ones and also provide mechanisms that define new document
```

required/latex-lab/latex-lab-block.dtx
```diff
@@ -4687,6 +7058,13 @@
 %
 % \end{implementation}
 %
+%
+% \Finale
+%
+% \endinput  ^^A don't typeset the first prototpye notes -------
```
mbertucci47:
```suggestion
% \endinput  ^^A don't typeset the first prototype notes -------
```

required/latex-lab/latex-lab-namespace.dtx
```diff
@@ -203,17 +207,28 @@ tables,       Sect,     pdf2,
 %<@@=tag>
 %    \end{macrocode}
 %
+% 
 % \begin{macro}{\AssignStructureRole}
-% We redefine \cs{AssignStructureRole} to setup that tagging rolemapping
+%    We redefine \cs{AssignStructureRole} to setup that tagging
+%    rolemapping.  The second argument needs expansion so that we do
```
mbertucci47: Not sure what the first sentence is saying

required/latex-lab/latex-lab-namespace.dtx
```diff
@@ -316,62 +330,60 @@ tables,       Sect,     pdf2,
...
+\AssignStructureRole{block/list/body}{itembody} 
+%    \end{macrocode}
+%    
+%    \begin{macrocode}
+\NewStructureName{block/quote}
+\AssignStructureRole{block/quote}{quote}
+%    \end{macrocode}
+%    
+%    For now we map \texttt{verse} to \texttt{quote} which means
+%    it will become of  BlockQuote in tagging.\fmi{decide on tagging
```
mbertucci47:
```suggestion
%    it will become BlockQuote in tagging.\fmi{decide on tagging
```

mbertucci47 (commented):
Some typos

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: guess that needs to be handled too :-(

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: this ends up being Lbl (and I know that).

required/latex-lab/latex-lab-block.dtx
FrankMittelbach: or maybe not ... we have to think about when we want that kind of expansion to happen.

required/latex-lab/latex-lab-namespace.dtx
FrankMittelbach: Made a changer but not sure either what you meant here @u-fischer.

car222222 (commented):
Unable to comment on changes to the main file now. 
Also, this has been moved to develop, and closed here.

car222222:
My GitHub connection is by now refusing to show me the edits and comments in here for this file:
…/latex-lab/latex-lab-block.dtx 
thus I am unable to make any further comments there.

It is claiming that this entry has become too big!! ?? And it will not show me such a large entry. 

—-
I also see that the latest version is now in “develop”. 
So maybe it would be best if I review this version of the whole file? 
Or should I wait a bit longer? 

Chris

PR #1948, opened 2025-12-24
Make \protected functions in templates

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [x] Test file(s) added

### DIFF
--- a/base/doc/ltnews43.tex
+++ b/base/doc/ltnews43.tex
@@ -223,6 +223,11 @@ relatives. This means you now \emph{can} type in an optional argument
 that is interpreted as \enquote{not present}, but this is not likely
 to happen by accident.
 
+\subsection{Revision of \cs{protected} status of functions in templates}
+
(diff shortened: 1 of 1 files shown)


josephwright:
Currently the full docs for templates are only in `lttemplates`, not in `clsguide` (or `usrguide`). In terms fo `ltnews, I wasn't sure this really was relevant - as you say, more or less all templates so far are our own.

FrankMittelbach:
> Currently the full docs for templates are only in `lttemplates`, not in `clsguide` (or `usrguide`). In terms fo `ltnews, I wasn't sure this really was relevant - as you say, more or less all templates so far are our own.

my mistake, yes it is not, but probably it should be at some point (in clsguide as you say.).  Still, in that case a note in ltnews it probably good even if it mainly used by us.

FrankMittelbach (approved):

PR #1951, opened 2025-12-31
Support all (new) babel-german options

jspitz:
As suggested [here](https://github.com/latex3/babel/issues/371#issuecomment-3699904775)

This adds support for Swiss Standard German (swissgerman, nswissgerman), which was missing, and for the forthcoming new babel-german options (german-de etc.). All of these use the same strings, since none of the spelling differences or different lexemes are used.

To prevent redundancy and ease maintenance, this introduces a new command (`\DeclareVrefOption`) for setting `\extras` to multiple (comma-separated) options in a row. I went for l3 for this. I could probably also do this in TeX or 2e, but without relying on 3rd party packages, it would be significantly more complex I think. So I hope this is OK.

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/required/tools/varioref.dtx
+++ b/required/tools/varioref.dtx
@@ -705,18 +705,13 @@
 %    with non-varying strings (such as the German varieties).
 % \changes{v1.6h}{2025/12/31}{Macro added}
 %    \begin{macrocode}
-\ExplSyntaxOn
-\NewDocumentCommand{\DeclareVrefOption}{ m m }
- {
-  \clist_clear_new:N \l__vref_lang_options
-  \clist_set:Ne \l__vref_lang_options { #1 }
-  \tl_set:Nn \l__vref_lang_args_tl { #2 }
-  \clist_map_inline:Nn \l__vref_lang_options
-    {
-    \DeclareOption{##1}{\exp_args:Nco \vref@addto {extras##1} { \l__vref_lang_args_tl } }
-    }
- }
-\ExplSyntaxOff
+\def\DeclareVrefOption#1#2{%
+  \def\vref@lang@list{#1}%
+  \@for\vref@list@item:=\vref@lang@list\do{%
+    \edef\vref@list@item@extras{extras\vref@list@item}
+    \DeclareOption{\vref@list@item}{\expandafter\vref@addto\csname\vref@list@item@extras \endcsname{#2}}
+   }
+}
 %    \end{macrocode}
 %  \end{macro}
 %


u-fischer:
Hm. I'm not sure if this is the way to go. If every package with language support has to fill 18 german `\extraXXX` with basically the same information something is wrong. These languages variants should have a root language that they call first.

@jbezos what do you think?

jspitz:
> Hm. I'm not sure if this is the way to go. If every package with language support has to fill 18 german `\extraXXX` with basically the same information something is wrong. These languages variants should have a root language that they call first.

This is not the case. If you want to modify extras for 1991 German German, redefining `\extrasngerman` is sufficient as this is inherited by all 1996 German varieties, likewise for the other national or spelling varieties (with the legacy `\extrasgerman`, `\extrasaustrian` etc.). There is also a very generic macro (`\@extrasgerman`) that can be changed if something should be applied to any and all German varieties. This is documented in the new manual.

What this is doing is simply making varioref understand the babel options.

jspitz:
> To prevent redundancy and ease maintenance, this introduces a new command (`\DeclareVrefOption`) for setting `\extras` to multiple (comma-separated) options in a row. I went for l3 for this. I could probably also do this in TeX or 2e, but without relying on 3rd party packages, it would be significantly more complex I think. 

Actually, not too difficult. I am not too versed in these matters, but the following would work as well, if you prefer to keep out l3 for now:
```tex
\def\DeclareVrefOption#1#2{%
  \def\vref@lang@list{#1}%
  \@for\vref@list@item:=\vref@lang@list\do{%
    \edef\vref@list@item@extras{extras\vref@list@item}
    \DeclareOption{\vref@list@item}{\expandafter\vref@addto\csname\vref@list@item@extras \endcsname{#2}}
   }
}
```

u-fischer:
> If you want to modify extras for 1991 German German, redefining \extrasngerman is sufficient

Then why does your code map over the language lists and does `\DeclareOption{##1}{\exp_args:Nco \vref@addto {extras##1}...`, which means that it adds code to all 18 different `\extrasXXX`. ? 

> What this is doing is simply making varioref understand the babel options.

But why is this needed? And is this needed for all other packages with some german language support, say cleveref, hyperref, biblatex?

jspitz:
> Then why does your code map over the language lists and does `\DeclareOption{##1}{\exp_args:Nco \vref@addto {extras##1}...`, which means that it adds code to all 18 different `\extrasXXX`. ?

Because it would be more complicated (and no gain) to differentiate it to `\extrasgerman`, `\extrasngerman`, `\extrasaustrian`, etc.

> > What this is doing is simply making varioref understand the babel options.
> 
> But why is this needed? 

Because otherwise something like
```
\documentclass[german-de]{article}
\usepackage{babel}
\usepackage{varioref}
```
would not work

> And is this needed for all other packages with some german language support, say cleveref, hyperref, biblatex?

Not all I think, but some (depending on how they implement l7n support).

jspitz:
BTW if these packages would employ `\BCPdata` rather than relying on language option names, things would be way easier (and one wouldn't need to consider all possible babel and polyglossia options)

jspitz:
> Not all I think, but some (depending on how they implement l7n support).

For `biblatex`, the fix would simply be to extend the list in `\DeclareRedundantLanguages` in `german.lbx`

jspitz:
An alternative approach on `babel-german` level that would work for `varioref` (and packages using a similar approach,) would be to iterate over class options and replace the new values with the known ones, as in:
```tex
\def\bbl@mk@class@alias#1#2{%
  \def\bbl@tmp@classoptionslist{}
  \bbl@foreach\@raw@classoptionslist{%
    \def\bbl@tmpa{#1}
    \ifx\CurrentOption\bbl@tmpa
      \edef\bbl@tmp@classoptionslist{\bbl@tmp@classoptionslist\zap@space#2 \@empty,}%
    \else
      \edef\bbl@tmp@classoptionslist{\bbl@tmp@classoptionslist\zap@space#1 \@empty,}%
    \fi%
  }
  \let\@raw@classoptionslist\bbl@tmp@classoptionslist
  \def\bbl@tmp@classoptionslist{}
  \bbl@foreach\@classoptionslist{%
     \def\bbl@tmpa{#1}
     \ifx\CurrentOption\bbl@tmpa
        \edef\bbl@tmp@classoptionslist{\bbl@tmp@classoptionslist\zap@space#2 \@empty,}%
     \else
        \edef\bbl@tmp@classoptionslist{\bbl@tmp@classoptionslist\zap@space#1 \@empty,}%
     \fi%
  }
  \let\@classoptionslist\bbl@tmp@classoptionslist
}
```
and then, e.g. in `german-de.ldf`, `\bbl@mk@class@alias{german-de}{ngerman}`

Biblatex could be made to work by adding
```tex
\expandafter\main@language\expandafter{#2}%
```
to the definition of the above function

Would you prefer me to do that?

Varioref still needed support for Swiss Standard German, of course.

--

Edit: fix code

jspitz:
I decided to do the latter (https://codeberg.org/jspitz/babel-german/commit/f57ca8fb0d21466873a6f548961fc7f05aaa8d85)

Closing this. I will open a new request for Swiss German

PR #1955, opened 2026-01-02
Make \DeclareMathScriptfontMapping mappings global

zauguin:
The mappings defined by `\DeclareMathScriptfontMapping` define global properties of font families, so it makes more sense for them to be global than to be local.

# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added


FrankMittelbach:
I can't look now but how are the other font declarations handled (because they are all similarly "global", aren't they?).  I think there should be consistency, so if they or some of them are local then it might need a bigger change.

zauguin:
The general font declarations which define something for a specific font (like `\DeclareFontShape` and `\DeclareFontFamily`) are already global.

The `\DeclareFontShapeChangeRule` and `\DeclareFontSeriesChangeRule` are defining local mappings through. As far as I can tell this doesn't really hurt since there is not much reason to use them in a group (especially they wouldn't usually appear in `.fd` files), but I think it would be more consistent to make these rules global.

FrankMittelbach:
> The `\DeclareFontShapeChangeRule` and `\DeclareFontSeriesChangeRule` are defining local mappings through. As far as I can tell this doesn't really hurt since there is not much reason to use them in a group (especially they wouldn't usually appear in `.fd` files), but I think it would be more consistent to make these rules global.

Thinking a bit more about this, I would leave these two alone. You need global with `\DeclareMathScriptfontMapping` for its use in .fd files, but the other do not belong in .fd files as you say. And I can imagine some odd situations where one might want to alter the change rules locally and if we make them global then that means one can't simply do this using a group.

base/doc/ltnews43.tex
```diff
@@ -242,6 +242,16 @@ \subsection{Improve transparency of \cs{label}, \cs{index}, and friends}
 %
 \githubissue{1910}
 
+\subsection{Global mappings for math script font families}
+
+The command \cs{DeclareMathScriptfontMapping} added in the previous
+release declares related font families used for scriptsize and
+scriptscriptsize mathematics. Previously this mapping was set locally,
+therefore it could not be used in \texttt{.fd} files and files used
```
FrankMittelbach: and files used?  Not sure I understand the sentence

FrankMittelbach (approved):
approved, but perhaps the text needs some clarification/correction

FrankMittelbach:
besides fixing the sentence in ltnews and resolving the conflict, I thin this could be merged

FrankMittelbach:
what's the reason for the test failures?

muzimuzhi:
> what's the reason for the test failures?

@FrankMittelbach Diff files shown different font IDs and new font info. Tested locally, after merging branch 'develop' into the head branch of this PR,
```
cd required/latex-lab
l3build check -cconfig-math
```
passed.

So probably just need a rerun, then the checks triggered by `pull_request` will pass.

zauguin:
The test differences were intended: The `.fd` files there assumed that this change is in effect, so they used this to set the script and scriptscript font. Therefore adding the change changed the fonts to use the correct script variants.

PR #1970, opened 2026-01-14
heading command implementation with templates

u-fischer:
# Internal housekeeping

## Status of pull request
- Ready to merge

- [x] Test file(s) added
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated  not yet,

### DIFF
--- a/.github/tl_packages
+++ b/.github/tl_packages
@@ -30,6 +30,7 @@ arydshln
 bidi
 bigfoot
 blindtext
+cellspace
 cleveref
 crop
 dinbrief
@@ -49,7 +50,6 @@ unicode-math
 varwidth
 microtype
 listings
-
 #
 # special testing for latex-lab
 abraces
--- a/base/build.lua
+++ b/base/build.lua
@@ -154,6 +154,7 @@ typesetdeps =
   {
     maindir .. "/required/graphics",
     maindir .. "/required/tools",
+    maindir .. "/required/firstaid",
     maindir .. "/required/amsmath"    -- for l3doc.cls :-(
   }
 unpackdeps  = {}
--- a/base/testfiles-lthooks/ltcmdhooks-001.tlg
+++ b/base/testfiles-lthooks/ltcmdhooks-001.tlg
@@ -235,6 +235,7 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid = 0 -> 
 [lthooks] Handled code for firstaid
+[lthooks] Update code for hook 'package/zref-perpage/after' on input line ...:
 [lthooks] Update code for hook 'class/smfart/after' on input line ...:
 Code labels for sorting:
  firstaid
@@ -271,6 +272,12 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid/aliascounter = 0 -> 
 [lthooks] Handled code for firstaid/aliascounter
+[lthooks] Update code for hook 'package/cellspace/after' on input line ...:
+Code labels for sorting:
+ firstaid
+Data structure for label rules:
+ firstaid = 0 -> 
+[lthooks] Handled code for firstaid
 [lthooks] Update code for hook 'cmd/foo/before' on input line ...:
 Code labels for sorting:
  x and y
@@ -313,6 +320,7 @@ Data structure for label rules:
  class/amsproc/after -> \__hook_toplevel class/amsproc/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}[firstaid/aliascounter]{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next class/amsproc/after  
  package/amsthm/after -> \__hook_toplevel package/amsthm/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}[firstaid/aliascounter]{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next package/amsthm/after  
  class/amsbook/after -> \__hook_toplevel class/amsbook/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next class/amsbook/after  
+ package/cellspace/after -> \__hook_toplevel package/cellspace/after \FirstAidNeededT {cellspace}{sty}{....-..-.. v... Spacing table cells (JN)}{\let \@startpbox@action \@startpbox }\__hook_next package/cellspace/after  
  cmd/foo/before -> before 1 before 2 \__hook_toplevel cmd/foo/before {#1}{#2}{#3}{#4}{#5}{#6}{#7}{#8}{#9}\__hook_next cmd/foo/before {#1}{#2}{#3}{#4}{#5}{#6}{#7}{#8}{#9} 
  cmd/foo/after -> \__hook_toplevel cmd/foo/after {#1}{#2}{#3}{#4}{#5}{#6}{#7}{#8}{#9}after 2 after 1 \__hook_next cmd/foo/after {#1}{#2}{#3}{#4}{#5}{#6}{#7}{#8}{#9} 
 No file ltcmdhooks-001.aux.
--- a/base/testfiles-lthooks/lthooks-000.tlg
+++ b/base/testfiles-lthooks/lthooks-000.tlg
@@ -171,6 +171,7 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid = 0 -> 
 [lthooks] Handled code for firstaid
+[lthooks] Update code for hook 'package/zref-perpage/after' on input line ...:
 [lthooks] Update code for hook 'class/smfart/after' on input line ...:
 Code labels for sorting:
  firstaid
@@ -207,6 +208,12 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid/aliascounter = 0 -> 
 [lthooks] Handled code for firstaid/aliascounter
+[lthooks] Update code for hook 'package/cellspace/after' on input line ...:
+Code labels for sorting:
+ firstaid
+Data structure for label rules:
+ firstaid = 0 -> 
+[lthooks] Handled code for firstaid
 [lthooks] All initialized (non-empty) hooks:
  begindocument -> \__hook_toplevel begindocument \__hook_next begindocument  
  enddocument/info -> \@dofilelist \@enddocument@kernel@warnings \__hook_toplevel enddocument/info \__hook_next enddocument/info  
@@ -233,6 +240,7 @@ Data structure for label rules:
  class/amsproc/after -> \__hook_toplevel class/amsproc/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}[firstaid/aliascounter]{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next class/amsproc/after  
  package/amsthm/after -> \__hook_toplevel package/amsthm/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}[firstaid/aliascounter]{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next package/amsthm/after  
  class/amsbook/after -> \__hook_toplevel class/amsbook/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next class/amsbook/after  
+ package/cellspace/after -> \__hook_toplevel package/cellspace/after \FirstAidNeededT {cellspace}{sty}{....-..-.. v... Spacing table cells (JN)}{\let \@startpbox@action \@startpbox }\__hook_next package/cellspace/after  
 No file lthooks-000.aux.
 LaTeX Font Info:    Checking defaults for OML/cmm/m/it on input line ....
 LaTeX Font Info:    ... okay on input line ....
--- a/base/testfiles-lthooks/lthooks-001.tlg
+++ b/base/testfiles-lthooks/lthooks-001.tlg
@@ -178,6 +178,7 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid = 0 -> 
 [lthooks] Handled code for firstaid
+[lthooks] Update code for hook 'package/zref-perpage/after' on input line ...:
 [lthooks] Update code for hook 'class/smfart/after' on input line ...:
 Code labels for sorting:
  firstaid
@@ -214,6 +215,12 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid/aliascounter = 0 -> 
 [lthooks] Handled code for firstaid/aliascounter
+[lthooks] Update code for hook 'package/cellspace/after' on input line ...:
+Code labels for sorting:
+ firstaid
+Data structure for label rules:
+ firstaid = 0 -> 
+[lthooks] Handled code for firstaid
 [lthooks] Update code for hook 'xxx' on input line ...:
 Code labels for sorting:
  label1 and label2
@@ -248,6 +255,7 @@ Data structure for label rules:
  class/amsproc/after -> \__hook_toplevel class/amsproc/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}[firstaid/aliascounter]{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next class/amsproc/after  
  package/amsthm/after -> \__hook_toplevel package/amsthm/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}[firstaid/aliascounter]{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next package/amsthm/after  
  class/amsbook/after -> \__hook_toplevel class/amsbook/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next class/amsbook/after  
+ package/cellspace/after -> \__hook_toplevel package/cellspace/after \FirstAidNeededT {cellspace}{sty}{....-..-.. v... Spacing table cells (JN)}{\let \@startpbox@action \@startpbox }\__hook_next package/cellspace/after  
  xxx -> foobar\__hook_toplevel xxx \__hook_next xxx  
 LaTeX hooks Warning: Cannot remove chunk 'label3' from hook 'xxx' because it does not exist in that hook.
 LaTeX hooks Warning: Cannot remove chunk 'label3' from hook 'yyy' because the hook does not exist.
@@ -320,12 +328,14 @@ The sequence \g__hook_all_seq contains the items (without outer braces):
 >  {package/chemnum/after}
 >  {package/cleveref/after}
 >  {package/arydshln/after}
+>  {package/zref-perpage/after}
 >  {class/smfart/after}
 >  {class/smfbook/after}
 >  {class/amsart/after}
 >  {class/amsproc/after}
 >  {package/amsthm/after}
 >  {class/amsbook/after}
+>  {package/cellspace/after}
 >  {xxx}.
 [lthooks] Update code for hook 'para/before' on input line ...:
 [lthooks] Update code for hook 'para/after' on input line ...:
@@ -498,6 +508,7 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid = 0 -> 
 [lthooks] Handled code for firstaid
+[lthooks] Update code for hook 'package/zref-perpage/after' on input line ...:
 [lthooks] Update code for hook 'class/smfart/after' on input line ...:
 Code labels for sorting:
  firstaid
@@ -534,6 +545,12 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid/aliascounter = 0 -> 
 [lthooks] Handled code for firstaid/aliascounter
+[lthooks] Update code for hook 'package/cellspace/after' on input line ...:
+Code labels for sorting:
+ firstaid
+Data structure for label rules:
+ firstaid = 0 -> 
+[lthooks] Handled code for firstaid
 [lthooks] Update code for hook 'xxx' on input line ...:
 Code labels for sorting:
  label1
@@ -566,5 +583,6 @@ Data structure for label rules:
  class/amsproc/after -> \__hook_toplevel class/amsproc/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}[firstaid/aliascounter]{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next class/amsproc/after  
  package/amsthm/after -> \__hook_toplevel package/amsthm/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}[firstaid/aliascounter]{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next package/amsthm/after  
  class/amsbook/after -> \__hook_toplevel class/amsbook/after \@amsthm@reset@ynthm \AddToHook {package/cleveref/after}{\let \@ynthm \@amsthm@saved@ynthm }\__hook_next class/amsbook/after  
+ package/cellspace/after -> \__hook_toplevel package/cellspace/after \FirstAidNeededT {cellspace}{sty}{....-..-.. v... Spacing table cells (JN)}{\let \@startpbox@action \@startpbox }\__hook_next package/cellspace/after  
  xxx -> foo\__hook_toplevel xxx \__hook_next xxx  
 The hook xxx contains the rules:
--- a/base/testfiles-lthooks/lthooks-002.tlg
+++ b/base/testfiles-lthooks/lthooks-002.tlg
@@ -178,6 +178,7 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid = 0 -> 
 [lthooks] Handled code for firstaid
+[lthooks] Update code for hook 'package/zref-perpage/after' on input line ...:
 [lthooks] Update code for hook 'class/smfart/after' on input line ...:
 Code labels for sorting:
  firstaid
@@ -214,6 +215,12 @@ Code labels for sorting:
 Data structure for label rules:
  firstaid/aliascounter = 0 -> 
 [lthooks] Handled code for firstaid/aliascounter
+[lthooks] Update code for hook 'package/cellspace/after' on input line ...:
+Code labels for sorting:
+ firstaid
+Data structure for label rules:
+ firstaid = 0 -> 
+[lthooks] Handled code for firstaid
 [lthooks] Update code for hook 'xxx' on input line ...:
 Code labels for sorting:
  label1 and label2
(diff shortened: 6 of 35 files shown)


required/latex-lab/latex-lab-firstaid.dtx
FrankMittelbach:
```suggestion
%% File: latex-lab-firstaid.dtx (C) Copyright 2023-2026 LaTeX Project
```

required/latex-lab/latex-lab-firstaid.dtx
```diff
@@ -120,6 +124,93 @@
...
+\cs_new_protected:Npn \@@_firstaid_amsbook_heading:
+ {
+  \DeclareInstance{heading}{chapter}{display}
+    {
+      , name          = chapter
+      , level         = 0
+      , placement         = top
+      , after-penalty-sep = 32pt
+      , after-sep     = 22pt
+      , number-format = \MakeUppercase{X\chaptername}\enspace \thechapter
```
FrankMittelbach: X\chaptername  ?

required/latex-lab/latex-lab-firstaid.dtx
```diff
@@ -120,6 +124,93 @@
...
+     , final-code=  \vfil\vfil\newpage\newpage\thispagestyle{empty}              
+     , number-format = \partname\nobreakspace\thepart
+     , decls         = \centering\bfseries
+     , number-decls  = \huge
+     , title-decls   = \Huge
+     , headformat-instance = display
+     , mark-cmd      = \partmark {##1}
+   }      
+  \DeclareDocumentCommand \part {s ={shorttitle}o m}
+      { \ParseLaTeXeHeading {part} {##1} {##2} {##3} }
```
FrankMittelbach: Not important for this pull, but I suggest that we adopt the convention to first define the top-level command or environment, and then the template instances that are used by it. Not the other way around.

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2821 @@
...
+% \tableofcontents
+% \medskip
+%
+%
+% \begin{documentation}
+%
+%
+%
+% \section{Introduction}
+%    This module reimplements tagging aware heading commands with templates.
```
FrankMittelbach, line 141:
```suggestion
%    This module implements tagging aware heading commands with templates. 
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2821 @@
...
+%     \DescribeMacro\theheading
+%     This command expands inside a heading to the current number.
+%     
+%     \DescribeMacro\partmark
+%     This replaces the fix \verb+\markboth{}{}+ or similar in the standard \cs{part}
+%    definition. 
+%    
+%                            
+%    \subsection{The optional argument} 
+%    The standard heading commands (re)defined with this module use the standard 2e syntax, e.g.,
```
FrankMittelbach:
```suggestion
%    The standard heading commands (re)defined by this module use the standard 2e syntax, e.g.,
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2821 @@
...
+%        
+%    The following keys can be used
+%    \begin{description}
+%      \item[\key{toc}] This sets the text for the table of contents. It also forces 
+%      that the entry is created. So
+%      \begin{verbatim}
+%      \section*[toc=Introduction]{Introduction to this document}
+%      \end{verbatim}
+%      will create an unnumbered entry \enquote{Introduction} in the table of contents.
+%      An empty value, so \texttt{toc=} will suppress the toc entry.
```
FrankMittelbach:
```suggestion
%      An empty value (i.e., \texttt{toc=}) will suppress the toc entry. 
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2821 @@
...
+%      the bookmark entry.
+%      
+%      \item[\key{nameref}] This allows to set the text used for a crossreference
+%      with \cs{nameref}.
+%      
+%      \item[\key{label}] This sets a label, so is an alternative to using a 
+%      \cs{label} command.
+%      
+%      \item[\key{subtitle}, \key{quote}] Theses keys will allow to pass a subtitle and
+%      a quote to the underlying code, but currently they are not yet used.
```
FrankMittelbach:
```suggestion
%      a quote to the underlying code, but currently they are not yet used by the implementations of the standard \LaTeX{} heading commands.
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2821 @@
...
+%      \section[short]{long title}
+%      \end{verbatim}
+%      is the same as 
+%      \begin{verbatim}
+%      \section[shortitle=short]{long title}
+%      \end{verbatim}     
+%      The key sets the toc, running, bookmark and nameref text.
+%      
+%      \item[\key{numbered}, \key{unnumbered}] This allow to control 
+%      the numbering without stopping toc, running head or bookmarks as would do the
```
FrankMittelbach:
```suggestion
%      the numbering without stopping toc, running head or bookmarks as it would happen with the
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2821 @@
...
+%      starred version of the heading command. 
+%      
+%      \item[template keys] Any other key present is assumed to be a key that should
+%      be passed to the heading instance to overwrite some of its
+%      settings. So, e.g.,
+%      \begin{verbatim}
+%      \section[placement=top,number-format=\fbox{\theheading}]{Title}
+%      \end{verbatim}
+%      will force the section to start a new page and put the number into an \cs{fbox}.
+%     For a list of supported keys, check the following documentation.
```
FrankMittelbach:
```suggestion
%     For a list of supported keys, check the following part of the documentation.  
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2821 @@
...
+%    \end{verbatim}
+%    The name of the instance is the same as the heading command.
+%    
+%    To support other classes which still setup their heading commands with 
+%    \cs{@startsection} the code has a legacy compatibility layer: the
+%    \cs{@startsection} command has been redefined to setup instances on-the-fly
+%    at the first use of a heading command. These instances have names using the
+%    the first argument of \cs{@startsection} with an attached \texttt{-@startsection}. 
+%    As they are created on-the-fly they can be only edited 
+%    \emph{after} the first use of the heading command. Also
```
FrankMittelbach: what is wrong with checking if the instance  `name-@startsection` already exist and if so not generate it on the fly? That way you can alter the layout in the preamble for classes tha use `\@startsection`.

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2821 @@
...
+%
+% \section{Setting up \LaTeXe{} heading commands in classes}
+%
+%  Heading commands are managed by defining an instance of the 
+%  \texttt{heading} template and then using through \cs{ParseLaTeXeHeading}, e.g.,
+% \begin{verbatim}
+%   \DeclareDocumentCommand \part {s ={shorttitle}o m}
+%       { \ParseLaTeXeHeading {part} {#1} {#2} {#3} }
+% \end{verbatim}
+%  The name of the \texttt{heading} template is given in the first argument
```
FrankMittelbach:
```suggestion
%  The name of the \texttt{heading} instance is given in the first argument
```

FrankMittelbach: this is an instance not a template here!

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2821 @@
...
+%  of \cs{ParseLaTeXeHeading} and it is recommended that classes use the 
+%  same name as the heading command, to make it easier for users to identify the
+%  instance to edit if they want to adapt the command. Examples of instances
+%  that mimic the behaviour of the heading commands in the standard classes can be found
```
FrankMittelbach:
```suggestion
%  that mimic the behavior of the heading commands in the standard classes can be found 
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2821 @@
...
+%  The name of the \texttt{heading} template is given in the first argument
+%  of \cs{ParseLaTeXeHeading} and it is recommended that classes use the 
+%  same name as the heading command, to make it easier for users to identify the
+%  instance to edit if they want to adapt the command. Examples of instances
+%  that mimic the behaviour of the heading commands in the standard classes can be found 
+%  below. 
+%
+%  Legacy setup using \cs{@startsection} remains supported albeit
+%  not that performant and with some restrictions for the users, 
+%  see section \ref{sec:startsection} above and in the documentation below.
```
FrankMittelbach, line 431: as I said I think we can and should remove that restriction.

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2821 @@
...
+%  
+%  In contrast \cs{secdef} is not supported. It delegates the
+%  layout to two commands which can contain arbitrary code (usually
+%  hardwired) so that there is no realistic chance to take this
+%  apart and figure out what values should be used for what template
+%  parameters. Classes using \cs{secdef} should setup suitable instances that
+%  mimic their current layout.
```
FrankMittelbach: I think we should try to offer a processing fallback nevertheless: something like \secdef if encountered drops its arguments and makes a guess on what kind of heading is being defined (should be possible most of the time by looking at the command names in the secdef). This way the classes would work even though with changed layout.

required/latex-lab/latex-lab-sec-template.dtx
u-fischer: but then we should also change the title which says "Reimplementation"?

FrankMittelbach: ok, maybe not. Or in both as you say.

required/latex-lab/latex-lab-sec-template.dtx
u-fischer: That is what we are doing, but it happens only if `\@startsection` is actually used, and that is only at the first call of a sectioning command. So `\def\section{\@startsection{...}`  (no instance created) `\section{blub}` (instance created) and now it can be edited.

FrankMittelbach: But that means, that If I define the instance (rather than only edit it) in the preamble I can do so and it works.

so the guidelines could explain that and how to do it (with an example). That would then still allow a user to overwrite the class. Doing it only after the first \section would mean the first section has the wrong settings and you can't alter them.

required/latex-lab/latex-lab-sec-template.dtx
u-fischer: well looking at the uses of `\secdef` the largest part of hits are for part and chapter and for that we could add a switch to \secdef to do the default definitions, but the rest imho is not worth the trouble.

required/latex-lab/latex-lab-sec-template.dtx
u-fischer: I added a guideline, does that look ok?

required/latex-lab/latex-lab-sec-template.dtx
FrankMittelbach: that was exactly my point. do the low-hanging fruits (with a warning about default design) and error on \secdef that is not recognized as incompatible with tagging

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+%    \cs{@startsection} the code has a legacy compatibility layer: the
+%    \cs{@startsection} command has been redefined to setup instances on-the-fly
+%    at the first use of a heading command. These instances have names using the
+%    the first argument of \cs{@startsection} with an attached \texttt{-@startsection}. 
+%    As they are created on-the-fly they can be only edited 
+%    \emph{after} the first use of the heading command. Also
+%    it is not possible to define two different heading commands
+%    of the same level with \cs{@startsection}. 
+%  
+% \subsection[label=sec:convert]{Converting heading commands}
```
FrankMittelbach:
```suggestion
% \subsection[label=sec:convert]{Converting heading commands defined with \cs{@startsection}}
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+% The sectioning commands use two template \emph{types}, \texttt{heading} and \texttt{headformat}. 
+% Compile the following document
+% 
+% \begin{verbatim}
+% \DocumentMetadata{}
+% \documentclass{ltugboat}
+%  
+% \begin{document}
+% 
+% \section{test}
```
FrankMittelbach:
```suggestion
% \section{test}  % <--- needed so that the instances get defined
%
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+% >  before-code  =>  \tubsecfmt 
+% >  after-code  =>  
+% >  number-title-sep  =>  1em
+% >  from template  =>  hang.
+% \end{verbatim}
+% 
+% \subsubsection{Step 2: Get the default template values}
+% 
+% The last line in both lists show after the \texttt{from template} the names of the 
+% templates of each type. That can be used to get the default values of these templates.
```
FrankMittelbach, line 341: why bother with the default values? I think it is simpler to describe the procedure without it (ie leaving out this step). At the very end of the guidelines you could say something like if you want to make use of the default values in the templates rather than setting everything explicitly, do the following as step 1b) ... and then insert this part.

Right now I think the guidelines are unnecessarily complicated and in most case it is probably even better if everything is specified.

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+% The template 'hang' of type 'headformat' has default values:
+% >  indent  =>  0pt
+% >  before-code  =>  
+% >  after-code  =>  
+% >  number-title-sep  =>  1em.
+% \end{verbatim}
+% 
+% \subsection{Step 3: Declare the instances}
+% By comparing the instance values with the default values one can see which values 
+% should be changed in the instance. The names of the new instances can be
```
FrankMittelbach, line 384: that would then change to in wording.

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+% >  indent  =>  0pt
+% >  before-code  =>  
+% >  after-code  =>  
+% >  number-title-sep  =>  1em.
+% \end{verbatim}
+% 
+% \subsection{Step 3: Declare the instances}
+% By comparing the instance values with the default values one can see which values 
+% should be changed in the instance. The names of the new instances can be 
+% freely chosen but normally the name of the heading command is a suitable choice.
```
FrankMittelbach:
```suggestion
% freely chosen but best practice is to use the name of the heading command (without a backslash). 
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+% This leads then to these declarations:
+% \begin{verbatim}
+% \DeclareInstance{heading}{section}{display} %#1=heading, #2=name, #3=template name
+%  {
+%    level       =  1,
+%    mark-cmd    =  \sectionmark{#1}, % remove second hash
+%    before-sep  =  8.0pt plus 2.0pt minus 2.0pt, 
+%    after-sep   =  4.0pt,   
+%    decls       =  \tubsecfmt,
+%    headformat-instance = section, % new name
```
FrankMittelbach, line 397: with my suggestion above we would have a few more lines here

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+% \end{verbatim} 
+% 
+% \subsubsection{Step 4 Redefine the \cs{section} command}
+% 
+% At last the heading should be defined to use the new interface. 
+% \begin{verbatim}
+% \DeclareDocumentCommand \section {s ={shorttitle}o m}
+%      { \ParseLaTeXeHeading {section} {#1} {#2} {#3} }
+% \end{verbatim}   
+%
```
FrankMittelbach, line 412: Here one could add a section about the possibility to make use of the default values. (or simply leave it out altogether).

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+% \section{Template types and templates for headings}
+% 
+%    The template(s) for headings expect(s) a large number of
+%    positional arguments containing document user data. The
+%    document-level command, e.g., \cs{section} may not offer all of
+%    them directly, but may do so via a key value interface or not at
+%    all. If no interface is provided then the template is passed
+%    \enquote{no value}. If it is offered via a key value interface,
+%    the template receives whatever is set by the user (and
+%    \enquote{no-value} otherwise).
```
FrankMittelbach: ditto here

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+%
+%    Whether or not the heading is numbered is governed through a
+%    boolean, e.g., result of an \texttt{s} specification in
+%    \cs{NewDocumentCommand} or equivalent.
+%
+%    If there is no subtitle or quotation then this is indicated
+%    with \cs{NoValue}.
+%
+%    Note: instead of unnumbered + counter we could just have a single
+%    argument containing the number representation (or \cs{NoValue} of
```
FrankMittelbach:
```suggestion
%    argument containing the number representation (or \cs{NoValue} if
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+%    not used). The reason that there ae two is that this allows us to
```
FrankMittelbach:
```suggestion
%    not used). The reason that there are two is that this allows us to
```

required/latex-lab/latex-lab-sec-template.dtx
u-fischer: Something like this? 
~~~~
\RenewDocumentCommand\secdef{mms={shorttitle}om}
 {
  \str_case:enF {\cs_to_str:N#1}
   {
    {@part}
     {
      \IfNoValueTF{#4}
       {\ParseLaTeXeHeading{part}{#3} {start-code=\relax} {#5}}
       {\ParseLaTeXeHeading{part}{#3} {start-code=\relax,#4} {#5}}
     }
    {@chapter}{\ParseLaTeXeHeading{chapter}{#3} {#4} {#5}}
   }
   {\ERROR}
 }   
 ~~~~

(One could also define a special instance without the start-code)

required/latex-lab/latex-lab-sec-template.dtx
u-fischer: I bothered with the defaults because I wanted to avoid expl3 syntax (the penalty), because I think it is educating to actually check the list, because it makes the instance declaration much shorter, because it avoids the  need of various "Nebensätze" which then would need explanations anyway: if I say somewhere that you can leave out the keys with defaults value I have to explain anyway how to get them. This here is longer but it is a straight way without detours and complicated decisions.

required/latex-lab/latex-lab-sec-template.dtx
FrankMittelbach: looks fine to me

FrankMittelbach: but of course doing that means documentation needs updating, and I think it would be okay to put in a warning into the log if it succeeds not only an error if it fails.

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+%    and passed to the table of contents, the running headers, the bookmarks and used
+%    with \cs{nameref}. Note that this means, that 
+%    \begin{verbatim}
+%    \section*[Title]{Title}
+%    \end{verbatim}
+%    will create a toc entry and a bookmark!
+%        
+%    The following keys can be used
+%    \begin{description}
+%      \item[\key{toc}] This sets the text for the table of contents. It also forces
```
josephwright, line 180: I'd make this a bit longer, `toc-entry` or `TOC-entry` (the latter as 'toc' is an abbreviation)

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,2953 @@
...
+%      with \cs{nameref}.
+%      
+%      \item[\key{label}] This sets a label, so is an alternative to using a 
+%      \cs{label} command.
+%      
+%      \item[\key{subtitle}, \key{quote}] Theses keys will allow to pass a subtitle and
+%      a quote to the underlying code, but currently they are not yet used 
+%      by the implementations of the standard \LaTeX{} heading commands.
+%      
+%      \item[\key{shorttitle}] There is the key which is used if the
```
josephwright, line 207: `short-title`?

josephwright (approved):

required/latex-lab/latex-lab-sec-template.dtx
FrankMittelbach: That's  a document-level key, those I would keep short personally. I think it is okay to have longer names for those that are typically set only in instances such as vspace-begin (even if they can also be set  on an individual environment) but here toc, bookmark,nameref,label are only for document use. But this is somethingg to look at again with more things being done in template style perhaps.

required/latex-lab/latex-lab-sec-template.dtx
FrankMittelbach: same reason: as short as reasonably possible was my thinking

josephwright: See `siunitx` v1 vs v2/3 :-)

required/latex-lab/build.lua
```diff
@@ -182,6 +183,7 @@ specialtypesetting["latex-lab-namespace.dtx"] = {format = "lualatex"}
 specialtypesetting["latex-lab-new-or-1.dtx"] = {format = "lualatex"}
 specialtypesetting["latex-lab-new-or-2.dtx"] = {format = "lualatex"} -- error
 specialtypesetting["latex-lab-sec.dtx"] = {format = "lualatex"}
+specialtypesetting["latex-lab-sec.template.dtx"] = {format = "lualatex"}
```
muzimuzhi:
```suggestion
specialtypesetting["latex-lab-sec-template.dtx"] = {format = "lualatex"}
```

required/latex-lab/build.lua
```diff
@@ -194,6 +196,7 @@ specialtypesetting["latex-lab-unicode-math.dtx"] = {format = "lualatex"}
 specialtypesetting["documentmetadata-support-doc.tex"] = {format = "lualatex"}
 specialtypesetting["documentmetadata-support-code.tex"] = {format = "lualatex"}
 specialtypesetting["latex-lab-enumitem.dtx"] = {format = "lualatex"}
+specialtypesetting["latex-lab-sec-template.dtx"] = {format = "lualatex"}
```
muzimuzhi: No longer needed as the typo on line 186 would be corrected.


```suggestion
```

required/latex-lab/latex-lab-sec-template.dtx
u-fischer: so I will leave it as toc for now.

required/latex-lab/latex-lab-sec-template.dtx
u-fischer: left as shorttitle for now

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,3036 @@
...
+%      {number-format=\fbox{\theheading}}
+%    \end{verbatim}
+%    The name of the instance is the same as the heading command.
+%    
+%    To support other classes which still setup their heading commands with 
+%    \cs{@startsection} the code has a legacy compatibility layer: the
+%    \cs{@startsection} command has been redefined to setup instances on-the-fly
+%    at the first use of a heading command. These instances have names using the
+%    the first argument of \cs{@startsection} with an attached \texttt{-@startsection}. 
+%    As they are created on-the-fly they can be only edited
```
FrankMittelbach: they can only be edited after the heading was used, but you can define the instance beforehand in the preamble and I think that is what should be mentioned at this point with a ref to the guideline section.

FrankMittelbach: > they can only be edited after the heading was used, but you can define the instance beforehand in the preamble and I think that is what should be mentioned at this point with a ref to the guideline section.

most people will understand "can't be edited" as it is impossible. And the guidelines talk about converting away from @startsection the connection is not that clear

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,3036 @@
...
+%    the first argument of \cs{@startsection} with an attached \texttt{-@startsection}. 
+%    As they are created on-the-fly they can be only edited 
+%    \emph{after} the first use of the heading command. Also
+%    it is not possible to define two different heading commands
+%    of the same level with \cs{@startsection}. 
+%  
+% \subsection[label=sec:convert]{Converting heading commands defined with \cs{@startsection}}
+% 
+% To convert a heading command defined with \cs{@startsection} to the new interface suitable 
+% instances must be declared. How this can be done is demonstrated here with the \cs{section}
```
FrankMittelbach:
```suggestion
% instances must be declared. The same needs to happen if one wants to alter the layout of a heading that was defined with \cs{@startsection} in the document preamble. How this can be done is demonstrated here with the \cs{section} 
```

required/latex-lab/latex-lab-sec-template.dtx
FrankMittelbach: but for many users it is a hurdle because more complicated than it has to be. They don't want you to force them to educate themselves. But I'm not too woried that this is ready by many  ... so fine :-)

required/latex-lab/latex-lab-sec-template.dtx
FrankMittelbach: I think it receives \cs{NoNalue} these days

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -0,0 +1,3036 @@
...
+% \DescribeMacro\secdef
+%    In contrast \cs{secdef} only does the parsing and delegates the
+%    layout to two commands which can contain arbitrary code (usually
+%    hardwired) so that there is no realistic chance to take this
+%    apart and figure out what values should be used for what template
+%    parameters.
+%
+%    We therefore do not attempt to model this, but instead just use
+%    the standard layout from \LaTeX's standard classes for
+%    \cs{chapter} and \cs{part} and error otherwise.
```
FrankMittelbach:
```suggestion
%    \cs{chapter} and \cs{part} if we can identify that the \cs{secdef} is used to define one of these 
%    heading commands. This means that taggingg works, but that most likely the layout is not truthfully 
%    replicated. Otherwise, if we don't know what the  \cs{secdef} was intended for, we generate an error. 
```

FrankMittelbach (approved):
a few suggestions for changes here and there, otherwise approved

required/latex-lab/latex-lab-sec-template.dtx
u-fischer: well the text doesn't say "can't be edited", it says they can "only be edited after the first use". But I added a sentence pointing to to the next section.

FrankMittelbach: my point was that you can "edit" them without first using the heading command if you provide the necessary instance, e.g., section-@startsection, in the preamble. That's a form of edit/change-of-layout and in that case (if that instance was specified in the premable nothing is generated on the fly.

FrankMittelbach: but I'm ok the way it is now

u-fischer: sure, but if you do the work to provide an instance, then I think it is better to go all the way and use also the new parse command, that is only three lines more and you get rid of the long instance name.  So I didn't mention this option.

PR #1980, opened 2026-01-22
Initial 'collect all the info' doc file

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/base/doc/ltnews43.tex
+++ b/base/doc/ltnews43.tex
@@ -275,7 +275,11 @@ be to include all of new features here. We also hope to collect up existing
 documentation in the longer-standing sources (going back to before \LaTeXe{}!).
 
 This is work in progress so we expect the document to grow and its structure
-(based on feedback and experience with it use) might change over over time.
+(based on feedback and experience with it use) might change over over time. In
+particular, the current content is made up of files that were written to be
+read independently: as such, there is some duplication, suboptimal ordering and
+formatting variation. That will all be addressed over time, as we aim for a
+single coherent document.
 
 \section{Bug fixes}
 


gucci-on-fleek:
Maybe it would be a good idea to add `usrguide`, `usrguide-historic`, `clsguide`, and `fntguide` here too?

josephwright:
@gucci-on-fleek Two reasons

1. The first pass here is going with docs that are in the same class already, so it's workable at low cost: that means things already in `l3doc`, which excludes the 'classical' files
2. I'm not sure that we want to put that info into a new doc - `usrguide`, etc, are written in a different style to something like `lthooks` - I was imagining more that we might move material around (in particular, I suspect we may want to redistribute the `ltcmd` docs and perhaps some of the case changing stuff)

gucci-on-fleek:
@josephwright 

> 1. The first pass here is going with docs that are in the same class already, so it's workable at low cost: that means things already in `l3doc`, which excludes the 'classical' files

Ok, well hard to argue with that :)

> 2. I'm not sure that we want to put that info into a new doc - `usrguide`, etc, are written in a different style to something like `lthooks` - I was imagining more that we might move material around (in particular, I suspect we may want to redistribute the `ltcmd` docs and perhaps some of the case changing stuff)

The advantage is discoverability—right now, people have no idea where to even look for official LaTeX documentation, so they just blindly Google things and copy whatever the first result says, or ask ChatGPT and do whatever it tells them to do. For example, if someone compiles the following document and wonders why it doesn't work

```tex
\documentclass{article}

\usepackage{amsmath}[leqno]

\begin{document}
    \begin{equation}
        a = b + c
    \end{equation}
\end{document}
```

they'd somehow need to know that `usrguide-historic` is where `\usepackage` is documented, whereas if everything was in one file, then they could just <kbd>Ctrl+F</kbd> for it. Similarly, how many people even know that `fntguide` exists? 

Right now, the best way for me to find LaTeX documentation is to `grep` through all the manual sources to figure out which PDF file I need to open, but this is probably beyond what most users would ever do. So even if the tone/themes aren't very cohesive, I'd argue that a single incohesive document that people can find is better than multiple cohesive documents that nobody ever reads (because they can't find them/don't know that they exist).

josephwright:
@gucci-on-fleek 

> The advantage is discoverability—right now, people have no idea where to even look for official LaTeX documentation, so they just blindly Google things and copy whatever the first result says, or ask ChatGPT and do whatever it tells them to do. For example, if someone compiles the following document and wonders why it doesn't work
> 
> ```
> \documentclass{article}
> 
> \usepackage{amsmath}[leqno]
> 
> \begin{document}
>     \begin{equation}
>         a = b + c
>     \end{equation}
> \end{document}
> ```

That's `amsmath` so doesn't 'belong' here anyway (at least, not for quite some time).

> they'd somehow need to know that `usrguide-historic` is where `\usepackage` is documented, whereas if everything was in one file, then they could just Ctrl+F for it. Similarly, how many people even know that `fntguide` exists?

They are all covered in TLC3 (and TLC2 and I suspect TLC1 - I don't have a copy). And yes, I know that saying 'read the book' is not a popular answer :-)

One thing we did say is 'to do' is cross-link between the core docs for exactly this reason: first I need some form of agreed text here (likely needs some intro as well as the current content), when we should add some more-or-less identical text at the top of each main doc file

> This file, `<name>`, covers `<one-liner description>`. You may also want to look at the following documentation files:
> 1. `<name-1>`: `<desc-1>`
> 2. `<name-2>`: `<desc-2>`

josephwright:
@gucci-on-fleek Key point: I did say 'initial' :-) I'd like to get _something_ into the next dev release, so we can for example look at removing the `X-doc` files and having some sense of _moving_ toward a predictable single location. But I can't do all of that in one go!

josephwright:
Oh, and the 'cross-refs' likely should include `interface3`.

josephwright:
An aside: I wonder if we might want to pencil in a split between 'user' and 'API' docs for '`ltx-doc`' (re-development of `l3doc`): currently we get both for example `\AddToHook` and `\hook_gput:nnn`, and it would be nice if they were separate. (In `siunitx`, I currently do the split by having an entirely separate `.tex` file for the user docs, and have the code API and code in the `.dtx` files - that works OK too, but would mean splitting the kernel code back out and I suspect we don't want to do that.)

gucci-on-fleek:
@josephwright 

> For example, if someone compiles the following document and wonders why it doesn't work
> > ```
> > \documentclass{article}
> > 
> > \usepackage{amsmath}[leqno]
> > 
> > \begin{document}
> >     \begin{equation}
> >         a = b + c
> >     \end{equation}
> > \end{document}
> > ```
> 
> That's `amsmath` so doesn't 'belong' here anyway (at least, not for quite some time).

The answer here doesn't involve `amsmath`; it involves knowing that package options go in square brackets _before_ the package name, while the square brackets _after_ the package name only take a date/version. And `usrguide-historic` is the (only) place where the syntax of `\usepackage` is documented, which is not exactly the first place a new user would think to look.

> > they'd somehow need to know that `usrguide-historic` is where `\usepackage` is documented, whereas if everything was in one file, then they could just Ctrl+F for it. Similarly, how many people even know that `fntguide` exists?
> 
> They are all covered in TLC3 (and TLC2 and I suspect TLC1 - I don't have a copy). And yes, I know that saying 'read the book' is not a popular answer :-)

For this specific example, TLC3 documents the second optional argument in the rollback section, but I couldn't find anywhere that explicitly defines the exact syntax of `\usepackage` like `usrguide-historic` does (but it's a _huge_ book, so I may have missed it somewhere).

> Key point: I did say 'initial' :-) I'd like to get _something_ into the next dev release, […] But I can't do all of that in one go!

Oh, fair enough! I just assumed that you'd just need to add `\include{usrguide.tex}` and everything would just magically work as expected, but if it's more complicated than that, then I agree that it definitely makes sense to delay that until later.

> An aside: I wonder if we might want to pencil in a split between 'user' and 'API' docs for '`ltx-doc`'

My personal preference is for everything in one document, since that makes it easier to search for things, but splitting things into separate 'user' and 'API' docs seems fairly reasonable to me. (But I really like big documents—I have a 17k page PDF containing every single TUGboat issue that I search through occasionally)

josephwright:
@gucci-on-fleek 
> The answer here doesn't involve `amsmath`; it involves knowing that package options go in square brackets _before_ the package name, while the square brackets _after_ the package name only take a date/version. And `usrguide-historic` is the (only) place where the syntax of `\usepackage` is documented, which is not exactly the first place a new user would think to look.

Ah, right - we might want to look at that :-) Largely what I was getting at with 'collect stuff up' - more-ro-less the plan is to go thorough the sources and make sure all of the document commands are covered. Some of that might be in `usrguide`, but I don't think it all is.

> > An aside: I wonder if we might want to pencil in a split between 'user' and 'API' docs for '`ltx-doc`'
> 
> My personal preference is for everything in one document, since that makes it easier to search for things, but splitting things into separate 'user' and 'API' docs seems fairly reasonable to me. (But I really like big documents—I have a 17k page PDF containing every single TUGboat issue that I search through occasionally)

I wasn't necessarily thinking it would mean the code API didn't show up, more that like the doc/implementation split, it would allow some control. I think it would make sense to have 'user' docs collected _then API docs, in the same way `source3` has API then implementation. But that needs the two parts to be marked independently.

dbitouze:
> [...] docs for 'ltx-doc'

BTW, why such an oddity of dropping the vowels of “latex”. New users would probably miss what “ltx” means

josephwright:
> > [...] docs for 'ltx-doc'
> 
> BTW, why such an oddity of dropping the vowels of “latex”. New users would probably miss what “ltx” means

We want a 'namespace' that we can use for out-of-the-box taggable classes which is not too long and which might at some future point be 'assumed'. So we've got `ltx-talk` and probably going to have `ltx-doc`. 'latex' here would likely be a bit odd, and it's definitely getting a little long.

base/doc/latex-cmds.tex
```diff
@@ -0,0 +1,163 @@
...
+\def\toclevel@part{-1}
+\def\maketitle{\chapter{\@title}}
+\let\thanks\@gobble
+\let\DelayPrintIndex\PrintIndex
+\let\PrintIndex\@empty
+\providecommand*{\hexnum}[1]{\text{\texttt{\char`\"}#1}}
+\makeatother
+
+\clearpage
+\tableofcontents
```
FrankMittelbach, line 143: there is no contents generated by \tableofcontents

base/doc/latex-cmds.tex
```diff
@@ -0,0 +1,163 @@
...
+\clearpage
+\pagenumbering{arabic}
+
+\DisableImplementation
+
+\part{Hooks}
+
+\DocInput{lthooks.dtx}
```
FrankMittelbach, line 154: there is also no contents generated if \tableofcontents is used as part of the individual files and I think it probably should too (or it should be fully silenced).

ltnews.tex has an implementation how to do that (it has a full table and individual tocs)

base/doc/latex-cmds.tex
```diff
@@ -0,0 +1,163 @@
...
+\DocInput{lttemplates.dtx}
+
+\clearpage
+
+\begingroup
+  \def\endash{--}
+  \catcode`\-\active
+  \def-{\futurelet\temp\indexdash}
+  \def\indexdash{\ifx\temp-\endash\fi}
+  \DelayPrintIndex
```
FrankMittelbach, line 177: The index lists none of the hooks (or sockets). In the original documentation some are in the index.
I think all should be in the index. In a separate branch I have a draft implementation to figure out how that could happen. That could be added when ready. But I'm a bit  surprised that currently none are showing at all. Why is that?

base/doc/ltnews43.tex
```diff
@@ -261,6 +261,19 @@ \subsection{Revision of \cs{protected} status of functions in templates}
...
+Adding new features to \LaTeX{} means writing new documentation. To date, each
+new source has had separate documentation, for example
+\texttt{lthooks-doc.pdf}. This is a convenient way for development to take
+place, particularly prior to code being stable. However, it makes it
+challenging to find information.
+
+We have therefore started to collect all of this information into a single
+documentation, currently called \texttt{latex-cmds.pdf}. The aim over time will
+be to include all of new features here. We also hope to collect up existing
+documentation in the longer-standing sources (going back to before \LaTeXe{}!).
```
FrankMittelbach, line 275: perhaps add something like

\par
This is work in progress so we expect the document to grow and its structure (based on feedback and experence with it use) might change over over time.

base/changes.txt
```diff
@@ -6,6 +6,9 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2026-01-22  Joseph Wright  <Joseph.Wright@latex-project.org>
+	* latex-cmds.tex: New file
```
FrankMittelbach: say what it is for in 10 words

base/doc/latex-cmds.tex
josephwright: Seems to work for me: this is the same setup as in `interface3`

base/doc/latex-cmds.tex
josephwright: Other than a stray `Contents` heading, this is doing what I'd expected (no TOC for each file, just one at the start)

base/doc/latex-cmds.tex
josephwright: Again, seems to be works-for-me

base/doc/latex-cmds.tex
FrankMittelbach: I have installed your branch and ran l3build doc latex-cmds inside

base/doc/latex-cmds.tex
FrankMittelbach: ah yes, looked at the wrong "Contents" ... there is a global one on top. But then perhaps you should set it explicitly to do nothing afterwards

josephwright: That's easy to arrange :-)

base/doc/latex-cmds.tex
```diff
@@ -0,0 +1,172 @@
...
+
+The development version of the bundle can be found at
+
+   https://github.com/latex3/latex2e
+
+for those people who are interested.
+
+\fi
+
+\documentclass{l3doc}
```
FrankMittelbach: actually I think you should  use l3in2edoc if that works

muzimuzhi:
Add also the `ltproperties` module?

BTW it seems a leftover that `ltproperties.dtx` is not marked as "L3 layer module" in `format.ins`
https://github.com/latex3/latex2e/blob/3c492b44f3ebaa6a8b741aab2cc00176f06fdeba/base/format.ins#L196

FrankMittelbach:
agree to both, please update format.ins in this pr as well

base/doc/ltnews43.tex
```diff
@@ -261,6 +261,22 @@ \subsection{Revision of \cs{protected} status of functions in templates}
...
+\section{New or improved documentation}
+
+Adding new features to \LaTeX{} means writing new documentation. To date, each
+new source has had separate documentation, for example
+\texttt{lthooks-doc.pdf}. This is a convenient way for development to take
+place, particularly prior to code being stable. However, it makes it
+challenging to find information.
+
+We have therefore started to collect all of this information into a single
+documentation, currently called \texttt{latex-cmds.pdf}. The aim over time will
```
cfr42: 'documentation' -> 'document'?

cfr42:
@gucci-on-fleek  I do not like huge documents much. It takes too long to search them. And unless they are optimised, very large PDFs are more-or-less useless online, which is the relevant place for an awful lot of users. E.g. some of the context docs just give me errors. So if part of the aim is to provide a better option than whatever-chatGPT-says, then I think you need to limit consolidation to more usable sizes than ConTeXt favours.

muzimuzhi:
I checked the compiled `latex-cmds.pdf` and here is it's first two level bookmarks:
```
Contents                                          Page 2
I Hooks                                           Page 5
  1 LaTeX's hook management                       Page 6
  2 The ltfilehook documentation                  Page 36
  3 The ltcmdhooks module                         Page 42
  4 The ltshipout documentation                   Page 48
II Page design                                    Page 56
  5 The ltmarks.dtx code                          Page 57
  6 The ltpara.dtx code                           Page 67
III Design-level tools                            Page 76
  7 LaTeX's socket management                     Page 77
  8 The lttemplates.dtx code                      Page 86
Index                                             Page 97
  ...
```

Probably we need to revise titles in these modules, or do tricks to replace them with new, more user-friendly titles in `latex-cmds.tex`.

<details><summary>Steps to extract bookmarks</summary>
<p>

1. Download the artifact named `Documentation-base-3`.
2. Extract `base/doc/latex-cmds.pdf`.
3. Save the following Python script as `extracttoc.py` and run it using `uv run extracttoc.py latex-cmds.pdf` ([`uv run`](https://docs.astral.sh/uv/reference/cli/#uv-run) doc):
   ```py
    # /// script
    # dependencies = [
    #   "pymupdf",
    # ]
    # ///
    import sys
    import fitz

    with fitz.open(sys.argv[1]) as doc:
        for level, title, page_num in doc.get_toc():
            if level < 3:
                print(f"{('  '*(level-1) + title).ljust(50)}Page {page_num}")
            if level == 1 and title.startswith("Index"):
                print("  ...")
                break
   ```

</p>
</details>

josephwright:
> agree to both, please update format.ins in this pr as well

I'll do that on `develop` as it's not directly part of the PR

josephwright:
> Probably we need to revise titles in these modules, or do tricks to replace them with new, more user-friendly titles in `latex-cmds.tex`.

Yes: I think if/when we drop e.g. `ltooks-doc`, we should revise the content of the sources. That's much the same as for `l3....dtx`: whilst the files can be compiled indepently, they are set up with the primary target `interface3`.

base/doc/latex-cmds.tex
```diff
@@ -0,0 +1,174 @@
...
+\pagenumbering{roman}
+
+\maketitle
+
+\begin{abstract}
+  This document contains user (and programmer) documentation for much of the
+  \enquote{new} code added to the \LaTeX{} kernel in recent years. Over time,
+  this coverage is expected to be expanded to include the majority of user
+  commands defined in \texttt{latex.ltx}.
+\end{abstract}
```
FrankMittelbach, line 59: I suggest that we also explain there

- that the individual parts are written at different times and were meant to stand on their own
- that therefore explanations are currently not necessary in the right order (or duplicated)
- headings are uniform, etc
- and that we expect this to be sorted out over time

for example the socket intro explains the relation between templates / hooks / sockets way after all the hook interfaces are discussed

FrankMittelbach: perhaps something similar should also be said in ltnews43.

josephwright:
Any thoughts on whether we should drop the `-doc` files at this stage or wait? I'm mindful we have the PDF info twice - both confusing and bloating.

FrankMittelbach:
> Any thoughts on whether we should drop the `-doc` files at this stage or wait? I'm mindful we have the PDF info twice - both confusing and bloating.

I would keep them for now, there are a bunch of dependencies, eg places where they are referenced. In fact, as you have heard opinions are divided on just one big doc or several small ones. I don't think the bloating is a big problem and we might end up with keeping one big and several small ones (with the small ones all having a box on top that states that this is an excerpt of a bigger documentation which is  "link").

davidcarlisle:
I'd keep the existing doc files for now, the new combined one is guarded by multiple warnings of it being an initial version with change expected, so it doesn't feel right to drop the older doc yet, at som epoint we might want to reassess.

Skillmon:
Sorry for being absolutely passive on this here (not reviewing or supporting this at all) after being so vocal about it in the SX chat. Anyway, I wanted to say thanks for the consideration and the work that went into this! I really think that this is a step in the right direction!

cfr42:
If you are really going for One-Doc-To-Rule-Them-All, can you please linearise it so it is actually usable online in, say, chromium? A whole lot of TeX docs just show no pages at all for me. The PDF never finishes downloading. I generally have it locally but that is not true for an increasing number of users. And many who do don't know they have the stuff or don't know how to access it. 

A solution which makes hard-to-find documentation impossible-to-read documentation will not solve anything.

To be clear: I'm not against consolidation - it would be wonderful to have fewer possible places something might be documented, if only I could guess where that is - but you can have too much of a good thing.

Perhaps bear in mind that a common response to suggesting somebody look at the tikz manual is that it is too long. That's the response even though the manual is very well-designed with introductory tutorial material to begin, tikz reference material next and lower-level details later. It also has a plethora of examples and very often, clicking on a command or key word in an example takes you to the page explaining that command or key word. So ctrl+f is generally fast because you usually only need to find the first result to be 1-click from the answer. 

The documentation you're preparing will obviously (understandably, inevitably, unavoidably) be much, much less user-friendly. So bear in mind the likely psychological/social effects fo clicking a link and finding 3 hours later you have 1,000 pages which might as well be written in Ancient Martian Runes. 

Note: bear in mind that analytic philosophy is essentially professional training in devils' advocacy.

gucci-on-fleek:
@cfr42 

> can you please linearise it so it is actually usable online in, say, chromium? A whole lot of TeX docs just show no pages at all for me. The PDF never finishes downloading.

Linearizing is good in general, but it only really matters for documents that are large in the sense of “large disk size”, not in the sense of “lots of pages”. The LaTeX manuals have no bitmapped images or vector graphics, and usually only use a single font, so the documents tend to be fairly tiny. I can't imagine any potential LaTeX manual having more pages than `source3.pdf`, and that only takes up 7.24 MB, which should download quick enough over most slow-but-not-dialup connections.

(I tested [`https://texdoc.org/serve/source3/0`](https://texdoc.org/serve/source3/0) just now and it loaded pretty quick in Chromium, even over a relatively slow network connection)

cfr42:
@gucci-on-fleek 

> Linearizing is good in general, but it only really matters for documents that are large in the sense of “large disk size”, not in the sense of “lots of pages”. The LaTeX manuals have no bitmapped images or vector graphics, 

That's true right now, but it really depends what ends up in the format and what documentation gets merged. Some of the documentation for tagging and font loading does have graphics, for example. 

Now, I admit, that the documentation seems unlikely to evolve into the equivalent of ConTeXt docs or the Ti*k*Z manual, so perhaps it is a pointless suggestion.

PR #1995, opened 2026-02-02
Convert blank arg to empty keyval list in ltcmd

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [x] Test file(s) added

### DIFF
--- a/required/latex-lab/testfiles/tagging-status.luatex.tlg
+++ b/required/latex-lab/testfiles/tagging-status.luatex.tlg
@@ -35,6 +35,7 @@ Don't change this file in any respect.
  ltxcmds.sty
  pdftexcmds.sty
  infwarerr.sty
+ hycolor.sty
  nameref.sty
  refcount.sty
  gettitlestring.sty
@@ -57,7 +58,6 @@ Don't change this file in any respect.
  --------------------------
  latex-lab-enumitem.sty
  latex-lab-testphase-sec-template.sty
- hycolor.sty
 ====================================
  End of status report
 ====================================
--- a/required/latex-lab/testfiles/tagging-status.tlg
+++ b/required/latex-lab/testfiles/tagging-status.tlg
@@ -35,6 +35,7 @@ Don't change this file in any respect.
  ltxcmds.sty
  pdftexcmds.sty
  infwarerr.sty
+ hycolor.sty
  nameref.sty
  refcount.sty
  gettitlestring.sty
@@ -57,7 +58,6 @@ Don't change this file in any respect.
  --------------------------
  latex-lab-enumitem.sty
  latex-lab-testphase-sec-template.sty
- hycolor.sty
 ====================================
  End of status report
 ====================================


base/doc/usrguide.tex
```diff
@@ -581,6 +581,10 @@ \subsection{Auto-converting to key--value format}
    (short-text = {Not = to a keyval!})
 \end{verbatim}
 
+If the argument is completely blank or is a brace group which is blank, the
+conversion results in an empty keyval list, not \verb*|key = |. This is the
```
muzimuzhi: Using the on-going `short-text` example,

```suggestion
conversion results in an empty keyval list, not \verb*|short-text = |. This is the
```

base/doc/ltnews43.tex
```diff
@@ -255,6 +255,19 @@ \subsection{Revision of handling of \enquote{no value} concept}
...
+
+A second revision to \cs{NewDocumentCommand}, etc., in this release applies to
+conversion of classical arguments to key--value form, for example
+\begin{verbatim}
+  \DeclareDocumentCommand \caption
+    {s ={short-text} +O{#3} +m}
+\end{verbatim}
+This will convert a classical optional argument of \verb|\caption| to a keyval
+one called \texttt{short-title}. In this release, if the optional argument is
+given but entirely empty, the result will be an entirely empty argument rather
```
Skillmon:
```suggestion
given but blank, the result will be an entirely empty argument rather
```

base/doc/usrguide.tex
```diff
@@ -581,6 +581,10 @@ \subsection{Auto-converting to key--value format}
    (short-text = {Not = to a keyval!})
 \end{verbatim}
 
+If the argument is completely blank, the conversion results in an empty keyval
+list, not \verb*|short-text = |. This is the desired outcome in almost all
```
Skillmon:
```suggestion
list, not \verb*|short-text = |. This is anticipated to be the desired outcome in almost all
```

I for one have no idea whether this is really the desired outcome in almost all cases. I wonder why you are convinced it is.

base/ltcmd.dtx
```diff
@@ -4721,7 +4722,9 @@
 %   the value for the text key. Alternatively, if the start of the input is
 %   |=,| then it is forced to be key--value. To avoid needing to worry about
 %   catcodes for this, and to allow spaces around the |=|, we use a
-%   series of steps rather than a delimited argument.
+%   series of steps rather than a delimited argument. We also deal with the
+%   case where the entire argument is empty (or is an entirely empty brace
```
Skillmon:
```suggestion
%   case where the entire argument is blank (or is an entirely empty brace
```

base/ltcmd.dtx
```diff
@@ -4730,15 +4733,15 @@
   }
 \cs_new_protected:Npn \@@_arg_to_keyvalue_braces:nnn #1#2#3
   {
-    \tl_if_head_is_group:nT {#1}
+    \bool_lazy_and:nnTF
+      { \tl_if_head_is_group_p:n {#1} }
+      { \tl_if_blank_p:o { \use_none:n #1 } }
+      { \tl_set:Nx \ProcessedArgument { #2 = { \exp_not:n #1 } } }
```
Skillmon, line 4739:
```suggestion
      { \tl_set:Ne \ProcessedArgument { #2 = { \exp_not:n #1 } } }
```

josephwright: This one is by-design: we need to wait long enough that we avoid the 'older `expl3`, newer kernel'. My assumption is we will do all of them in one go, perhaps around. years after the `x` to `e` change so late-2028.

base/doc/usrguide.tex
josephwright: Two reasons: 1) It's very unlikely someone would do `\secton[]{foo}` and want say an empty TOC line 2) Several existing environments ignore an entirely empty optional arg. even if it normally means 'short version'.

Skillmon: 1) What do you reckon would someone want if `\section[]{foo}` is input if not an empty ToC entry? There ought to be a reason for the empty optional argument.

2) Because historically it was easier to parse for an empty argument instead of what `o` offers. Established behaviour of commands and environments doing that shouldn't be interpreted as being the same as this.

FrankMittelbach: > 1. What do you reckon would someone want if `\section[]{foo}` is input if not an empty ToC entry? There ought to be a reason for the empty optional argument.

With the first argument of commands and environments becoming "key-val" in many cases you have the typical situation with key-val arguments that people have a list of key-vals with some of which being commented out (but kept available for easy activation. You see that currently with `\documentclass` or `\usepackage`. In such a case commenting out the last key-val shouldn't turn that into `default-key=<empty>`.

For existing optional arguments where we extend to allow key/val or simple string making an empty optional argument mean `default-key=`made little sense in the past, eg what good is an empty line in the TOC, or what should be the meaning of `Theorem 1 ().` in the case of theoremlike envs, plain LaTeX behaved like this, but amsthm interpreted an empty optional as non-existent, so some inconsistency there.

Our plan is to normalize that and always interpret in case of `={default}o` a given `[]` as an empty keyval and not as an empty string to be passed on as `default=`. Both choices are possible, but after some experimentation with the approach we decided for going this way. If a user really wants to pass an empty string, then `[default=]` would do the trick.

Skillmon: I still don't see the "almost all cases" but you're the team. I think special casing too much is bound to bite you in the long run (hence I don't really understand this change at all, and it won't be expected behaviour for me, certainly) but I trust your judgement.

u-fischer: @Skillmon there is no special casing. In a normal key-value argument an empty argument is an empty key-val, e.g. `\includegraphics[]{example-image}` calls `\setkeys{Gin}{}`.  If you have a non-key-val argument you then have an empty text argument. In the handling of the `={shorttitle}o` argument type there is an ambiguity:  an empty argument could mean an empty list or an empty value for the `shorttitle`  key. So you have to make a choice. There are arguments for both but non is more special than the other.

davidcarlisle: @Skillmon  it's hard to predict the future, but for cases where the expected use is kv and the `=` processing of a non-kv string is just legacy compatibility, it's a bit odd if the not unlikely empty argument triggers the compatibility processing rather with what is the expected behaviour for an empty kv list .

So yes, now, if you are used to `\caption` having a non kv argument for the short form then you may expect `\caption[]{wibble}` to make an empty list of tables entry (but why would you do that?) but once all normal tutorials have only described the usable keys and the legacy syntax is only described in some footnote on historic usage, the expected interpretation of `\caption[]` would (possibly?:-) be that of an empty key list.

Skillmon: @u-fischer the difference between interpreting it as the value for the specified fallback key and interpreting it as an empty key=value list is that the latter *requires* a sentence more in any documentation of the feature, so evidently is a special case, whereas saying "If it's used and no `=` is found, it's handled as if it was the argument to the key `shorttitle`." would suffice in the former.

@davidcarlisle the "why would you do that" doesn't matter. It can happen, though obviously `\caption`, `\chapter`, etc. are poor examples in other macros the purposeful specification of something empty might make much more sense. This is about expected behaviour from a user's point of view. If I know a macro will use `bar` if I use it as `\foo[bar]{baz}` for some specific component of its behaviour it's perfectly reasonable to expect it to use an empty value for that same component if I use `\foo[]{baz}`.

But as I said, it's the team's call not mine so whatever you decide goes anyway. So we can stop this discussion (at least on my end, though I might reply to any further messages :)).

josephwright:
@Skillmon  You might guess that the team had some discussions about this PR :-) I'll adjust the news to try to reflect the discussion.

PR #2003, opened 2026-02-11
Remove 'fake math' from \text(sub|super)script

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [n/a] Test file(s) added - enough tests already :-)

## Notes

- I've not added specific tests as the changes show in several places already
- I've set the default offsets based on Appendix G, so 'no change' for existing documents
- I'm not totally sure about the naming of commands holding the offset and spacing
- Quite where the docs should go is open: I wondered about `latex-cmds`?

### DIFF
--- a/base/ltfloat.dtx
+++ b/base/ltfloat.dtx
@@ -1367,7 +1367,10 @@
 %   default for |\textsuperscript@offset| arises so we get the same result
 %   as from the math mode predecessor. The one for |\textsuperscript@space|
 %   is needed as in math mode, |\scriptspace| is not dropped at a line break
-%   (as it's a box adjustment).
+%   (as it's a box adjustment). Lua\TeX{} has rather easier to understand
+%   names for the various parameters, and these will be correct whether
+%   classical or OpenType fonts are in use for math, so we set up with
+%   these if available.
 %    \begin{macrocode}
 %</2ekernel>
 %<latexrelease>\IncludeInRelease{2026/06/01}%


josephwright:
This is step 1 of removing fake math: that's reflected in the LaTeX News text, which is intended to stay more or less the same as additional parts of the task are done.

base/doc/ltnews43.tex
```diff
@@ -278,6 +278,30 @@ \subsection{Revision of \cs{protected} status of functions in templates}
...
+  \item \cs{textsuperscript} and \cs{textsubscript}
+\end{itemize}
+
+To avoid as far as possible changes to existing output, the standard settings
+continue to use math font dimension values for placing super and
+subscripts.\footnote{The calculation used for placement of math mode sub- and
+superscripts is described in Appendix~G of \emph{The TeX{}book}, and relies on
+multiple math font dimensions.} For new documents, we suggest setting to a
+predictable offset based on text font dimensions (for example a fixed
+percentage of the x-height).
```
FrankMittelbach: We recommend basing the values on text font sizes, but we nowhere give an example how that should be done. I think this should be described either here or possibly in usrguide below.

base/ltfloat.dtx
```diff
@@ -31,7 +31,7 @@
 %
 %<*driver>
 % \fi
-\ProvidesFile{ltfloat.dtx}[2024/04/22 v1.2g LaTeX Kernel (Floats)]
+\ProvidesFile{ltfloat.dtx}[2026/02/10 v1.2h LaTeX Kernel (Floats)]
```
FrankMittelbach: The macros have been in ltfloat in th past, but why? I don't think they belong there do they? I wonder if we should  take the opportunity to move them. Granted, we don't yet have something like lttext.dtx, but I still wonder if that wouldn't be the right  time to start something  like that. We could of course do this in a separate pr but might as well start it here (perhaps).

josephwright: Sure, not an issue

base/doc/ltnews43.tex
josephwright: I was not sure what the official position is here: do we show `\textfont`, etc. at all? Also, there's an OpenType parameter that's likely appropriate - I'm not totally sure how that's accessed!

FrankMittelbach: If you give a teaser like the footnote then I'm sure we will get the question, how to do that, so we might as well give some advice (at least in the code documentation). Perhaps @zauguin has some advice on how to do that with OpenType fonts.

base/doc/ltnews43.tex
```diff
@@ -278,6 +278,30 @@ \subsection{Revision of \cs{protected} status of functions in templates}
 necessary. As part of this, we have adjusted templates such that keys
 declared as functions are stored with \cs{protected} status.
 
+\subsection{Removal of \enquote{fake math} from the kernel}
+
+Internally, some text constructs in \LaTeX{} have used math mode internally, as
+this allowed use of the built-in positioning of material with limited macro
```
u-fischer, line 284: "Internally"  used twice

base/doc/usrguide.tex
```diff
@@ -1454,6 +1454,38 @@ \section{Case changing}
...
+commands: previous versions were implemented using math mode internally for
+placement.} Two aspects of placement can be adjusted by setting the control
+commands to updated values, both dimension expression (see \cs{dimeval}):
+\begin{itemize}
+  \item The vertical offset of the sub- or superscript. The standard setting
+    uses the font dimensions of the current math font, but for new documents,
+    using a value based on the text font is recommended. For example
+    |\renewcommand*\textsuperscript@offset{0.86\fontdimen5\font}| will set
+    the vertical offset to $86\,\mathrm{\%}$ of the x-height of the current
+    font.
```
davidcarlisle: rather than using fontdimen5 and calling it a percentage of the ex-height isn't it simpler to use ex units and show it set to .86ex ? expecially coming from opentype fonts ex has a natural meaning elsewhere but fontdimen5 is ...

base/ltfloat.dtx
davidcarlisle: agree with moving this, perhaps ltboxes ?

base/doc/usrguide.tex
```diff
@@ -1454,6 +1454,38 @@ \section{Case changing}
...
+  \item The space added after the sub- or superscript: the standard setting is
+    a fixed distance which is also applied in math mode (\cs{scriptspace}).
```
FrankMittelbach: maybe something like 

a fixed distance; by default we use the one that would also be applied in math mode (\cs{scriptspace}).

Otherwise I think mentioning mathmode is not understandable.

base/doc/usrguide.tex
FrankMittelbach: sounds like a good idea

FrankMittelbach:
don't really think ltboxes is better than ltfloat (this is text processing not anything to do with boxes really

josephwright:
Other than the possible move, I think I've addressed things

base/doc/usrguide.tex
davidcarlisle: currently we follow math mode and don't actually add a space but enlarge the box, this caught me out when reviewing yesterday as it doesn't appear as you'd expect in box logging, but I wonder if we should change this. This description makes it sound like a skip is being added, you might usefully want to add a font-dependent stretch space such as a thinspace, but it has to be a dimen if we just adjust the box width (although you could define it to be font depenedent eg `0.1em` which is already a slight improvement on `\scriptspace`)

FrankMittelbach: that too seems like a good suggestion, what do you think @josephwright ?

base/ltfloat.dtx
```diff
@@ -1350,32 +1350,69 @@
...
-%<latexrelease>                 {\@textsuperscript}{superscript baseline}%
-\def\@textsuperscript#1{%
-  {\m@th\ensuremath{^{\mbox{\fontsize\sf@size\sf@size#1}}}}}
+\protected\def\@textsuperscript#1%
+  {%
+    \check@mathfonts
+    \leavevmode
+    \begingroup
+      \sbox0{\fontsize\sf@size\sf@size#1}%
+      \wd0\dimexpr\wd0+\textsuperscript@space\relax
```
davidcarlisle: as commented in the doc we could avoid extending the box and instead add a  \hskip\textsuperscript@space  after the box (with or without a nobreak penalty)

base/doc/usrguide.tex
josephwright: I was mainly aiming to get minimal changes in the `.tlg` files, or at least ones I could understand :-) I'm quite happy making this more flexible - really depends if you want the space _in_ the box or outside of it (I guess I weakly favour keeping the space inside the box).

davidcarlisle: yes the space would need to be outside the box to usefully scale (or be dropped at a linebreak)

josephwright: I'm not sure if we want to drop at a linebreak: doesn't happen at the present so would affect existing docs

FrankMittelbach: can there be a break currently? I mean `\textsuperscript{foo}bar` would not have a possible line break before `bar` and so I think we shouldn't add one, so `\nobreak` I guess

FrankMittelbach: > I was mainly aiming to get minimal changes in the `.tlg` files, or at least ones I could understand :-) I'm quite happy making this more flexible - really depends if you want the space _in_ the box or outside of it (I guess I weakly favour keeping the space inside the box).

the test changes so far prove that you succeded in that. So a change now would change the test files but with a hopefully understandable pattern. And yes, we could move the space inside the box I guess, perhaps that's better.

FrankMittelbach: > yes the space would need to be outside the box to usefully scale (or be dropped at a linebreak)

ah yes, didn't think about that, so forget previous remark

davidcarlisle: > can there be a break currently? I mean `\textsuperscript{foo}bar` would not have a possible line break before `bar` and so I think we shouldn't add one, so `\nobreak` I guess

@FrankMittelbach 
well that's true but very unusual input, if possible I think it would be good to allow

```
this\textsuperscript{a} that
```

and if a linebreak happens at the space you also drop the scriptspace.  No problem with making it non breakable by default.

FrankMittelbach: yes, probably an improvement even though in theory it could lead to slightly different line breaks

josephwright: If we want to have a `\nobreak`, I can't have a simple 'space' in the var - I have to have typesetting commands instead.

davidcarlisle: Yes exactly, a command with initial def `\nobreak\kern\scriptspace` or similar

base/doc/usrguide.tex
```diff
@@ -1454,6 +1454,42 @@ \section{Case changing}
...
+  |\textsuperscript@space|
+\end{decl}
+The standard settings for these commands use the same placement parameters as
+\TeX{} would use in math mode.\footnote{This reflects the history of these
+commands: previous versions were implemented using math mode internally for
+placement.} Two aspects of placement can be adjusted by setting the control
+commands to updated values:
+\begin{itemize}
+  \item The vertical offset of the sub- or superscript: this is a dimension
+    expression as described for \cs{dimeval}. The standard setting
```
FrankMittelbach:
```suggestion
    expression (as described for \cs{dimeval}), stored in a command so that it is evaluated at the time of use. The standard setting
```

base/doc/usrguide.tex
```diff
@@ -1454,6 +1454,42 @@ \section{Case changing}
...
+    uses the font dimensions of the current math font, but for new documents,
+    using a value based on the text font is recommended. For example
+    |\renewcommand*\textsuperscript@offset{0.86ex}| will set
+    the vertical offset to $86\,\mathrm{\%}$ of the x-height of the current
+    font.
+  \item The space added after the sub- or superscript: the standard setting is
+    a non-breakable space of the same size as would be used in math mode
+    (\cs{scriptspace}). It may be preferable in new documents to allow a break
+    here, as the space is then discarded, for example
```
FrankMittelbach:
```suggestion
    here, as the space is then discarded, for example,
```

base/ltfloat.dtx
```diff
@@ -1350,32 +1350,69 @@
...
+    \endgroup
+  }
+\def\textsuperscript@offset{%
+  \ifdim\fontdimen14\textfont2 <\dimexpr\dp0 + 0.25\fontdimen5\textfont2 \relax
+    \dp0 + 0.25\fontdimen5\textfont2 %
+  \else
+    \fontdimen14\textfont2 %
+  \fi
+}
+\def\textsuperscript@space{\nobreak\hskip\scriptspace}
```
davidcarlisle: This was possibly my suggestion but I think it needs a trailing `\null` (or some non-discardable node) otherwise a space following `\textsuperscript{x}`  will be part of the non-discardable sequence started by the penalty and a linebreak won't be  possible.


```
\documentclass{article}

\textwidth70pt

\begin{document}

\hrule


1) xxx aaaa\raisebox{5pt}{x} bbb

2) xxx aaaa\raisebox{5pt}{x\kern5pt} bbb


3) xxx aaaa\raisebox{5pt}{x}\nobreak\hskip5pt\relax{} bbb


4) xxx aaaa\raisebox{5pt}{x}\nobreak\hskip5pt\null\relax{} bbb


\end{document}
```

the old version is more or less (2) (although the box is made wider without the kern) 

this PR is showing (3) but no line break happens

4 adds the non discardable box and you again allow a linebreak at the space before `bbb`

josephwright: Wasn't that what you asked for? Current behaviour means no break so the space is retained ...

josephwright: I'm a bit concerned we are adding quite a lot to the output here - if the aim is to avoid change for the default case, perhaps I should find an interface that sticks to adjusting the box width? That said, if we do want something additional here, my preferences is a zero-width kern :-)

FrankMittelbach: > Wasn't that what you asked for? Current behaviour means no break so the space is retained ...

without the \null a normal space following the \nobreak\hskip... would also be retained which is certainly not what we want.

davidcarlisle: it's what I asked for but it's wrong (shocking but true:-)  adding the `\null` is a bit more change in the test suite but if we are changing the command for the first time since 1995, it's probably worth a bit of test suite churn to get the required features in

davidcarlisle: "my preferences is a zero-width kern" a zero width kern is not enough, it needs to be a non-discardable item to avoid the `\nobreak` applying to a following space.

base/changes.txt
```diff
@@ -6,6 +6,9 @@ to completeness or accuracy and it contains some references to files that are
 not part of the distribution.
 ================================================================================
 
+2026-02-11 Joseph Wright  <Joseph.Wright@latex-project.org>
+	* ltfloat.dtx, usrguide.tex:: Avoid 'fake math' in \textsubscript and \textsuperscript
```
muzimuzhi:
```suggestion
	* ltfloat.dtx, usrguide.tex: Avoid 'fake math' in \textsubscript and \textsuperscript
```

base/ltfloat.dtx
```diff
@@ -1350,32 +1350,69 @@
...
+% \changes{v1.2h}{2026-02-11}{Command added}
+%  \begin{macro}{\textsuperscript@space}
+% \changes{v1.2h}{2026-02-11}{Command added}
 %    This command should not be used directly, but may be used to define
 %   other commands |\textsuperscript|, |\@makefnmark|. |#1| should
 %   always start with a font selection command, to activate the font
-%   size switch.
+%   size switch. Doing everything in text mode using the classical dimensions
+%   means using the settings from math mode: to allow the user to deviate
+%   from those, we use a one-elevel indirection. Notice that the script space
```
muzimuzhi:
```suggestion
%   from those, we use a one-level indirection. Notice that the script space
```

muzimuzhi (changes requested):
Two typos.

base/ltfloat.dtx
josephwright: @davidcarlisle I tried `5) xxx aaaa\raisebox{5pt}{x}\nobreak\hskip5pt\kern0pt\relax{} bbb` and the output shows
```
......\TU/lmr/m/n/10 x
....\penalty 10000
....\glue 5.0
....\kern0.0
....\glue(\rightskip) 0.0
...\penalty 300
```
so no space ...

josephwright: But the real question is what makes sense as an interface and to document here: I'm not sure it's that easy.

davidcarlisle: ah I suppose the kern there is a "non discardable item" as it's preceded by the nobreak penalty even though kerns are discardable items in general

FrankMittelbach: > But the real question is what makes sense as an interface and to document here: I'm not sure it's that easy.

I don't see that as problematical. This is an interface mainly for classes (including to keep or drop the space at a line break) and offering there a command that contains an \hspace does't seem problematical. Keeping the space is really only a backward comatibility thing as it makes much more sense to drop it in which case the definition of the command would normally simply contain some `\hspace{xx em}` to be font dependent. True, the vertical shift would then be just a dimension and the horizontal one a space command, but on the class level (or even in the preamble) I don't think that is an issue.

josephwright: @FrankMittelbach  Problematic in the sense if one was starting today, you'd only have a skip here and document you can provide a suitable expression to be inserted: needing a non-discardable space is what makes it to me) more awkward. But hopefully I've expressed it in a way that is understandable.

FrankMittelbach: okay, yes. Correct would probably be to provide a template for raising and lowering text with a bunch of parameters that you can set including dropping or keeping that kern etc. But I don't think we are really advanced enough to start doing meaningful micro templates. So one other way to handle that is could be to only changed from math to non-math usage and mark your four internal parameter macros as "internal subject to change", ie not document them in usrguide and put the word "kernel" into their name andwith a warning in the code that they might get replaced.

josephwright: Can certainly do that: I was only documenting as it then offered something beyond 'no math mode'. If we decide we don't want to document flexibility ATM, I'd favour simply avoiding any vars at all - 'kernel' or not, they'll get used otherwise!

base/doc/ltnews43.tex
```diff
@@ -278,6 +278,30 @@ \subsection{Revision of \cs{protected} status of functions in templates}
...
+  \item \cs{textsuperscript} and \cs{textsubscript}
+\end{itemize}
+
+To avoid as far as possible changes to existing output, the standard settings
+continue to use math font dimension values for placing super and
+subscripts.\footnote{The calculation used for placement of math mode sub- and
+superscripts is described in Appendix~G of \emph{The TeX{}book}, and relies on
+multiple math font dimensions.} For new documents, we suggest setting to a
+predictable offset based on text font dimensions (for example a fixed
+percentage of the x-height).
```
Skillmon: "Setting" what exactly? I think there's an object missing in this sentence.

base/ltfloat.dtx
```diff
@@ -1350,32 +1350,71 @@
...
+    \endgroup
+  }
+\def\textsuperscript@offset{%
+  \ifdim\fontdimen14\textfont2 <\dimexpr\dp0 + 0.25\fontdimen5\textfont2 \relax
+    \dp0 + 0.25\fontdimen5\textfont2 %
+  \else
+    \fontdimen14\textfont2 %
+  \fi
+}
+\def\textsuperscript@space{\nobreak\hskip\scriptspace\kern0pt }
```
Skillmon:
```suggestion
\def\textsuperscript@space{\nobreak\hskip\scriptspace\kern\z@}
```

(faster and fewer tokens)

base/ltfloat.dtx
```diff
@@ -1414,20 +1453,50 @@
...
+    \endgroup
+  }
+\def\textsubscript@offset{%
+  \ifdim\fontdimen16\textfont2 <\dimexpr\ht0 - 0.8\fontdimen5\textfont2 \relax
+    \dp0 + 0.25\fontdimen5\textfont2 %
+  \else
+    \fontdimen16\textfont2 %
+  \fi
+}
+\def\textsubscript@space{\nobreak\hskip\scriptspace\kern0pt }
```
Skillmon:
```suggestion
\def\textsubscript@space{\nobreak\hskip\scriptspace\kern\z@}
```

Skillmon (commented):
The 0 as in `\box0`/`\wd0` etc. might be changed to using `\z@` as well (fits better into the historical LaTeX code).

josephwright:
> The 0 as in `\box0`/`\wd0` etc. might be changed to using `\z@` as well (fits better into the historical LaTeX code).

I'll defer to @davidcarlisle and @FrankMittelbach here!

FrankMittelbach:
> > The 0 as in `\box0`/`\wd0` etc. might be changed to using `\z@` as well (fits better into the historical LaTeX code).
> 
> I'll defer to @davidcarlisle and @FrankMittelbach here!

I agree with @Skillmon here. In this contet `\z@` seems natural.

davidcarlisle:
> > > The 0 as in `\box0`/`\wd0` etc. might be changed to using `\z@` as well (fits better into the historical LaTeX code).
> > 
> > 
> > I'll defer to @davidcarlisle and @FrankMittelbach here!
> 
> I agree with @Skillmon here. In this contet `\z@` seems natural.

also ` \textfont\tw@` rather than `2 ` (if you are changing these anyway)

base/doc/ltnews43.tex
car222222: Agreed: it could be simply "this", but something more explicit and maybe detailed is better.

base/ltfloat.dtx
```diff
@@ -1350,32 +1350,71 @@
...
+    \check@mathfonts
+    \leavevmode
+    \begingroup
+      \sbox\z@{\fontsize\sf@size\sf@size#1}%
+      \raise\dimexpr\textsuperscript@offset\relax\box\z@
+      \textsuperscript@space
+    \endgroup
+  }
+\def\textsuperscript@offset{%
+  \ifdim\fontdimen14\textfont\tw@<\dimexpr\dp\z@ + 0.25\fontdimen5\textfont\tw@\relax
```
davidcarlisle: using numbered font dimens by default might be problematic in luatex which doesn't set them up if using opentype math. We could use the equivalent named parameters or perhaps in luatex it may be simpler just to drop the math dependency and use `.8ex` or whatever by default.

unicode-math puts the opentype font in fam2 and uses the luaotfload mathfontdimen feature to emulate numeric fontdimens, but lua-unicode-math doesn't currently (I think)

khaledhosny: FWIW, text fonts have subscript and superscript offsets and sizes in the [OS/2 table](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#ysubscriptysize). Some applications use them, but most mainstream applications don’t use them so some fonts can have non-sensible values.

zauguin: > unicode-math puts the opentype font in fam2 and uses the luaotfload mathfontdimen feature to emulate numeric fontdimens, but lua-unicode-math doesn't currently (I think)

`lua-unicode-math` very intentionally doesn't set these since it requires loading fonts (which makes it slower) and the advantages are rather minimal in most documents.

> using numbered font dimens by default might be problematic in luatex which doesn't set them up if using opentype math. We could use the equivalent named parameters or perhaps in luatex it may be simpler just to drop the math dependency and use .8ex or whatever by default.

By default I think we should use the named parameters to use the same offsets as in math (otherwise it might look very unexpected in documents using both) but it would be nice to have a simple interface to switch to the font dimensions mentioned by @khaledhosny.

davidcarlisle: @zauguin  what I was wondering is whether the format definition of `\textsuperscript` has a good way of knowing if the math layout is using OpenType or tfm fonts, especially as using math positioning is something of a historic legacy thing anyway using the text font metrics (either the specific superscript offsets or simply a fraction of x height might be reasonable and a lot simpler and easier to document)

josephwright: The only reason for using the math values here is document stability - if we feel that the linebreaking impact is low, we could deliberately make a change here. But the Appendix G rule is awkward and I suspect will show up if we change things.

zauguin: I don't see a need to know what kind of math layout is used. When accessing `\Umathsupshiftup\textstyle` and friends LuaTeX will provide the right dimensions autmatically, whether they end up coming from the legacy fontdimen or the OpenType named math parameter.

davidcarlisle: ah, so it will, Ok then so named params in luatex and numbers elsewhere that seems quite reasonable

josephwright: Presumably there are two named parameters?

davidcarlisle: well fontdimen5 is x-height   but there are pre-calculated named parameters such as `\Umathsupbottommin`  which is `abs(math_x_height/4)` which might also prove useful

FrankMittelbach: wouldn't it be better then to abstract from  the engine specific solutions as we do in other cases and provide a comon interface to the parameters?

josephwright: Feels like something that lives nicely in `expl3`: `\font_superscript_offset:`, etc. :-) More seriously, I guess more-or-less the LuaTeX primitive names with some kernel-specific prefix?

josephwright: > FWIW, text fonts have subscript and superscript offsets and sizes in the [OS/2 table](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#ysubscriptysize). Some applications use them, but most mainstream applications don’t use them so some fonts can have non-sensible values.

Definitely use of text mode values is down to the user

PR #2010, opened 2026-02-25
\pdfglyphtounicode

u-fischer:
This defines \pdfglyphtounicode and \pdfgentounicode for all engines (as noop when they do not support the concept) and loads and activates glyphtounicode and glyphtounicode-cmex for pdftex and luatex (in pdf mode).

Open question
* where should glyphtounicode-cmex live? (currently it is in latex-lab as there we used it for legacy math)

## Status of pull request

- Ready to merge

- [x] Test file(s) (only for pdftex and luatex in latex lab as PDF based test)

### DIFF
--- a/base/doc/ltnews43.tex
+++ b/base/doc/ltnews43.tex
@@ -337,7 +337,7 @@ from the file \file{glyphtounicode.tex}
 in order to improve copying from, and searching in,
 text for, e.g., common ligatures.
 
-We now extended this support and load additionally a file 
+We now extend this support and load additionally a file 
 \file{glyphtounicode-cmex.tex}
 which improves copying of mathematical symbols.
 


base/doc/ltnews43.tex
```diff
@@ -318,12 +318,29 @@ \section{New or improved documentation}
...
 formatting variation. That will all be addressed over time, as we aim for a
 single coherent document.
 
+\section{Glyphs, characters \& encodings}
+
+\subsection{Improved copy\,\&\,paste for \pdfTeX{} and \LuaTeX{} documents}
+
+In 2021 when compiling with \pdfTeX{} additional information
+(from the file \file{glyphtounicode.tex}) was added automatically
+to the PDF file in order to improve copying from, and searching in,
```
josephwright:
```suggestion
In 2021, additional information was added to the PDF file,
when compiling with pdfTeX{}, 
from the file \file{glyphtounicode.tex}
in order to improve copying from, and searching in,
```

josephwright: One might put the pdfTeX bit in bracket or omit, or even put before the date

base/testfiles/tlb-latexrelease-rollback-003-often.luatex.tlg
```diff
@@ -1864,12 +1865,41 @@ Already applied: [....-..-..] XeTeX character classes on input line ....
 Applying: [....-..-..] Save language for hyphenation on input line ....
 Already applied: [....-..-..] Save language for hyphenation on input line ....
 Skipping: [....-..-..] Preload glyphtounicode on input line ....
+Skipping: [....-..-..] Preload glyphtounicode on input line ....
 Applying: [....-..-..] Preload glyphtounicode on input line ....
+! Undefined control sequence.
+l. ... \pdfgentounicode
+                         =0
+The control sequence at the end of the top line
```
FrankMittelbach: that doesn't seem right. I think you should define the commands as noops and pretend that the noops have always been there

base/ltfinal.dtx
```diff
@@ -683,28 +690,89 @@
...
+%<latexrelease>\IncludeInRelease{2021/06/01}%
+%<latexrelease>                 {\pdfgentounicode}{Preload glyphtounicode}%
+%<latexrelease>\ifx \directlua\undefined \else
+%<latexrelease>  \let\pdfgentounicode \@undefined
+%<latexrelease>  \let\pdfglyphtounicode \@undefined
+%<latexrelease>  \pdfvariable gentounicode=0
+%<latexrelease>\fi
+%<latexrelease>\ifx \pdfglyphtounicode\@gobbletwo 
+%<latexrelease> \let\pdfglyphtounicode\@undefined
+%<latexrelease> \let\pdfgentounicode\@undefined
```
FrankMittelbach: I wouldn't go back to undefined because that essentially means any package using these commands will break if you roll back before 2021

base/doc/ltnews43.tex
mrpiggi: was `\pdfTeX{}` > `pdfTeX{}` changed deliberately?

FrankMittelbach (changes requested):
basically fine, but I would handle rollback differently.

base/doc/ltnews43.tex
```diff
@@ -318,12 +318,37 @@ \section{New or improved documentation}
...
+in order to improve copying from, and searching in,
+text for, e.g., common ligatures.
+
+We now extended this support and load additionally a file 
+\file{glyphtounicode-cmex.tex}
+which improves copying of mathematical symbols.
+
+As starting with version 1.24 \LuaTeX{} will support an extended syntax 
+of \cs{pdfglyphtounicode} that makes it possible to improve copy\,\&\,paste 
+of symbol fonts we load and activate the files also with this engine if PDF mode
```
mrpiggi: maybe
> [...]of symbol fonts. Therefore, we load and activate the files [...]

base/ltfinal.dtx
u-fischer: well they were undefined (in engines like xetex) before 2021 and shouldn't this be the case too if someone rolls back?

FrankMittelbach: yes and no. For extension that we expect to be provided only by us, we do this but for core functionality we always added them pretenting that noop versions have been around earlier if that made any sense (or was likely to work. Otherwise you basically stop rolling back from working. And this is a typical case of the latter. A symbol package would have work 10 years ago but the moment you add tounicode commands to it it would stop working in rollback

base/testfiles/tlb-latexrelease-rollback-003-often.luatex.tlg
u-fischer: well I have no idea if the result is right or not, the testfile is doing so curious things that everything could be expected ;-). But if I do not have to undefine the commands it looks better so ....

base/doc/ltnews43.tex
```diff
@@ -318,12 +318,38 @@ \section{New or improved documentation}
...
+
+\subsection{Improved copy\,\&\,paste for \pdfTeX{} and \LuaTeX{} documents}
+
+In 2021, additional information was added to the PDF file,
+when compiling with \pdfTeX{}, 
+from the file \file{glyphtounicode.tex}
+in order to improve copying from, and searching in,
+text for, e.g., common ligatures.
+
+We now extended this support and load additionally a file
```
muzimuzhi: Use the same tense? Either "extended ... loaded" or "extend ... load".

base/doc/ltnews43.tex
```diff
@@ -318,12 +318,38 @@ \section{New or improved documentation}
...
+in order to improve copying from, and searching in,
+text for, e.g., common ligatures.
+
+We now extended this support and load additionally a file 
+\file{glyphtounicode-cmex.tex}
+which improves copying of mathematical symbols.
+
+As starting with version 1.24 \LuaTeX{} will support an extended syntax 
+of \cs{pdfglyphtounicode} that makes it possible to improve copy\,\&\,paste 
+of symbol fonts. Therefore, we load we load and activate the files also with
```
muzimuzhi: Dup words.

```suggestion
of symbol fonts. Therefore, we load and activate the files also with 
```

base/doc/ltnews43.tex
u-fischer: not sure. My feeling is that  we extended *once* and so load *always*.

FrankMittelbach: I would go for extend + load.  Its subtle I guess :-)

PR #2016, opened 2026-03-10
Some docu corrections

u-fischer:
## Status of pull request
- Ready to merge

dates changes.

### DIFF
--- a/base/ltcounts.dtx
+++ b/base/ltcounts.dtx
@@ -33,7 +33,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{ltcounts.dtx}
-             [2026/030/10 v1.2c LaTeX Kernel (Counters)]
+             [2026/03/10 v1.2c LaTeX Kernel (Counters)]
 % \iffalse
 \documentclass{ltxdoc}
 \GetFileInfo{ltcounts.dtx}


base/ltcounts.dtx
```diff
@@ -166,7 +166,7 @@
 %  |\cl@foo| & List of counters to be reset when foo stepped.
 %              Has format
 %                  |\@elt{countera}\@elt{counterb}\@elt{counterc}|.\\
-%  |\alias@cnt@foo| & if foo is an alias counter created with |\newcounteralias| this
+%  |\alias@ctr@foo| & if foo is an alias counter created with |\newcounteralias| this
```
FrankMittelbach, line 169: really "ctr"? I thought we use "cnt" more often or generally

base/ltcounts.dtx
```diff
@@ -33,7 +33,7 @@
 %<*driver>
 % \fi
 \ProvidesFile{ltcounts.dtx}
-             [2026/02/08 v1.2c LaTeX Kernel (Counters)]
+             [2026/030/10 v1.2c LaTeX Kernel (Counters)]
```
josephwright: Stray digit

base/ltsockets.dtx
```diff
@@ -520,7 +520,12 @@
 % 
 % \end{itemize}
 % 
+% \subsection{Socket and plug names}
 % 
+% The \meta{socket-name}  and \meta{socket-plug-name} are always
+% expanded using \TeX's \cs{csname}\ldots\cs{endcsname}, as such,
+% (expandable) commands and Unicode characters are allowed in
```
FrankMittelbach, line 527: no sure every unicode char would be possible, would it? but yes  on the whole that's kind of true

base/ltcounts.dtx
u-fischer: Well we have also ctr e.g. `\@newctr`  and `|\newcounter|\marg{newctr}\oarg{oldctr}`, and the code does use the name and nobody complained about it ;-). I only made a typo in the docu.

base/ltsockets.dtx
u-fischer: The text is more or less copied from a similar section in lthooks.dtx ;-)

PR #2027, opened 2026-03-19
Add \DebugTemplates(On|Off) to kernel

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added


u-fischer:
In some places `\[Template] ==> primary context is 'float'` is now missing. 

The question is if the context debug messages should be activated with `\DebugTemplatesOn` or if they should get their own switch?

josephwright:
@u-fischer Corrected

required/latex-lab/testfiles-block/blocks-alltt.luatex.tlg
```diff
@@ -1,20 +1,20 @@
 This is a generated file for the l3build validation system.
 Don't change this file in any respect.
-[Template] ==> Use 'blockenv' instance: alltt on input line ...
+[Template] ==> Using instance 'alltt' of type 'blockenv' on line ...
 [Template] ==>   template: 'std; arguments: |\NoValue |\BooleanFalse |\NoValue |\NoValue |
-[Template] ==> Use 'block' instance: verbatim-1 on input line ...
+[Template] ==> Using instance 'verbatim-1' of type 'block' on line ...
 [Template] ==>   template: 'std'; argument: ||
```
FrankMittelbach, line 6: In the block code I also show the template and its arguments which is really helpful. Now that may has to happen in the individual templates (or could it be picked up somehow?). If it has to happen in the individual templates then  there needs to be a debug command which is not internal to the module but could be used in other modules, but being ables to generate that data in the debugging situation directly would be better.

josephwright: I'm not following here: I don't think I've lost any information in the `.tlg` files, only slightly reformatted.

FrankMittelbach: no you haven't. All I'm musing about is the following: At the moment I have lines like
```
  \template_debug_typeout:n{~\space template:~ 'std';~
                            arguments:~ \exp_not:o{\exp_after:wN |#1|#2|#3|#4|}}
```
at the start of some of the template code (which then comes out directly after the instance is called and generates the debugging code) but it would be much nicer if that could be done automatically by the debugging code. Not sure it could though.

FrankMittelbach:
> In some places `\[Template] ==> primary context is 'float'` is now missing.
> 
> The question is if the context debug messages should be activated with `\DebugTemplatesOn` or if they should get their own switch?

I think they should appear when \TemplateDebugOn is set. Don't see that it makes sense to have a separate switch for them.

required/latex-lab/testfiles-block/blocks-alltt.luatex.tlg
josephwright: So anything to do now?

FrankMittelbach: well, if you can think of a way to make this automatically work, it would be great. If not then one has to do it inside for each and every template, which is not really a good approach. But if you don't have a good idea how that could work (I haven't) the we should merge it (and perhaps make a note that somehting like this should be considered as an extension

josephwright: Do you mean saving the template info to the code when we create an instance?

FrankMittelbach: That might be a good option, actually. I was originally more thinking that if template Debugging is on then \UseInstance could have a special definition that picks up the instance arguments and does it on that level. But perhaps that difficult to do in which case your suggestion would be a simple way out.

josephwright: Merge this PR then I circle back for that?

FrankMittelbach: fine with me

PR #2031, opened 2026-03-29
fix for https://github.com/latex3/tagging-project/issues/1279

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [x] Test file(s) added

latex3/tagging-project#1279


u-fischer:
Looks ok, but tagging wise one should perhaps note that two consecutive block environments can not be inside the same semantic paragraph, the second block always start a new one (that was already the case before, so not due to this code).

u-fischer (approved):

FrankMittelbach:
> Looks ok, but tagging wise one should perhaps note that two consecutive block environments can not be inside the same semantic paragraph, the second block always start a new one (that was already the case before, so not due to this code).

hmm, is that the case? I don't really think it should even if it was always the case before. Are you saying the `<BlockQuote>` is not allowed to immediately follow another block quote? 
I would have expected to see
```
<text-unit>
   <text> ...</text>
  <BlockQuote> ... </BlockQuote>
%   <text> ...</text>                                  % <-- with this simply missing
  <BlockQuote> ... </BlockQuote>
   <text> ...</text>
</text-unit>
```

u-fischer:
> Are you saying the <BlockQuote> is not allowed to immediately follow another block quote?

it is allowed but our code doesn't do it. 

~~~~
\DocumentMetadata{uncompress,tagging=on}
\documentclass{article}

\begin{document}
some text
\begin{center}
abc
\end{center}
\begin{center}
efg
\end{center}
some text
\end{document}
~~~~

produces this

image

FrankMittelbach:
> Are you saying the
> 
> > is not allowed to immediately follow another block quote?
> 
> it is allowed but our code doesn't do it.

true. This is due to the tagging code ending the para structure unconditionally in `\begin` if `@endpe=true` reusing the `kernel/endpe/vmode` instead of delaying to the receipe code. That actually renders part of that receipe code useless as it expects to be able to test for `@endpe`. Guess that needs to be looked at and the logic slightly changed.

FrankMittelbach:
This code now also handles `@endpe` in a way the consecutive envs will not wrap themselves up into individual semantic paragraphs but will continue the currently open paragraph. In particular `\begin` will not unconditionally end the semantic paragraph if it is seen immediately after an env requesting `@endpe` handling.

u-fischer (approved):

PR #2033, opened 2026-04-05
Update ltxdoc.dtx fix typos (missing space, uppercase instead of lowe…

quark67:
…rcase)

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- Ready to merge

- [ ] Test file(s) added

No, in fact, I have not added test files etc., it's only a typo fix proposal.

### DIFF
--- a/base/ltxdoc.dtx
+++ b/base/ltxdoc.dtx
@@ -30,7 +30,7 @@
 %
 %<class>\NeedsTeXFormat{LaTeX2e}
 %<class>\ProvidesClass{ltxdoc}
-%<class>         [2026/01/16 v2.1l Standard LaTeX documentation class]
+%<class>         [2026/04/06 v2.1l Standard LaTeX documentation class]
 %
 %<*driver>
 \documentclass{ltxdoc}
@@ -71,7 +71,7 @@
 %\begin{verbatim}
 % latex2e docclass.dtx
 %\end{verbatim}
-% would produce the documentation of the Class and package interface.
+% would produce the documentation of the class and package interface.
 %
 % Each file that is used in producing the \LaTeXe\ format (ie not
 % including the standard class and packages) will be printed together in
@@ -127,7 +127,7 @@
 %
 % As an example, consider |ltclass.dtx|, which contains the sources for
 % the new class and package interface commands.  With no |cfg|
-% file, a 19~page document is produced. With the above configuration
+% file, a 19~pages document is produced. With the above configuration
 % a slightly more readable document (4~pages) is produced.
 %
 % Conversely, if you really want to read the source listings in detail,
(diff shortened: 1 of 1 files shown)


muzimuzhi:
By convention please also bump the file date. File version does not need to be changed for doc-only changes like this.

https://github.com/latex3/latex2e/blob/c5625715dc1b0bf17f280da84f47948c1acee749/base/ltxdoc.dtx#L33

quark67:
Done, thanks, I was unaware of that. I also added more typos fixes.

FrankMittelbach:
> Done, thanks, I was unaware of that. I also added more typos fixes.

changes in files need to be identifiable which is the case if the announcement strng changes. By convention we only change the date when there are only documentation changes.

This is also why you shouldn't checkmark the items in the pr list if a you actually  have't done any of them (as it was the case in your initial pull request. Not all of them are necessary depending on the change, so sometimes we use n/a, but do not check them if the checkmark is actually not correct.

quark67:
I don't understand what I have to do more. Probably it would be easier if the next time I open an issue and say what line to change in the file? I'm not an English speaker. 
What is the right way for contributing about typo fixes?
I have checked all boxes because I thought that if no box is checked, the pull request is automatically blocked.
If I need to do more action, please explain what.

FrankMittelbach:
you don't have to do more. For just typos it is is enough to  alter the date as you did now (and I marked the checkbox). The others are there for more complex changes that require \changes entries or a new test file or an ltnews entry. We tell people if we need more. I was just saying that marking them as done if they aren't done is not the right approach.

quark67:
Ah OK, I have need to do [x] Version and date string updated in changed source files (and not only change the date in the file). 
Sorry I haven't understand this. Thanks for your explanations. 
By the way, which date needs to be added? For example, I'm in Europe, and in 1h30, the date becomes the 7th April. But in USA, the date is still the 6th April. So (if I do something similar with an other file — without the previous errors), the date needs to be the local date of the author of the pull request or the date in the USA? (and if so, East time or Pacific time?).

FrankMittelbach:
> Ah OK, I have need to do [x] Version and date string updated in changed source files (and not only change the date in the file). Sorry I haven't understand this. Thanks for your explanations. 

More or less all PR are internal (because as we say  at the topit is not easy for others to do everything that is necessary and we therefore suggest that people contact us first in case of pulls that involve coding). Something like typos is different and there from the list normally all that is necessary is to change the identification string, i.e., we don't need \changes entries or changes.txt updates or rollback etc. And if we do for some reason, we would ask for it. 

> By the way, which date needs to be added? For example, I'm in Europe, and in 1h30, the date becomes the 7th April. But in USA, the date is still the 6th April. So (if I do something similar with an other file — without the previous errors), the date needs to be the local date of the author of the pull request or the date in the USA? (and if so, East time or Pacific time?).

No perfect answer to that one, but as the idenfication string is there just to make different version distinguishable, we use the date that matches the situation in Europe as most of us are from here too. so UTC really.

quark67:
Thanks to have taken the time for a complete answer. Have a nice day!

FrankMittelbach:
and thanks for supplying corrections.

Issue #793, opened 2022-03-15
`amsmath`: `multline` fails to compensate baseline open-up


RuixiZhang42:
## Brief outline of the bug

Within multi-line display math such as `align` and `gather`, the distance between baselines are opened up by `\jot` (e.g., 12pt baseline distance in text, 15pt baseline distance in display math). As pointed out in *The TeXbook*, care should be taken for the first line in the display: The baseline is now 15pt below that of the last line from the proceeding paragraph, so we should &ldquo;back up&rdquo; by the open-up amount of 3pt to keep the distance at 12pt. The backing up is needed because the distance is also 12pt between the baseline of the display&rsquo;s last line and that of the following paragraph&rsquo;s first line.

It appears that `align` and `gather` take care of the backing up, but `multline` fails to do so.

This report originates from the question: https://tex.stackexchange.com/q/636672

## Minimal example showing the bug

```latex
\RequirePackage{latexbug}
\documentclass{article}
\usepackage{amsmath}

\begin{document}

\setlength\abovedisplayskip{0pt}
\setlength\belowdisplayskip{0pt}

\noindent
Lorem ipsum dolor sit amet, consectetur adipiscing
elit. \texttt{equation*}:
\begin{equation*}
  \begin{aligned}
  \smash{\Big|} a_{n+1} &= [2(n+1)][(n+1)^2-1]^{-1}
   = n^{-1} + (n+2)^{-1}
   \\
   &< (n-1)^{-1} + (n+1)^{-1}
   = (2n)(n^2 - 1)^{-1} = a_n.
  \end{aligned}
\end{equation*}
Lorem ipsum dolor sit amet, consectetur adipiscing
elit. \texttt{align*}:
\begin{align*}
 \smash{\Big|} a_{n+1} = [2(n+1)][(n+1)^2-1]^{-1}
 = n^{-1} + (n+2)^{-1}
 \\
 < (n-1)^{-1} + (n+1)^{-1}
 = (2n)(n^2 - 1)^{-1} = a_n.
\end{align*}
Lorem ipsum dolor sit amet, consectetur adipiscing
elit. \texttt{multline*}:
\begin{multline*}
 \smash{\Big|} a_{n+1} = [2(n+1)][(n+1)^2-1]^{-1}
 = n^{-1} + (n+2)^{-1}
 \\
 < (n-1)^{-1} + (n+1)^{-1}
 = (2n)(n^2 - 1)^{-1} = a_n.
\end{multline*}
[...]

PR #2061, opened 2026-04-28
Safely override `\everycr` inside `multline` measuring phase

peaR-red:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [x] Test file(s) added


peaR-red:
Requesting reviews...

Also, it seems to me that this is quite a major update as I had to update various existing test files. So perhaps this needs an `ltnews` entry? Maybe even a rollback (though I'm not sure why would one roll back to the previous `multline` that lacked an `\everycr` assignment)?

FrankMittelbach:
Please always link in the issue that a pull is attempting to fix, i.e., https://github.com/latex3/latex2e/issues/793 here. It is referred to in the code (and in the name of your branch) but a direct link is helpful.

peaR-red:
> Please always link in the issue that a pull is attempting to fix, i.e., #793 here. It is referred to in the code (and in the name of your branch) but a direct link is helpful.

OK, will keep that in mind.

Going back to my previous comment, I suppose this change warrants an `ltnews` entry (given the extensive modifications to existing test files to accommodate the changed behavior). Thoughts?

FrankMittelbach:
> Going back to my previous comment, I suppose this change warrants an `ltnews` entry (given the extensive modifications to existing test files to accommodate the changed behavior). Thoughts?

definitely, as a minimum. It changes the vertical spacing and hence thepagination if you are unlucky in existing documents and that may mean we have to decide if there is the need for a "undo" of some sort, though perhaps not. @davidcarlisle ?

peaR-red:
> definitely, as a minimum.

OK. Are there any refinements I should make to the current draft?
```latex
\subsection{Correct vertical spacing above \env{multline}}

In multi-line display math environments such as \env{align} and \env{gather},
the distance between lines is slightly increased (by an amount controlled by
\cs{jot}) to ensure that tall mathematical symbols do not clash.

It was then discovered that the \env{multline} environment failed to apply
this compensation. This resulted in \env{multline} displays appearing slightly
further away from the preceding paragraph than other multi-line equations.
The cause was identified as a conflict during the environment's internal
measurement phase, which has now been corrected.
%
\githubissue{793}
```

> It changes the vertical spacing and hence thepagination if you are unlucky in existing documents and that may mean we have to decide if there is the need for a "undo" of some sort, though perhaps not. @davidcarlisle ?

Hmm, I'm probably not authoritative enough to decide on whether there should be a rollback provided. I'd be happy to hear the team's thoughts! :-)

FrankMittelbach:
> ```
> The cause was identified as a conflict during the environment's internal
> measurement phase, which has now been corrected.
> %
> \githubissue{793}
> ```

Cause isn't really important for ltnews I think (should be covered in comments in the code as appropriate, though). So I would say something like
```
This has been corrected.

As a result more material might fit on a page potentially changing the pagination. If that happens, an explict
 \cs{pagebreak} will restore the original pagination in old documents.
%
\githubissue{793}
```

> It changes the vertical spacing and hence thepagination if you are unlucky in existing documents and that may mean we have to decide if there is the need for a "undo" of some sort, though perhaps not. @davidcarlisle ?
> 
> Hmm, I'm probably not authoritative enough to decide on whether there should be a rollback provided. I'd be happy to hear the team's thoughts! :-)

Given that it shortens the page I think we are probably good without it. If it would lengthen the page that would be different as that is less easy to correct.

ps I still think it is fine/better if you simply make your suggestion for text as part of the pull. That way any comments can be written as part of a review in a way that they can be directly applied instead of this somewhat roundabout way.

peaR-red:
> Cause isn't really important for ltnews I think (should be covered in comments in the code as appropriate, though).

I think it'd be best if this had a news entry, because like I said I had to modify quite a lot of test files that use the old `multline` behavior. And I changed the wording a bit as per your suggestions.

FrankMittelbach:
> > Cause isn't really important for ltnews I think (should be covered in comments in the code as appropriate, though).
> 
> I think it'd be best if this had a news entry, because like I said I had to modify quite a lot of test files that use the old `multline` behavior. And I changed the wording a bit as per your suggestions.

yes, of course, it should have a news entry and all your text there is fine. I would only change the end because what the technical reason for the failure is is not important in ltnews, but the fact that it might change the pagination of existing documents is.

FrankMittelbach:
> Seems like the workflow has some problems... It's been stuck for a while now.

occasionally starting up ubuntu has problems. I canceled and restarted the hanging process

peaR-red:
@FrankMittelbach Requesting updates on the review process... Are there any improvements I could make to the existing commits (e.g., docs, test files, and whatnot)?

FrankMittelbach:
> @FrankMittelbach Requesting updates on the review process... Are there any improvements I could make to the existing commits (e.g., docs, test files, and whatnot)?

busy marrying off one of our sons, no time for LaTeX work before next week really, but was anyway waiting for @davidcarlisle  to take a look.

required/amsmath/changes.txt
```diff
@@ -13,6 +13,10 @@
 All changes above are only part of the development branch for the next release.
 ================================================================================
 
+2026-04-28 peaR <pearatlatex@gmail.com>
+	* amsmath.dtx (subsection{The \env{multline} environment}):
+	 Safely sequester the \everycr logic inside \mmeasure@ (gh/793)
```
FrankMittelbach, line 18: "sequester" seems to be rather an odd word here. You really mean that?

FrankMittelbach (approved):
I guess it is fine (though I would change thee  change entry).
@davidcarlisle could you also have a look?

davidcarlisle:
> > Going back to my previous comment, I suppose this change warrants an `ltnews` entry (given the extensive modifications to existing test files to accommodate the changed behavior). Thoughts?
> 
> definitely, as a minimum. It changes the vertical spacing and hence thepagination if you are unlucky in existing documents and that may mean we have to decide if there is the need for a "undo" of some sort, though perhaps not. @davidcarlisle ?



> > Going back to my previous comment, I suppose this change warrants an `ltnews` entry (given the extensive modifications to existing test files to accommodate the changed behavior). Thoughts?
> 
> definitely, as a minimum. It changes the vertical spacing and hence the pagination if you are unlucky in existing documents and that may mean we have to decide if there is the need for a "undo" of some sort, though perhaps not. @davidcarlisle ?

@FrankMittelbach  I think we can (hopefully) avoid a new rollback checkpoint there is already a 2018 rollback so legacy documents would be covered if necessary. If testing in the wild shows the need, adding a 2026 dated rollback later would be possible without any major change, just adding the declaration and distributing the dated snapshot.

davidcarlisle (approved):

required/amsmath/changes.txt
davidcarlisle: I think "embed" would be better, we can change it on develop branch directly, I think @FrankMittelbach

FrankMittelbach: please do that then.

PR #2120, opened 2026-06-15
Fix language default

u-fischer:
This corrects the language default.

## Status of pull request

- Ready to merge

- [ ] Test file(s) added

### DIFF
--- a/required/latex-lab/documentmetadata-support.dtx
+++ b/required/latex-lab/documentmetadata-support.dtx
@@ -18,8 +18,8 @@
 % for those people who are interested or want to report an issue.
 %
 %    \begin{macrocode}
-\def\documentmetadatasupportversion{1.1c}
-\def\documentmetadatasupportdate{2026-05-27}
+\def\documentmetadatasupportversion{1.1d}
+\def\documentmetadatasupportdate{2026-06-15}
 %    \end{macrocode}
 %
 %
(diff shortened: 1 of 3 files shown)


required/latex-lab/documentmetadata-support.dtx
```diff
@@ -488,14 +488,14 @@
         \keys_set_exclude_groups:nnn  { document / metadata } { init,pdf } { #1 }
 %    \end{macrocode}
 % Finally we set up the language default.
-% If the \texttt{language} key hasn't be set yet, we set it directly
-% either to the value of the \texttt{lang} key or to en-US.
+% If the \texttt{language} key hasn't be set yet, we set it
```
mbertucci47:
```suggestion
% If the \texttt{language} key hasn't been set yet, we set it 
```

josephwright (approved):

required/latex-lab/documentmetadata-support.dtx
```diff
@@ -488,14 +488,14 @@
         \keys_set_exclude_groups:nnn  { document / metadata } { init,pdf } { #1 }
 %    \end{macrocode}
 % Finally we set up the language default.
-% If the \texttt{language} key hasn't be set yet, we set it directly
-% either to the value of the \texttt{lang} key or to en-US.
+% If the \texttt{language} key hasn't be set yet, we set it 
+% either to the value of the \texttt{lang} key or leave it empty.
```
FrankMittelbach: What the code does (as far as I can see) is to set language:=lang always (if lang is empty then language will also be empty). All the extra stuff the code does is to avoid to unnecessarily set the document/language metadata from empty to empty. So perhaps it is clearer to say

... we set it to the value of \texttt{lang} (if that is also empty then nothing needs doing). This way a language package like \pkg{babel} can correctly react to the situation that neither language nor lang has been explicitly set.

FrankMittelbach (changes requested):
approved with a suggestion.

FrankMittelbach (approved):

PR #2133, opened 2026-06-25
Remove fake math from \underline

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [ ] Test file(s) added


base/doc/ltnews44.tex
```diff
@@ -179,7 +179,13 @@ \subsection{Revisiting the para-handling logic for block environment}
 %
 \taggingissue[s]{1402 and 1415 among others}
 
-%\section{Code improvements}
+\section{Code improvements}
+
+\subsection{Further removal of \enquote{fake math} from the kernel}
+
+In the previous release, we removed most use of \enquote{fake math} from the
```
FrankMittelbach:
```suggestion
In the previous release, we removed most use of \enquote{fake math} (math mode not denoting a formula but used for positioning or manipulating text material) from the
```

base/doc/ltnews44.tex
```diff
@@ -220,7 +226,7 @@ \subsection{US Spelling}
 \newblock Reprinted with corrections in 1996.
 
 \bibitem{44:ltnews} \LaTeX{} Project Team.
-  \emph{\LaTeXe{} news 1--44}. June, 2026.
+  \emph{\LaTeXe{} news 1--43}. June, 2026.
```
FrankMittelbach:
```suggestion
  \emph{\LaTeXe{} news 1--44}. November, 2026.
```

By the time this goes out it is 1-44

base/testfiles-doc/tlb-dox004.luatex.tlg
```diff
@@ -1429,7 +1429,7 @@ Completed box being shipped out [4]
 .......\glue(\parskip) 0.0 plus 1.0
 .......\glue(\parskip) 0.0
 .......\glue(\baselineskip) 0.0
-.......\hbox(7.69997+3.30003)x155.0, glue set 101.6839fill, direction TLT
+.......\hbox(7.69997+3.30003)x155.0, glue set 97.05893fill, direction TLT
```
FrankMittelbach: a bit surprising that the glue setting changes

davidcarlisle (approved):

davidcarlisle:
approved once the doc tests pass:-)

FrankMittelbach (changes requested):
one comment was outdated while I was writing it

base/doc/ltnews44.tex
josephwright: Don't follow - this should be looking back, surely?

base/testfiles-doc/tlb-dox004.luatex.tlg
josephwright: We might need to tr to pull this apart: we also gain some other stuff, I wonder if it's some form of unexpect hook? (It's only showing with LuaTeX, which is odd.)

u-fischer: There seem to be links created twice. This should be investigated.

u-fischer: ah, no, I was looking at an outdated version.

base/doc/ltnews44.tex
FrankMittelbach: look at it from the official release of ltnews44 in November. Then ltnews.pdf has exactly that title. In fact ltnews.pdf in -dev already has this title.

We always handled it this way.

josephwright:
The parameters here need thought: `thickness` seems right, the other two are more troublesome.

josephwright:
I'm looking back here at the parameter names. I wonder if we should base them on the OpenType ones, in order below the text:

- UnderbarVerticalGap (`\Umathunderbarvgap`)
- UnderbarRuleThickness (`\Umathunderbarrule`)
- UnderbarExtraDescender (`\Umathunderbarkern`)

required/tools/testfiles/sx231401.tlg
```diff
@@ -92,20 +92,21 @@ Completed box being shipped out [1]
 .......\hbox(0.0+0.0)x0.0
 .......\glue 5.0
 ......\penalty 0
-......\mathon
 ......\vbox(6.83331+1.9999)x72.30565
-.......\hbox(6.83331+0.0)x72.30565
+.......\hbox(6.83331+0.0)x72.30565, glue set 40.00006fil
+........\glue(\leftskip) 0.0 plus 1.0fil
```
FrankMittelbach: this seems to shift text from left to right by 40 points. why?

required/tools/testfiles/sx231401.tlg
```diff
@@ -92,20 +92,21 @@ Completed box being shipped out [1]
...
 ........\OT1/cmr/m/n/10 A
 ........\OT1/cmr/m/n/10 n
 ........\OT1/cmr/m/n/10 s
 ........\OT1/cmr/m/n/10 w
 ........\kern-0.27779
 ........\OT1/cmr/m/n/10 e
 ........\OT1/cmr/m/n/10 r
-........\glue 40.00006
-.......\kern1.19994
```
FrankMittelbach, line 106: here the 40 pt are after the text, now they are as a fill before?

required/tools/testfiles/sx231401.tlg
```diff
@@ -156,20 +157,21 @@ Completed box being shipped out [1]
 .......\hbox(0.0+0.0)x0.0
 .......\glue 5.0
 ......\penalty 0
-......\mathon
 ......\vbox(6.83331+1.9999)x72.30565
-.......\hbox(6.83331+0.0)x72.30565
+.......\hbox(6.83331+0.0)x72.30565, glue set 40.00006fil
+........\glue(\leftskip) 0.0 plus 1.0fil
```
FrankMittelbach: that seems to happen several times in this tlg

required/tools/testfiles/tlb-multicol-eisuke.luatex.tlg
```diff
@@ -24,11 +24,17 @@ LaTeX Font Info:    External font `cmex10' loaded for size
...
 % t=96.5 plus 14.0 plus 2.0fill minus 14.0 g=1601.80013 b=0 p=-51 c=-51#
 Overfull \hbox (0.97234pt too wide) in paragraph at lines ...
- []$
+ []
 \hbox(6.83331+1.9999)x61.3333, direction TLT
 .\glue(\leftskip) 0.0 plus 1.0fil
 .\vbox(6.83331+1.9999)x62.30563, direction TLT
-..\hbox(6.83331+0.0)x62.30563, direction TLT
+..\hbox(6.83331+0.0)x62.30563, glue set 30.00005fil, direction TLT
+...\glue(\leftskip) 0.0 plus 1.0fil
```
FrankMittelbach: looks like there is a similar change here in this file

required/tools/testfiles/tlb-multicol-eisuke.tlg
```diff
@@ -24,22 +24,24 @@ LaTeX Font Info:    External font `cmex10' loaded for size
...
 % t=96.5 plus 14.0 plus 2.0fill minus 14.0 g=1601.80013 b=0 p=-51 c=-51#
 Overfull \hbox (0.97234pt too wide) in paragraph at lines ...
- []$
+ []
 \hbox(6.83331+1.9999)x61.3333
 .\glue(\leftskip) 0.0 plus 1.0fil
 .\vbox(6.83331+1.9999)x62.30563
-..\hbox(6.83331+0.0)x62.30563
+..\hbox(6.83331+0.0)x62.30563, glue set 30.00005fil
+...\glue(\leftskip) 0.0 plus 1.0fil
```
FrankMittelbach: and in this one

FrankMittelbach (commented):
before approving I would like to understand what is going on with these \leftskip appearing as a "fill" suddenly. Seems to me that this indicates a layout change in the test files

u-fischer:
There is a shift. If I compile
~~~~
\documentclass{article}

\begin{document}
x

\begin{flushright}\mbox{Answer\hspace{4em}}\end{flushright}

\begin{flushright}\underline{Answer\hspace{4em}}\end{flushright}
\end{document}
~~~~

then I get with the old code

image

and with the new one

image

FrankMittelbach:
> There is a shift. If I compile

so there is a bug in the code, thanks for such a good example @u-fischer. Means back to the drawing bord @josephwright

FrankMittelbach:
> I'm looking back here at the parameter names. I wonder if we should base them on the OpenType ones, in order below the text:
> 
>     * UnderbarVerticalGap (`\Umathunderbarvgap`)
> 
>     * UnderbarRuleThickness (`\Umathunderbarrule`)
> 
>     * UnderbarExtraDescender (`\Umathunderbarkern`)

If those are the OpenType names then I think we should use those.

josephwright:
Right, the issue with the shift is because I've tried to avoid adding an extra box level, but that fails if there is glue in the content. The PR currently uses `\unhcopy`: if we switch to `\copy`, then we deal with the spacing but have to live with another box level. Anyone see a way to avoid the extra box but preserve glue setting (I don't).

josephwright:
I'm going to update and force-push to give a clear idea of where we are now.

josephwright:
> Anyone see a way to avoid the extra box but preserve glue setting (I don't).

Ah, I'm being daft - we can of course just re-typeset the content.

u-fischer:
> Ah, I'm being daft - we can of course just re-typeset the content.

If you typeset twice you have to consider tagging and counters.

josephwright:
> > Ah, I'm being daft - we can of course just re-typeset the content.
> 
> If you typeset twice you have to consider tagging and counters.

It's OK, doesn't work anyway - you still have the glue setting business anyway. So I really don't see a way out of adding a box level :-(

josephwright:
> > Ah, I'm being daft - we can of course just re-typeset the content.
> 
> If you typeset twice you have to consider tagging and counters.

It's OK, doesn't work anyway - you still have the glue setting business anyway. So I really don't see a way out of adding a box level :-(

FrankMittelbach:
> Right, the issue with the shift is because I've tried to avoid adding an extra box level, but that fails if there is glue in the content. The PR currently uses `\unhcopy`: if we switch to `\copy`, then we deal with the spacing but have to live with another box level. Anyone see a way to avoid the extra box but preserve glue setting (I don't).

I guess I wouldn't worry about the extra box level too much.

FrankMittelbach (approved):
much better :-)

 also the tlg diffs now only have the minimally expected differences (mathon/off gone) and an implicit kern move to explicit (I think)

davidcarlisle (approved):
merge conflict in changes but otherwise looks good now

PR #2140, opened 2026-07-03
Avoid confusion due to "{c}" at end of arg spec

### LINKED ISSUE #2139: ltcmd environment with default argument `c' triggers verbatim
## Brief outline of the bug

When using `\NewDocumentEnvironment` with an `O`-type argument that has default value `c`, and the argument is not present, the next character is made verbatim.

It seems that [`\__cmd_prepare_signature_verb_chk:n`](https://github.com/latex3/latex2e/blob/develop/base/ltcmd.dtx#L2561) checks for the last item in the arg spec, which is the singleton `c` in this case. This causes `\__cmd_grab_D_verb_safe` to be used, as shown by the typeout below.

```latex
\RequirePackage{latexbug}
\documentclass{article}

\NewDocumentEnvironment{myAenv}{O{A}}{BEGIN(#1)}{END(#1)}
\NewDocumentEnvironment{mycenv}{O{c}}{BEGIN(#1)}{END(#1)}

\begin{document}

\begin{myAenv}%comment
hello %comment
\end{myAenv}

\begin{myAenv}
[arg]
hello %comment
\end{myAenv}

\begin{mycenv}%comment
hello %comment
\end{mycenv}

\begin{mycenv}
[arg]
hello %comment
\end{mycenv}

\begin{mycenv}[arg]%comment
hello %comment
\end{mycenv}

\typeout{\string\myAenv=\meaning\myAenv}
\typeout{\string\mycenv=\meaning\mycenv}

\end{document}
```

[tmp-envtest.log](https://github.com/user-attachments/files/29616940/tmp-envtest.log)

josephwright:

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Under development
- [x] Ready to merge

- [x] Test file(s) added

### DIFF
--- a/base/ltcmd.dtx
+++ b/base/ltcmd.dtx
@@ -1032,7 +1032,7 @@
 %    \end{macrocode}
 %
 %    \begin{macrocode}
-\cs_generate_variant:Nn \tl_head:n { e }
+\cs_generate_variant:Nn \tl_if_in:nnT { e }
 %    \end{macrocode}
 %
 % \subsection{Variables and constants}
@@ -2563,7 +2563,8 @@
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_prepare_signature_verb_chk:n #1
   {
-    \tl_if_in:enTF { \tl_trim_space:e { \tl_to_str:n {#1} } \exp_not:N \q_stop }
+    \tl_if_in:enT
+      { \tl_trim_spaces:e { \tl_to_str:n {#1} } \exp_not:N \q_stop }
       { c \q_stop }
       { \bool_set_true:N \l_@@_final_verb_bool }
   }
--- a/base/ltcmd.dtx
+++ b/base/ltcmd.dtx
@@ -1032,7 +1032,7 @@
 %    \end{macrocode}
 %
 %    \begin{macrocode}
-\cs_generate_variant:Nn \tl_if_in:nnT { e }
+\cs_generate_variant:Nn \tl_if_in:nnT { ee }
 %    \end{macrocode}
 %
 % \subsection{Variables and constants}
@@ -2563,9 +2563,9 @@
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_prepare_signature_verb_chk:n #1
   {
-    \tl_if_in:enT
+    \tl_if_in:eeT
       { \tl_trim_spaces:e { \tl_to_str:n {#1} } \exp_not:N \q_stop }
-      { c \q_stop }
+      { \tl_to_str:n { c } \exp_not:N \q_stop }
       { \bool_set_true:N \l_@@_final_verb_bool }
   }
 %    \end{macrocode}
--- a/base/ltcmd.dtx
+++ b/base/ltcmd.dtx
@@ -35,7 +35,7 @@
 %
 %    \begin{macrocode}
 \def\ltcmdversion{v1.3n}
-\def\ltcmddate{2026-07-03}
+\def\ltcmddate{2026-07-28}
 %    \end{macrocode}
 %
 %<*driver>
@@ -2530,7 +2530,7 @@
 % \begin{macro}{\@@_prepare_signature:n}
 % \changes{v1.3a}{2025-01-23}{Extend to cover \texttt{c}-type grabbing}
 % \begin{macro}{\@@_prepare_signature_verb_chk:n}
-% \changes{v1.3n}{2026-07-03}{Test last character as a string}
+% \changes{v1.3n}{2026-07-28}{Test last character as a string}
 % \begin{macro}{\@@_prepare_signature:N}
 % \begin{macro}{\@@_prepare_signature_bypass:N}
 %   Actually creating the signature uses the same loop approach as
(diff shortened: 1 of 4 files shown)


base/ltcmd.dtx
```diff
@@ -2551,18 +2552,20 @@
     \tl_clear:N \l_@@_process_one_tl
     \bool_set_false:N \l_@@_process_some_bool
     \tl_clear:N \l_@@_signature_tl
+    \bool_set_false:N \l_@@_final_verb_bool
```
FrankMittelbach, line 2555: why set unconditionally here and then overwrite later? Instead of the TF as before? Seems clear to me (not sure which is faster: two assignments or  droppening an F argument)

base/ltcmd.dtx
```diff
@@ -2551,18 +2552,20 @@
...
     \@@_prepare_signature_verb_chk:n {#1}
     \@@_prepare_signature:N #1 \q_recursion_tail \q_recursion_stop
     \bool_if:NF \l_@@_expandable_bool { \@@_flush_m_args: }
   }
 %    \end{macrocode}
-%   A quick check on the final arg.~type.
+%   A quick check on the final arg.~type, done using a string approach to
```
FrankMittelbach:
```suggestion
%   A quick check on the final arg.~type, done using a string approach  (that retains braces) to
```

FrankMittelbach (approved):
2 minor suggestions

base/ltcmd.dtx
josephwright: For all of the other control vars, we set up front then just reset where required - it seemed more natural to bring this into line whilst looking at the code.

base/ltcmd.dtx
josephwright: Thought that was obvious, but fair enough

base/ltcmd.dtx
FrankMittelbach: fair point. I was not looking at the overall code, just was a bit surprised to see if being moved out.

base/ltcmd.dtx
FrankMittelbach: obvious only if you do a lof of such coding. I had to think twice why it solves the problem.

base/ltcmd.dtx
```diff
@@ -2551,18 +2552,20 @@
...
 %    \end{macrocode}
-%   A quick check on the final arg.~type.
+%   A quick check on the final arg.~type, done using a string approach  (that retains braces) to
+%   avoid an issue with \verb|{c}| as the last balanced text.
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_prepare_signature_verb_chk:n #1
   {
-    \str_if_eq:eeTF { \tl_head:e { \tl_reverse:n {#1} } } { c }
+    \str_if_eq:eeT
+      { \tl_head:e { \tl_reverse:e { \tl_to_str:n {#1} } } } { c }
```
Skillmon: Could also use

```suggestion
    \exp_last_unbraced:Ne \token_if_eq_charcode:NNTF
      { \tl_head:e { \tl_reverse:e { \tl_to_str:n {#1} } . } } c
```

one could also use the faster `\tl_head:w` but that would make the code more complicated. Also, if the input is stringified getting the last non-space token of the string could be done in a single sweep through the list instead of with `\tl_reverse:n`, but that seems to be too much hassle as an optimisation for `ltcmd` :)

josephwright: I did think about performance here - but as it's in the setup phase, I think overall we are best going with clarity

Skillmon: I have another idea that has much better performance for not trivially short argument specs:

```
\tl_if_in:enTF { \tl_trim_space:e { \tl_to_str:n {#1} } \exp_not:N \q_stop } { c \q_stop }
```

davidcarlisle: this looks good to me

davidcarlisle (changes requested):
I think I'd use

https://github.com/latex3/latex2e/pull/2140#discussion_r3566834979

rather than reversing the list, otherwise looks good.

base/ltcmd.dtx
```diff
@@ -2551,18 +2552,20 @@
...
   }
 %    \end{macrocode}
-%   A quick check on the final arg.~type.
+%   A quick check on the final arg.~type, done using a string approach  (that retains braces) to
+%   avoid an issue with \verb|{c}| as the last balanced text.
 %    \begin{macrocode}
 \cs_new_protected:Npn \@@_prepare_signature_verb_chk:n #1
   {
-    \str_if_eq:eeTF { \tl_head:e { \tl_reverse:n {#1} } } { c }
+    \tl_if_in:enTF { \tl_trim_space:e { \tl_to_str:n {#1} } \exp_not:N \q_stop }
```
Skillmon: You're only providing a `T` branch here. Should use that variant instead.

Also for code readability I'd either put both first arguments on the same line as `\tl_if_in:enT` or both below it in separate lines.

josephwright: Try now :-)

base/ltcmd.dtx
```diff
@@ -1032,7 +1032,7 @@
 %    \end{macrocode}
 %
 %    \begin{macrocode}
-\cs_generate_variant:Nn \tl_head:n { e }
+\cs_generate_variant:Nn \tl_if_in:nnT { ee }
```
Skillmon, line 1035: New variant defined in the kernel, maybe define the whole bunch of variants and document those in `l3tl`?

Skillmon (commented):
LGTM

base/ltcmd.dtx
josephwright: I move variants periodically to `l3kernel`, but I can't assume that they are available immediately, so this will need to stay for 'a while' (probably several years).

FrankMittelbach: I understand that for use of variants in packages, but the kernel does require an uptodate l3kernel so I would assume that you can do that do that fairly directly.

PR #2153, opened 2026-07-28
GHA: Timeout after 15 minutes in CI tests to avoid blocking runners

zauguin:

Currently our GitHub Actions tasks can run for up to 6 hours before they timeout. If they take a lot of time they occupy GitHub Runners which are then not available to execute other CI pipelines. Since our tests usually don't take too long, this reduces the timeout for one l3build run to 15 minutes to fail faster and free resources if things are broken.

# Internal housekeeping

## Status of pull request

- Feedback wanted 

- Ready to merge

- n/a Test file(s) added
- n/a Version and date string updated in changed source files
- n/a Relevant `\changes` entries in source included
- n/a Relevant `changes.txt` updated
- n/a Rollback provided (if necessary)?
- n/a `ltnewsX.tex` (and/or `latexchanges.tex`) updated

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -171,6 +171,7 @@ jobs:
           ${{ matrix.engine && format('-e {0}', matrix.engine) || ''}}
           ${{ matrix.first && format('--first {0}', matrix.first) || ''}}
           ${{ matrix.last && format('--last {0}', matrix.last) || ''}}"
+        timeout-minutes: 15
       # "/" is one of invalid characters in artifact names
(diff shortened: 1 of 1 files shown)


FrankMittelbach:
I don't think 15min is enough. I think I have seen tests running much longer on bad days certainly above 15 if I recall correctly.

zauguin:
I ran some analysis over the last 1000 workflow runs (that's almost exactly since the beginning of the year) with a total of ~40000 workflow jobs. The only jobs which took longer than 10min in the l3build step were the two runs yesterday which were cancelled after 3h. 

The jobs sometimes took much longer, but the timeout here is only applied to the l3build step while the jobs taking longer (up to ~30min) took a long time in the ghostscript installation step instead which is not affected by this change.

FrankMittelbach:
fine then I guess

Issue #432, opened 2020-11-22
amsmath.dtx: LuaTeX version of \genfrac@rule and \genfrac@choice are buggy


RuixiZhang42:
## Brief outline of the bug

The traditional TFM-based math symbol fonts (family&nbsp;2) are required to have `\fontdimen20` and `\fontdimen21` when constructing generalized fractions with delimiters. These two font dimensions do not have correspondences in the OpenType Math Table. So, in order to support OpenType math fonts under XeLaTeX/LuaLaTeX, `amsmath` provides different implementations for `\genfrac`.

However, there seems to be illogical code in the LuaTeX version.

## Minimal example showing the bug, Log file (required) and possibly PDF file

Not applicable. The use of OpenType math fonts needs the `unicode-math` package, which is forbidden by `\RequirePackage{latexbug}`. And it’s too much work to write `\textfont1=<some OpenType math font> \textfont2=<some OpenType math font> \textfont3=<some OpenType math font> ...` to produce a minimal broken example.

## The problem

https://github.com/latex3/latex2e/blob/e09c9d47477e8966bdb53b385797f7d92dbb34ef/required/amsmath/amsmath.dtx#L732

When `\Umathfractiondelsize\displaystyle=\z@`, the above expands to `2.39\fontdimen62.39\tw@`, which is ridiculous.

## The fix

```latex
% LuaTeX version
% \changes{...}{2020/11/..}{correct fontdimen syntax}
% |#1| is the delimiter, |#2| is the math style, |#3| is the math font,
% and |#4| is the fontdimen parameter.
%    \begin{macrocode}
\def\genfrac@rule#1#2#3#4{%
\hbox{$\left#1\vcenter{\hrule \@width\z@
                       \@height
                       \ifdim\Umathfractiondelsize#2=\z@
                       #4\fontdimen6#3\tw@
                       \else
                       \Umathfractiondelsize#2%
                       \fi
                      }\right.$}}
%    \end{macrocode}
%
%    \begin{macrocode}
\def\genfrac@choice#1#2{%
\ifx @#2@\else
%    \end{macrocode}
% \changes{v2.17a}{2017/09/02}{move \cs{nulldelimiterspace} correction}
%    \begin{macrocode}
\ifx c#1\kern-\nulldelimiterspace\fi
{\delimitershortfall\z@\delimiterfactor\@m
 \mathsurround\z@\nulldelimiterspace\z@
%    \end{macrocode}
[...]

PR #2174, opened 2026-08-26
Remove special handling for LuaTeX in \genfrac

zauguin:

In \genfrac there is special handling for XeTeX and LuaTeX to avoid `\overwithdelims` and friends. The LuaTeX replacement is provided since "the relevant font dimen parameters are often not set in luatex as there are no matching values in the OpenType Math table". The replacement code conditionally uses the relevant font dimensions and tries to fallback to other logic if they are unset. The fallback code is completely broken. Triggering it manually fails with `! Missing font identifier` since a scaling factor is passed when a font identifier is expected. Looking at the git history suggests that it never worked. We could try to fix it, but since noone complained about `\binom` and friends being completely broken in LuaTeX in the last decade it seems safe to assume that the font parameter is always set in practice. Therefore we can use `\overwithdelims` and friends directly. (Also if the font parameter does turn out to be missing in many cases we can start generating it in luaotfload.)

# Internal housekeeping

## Status of pull request

- Ready to merge

- n/a Test file(s) added


mbertucci47:
This would be related to #432 I suppose

josephwright (approved):
I wonder if we need a `ltnews` entry? But otherwise good to go

FrankMittelbach (approved):
but with a short news entry I think

PR #2177, opened 2026-09-01
latex-lab-footnotes: report a real error when \@makefntext cannot be patched

makoshark:
When `\@makefntext` uses none of the four forms the patching code recognises, it falls through to a bare `\ERROR`, which is undefined. The author sees

```
! Undefined control sequence.
<argument> \ERROR
l.3 \begin{document}
```

with nothing naming the command at fault or the class responsible, at the one point in the run where a document with tagging enabled cannot proceed. The source already marks this as unfinished: "gives up with an \ERROR (which needs to get a proper definition)".

This (attempts!) to replace it with a message that says which command could not be patched, which forms are recognised, and what to change:

```
! Package fnote Error: Cannot patch \@makefntext for footnote tagging.

The tagging code adds its sockets by patching the current definition of
\@makefntext, which it can only do when that definition places the mark with
\hbox, \hb@xt@, \makebox or \@makefnmark.
The definition in force uses none of them, so footnotes will not be tagged.
Redefining \@makefntext to set the mark with \@makefnmark avoids this.
```

`\@@_tmp:w` is already set to `\exp_stop_f:` on this branch, so the run continues with footnotes untagged rather than stopping.

I stumbled onto this bug using the apa7 class, whose `\@makefntext` uses `\textsuperscript{\@thefnmark}`; reported as latex3/tagging-project#1577.

No `changes.txt` entry, since those all appear to be written by the team and reference a tagging-project issue number. Happy to add one.


FrankMittelbach (changes requested):
- please update version number and date
- also add a \changes entry
- and a changes.txt entry
thanks

makoshark:
@FrankMittelbach Done in 829de49, a new commit on the branch: version and date bumped to v0.8q/2026-09-02, plus a `\changes` entry and a `changes.txt` entry. Thanks for the review.

FrankMittelbach (approved):

FrankMittelbach:
Actually what would be good would be also a test file that triggers the error, preferably one that doesn't require an additional external class but works by mimiking a code change in that part.

makoshark:
@FrankMittelbach Test added in https://github.com/latex3/latex2e/pull/2177/commits/d85bd95ce282eea58fc6d3fea215732fa02412b2: `testfiles-footnote/footnote-patches-failed.lvt`, an `article` with `\@makefntext` redefined to place the mark with none of the four recognised forms, so no external class is needed. `\START` is in the preamble because the patching happens in the `begindocument` hook.

PR #2182, opened 2026-09-09
Extend support for nameref  

u-fischer:
This resolved most of tagging issue https://github.com/latex3/tagging-project/issues/1538

* Theorems (and proofs) store the content of the note into `\@currentlabelname`. If the note is empty they set `\@currentlabelname` to empty too. 

* Lists do not set `\@currentlabelname` by the default, but this can be activated (e.g. by nameref) with 

~~~~
\EditInstance{item}{description}
  {label-ref=\def\@currentlabel{#1}\def\@currentlabelname{#1}}
\EditInstance{item}{basic}   
  {label-ref=\def\@currentlabel{#1}\def\@currentlabelname{#1}}
~~~~

I'm unsure if it should be activated by default, and if yes, if it should also be activated for enumerate and itemize. 

Note that then both `\item[text\label{foo}]` and `\item[text] \label{foo}` would work (nameref supports only the first syntax).  `\item[text\label{foo}]` would lead to a `\newlabel` with `\label` in the argument, but nameref handles that as the reference commands redefines `\label` locally to do nothing.

~~~~
\newlabel{desc:foo}{{some label text}{2}{some label text\label {desc:foo}}{description*.1}{}}
~~~~

The testfile `nameref-support` in `testfiles-block` demonstrates the options and can be use for tests.

## Status of pull request

Mostly ready, but feedback how to handle lists by default is wanted.

- Feedback wanted 
- Ready to merge

- [x] Test file(s) added


required/latex-lab/latex-lab-block.dtx
```diff
@@ -5633,6 +5638,13 @@
 %    \begin{macrocode}
     \IfValueT{#3} { \tl_set:Nn \l_@@_note_tl {#3} }
 %    \end{macrocode}
+% We set \cs{@currentlabelname} to the current note or to empty.
```
FrankMittelbach, line 5645: why to empty? wouldn't it be better to use something like "Theorem 2" sas the name if there is no note? I mean empty is always kind of wrong

required/latex-lab/latex-lab-block.dtx
```diff
@@ -7825,7 +7844,9 @@
 % \item References: a function \texttt{counter-ref} for how the label
 %   should be referred to when it is constructed from the counter,
 %   \texttt{label-ref} and \texttt{label-autoref} used when \tn{item}
-%   has an optional argument.
+%   has an optional argument. The functions should not typeset something
+%   but e.g. define how \cs{@currentlabel} and \cs{@currentlabelname} should
```
FrankMittelbach:
```suggestion
%   but, e.g., define how \cs{@currentlabel} and \cs{@currentlabelname} should 
```

required/latex-lab/latex-lab-sec.dtx
```diff
@@ -258,7 +258,6 @@
     \seq_get:NN \g_@@_sec_stack_seq \l_@@_sec_tmpa_tl
     \int_compare:nNnT {#1}<{\exp_last_unbraced:NV\use_ii:nnn\l_@@_sec_tmpa_tl+1}
       {
-        \seq_get:NN\g_@@_struct_tag_stack_seq \l_@@_sec_tmpb_tl
```
FrankMittelbach, line 261: what's happening here? what is that fixing?

FrankMittelbach (changes requested):
Can't see much use for enabling \nameref support for \item of enumerate or itemize. On the other hand it probably wouldn't hurt either. And not sure it is helpful to enable the support only inside the nameref package.

required/latex-lab/latex-lab-block.dtx
u-fischer: well if a section has no title (`\section{}`) we are not setting `\@currentlabelname` to the number too. If the title is empty it is empty, why filling it up with some arbitrary other data?

FrankMittelbach: That's different in my opinion. a) I wasn't suggesting to set it to the number, I was suggesting to set it to the name, i.e., "prefix + number" which seems to be a useful value for a nameref in my opinion and b) in real life heading aren't normally empty so the question doesn't really arise.

But let's see what others think.

required/latex-lab/latex-lab-sec.dtx
u-fischer: I forgot to delete the line in the last pr, it does nothing (the next line sets the variable again).

u-fischer:
@FrankMittelbach 

> On the other hand it probably wouldn't hurt either. 

It changes some labels and different behaviour can always harm someone ;-)  

But I'm not against activating it by default for both if everyone agree, it would be less work as nameref wouldn't need an update.

> And not sure it is helpful to enable the support only inside the nameref package.

Until now only nameref (and zref-titleref) did support that, so anyone needing it had to use these packages.

davidcarlisle:
> Until now only nameref (and zref-titleref) did support that, so anyone needing it had to use these packages.

true but the direction of travel in recent releases has been to move more extended label handling into the kernel so I think that would be OK here too

base/doc/ltnews44.tex
```diff
@@ -462,8 +462,17 @@ \subsection{\cs{newlist} failed for \texttt{description} lists}
 \taggingissue{1518}
 
 
+\subsection{Storing more titles for named references}
 
-
+When using the \pkg{nameref} package it is possible to reference the names of theorems (as given by the optional argument) and to reference the optional argument of items of descriptions list. \pkg{nameref} made this possible by patching the internal commands of theorems and lists. These patches no longer worked with the new theorem and list code when \cs{DocumentMetadata} is used. This has now been repaired: Theorems and items with optional argument will update \cs{@currentlabelname} and so allow \cs{label} to store this value for later reference. Items will also store the optional argument in \cs{@currentlabel} and so \cs{ref} will reference this value:
```
mbertucci47:
```suggestion
When using the \pkg{nameref} package it is possible to reference the names of theorems (as given by the optional argument) and to reference the optional argument of items of description lists. \pkg{nameref} made this possible by patching the internal commands of theorems and lists. These patches no longer worked with the new theorem and list code when \cs{DocumentMetadata} is used. This has now been repaired: Theorems and items with optional arguments will update \cs{@currentlabelname} and so allow \cs{label} to store this value for later reference. Items will also store the optional argument in \cs{@currentlabel} and so \cs{ref} will reference this value:
```

FrankMittelbach (approved):

davidcarlisle (approved):

PR #2186, opened 2026-09-14
Use lualatex to typeset ltnews

josephwright:
Build system changes only


base/build.lua
```diff
@@ -227,9 +227,21 @@ function bundleunpack ()
   return 0
 end
 
+-- Due to our need to set up the format, we pass the exe here not the cmd
+specialtypesetting = specialtypesetting or {}
+specialtypesetting["ltnews.tex"] = {cmd = "luatex"}
+-- We have to do this for all the numbered issues as well, as they are not
+-- generated from the main ltnews.tex file but are separate files
+for i = 1,44 do
+  specialtypesetting["ltnews" .. string.format("%02d",i) .. ".tex"] = {cmd = "luatex"}
```
FrankMittelbach: don't like :-) that means for each new one we have to update the loop here

(and I think one of them doesn't exist (18 or so))

Can't you make the loope a bit longer so that this doesn't need to be touched while we still have the separation?

build-config.lua
```diff
@@ -42,14 +42,20 @@ typesetsuppfiles = typesetsuppfiles or
...
-  return runcmd(
-    'pdftex -fmt=pdflatex -interaction=' .. mode .. ' -jobname="' ..
-      string.match(file,"^[^.]*") .. '" "\\input ' .. file .. '"',
+  return runcmd(cmd .. " -fmt=" .. fmt ..
+    ' -interaction=' .. mode ..
+    ' -jobname="' .. string.match(file,"^[^.]*") ..
+    '" "\\input ' .. file .. '"',
     dir,{"TEXINPUTS","TEXFORMATS","LUAINPUTS"})
 end
+-- This means we need to use "pdftex" not "pdftex"
```
FrankMittelbach: staring at this  line: what is there difference between `"pdftex" not "pdftex"`?

FrankMittelbach (approved):
very far in the future :-)

base/build.lua
davidcarlisle: It would be better here to use  a pattern such as `ltnews%d%d.tex`  and then handle matches not just strings in  `specialtypesetting` , as we do in other l3build tables such as `tagfiles  {"*.cls","*.dtx","*.fdd","*.ins","*.tex","README.md"}` it doesn't matter really if it's "just us" but possibly other projects may want to choose which engine to use for different classes of file.

josephwright: `specialtypesetting` is a table lookup not a loop, as it's meant to be for, well, one-offs :-)

davidcarlisle: yes I know, but perhaps "needing to use lualatex for some docs" isn't so special, and we should perhaps change that or you could still use a table and use a file glob not an integer loop to get the list of names. (Either way it does not block this PR just something to consider in l3build, I think)

josephwright: I can look at that - it's I think something we can change without affecting stablity

u-fischer (approved):

base/doc/ltnews44.tex
```diff
@@ -536,6 +527,12 @@ \subsection{Drop special handling for \LuaTeX\ in \cs{genfrac}}
 
 \section{Documentation changes}
 
+\subsection{Tagging \LaTeX{} News}
+
+We have adjusted the sources for \emph{\LaTeX{} News} (including this news!) to
+enabling tagging (UA/2) automatically. This applies to the combined newsletters
```
FrankMittelbach:
```suggestion
enabling tagging (PDF/UA-2) automatically. This applies to the combined newsletters
```

base/ltnews.cls
```diff
@@ -34,14 +34,14 @@
 % newsletters, or as an example of how simple newsletters can be set
 % with LaTeX.  The changes log is at the end of the file.
 
-% Copyright 1994-2004
+% Copyright 1994-2006
```
FrankMittelbach:
```suggestion
% Copyright 1994-2026
```

FrankMittelbach (changes requested):

base/doc/ltnews44.tex
```diff
@@ -517,16 +517,7 @@ \subsection{Further removal of \enquote{fake math} from the kernel}
...
-\section{Changes to packages in the \pkg{amsmath} category}
-
-\subsection{Drop special handling for \LuaTeX\ in \cs{genfrac}}
-
-The \cs{genfrac} command relies on engine functionality to place delimiters
-around fractions in traditional \TeX\ engines but emulated this functionality
-in \XeTeX\ and \LuaTeX. For \LuaTeX\ this emulation was broken and is not
-needed and interferes with tagging, therefore it has been removed.
-%
-\githubissue{432}
```
FrankMittelbach, line 588: this looks suspicious. Is that the latest file or is this overwriting an earlier version?

base/doc/ltnews44.tex
```diff
@@ -536,6 +527,12 @@ \subsection{Drop special handling for \LuaTeX\ in \cs{genfrac}}
 
 \section{Documentation changes}
 
+\subsection{Tagging \LaTeX{} News}
+
+We have adjusted the sources for \emph{\LaTeX{} News} (including this news!) to
+enabling tagging (PDF/UA-2) automatically. This applies to the combined newsletters
+and each individual file.
+
```
FrankMittelbach: This is part of our effort to make all \LaTeX{} documentation accessible over time.

or some such text

base/doc/ltnews44.tex
u-fischer: This must be from Marcel's PR, it could be that there was an overlap and that my version hadn't his changes.

josephwright: Sorted - I will rebase before merged to get rid of this change

FrankMittelbach: in that case consider approved

FrankMittelbach (approved):

Issue #2191, opened 2026-09-16
Add a simple wrapper around token.set_lua


Udi-Fogiel:
## Brief outline of the enhancement

**LaTeX2e generally cannot add new features without an extreme amount of care to accommodate backwards compatibility. Please do not be offended if your request is closed for being infeasible.**

I wish that ltluatex will define a more convenient way to define TeX commands from Lua.
Currently you need to manually get the index from `luatexbase.new_luafunction` and insert it
in the `functions_table`.

Suggested definition:

```lua
function luatexbase.new_luacommand(csname, fn, ...)
  local fntable = lua.get_functions_table()
  local fnalloc = luatexbase.new_luafunction(csname)
  fntable[fnalloc] = fn
  token.set_lua(csname, fnalloc, ...)
end
```

I have no strong opinion about the name.

## Minimal example showing the current behaviour

```latex
\RequirePackage{latexbug}
\documentclass{article}
\directlua{
local function foo()
    print("\string\n=====\string\n foo\string\n=====\string\n")
end

local fnalloc = luatexbase.new_luafunction("foo")
print(fnalloc)
local fntable = lua.get_functions_table()
fntable[fnalloc] = foo
token.set_lua("foo", fnalloc)
}
\begin{document}
\foo
\end{document}
```

## Minimal example showing the desired new behaviour

```latex
\RequirePackage{latexbug}
\documentclass{article}
\directlua{
local function foo()
    print("\string\n=====\string\n foo\string\n=====\string\n")
end

luatexbase.new_luacommand("foo",foo)
}
\begin{document}
\foo
\end{document}
```

gucci-on-fleek:
A strong 👍 from me. And I also like your current name ~~`luatexbase.new_luafunction()`~~ `luatexbase.new_luacommand()`.

davidcarlisle:
> A strong 👍 from me. And I also like your current name `luatexbase.new_luafunction()`.

@gucci-on-fleek  `new_luafunction` is the existing name  the new wrapper is probably best seen as a lua (and easier to use) version of `\newluacmd` (see ltnews35)

Udi-Fogiel:
> > A strong 👍 from me. And I also like your current name `luatexbase.new_luafunction()`.
> 
> [@gucci-on-fleek](https://github.com/gucci-on-fleek) `new_luafunction` is the existing name

I think he meant to write `luatexbase.new_luacommand`.

[...]

PR #2192, opened 2026-09-16
Add luatexbase.new_luacommand (#2191)

Udi-Fogiel:
Wraps new_luafunction and token.set_lua into a single call for binding a Lua function to a control sequence.

# Internal housekeeping

## Status of pull request

- Feedback wanted 

- [ ] Test file(s) added


davidcarlisle:
I think I'm generally in favour of doing something like this, back when we added `new-luafunction` in `lt;uatex` it was rather new in LuaTeX and we didn't really have any examples and I've thought before the interface in `ltluatex` wasn't quite right having to do more than is ideal with managing the underlying table index.

FrankMittelbach (commented):
no real opinion, that is for others to approve or disaprove, but as far as I can see it looks fine (which name to use may be a question).

mbertucci47 (approved):

PR #2194, opened 2026-09-18
fix for latex3/tagging-project#1600  (on the correct branch)

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [ ] Test file(s) added


davidcarlisle (approved):

josephwright (approved):

u-fischer:
\part is noticibly lower on the page then with legacy (add `\usepackage{graphicx}` to ensure that the mediabox are the same) and in legacy the header shows a number from a chapter or section which is missing now:

image

image

FrankMittelbach:
> \part is noticibly lower on the page then with legacy (add `\usepackage{graphicx}` to ensure that the mediabox are the same) and in legacy the header shows a number from a chapter or section which is missing now:

I have no idea what you are refering to here.  Can I have an MWE?

u-fischer:
~~~~
\DocumentMetadata{}
\documentclass{amsbook}
\usepackage{graphicx}
\begin{document}

\part{Introduction}

\chapter[a short chapter]{A long chapter}
Some words

\section[A short section]{A long section}
Some other words

\newpage

Yet some other words

\newpage

And some more

\newpage
\end{document}
~~~~

left new, right legacy:

image


image

FrankMittelbach:
> and in legacy the header shows a number from a chapter or section which is missing now:

This is unrelated to the PR (it never worked with the new tagging code and requires also adjusting the class definitions `\chaptermark` etc. ). I'm inclined to leave it as is for now and only add a note that the running headers need further adjustments.

FrankMittelbach:
@u-fischer okay now?

u-fischer:
The header shows now "Chapter 1. A short chapter" instead of "1. A short chapter"? And the testfile fails?

u-fischer:
beside the headers I still see a difference in the height of \part (the lower version if from the new code):

image

FrankMittelbach:
> The header shows now "Chapter 1. A short chapter" instead of "1. A short chapter"? And the testfile fails?

that should be fixed with the next upload.

FrankMittelbach:
> beside the headers I still see a difference in the height of \part (the lower version if from the new code):

possible, but this PR is not about producing a amsbook version that is 100% identical in design as the old one (it was only intended to fix the headings so that they do not produce errors. This was not idential before and and the slight differences on that page do not affect later pagination.

PR #2197, opened 2026-09-19
Update testfiles

muzimuzhi:
To track updates from `lua-unicode-math` 0.11 (2026-09-14).

# Internal housekeeping

## Status of pull request

- Ready to merge

## Checklist of required changes before merge will be approved
- [ ] Test file(s) added
- [ ] Version and date string updated in changed source files
- [ ] Relevant `\changes` entries in source included
- [ ] Relevant `changes.txt` updated
- [ ] Rollback provided (if necessary)?
- [ ] `ltnewsX.tex` (and/or `latexchanges.tex`) updated

PR #2198, opened 2026-09-20
Fix tagging 1603

mbertucci47:

# Internal housekeeping

## Status of pull request

- Feedback wanted 

- [x] Test file(s) added


mbertucci47:
Sorry, meant to do this in my own fork :) Since I've already done it, I'll leave it here for easy viewing of the diffs.

All but one of the test changes are due to the lua-unicode-math update and also handled in #2197.

The only real test change is mtag-002. I don't know what is expected there.

FrankMittelbach:
> The only real test change is mtag-002. I don't know what is expected there.

well, just from the diff, I would presume that the test tries to verify that `\[ a \[ b \] c \]` gives a suitable error message, but that is no longer the case after you change. Instead it seems to compile (which it shouldn't) but I haven't analysed it further.

mbertucci47:
I added a test file but this shouldn't be merged as is since it just makes `\math_register_env:nn` equivalent to `\math_register_halign_env:nn`, so probably one of those should be deprecated.

mbertucci47:
Hmm this definitely still errors:
```tex
\DocumentMetadata{tagging=on}
\documentclass{article}

\ExplSyntaxOn
\cs_set_eq:NN \math_register_env:nn \math_register_halign_env:nn
\ExplSyntaxOff

\begin{document}

\[ a \[ b \] c \]

\end{document}
```
The error is
```
! LaTeX Error: Bad math environment delimiter.
```

FrankMittelbach:
> ```
> 
> \ExplSyntaxOn
> \cs_set_eq:NN \math_register_env:nn \math_register_halign_env:nn
> \ExplSyntaxOff
> ```

your assignment happens too late in this examples I think. `\math_register_env:n`was already applied to equation* in that case

mbertucci47:
@FrankMittelbach I don't think so. latex-lab-math.ltx has
```tex
\tl_gput_right:Nn \@kernel@before@begindocument
  {
    \math_register_env:n { equation }
    \math_register_env:n { equation* }
    \RenewDocumentEnvironment{math} {b}{$#1$}{}
    \RenewDocumentEnvironment{displaymath} {b}{\[#1\]}{}
  }
```
so the registration should be happening at begindocument. Adding `\ShowEnvironment{equation*}` after `\begin{document}` shows
```
> \begin{equation*}=document environment:
  #1:b
->\bool_if:NTF \l__math_collected_bool {}{\__math_process:nn {equation*}{#1}\@k
ernel@math@registered@begin \bool_set_true:N \l__math_collected_bool }\__math_e
nv_equation*_begin: #1.

> \end{equation*}:
->\__math_env_equation*_end: .
<recently read> }
                 
l.11 \ShowEnvironment{equation*}
```
which is what I'd expect. Compare without the `\cs_set_eq:NN`:
```
> \begin{equation*}=document environment:
  #1:b
->\bool_if:NTF \l__math_collected_bool {}{\__math_process:nn {equation*}{#1}\@k
ernel@math@registered@begin \bool_set_true:N \l__math_collected_bool }\__math_e
nv_equation*_begin: #1\__math_env_equation*_end: .

> \end{equation*}:
->.
<recently read> }
                 
l.11 \ShowEnvironment{equation*}
```

FrankMittelbach:
hmm, yes, so it seems. but the diff showed that the error vanished ... that should have been triggered vanished (or some of the errors at least. Guess that means one has to look closer at what the test produces.

mbertucci47:
mtag-002.tlg still contains
```
! LaTeX Error: Bad math environment delimiter.
See the LaTeX manual or LaTeX Companion for explanation.
Type  H <return>  for immediate help.
 ...                                              
l. ...\[ a \[ b \] c \]
Your command was ignored.
Type  I <command> <return>  to replace it with another command,
or  <return>  to continue without it.
! Missing \endgroup inserted.
<inserted text> 
                \endgroup 
l. ...\[ a \[ b \] c \]
I've inserted something that you may have forgotten.
(See the <inserted text> above.)
With luck, this will get me unwedged. But if you
really didn't forget anything, try typing `2' now; then
my insertion and my current dilemma will both disappear.
```
It just seems like some of the errors when run in nonstopmode went away.

Here's a diff of the logs: https://www.diffchecker.com/35hY1B7H/

FrankMittelbach:
ah, I see now. It was triggered 3 times before and now it is triggered twice (which is kind of right: once for the second `\[` and the the first `\]` closes the formula and the second then triggers the error again). So all in all it looks as if this is a) improving this test and b) it is also not showing any other defects elsewhere.

Thus, we should probably drop `\math_register_halign_env:nn` and just use `\math_register_env:nn` throughout.
(+ docu + ...)

mbertucci47:
Sounds good

FrankMittelbach:
> Sounds good

yep we overlpped with out analysis (and you were right from the start). Are you going to update this PR to become mergeable or want me to do it?

mbertucci47:
I can work on it.

mbertucci47:
Should `\math_register_halign_env:nn` be kept for compatibility? It looks like the kernel is the only place it is used.

mbertucci47:
I just removed `\math_register_halign_env:nn` since it isn't being used anywhere. All the tests pass so should be good to go now.

FrankMittelbach:
> I just removed `\math_register_halign_env:nn` since it isn't being used anywhere. All the tests pass so should be good to go now.

that would have been my suggestion too, given that it isn't used anywhere on CTAN (and quite unlikely to be used in the wild in documents).

FrankMittelbach:
It needs an ltnews entry (linking to the bug)

FrankMittelbach:
Finally I think we should add a new test that tests using the env/.../end hook on a selection of math display env

mbertucci47:
I did that for `equation*` in the tagging-1603 test but I'll add more environments.

FrankMittelbach (approved):

josephwright (approved):
Looks good but I wonder if here or elsewhere we need a comment about end-of-env hooks at the `b`- (and perhaps `c`-) type grabbers. We have typically said that you don't need any code in the end-of-env part of the environment setup with these, but clearly that isn't the case here.

PR #2204, opened 2026-09-23
Tagging1617

FrankMittelbach:
# Internal housekeeping

## Status of pull request

- Ready to merge

- [n/a] Test file(s) added

Changes in ltnews44 are mostly reformatting the paragraphs without other changes. The only changes is starting at around 418 adding a new secrtion.


u-fischer:
The support for the deprecated keys doesn't work properly.  I get a diff in the \showoutput in a test of tocloft. And e.g. this gives different behaviour:

~~~~
\DocumentMetadata{}
\documentclass{book}

\EditInstance{heading}{chapter}{final-code=\hrule}
%\EditInstance{heading}{chapter}{placement-end-code=\hrule}
\begin{document}
\chapter{First chapter} 
\end{document}
~~~~

base/doc/ltnews44.tex
```diff
@@ -472,12 +475,40 @@ \subsection{Heading template instances for AMS classes}
 \taggingissue{1477}
 
 
+
+\subsection{Replaced \key{start-code} and \key{final-code} keys in heading templates}
+
+The \key{placement} key in heading templates is responsible for
+defining the pagination behavior of a heading, e.g., \texttt{top}
+makes the heading starts on a new page, \texttt{normal} allows the
```
davidcarlisle:
```suggestion
makes the heading start on a new page, \texttt{normal} allows the
```

base/doc/ltnews44.tex
```diff
@@ -472,12 +475,40 @@ \subsection{Heading template instances for AMS classes}
...
+heading anywhere on a page but requires at least two lines of text
+afterwards, etc.
+
+Internally, it set two other keys (\key{start-code} and
+\key{final-code}) to carry out the necessary actions. These keys
+coulds als be explicitly set if the available options for
```
davidcarlisle:
```suggestion
could also be explicitly set if the available options for
```

base/doc/ltnews44.tex
```diff
@@ -472,12 +475,40 @@ \subsection{Heading template instances for AMS classes}
...
+\key{placement} needed further adjustments (in which case
+\key{placement} was ignored). Thus, these keys are strongly tied to
+\key{placement} but that wasn't obvious from their name, e.g., you
+couldn't just set \key{final-code} without breaking the layout of the
+heading in surprising ways.
+
+Therefore, we changed the key names to \key{placement-start-code} and
+\key{placement-end-code} to make this dependency clearer and also
+updated the documentation of all three keys. The old key names will
+remain usable for a while, but users who had used the in instances
```
davidcarlisle:
```suggestion
remain usable for a while, but users who had used them in instances
```

base/doc/ltnews44.tex
```diff
@@ -615,10 +663,11 @@ \section{Changes to packages in the \pkg{amsmath} category}
 
 \subsection{Drop special handling for \LuaTeX\ in \cs{genfrac}}
 
-The \cs{genfrac} command relies on engine functionality to place delimiters
-around fractions in traditional \TeX\ engines but emulated this functionality
-in \XeTeX\ and \LuaTeX. For \LuaTeX\ this emulation was broken and is not
-needed and interferes with tagging, therefore it has been removed.
+The \cs{genfrac} command relies on engine functionality to place
```
davidcarlisle:
```suggestion
The \cs{genfrac} command relied on engine functionality to place
```

davidcarlisle (approved):
some minor typos flagged inline which I could self accept from this interface but otherwise approve

josephwright (approved):

FrankMittelbach:
> The support for the deprecated keys doesn't work properly. I get a diff in the \showoutput in a test of tocloft. And e.g. this gives different behaviour:
> 
> ```
> \DocumentMetadata{}
> \documentclass{book}
> 
> \EditInstance{heading}{chapter}{final-code=\hrule}
> %\EditInstance{heading}{chapter}{placement-end-code=\hrule}
> \begin{document}
> \chapter{First chapter} 
> \end{document}
> ```

that wasn't supposed to work before. one should either use placement or both start-code and final-code. Do you have an non-working example in which both are set?

u-fischer:
> that wasn't supposed to work before. one should either use placement or both start-code and final-code. Do you have an non-working example in which both are set?

Well it did work and nothing in the documentation implied that it shouldn't.  The code handled the two values independently  and you could overwrite only the final code while keeping the default from the placement key for the start code.  

And this here is different too:

~~~~
\DocumentMetadata{}
\documentclass{book}

%\EditInstance{heading}{chapter}{start-code={},final-code=\hrule}
\EditInstance{heading}{chapter}{placement-start-code={},placement-end-code=\hrule}
\begin{document}
\chapter{First chapter} 
\end{document}
~~~~

davidcarlisle:
@u-fischer  hmm with `placement-end-code` you get a rule from `\hrule` but from `final-code`  (emulated) you don't,

davidcarlisle:
you get no error from

`\EditInstance{heading}{chapter}{start-code={},final-code=x\wibble\hrule}`

is the final code executed at all?

FrankMittelbach:
>  `\EditInstance{heading}{chapter}{start-code={},final-code=\hrule}`

Setting to somehting non-empty, e.g. `\EditInstance{heading}{chapter}{start-code=\relax,final-code=\hrule}`.

I grant you that it wasn't well documented but it is now. The keys had  a dependency and for speeding things up it triggered by just checking the value for the start key, the fact  you could get a mix by only setting the final-key was not intended to be an interface and I don't plan to make it one.

FrankMittelbach:
I can of course test of either key is non-empty but that just makes the processing slower.

davidcarlisle:
ah no it isn't executed as you have

```
\tl_if_empty:oF \l_@@_deprecated_start_code_tl
```

and @u-fischer  set the start to empty so the final wasn't used either.
I think you need to check for each of them separately

u-fischer:
the tocloft code set `placement=normal,final-code=\@cftpagestyle\@afterheading`, as it wants the default start code but set its own final code. Why do you want to force users, to copy and so duplicate the default start-code  in their settings?

davidcarlisle:
> I can of course test of either key is non-empty but that just makes the processing slower.

it's only two tests and if speed is an issue you could use the L3 name for `\ifx\@empty\l_@@_deprecated_start_code_tl`  rather than `\tl_if_empty:oF` which is doing rather more than you need here.

FrankMittelbach:
for the compatibility stuff I don't care as this goes out the moment Joseph implements the deprecated key interface. I have already adjusted to two tests.

u-fischer:
I don't think that the new text describes what is happening 

> However, if the \key{placement-start-code} key has been set to  be non-empty, then the \key{placement} key setting is fully ignored.

From this I would have expected the effect of an empty placement-end-code and to get everything on one page instead the final code from the page placement seems still to be in effect:

~~~~
\DocumentMetadata{}
\documentclass{book}

\EditInstance{heading}{chapter}
 {placement=page,placement-start-code=\hrule}
 
%\EditInstance{heading}{chapter}
% {placement-end-code={},placement-start-code=\hrule}
 
\begin{document}
some text 
\chapter{First chapter} 
next text
\end{document}
~~~~

u-fischer:
Why does line 2379 use `\str_case:Vn \l_@@_placement_tl` and 2431 `\str_case:VnF \l_@@_placement_tl`  (with an F branch)?

u-fischer (approved):

required/latex-lab/changes.txt
```diff
@@ -1,3 +1,10 @@
+2026-09-23  Frank Mittelbach  <Frank.Mittelbach@latex-project.org>
+
+	* latex-lab-sec-template.dtx (subsubsection{Templates of type \insttype{heading}}):
+	Renamed start-code key to placement-start-code (tagging/1617)
+	Renamed final-code key to placement-end-cod} (tagging/1617)
```
mbertucci47, line 5:
```suggestion
	Renamed final-code key to placement-end-code (tagging/1617)
```

required/latex-lab/changes.txt
```diff
@@ -1,3 +1,10 @@
...
+	Upadate documentation for these keys and placement key.
```
mbertucci47, line 6:
```suggestion
	Update documentation for these keys and placement key.
```

required/latex-lab/latex-lab-sec-template.dtx
```diff
@@ -668,30 +672,58 @@
...
+%               \key{placement-end-code} to a non-empty value, so that
+%               the \key{placement} key is fully ignored.
+%              }
+%              {\texttt{normal}}
+%
+%  \TemplateKey{placement-start-code}{tokenlist}
+%              {The value for this key is normally set by the
+%               \key{placement} key. Executed before the heading
+%               starts, so can issue, for example, a \cs{clearpage}.
+%               However, it is \emph{not} a user key that can ba
```
mbertucci47, line 705:
```suggestion
%               However, it is \emph{not} a user key that can be
```

mbertucci47 (changes requested):
Sorry, a bit late

required/latex-lab/changes.txt
FrankMittelbach: you don't like upadate?

PR #2206, opened 2026-09-23
Put math into artifacts in MathML SE mode unless overridden by luamml

zauguin:
When generating MathML structure elements with luamml the marked content attributes of all nodes know to luamml get overwritten. The only nodes which are affected by the mc in effect at the beginning are nodes which have been replaced in luamml. These should be marked as artifacts to avoid them showing up out of place at the beginning of the `Formula` structure element.

# Internal housekeeping

## Status of pull request

- Feedback wanted 
- Ready to merge

- [ ] Test file(s) added


davidcarlisle (approved):

davidcarlisle:
perhaps you should only do this for luatex. As has come up elsewhere `mathml-SE` with pdftex means "no MathML at all" so currently `mathml-SE`  with pdftex `a+b` just falls through but might be better than nothing, but this diff artifacts it so you get nothing at all (test latex-lab tagging-1037)

davidcarlisle (approved):
restoring the tick now it's luatex-only

davidcarlisle:
I hate reading pdf based tests `testfiles-math-luatex/math-20-structelem` is failing it enables SE via `\tagpdfsetup{math/mathml/structelem}`

zauguin:
I consider the pdfTeX behavior to be rather questionable and would prefer to how an error instead when mathml-SE is enabled there, but that's separate from this change.

The LuaTeX PDF test change is a noop: An empty Formula MC became an empty Artifact MC (and since the Formula MC got an MCID and the Artifact doesn't this moves the MCIDs around). We should expect similar test changes in other tests using mathml-SE though.

davidcarlisle:
yes I saw later it's just artifacting an empty MC,


Yes we were discussing just yesterday whether mathml-SE should be a no-op or explicitly be same as mathml-AF  with pdftex in the context of the tagging-status test suite , but at least for now not changing the pdftex output seems best.  so all  good as test updated.

